PR #155, opened 2015-03-17
New mapping

dangunter:
Not quite done with actual logic, but now is a good time to start merging this into develop.


meghasandesh:
The dataset_service seems to raise the 500 error when "GET /app/get_datasets/?organization_id=1 HTTP/1.1" is called. Any ideas on how to solve this?

mmclark:
Dan - There was some discussion in yesterday's call that merging this 
branch may break develop for other developers, due to some configuration 
changes you made for your environment.  In your view, what risk is there 
of that happening?

meghasandesh:
I tried a few things over the past few days - having a conditional dev.py setup, moving to Dan's setup (unsuccessfully because I couldn't set up the env correctly), and adding changes to develop one file at a time. One thing I noticed is that the program returns an error on adding the new models.py file. Specifically, these two lines are causing the HTTP 500 error I have been seeing so far:

# program names should match a value in common.mapper.Programs
source_prog = models.CharField(blank=True, max_length=80,)
# program version is in format "x.y[.z]"
source_program_version = models.CharField(blank=True, max_length=40,)

Commenting out these two lines seems to maintain the functioning of the program. If I understood the context of the variables above, they are defined as part of the file metadata which will be stored in the DB. So we would need to create new columns in the AWS in order for the program to work correctly.

Dan and Miles,

Let me know if you think this is the right step. If it turns out to be correct, can we have the AWS DB updated with the new columns?

Thanks

Megha

dangunter:
I think commenting them out is a hack. I thought the point of a Django model was that we didn't have to repeat ourselves for every data sink. I think you should just merge the whole thing and test it as a whole -- there are dependencies that are going to break if you merge one file and not another.

dangunter:
re: merging breaking the configs, that's possible. I assume they could verify and fix this?

RDmitchell:
Not completely understanding the context, seeing the words "add new
columns" "in the AWS DB" in this thread makes me a bit nervous.

Would we need to write a DB migration for the existing db on our instance
(and for everyone else who would need to make this change)?

Robin

On Fri, Mar 20, 2015 at 11:46 AM, meghasandesh <notifications@github.com>
wrote:

> I tried a few things over the past few days - having a conditional dev.py
> setup, moving to Dan's setup (unsuccessfully because I couldn't set up the
> env correctly), and adding changes to develop one file at a time. One thing
> I noticed is that the program returns an error on adding the new models.py
> file. Specifically, these two lines are causing the HTTP 500 error I have
> been seeing so far:
> program names should match a value in common.mapper.Programs
>
> source_prog = models.CharField(blank=True, max_length=80,)
> program version is in format "x.y[.z]"
>
> source_program_version = models.CharField(blank=True, max_length=40,)
>
> Commenting out these two lines seems to maintain the functioning of the
> program. If I understood the context of the variables above, they are
> defined as part of the file metadata which will be stored in the DB. So we
> would need to create new columns in the AWS in order for the program to
> work correctly.
>
> Dan and Miles,
>
> Let me know if you think this is the right step. If it turns out to be
> correct, can we have the AWS DB updated with the new columns?
>
> Thanks
>
> Megha
>
> —
> <https://github.com/SEED-platform/seed/pull/155#issuecomment-84098229>.
>



-- 
Robin Mitchell
Environmental Energy Technologies Division

meghasandesh:
Confirming that the branch works. Please execute these steps inside your vagrant box inside the directory where the SEED code resides:

workon seed
export DJANGO_SETTINGS_MODULE=BE.settings.dev
./manage.py syncdb --migrate

Issue #131, opened 2015-02-24
Hide the concatenation button


RDmitchell:
Until we can figure out what the issue is with the concatenation button sometimes working and sometimes not working, I would like to hide it so that the functionality is not available.

meghasandesh:
I will cover this bit. It will be straightforward

RDmitchell:
I have tested this on seedtest and it is done. I will test again when it is on the production server, and then will close this issue.

RDmitchell:
I am moving this to Miles issue list because it is Ready for Production.

RDmitchell:
tested and closed

PR #158, opened 2015-03-18
[Issue #131] Concatenation button hidden

meghasandesh:
Concatenation button hidden.


RDmitchell:
This is tested and can be moved into the public repository and the production server.

PR #165, opened 2015-03-31
Fix for issue where some records may have leading zeros in address causing match to fail.

StephenCzarnecki:
There was an issue with some addresses having leading zeros.  After talking with Robin added a change to _normalize_address_str to do a lstrip(0) on the house field to remove leading zeros from the normalized address.  Leading zeros still appear in building data just not used for matching.  Added a test to test_strip_leading_zeros to test for this: test_strip_leading_zeros

### DIFF
--- a/seed/tasks.py
+++ b/seed/tasks.py
@@ -888,7 +888,7 @@ def _normalize_address_str(address_val):
         return None
 
     if 'house' in addr and addr['house'] is not None:
-        normalized_address = addr['house']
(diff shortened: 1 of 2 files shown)


cycustodio:
Uploaded sample covered buildings and portfolio manager. 512 building matched, as expected. 

Buildings with trailing zeros in the addresses are matching with addresses with no trailing zeros.
address_trailing_zeros

PR #166, opened 2015-03-31
Fix for issue 24: set field order in mapping based on imported data.

StephenCzarnecki:
Now uses the column order in the original file for the order of the headers.  Made the change to cache_first_line.  See notes there.  It would be better to patch MCM per the comments I left in the code but this should fix the issue until that can be done.

### DIFF
--- a/seed/tasks.py
+++ b/seed/tasks.py
@@ -630,13 +630,28 @@ def cache_first_rows(import_file, parser):
         if row:
             validation_rows.append(row)
 
-    import_file.cached_second_to_fifth_row = "\n".join(
-        [
-            ROW_DELIMITER.join(map(lambda x: str(x), r.values()))
-            for r in validation_rows
-        ]
-    )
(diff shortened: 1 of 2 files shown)


cycustodio:
Uploaded sample covered buildings and portfolio manager files on seedtest. 

In the covered buildings file the columns are as follows:
UBI, GBA, BLDGS, Address, Owner, City, State, Zip, Property Type, AYB_YearBuilt

cb_column_order

Portfolio manager files columns are as follows:
Property Id, Property Name, Year Ending, Property Floor Area (Buildings and Parking) (ft2),
Address 1, City, State/Province, Postal Code, Year Built, ENERGY STAR Score, Site EUI (kBtu/ft2),
Total GHG Emissions (MtCO2e), Organization, Generation Date, Release Date
pm_column_order

The headers on the left side are not showing the correct field names (outside the scope of this fix, but something to note). Data is stored with the correct mapping in the building view.

PR #167, opened 2015-03-31
Avoiding duplicate headers

meghasandesh:
Preventing two sets of headers from appearing in the mapping table


meghasandesh:
Thank you Miles

PR #191, opened 2015-04-10
Re-enabled code to do fuzzy mapping on non-pm files.

StephenCzarnecki:
Seems to also deal with remembering the mapping for non-pm files.  Tested by taking a file and changing a header slightly (Address -> Addresz) and seeing that when imported as a CSV it suggests "Address Line 1" and when imported as a PM file it suggests "Addresz".  Verified matching memory for CSV files by importing a file and defining a strange mapping (UBI -> Bob) and then re-importing the file and seeing that the mapping suggestion is preserved.

### DIFF
--- a/seed/views/main.py
+++ b/seed/views/main.py
@@ -1181,38 +1181,34 @@ def get_column_mapping_suggestions(request):
         }
     column_types.update(db_columns)
 
-    # All inputs
+    # Portfolio manager files have their own mapping scheme
     if import_file.from_portfolio_manager:
         _log.info("map Portfolio Manager input file")
-    suggested_mappings = {}
-    ver = import_file.source_program_version
-    for col, item in simple_mapper.get_pm_mapping(
-            ver, import_file.first_row_columns,
-            include_none=True).items():
-        if item is None:
-            suggested_mappings[col] = (col, 0)
-        else:
-            cleaned_field = item.field.replace('-', ' ')
-            suggested_mappings[col] = (cleaned_field, 100)
+        suggested_mappings = {}
+        ver = import_file.source_program_version
+        for col, item in simple_mapper.get_pm_mapping(
+                ver, import_file.first_row_columns,
+                include_none=True).items():
+            if item is None:
+                suggested_mappings[col] = (col, 0)
+            else:
+                cleaned_field = item.field.replace('-', ' ')
+                suggested_mappings[col] = (cleaned_field, 100)
 
-    # This old, "fuzzy" matching procedure has been completely
-    # removed. However, the previous_mapping part of it will be
-    # restored in future versions.
-    # else:
+    else:
     #     # All other input types
-    #     suggested_mappings = mapper.build_column_mapping(
-    #         import_file.first_row_columns,
-    #         column_types.keys(),
-    #         previous_mapping=get_column_mapping,
-    #         map_args=[import_file.import_record.super_organization],
-    #         thresh=20  # percentage match we require
-    #     )
-    #     # replace None with empty string for column names
-    #     for m in suggested_mappings:
-    #         dest, conf = suggested_mappings[m]
-    #         if dest is None:
-    #             suggested_mappings[m][0] = u''
-
+        suggested_mappings = mapper.build_column_mapping(
+             import_file.first_row_columns,
+             column_types.keys(),
+             previous_mapping=get_column_mapping,
+             map_args=[import_file.import_record.super_organization],
+             thresh=20  # percentage match we require
+         )
+        # replace None with empty string for column names
+        for m in suggested_mappings:
+            dest, conf = suggested_mappings[m]
+            if dest is None:
+                suggested_mappings[m][0] = u''
     result['suggested_column_mappings'] = suggested_mappings
     result['building_columns'] = building_columns
     result['building_column_types'] = column_types


spencerdutton:
so did we (and by we I mean you) manage to work out what was happening with
the files not loading?

Spencer

On Fri, Apr 10, 2015 at 1:43 PM, StephenCzarnecki <notifications@github.com>
wrote:

> Seems to also deal with remembering the mapping for non-pm files. Tested
> by taking a file and changing a header slightly (Address -> Addresz) and
> seeing that when imported as a CSV it suggests "Address Line 1" and when
> imported as a PM file it suggests "Addresz". Verified matching memory for
> CSV files by importing a file and defining a strange mapping (UBI -> Bob)
> and then re-importing the file and seeing that the mapping suggestion is
> preserved.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/SEED-platform/seed/pull/191
> Commit Summary
>
>    - Re-enabled code to do fuzzy mapping on non-pm files. Seems to also
>    deal with remembering the mapping for non-pm files.
>
> File Changes
>
>    - *M* seed/views/main.py
>    <https://github.com/SEED-platform/seed/pull/191/files#diff-0> (52)
>
> Patch Links:
>
>    - https://github.com/SEED-platform/seed/pull/191.patch
>    - https://github.com/SEED-platform/seed/pull/191.diff
>
> —
> <https://github.com/SEED-platform/seed/pull/191>.
>



-- 
Dr Spencer Dutton
Building 90 R 3058
Indoor Environment
Commercial Ventilation and Indoor Air Quality Group
1 Cyclotron Rd, Berkeley, California, USA
Tel:  510 4867179
Email: smdutton@lbl.gov

RDmitchell:
No

cycustodio:
Tested by using the same building file (Boston's building file); uploaded as a regular spreadsheet and again as a portfolio manager file. The column suggestions differed (fuzzy matching vs copying over the column names) along with the confidence value(?), which means the fuzzy matching is working on the regular spreadsheet (varies) but not for portfolio manager file (set to 100). 

Suggestions for the regular file. 
regular_file

Suggestions for portfolio manager file. 
portfolio_manager_file

RDmitchell:
Hmmm. 

It does seem pretty clear that the file when imported as a non-PM file is doing the (not very accurate) fuzzy matching to suggest a name. 

However, for the case where you imported the same file as a PM file, I am not sure what is happening. 

The PM fixed mapping file doesn't have fields in it like S_EXT_FIN, so I am not sure how it figured out to use the name Ext Fin, unless that mapping was somehow already remembered (which it isn't supposed to do for PM files). 

Maybe you should do this same test with a PM file (maybe the Boston PM file), importing it both as a non-PM and then a PM file and see how the mappings differ. Presumably importing it as a non-PM file, you will see the same basic behavior, ie, it does the fuzzy matching and doesn't come up with very good suggestions. And then for the PM file, it should do the mapping in the json file for all fields that are in that file.

RDmitchell:
Also, it looks like all the PM fields are being set to confidence = 100? 
Megha, Stephen or Dan, can you confirm that is the case, ie, that none are being set to 0 any more?

StephenCzarnecki:
No, looks like it is setting values to 0 to me.  If I try to upload a file
I have that has one field really changed so I know nothing
matches  ("Address" -> "QQQWWWZZZ") I am seeing this output from
the backend.

{u'City': (u'Premises City', 100), u'Zip': (u'Zip', 0), u'GBA': (u'GBA',
0), u'BLDGS': (u'BLDGS', 0), u'UBI': (u'UBI', 0), u'State': (u'State of
Washington Unique Facilities Identifier (UFI)', 100), u'QQQWWWZZZZ':
(u'QQQWWWZZZZ', 0), u'Owner': (u'Owner', 0), u'Property Type': (u'Property
Type', 0), u'AYB_YearBuilt': (u'AYB_YearBuilt', 0)}

For comparison if I try the same file as csv (not PM) I see the following
(the "Jim" mapping is something I did before to test the saved mapping,
mapped GBA -> Jim) :

{u'City': [u'Premises City', 100], u'Zip': [u'postal_code', 100], u'GBA':
[u'Jim', 100], u'BLDGS': [u'building_count', 100], u'AYB_YearBuilt':
[u'year_built', 100], u'State': [u'state_province', 100], u'QQQWWWZZZZ':
[u'Window Width', 51], u'Owner': [u'owner', 100], u'Property Type':
[u'use_description', 100], u'UBI': [u'tax_lot_id', 100]}

meghasandesh:
It is indeed setting the zero values. I am looking at a solution to address
the greying out issue.

Megha

RDmitchell:
But the logic behind how the confidence level is being set does not necessarily make sense to me. 

I would like someone to walk me through the logic in the code about how this all works.

cycustodio:
Confirmed that fuzzy mapping and saving mapped fields are enabled for raw files. 
For portfolio manager files, the mapping depends on the BEDES list and does not save the mapping. 

Test: Uploaded and saved "2014 Reporting List for Upload 12102014" as raw file and pm file separately, without changing the field names. 

Then uploaded a subset of the file with field names modified. 
```
"tax_lot_id" ==> "TAX_LOT_ID_1234"
"Parcel ID As Number" ==> "Parcel_ID_As_Number_XXXX"]
"10-Digit Parcel ID" ==> "10-Digit Parcel ID"
"2013 MAIL ADDRESS" ==> "2000 mail address"
```
When I upload this subfile as raw file, the confidence should vary. The mapping suggestions should be the saved values.  
**Result**
as_rawfile

When I upload this subfile as portfolio manager file, the confidence should be 0. The mapping suggestions should be the field names copied over. A confidence of 100 on "10-Digit Parcel ID" means the field name is saved.  
**Result**
as_portmngr

PR #204, opened 2015-04-16
Fixed a number of very small issues

axelstudios:


RDmitchell:
Alex -- Most of what you fixed were typos -- do you think any of the changes would result in something that I should test as I am going through my standard (non-automated) GUI testing?

axelstudios:
Sorry @RDmitchell, I just now saw your message.  These changes should have no effect on the user experience, I just noticed a pattern of them as I was looking through the codebase during a meeting :)

This commit also cleaned up a number of JavaScript files, particularly issues with trailing commas.  Any modern browser understands them fine, but tools like JSLint will generally flag them as compatibility errors since <IE9 will interpret them very differently.

axelstudios:
I should also mention that I noticed typos in these two BEDES fields, but I did not change them because it would have larger repercussions since they're in a database migration step: https://github.com/SEED-platform/seed/blob/master/seed/data/bedes.py#L42-43

PR #216, opened 2015-04-25
Added fields to the bedes.py file and changed pm-mapping.json based on files from Robin.

StephenCzarnecki:
Also the PM mapping code was removing hyphens (-) from the suggested pm mappings which breaks some of the mappings in the pm-mappings.json file, removed this.  Do not know if this has any other implications.


RDmitchell:
Thanks Stephen!

Issue #86, opened 2015-01-10
DATA IMPORT: File will not load


RDmitchell:
There is something wrong with a file that the program does not seem to be able to load it.

RDmitchell:
The file can be found in the gdrive folder for this issue.

mmclark:
The first problem is that SEED can't load files with less then 6 data rows.  The sample file only has one row.  I'm going to enter an issue for this.

RDmitchell:
I am closing this issue as there are other issues for dealing with the problems in this file:

Extra lines in header - https://github.com/SEED-platform/seed/issues/96

PR #224, opened 2015-04-29
Upload error handling

### LINKED ISSUE #33: Display error message for Data loading status / problems
Show Upload Errors
Program should display an error message when there is an uploading problem. For example, the server reports an error that a field has > 256 characters, and the upload process will hang. But the user does not know that there is a problem because the Progress Bar is still displayed and no error message is displayed.

### LINKED ISSUE #88: DATA IMPORT: Import file even if there is only one record
Currently the program expects 6 rows to be in a data file that is imported. We need to allow data that only has one record to be imported (for use cases where data will be imported one record at a time). 

Explore the implications of only having one record before making the change. It may affect the mapping UI (?)

### LINKED ISSUE #100: Put up error message saying Process cannot be completed
It doesn't have to say what the cause is, just that the server has crashed or whatever. So that the user doesn't just see the 25% progress bar for three hours thinking that the program is still working. 

Josh and Dan had ideas about how to implement this if you need to talk to them. 

Probably the message should say something like "The program has experienced a severe internal error. Please contact seed-support@lists.lbl.gov" or something like that.

axelstudios:
This resolves #100, resolves #33, and fixes #88.  #86 is related and should be retested (I don't have that file).

This also disables the `Data Mapping` and `Data Matching` buttons if the imported data file had 0 buildings, like when there's a file parsing error.

Let me know if you find any issues during testing/code review, or if something is unclear.


RDmitchell:
Alex

We merged this pull request into master last night and then moved it onto our test server and now it doesn't seem like we can get any of the sample files to load. When I try loading the sample covered buildings file (covered-buildings-sample.csv), it seems to hang at 45%, and when I try to load the sample PM file (portfolio-manager-sample.csv) I get the following error message:

image

axelstudios:
Very strange, I'm not able to reproduce this in my development environment.  I'm going to try using S3 instead of my local file system to more closely match your test environment.

mmclark:
I've duplicated it in my local environment using local file storage.  The problem appears to be that seed.tasks.save_raw_data() is returning an invalid value. 

https://github.com/SEED-platform/seed/pull/224/files#diff-beec97433f86a3d4e2e6f44a030ed583L732

I'm going to revert the PR so you can take a deeper look.

axelstudios:
Ah, I know what's wrong - I've been using `CELERY_ALWAYS_EAGER` to make debugging easier, and that made this asynchronous call blocking.  Working on a fix.

RDmitchell:
Do you think it is possible to get this fixed sometime early next week (5/4 or 5/5)? Thx.

axelstudios:
Definitely - Nick and I have made some changes so it's working locally, so I think we'll have it polished today.

Issue #20, opened 2014-10-28
Add filter capability for exact match or fields without data


RDmitchell:
FILTERING: Can’t specify exact match
When doing the filtering, can't specify an "exact match" -- so there is not a way to see only records that have Use Description set to "Multi-Family", not including "Multi-Family Condominium". Could specify exact matches as surrounded in "quotes" (?) or some other scheme. This might enable blank fields to be filtered by “”

(LBNL Ticket # 124)

RDmitchell:
See also Issue # 64, which I'm not sure I completely understand. It is related but not the same. We may have to talk to Marshall (DC) when we get to this issue.

RDmitchell:
Megha -- I am assigning this issue to you to investigate whether this is something that you can do through the UI or whether it is "back end" work.

RDmitchell:
Here is a screen shot that shows the issue. In the Building List, you can see that some records have no data in the Energy Star Score field. The users would like to know which records have no data in this field, but right now there is now way to filter on "null" data. You can sort by a field and have the blanks turn up and the beginning or the end, but that is not an efficient way for users to deal with these records when they may have hundreds of records with this issue. 

One way to do this would be to have the user enter two double quotes together (?) "" if they want the program to return records where the fields with this "" filter have null data. There is probably a nicer UI option, but this could be a quick fix for now, with the idea of making it more user friendly later.

![image](https://cloud.githubusercontent.com/assets/6314950/7169769/4d725f30-e37d-11e4-8631-050dac549ff4.png)

Then there is the issue of being able to filter on an exact match. Right now when you type something into the filtering field at the top of the header, the program looks for records that contain what was typed. So for example, if you wanted to filter on the Premises Occupancy Classification field, and just return "Commercial Condo", the program returns records that contain both just Commercial and just Condo, as well as those that contain Commercial Condo. So we need a way for users to specify that they want to search for an exact match on Commercial Condo.

Seems like one simple approach would be to do an exact match if the search term was enclosed in double quotes? So typing in "Commercial Condo" in the filter field would return an exact match on the combination of words. Again, we could add a more user friendly UI option later if we thought that was needed, but this would be potentially a quick way to get the functionality into the program.

![image](https://cloud.githubusercontent.com/assets/6314950/7169918/c36423f8-e37e-11e4-95a2-7a240ce3834d.png)

nicholasserra:
[...]

PR #233, opened 2015-05-11
Issue 20 - JSON field exact and null searches

nicholasserra:
Enable exact and "null" searches on the Building extra_data json blob.

This fix needed a less than ideal hack for the "null" filtering. When fields are mapped and added to the building json blob, they are only added if they contain data. So i'm doing a `contains` query on the JSONField to see if it contains what looks like a key for that field. 

When we bump to Django 1.7, we can update the postgres JsonField package, and make use of postgres's "has" filter. This note was added as a comment in the code.

Refs #20

### DIFF
--- a/seed/search.py
+++ b/seed/search.py
@@ -185,21 +185,30 @@ def filter_other_params(queryset, other_params, db_columns):
         sanitized = strip_suffixes(k, ['__lt', '__gt', '__lte', '__gte'])
         if sanitized in columns:
             return True
-        else:
-            return False
+        return False
+
+    def is_string_query(q):
+        return isinstance(q, basestring)
+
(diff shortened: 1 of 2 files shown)


RDmitchell:
Yes it’s a problem in the JSON field that if there is no data in the original imported data, the field is not there. It is on the list to deal with, ie, maybe we need to force the fields without data to still be written into the JSON blob field with “null” data.

nicholasserra:
Yeah that's an option. I wanted this to be backwards compatible until we upgrade package versions. 

via Hiptop

RDmitchell:
Right. We won’t implement that fix now.

RDmitchell:
This PR has been tested on seedtest and is working

PR #234, opened 2015-05-12
Making the progress API backwards compatible

### LINKED ISSUE #232: Process stalling during remap_buildings
I uploaded one of the sample files to seedtest and when I click on "Map Your Data" the progress bar moves to 100% then stops. The last API call returns an internal server error (500) after several progress queries.

axelstudios:
@RDmitchell 

I didn't previously notice our changes affecting the progress requests of any actions other than uploading, but this change will allow any functions expecting the old format to continue to function.

Basically, the cache previously stored only a number to represent progress, e.g. 100, but now it stores a dictionary in order to communicate additional information asynchronously to the front end.  This change allows it to continue to accept ints or floats for the time being.

Fixes #232

### DIFF
--- a/seed/views/main.py
+++ b/seed/views/main.py
@@ -1930,6 +1930,9 @@ def progress(request):
 
     if cache.get(progress_key):
         result = cache.get(progress_key)
+        # The following if statement can be removed once all progress endpoints have been updated to the new json syntax
+        if type(result) != dict:
+            result = {'progress': result}
         result['progress_key'] = progress_key
         return result
     else:


RDmitchell:
Is there anything in particular that we should test (on our seedtest server) in order to make sure this functionality is working?

axelstudios:
If clicking data matching is working normally, then everything should be all set.  This will support the new progress cache data (that carries asynchronous task status back to the front end), as well as the old ones (just a numeric progress percent) until they're all updated.

PR #242, opened 2015-05-15
8 show green button file

philngo:
Fixed merge conficts, indentation.
Adds support for an Energy data tab showing uploaded and matched green button files, and shows matched green button files on the Building info tab of the building detail view.


mmclark:
Merged and posted to seedtest - thanks!

RDmitchell:
Phil -- when you tested this feature, how did you test it? I have tried a few different methods and don't seem to see the Green Button file info under the Energy Tab. 

I don't think SEED actually currently has the capability to match the GB data to an existing building record, so I am just importing the GB XML file by itself in an otherwise empty account. I am using a GB XML file from PGE (my personal residence), and it seems to import ok. But then when look at the Building Detail screen for that record, I don't see the name of the file listed under the Energy Tab. 

Here are my steps:

image

I upload a Green Button file using the green Upload Green Button Data button
image

I select my personal PGE Green button electricity file
image

Here is a link to the google drive folder with these sample PGE GB files (I know that the PGE files are not strictly compliant with the GB standard -- I wouldn't think that's the problem but maybe it is?)
https://drive.google.com/a/lbl.gov/folderview?id=0B3fTKpZ9Dx7LTFpnRkdmWjQxazQ&usp=sharing

It takes a while, but eventually the program reports that it successfully matched (but really it is not matched but uploaded) the XML file
image

If I go to the Building List, I see the one record created by this import
image

If I click on that record, I see that there is one field in the Building Info screen, which is the Address, which is as it should be
image

If I then click on the Energy link on the right hand side of the links, the program reports that there are no green button files for this building. 
image

If it is an issue of not having a GB compliant XML file, can you point me to a good file to test with (maybe you have already put that in the repo in you testing -- just point me where to look).

Thanks

philngo:
The problem does not seem to have to do with the GB XML. It looks like the problem was that the "get_building" API endpoint didn't handle the case in which there was only one import associated with only one building - so a composite building snapshot was never created. Issued another pull request (#254) to handle this case.

RDmitchell:
There are already issues relating to this problem but maybe they don't
address it directly: # 88, # 224, # 230.


On Mon, May 18, 2015 at 9:23 AM, Phil Ngo <notifications@github.com> wrote:

> The problem does not seem to have to do with the GB XML. It looks like the
> problem was that the "get_building" API endpoint didn't handle the case in
> which there was only one import associated with only one building - so a
> composite building snapshot was never created. Issued another pull request (
> #254 <https://github.com/SEED-platform/seed/pull/254>) to handle this
> case.
>
> —
> <https://github.com/SEED-platform/seed/pull/242#issuecomment-103118219>.
>



-- 
Robin Mitchell
Environmental Energy Technologies Division

PR #299, opened 2015-07-21
Issue 247 - Rework of JSON Blob extra_data field

### LINKED ISSUE #247: Rework of JSON Blob extra_data field
When records are imported and loaded into the database, only some of the fields are stored as actual fields in the BuildingSnapShot table. Most of the fields are put into a JSON Blob field called extra_data. However, if the field in the original record does not have any data in it, that field is not written into the extra_data blob. In essence, this means that the program is losing important data, ie, the fact that a field does not have data, when is should. 

This causes the sorting problems listed in Issue # 136.

In addition, for the fields that are written into the blob, some of them so not sort properly (doing an alpha sort on numbers for example) because the type of data is not properly defined. (See Issue # 187)

Seems like at least two things need to be added to the JSON blob extra_data field:

* add all fields in a record into the JSON Blob extra_data field, even if the value of the field is null. Have a method for indicating in the blob that the field value is null, which then represents itself as blank in the various displays of that data in the GUI
* add a field type characterization (such as numeric, string, date, etc) to the fields stored in the extra_data field. This might not be completely straightforward, if there is not a way to determine the field type. We might need some discussion about how data is imported and how field types are defined. 

The first step for this issue is to develop a proposal for the changes, which can then be vetted with the other developers through this issue. Then the proposal can be implemented.

nicholasserra:
Add blank fields to building extra data json data. Previously, empty fields were excluded from the json data. This patch also updates the null filtering on json data to handle this new case and be backwards compatible with old data.

Fixes #247

### DIFF
--- a/seed/lib/mcm/mapper.py
+++ b/seed/lib/mcm/mapper.py
@@ -179,7 +179,7 @@ def map_row(row, mapping, model_class, cleaner=None, concat=None, **kwargs):
         # then, send_apply_func will reference this function and be sent
         # to the ``apply_column_value`` function.
         send_apply_func = apply_func if item in apply_columns else None
-        if value and value != '':
+        if value != None:
             model = apply_column_value(
                 item, value, model, mapping, cleaner,
                 apply_func=send_apply_func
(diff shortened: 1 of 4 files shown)


nicholasserra:
Also seeing as this is an architectural change on the BE, I would recommend this be tested before merged. I did the basic run through, but there may be a more rigorous test you do @RDmitchell .

PR #300, opened 2015-07-22
Uses cookies to set expanded/collapsed state of sidebar nav

danielmcquillen:
This update adds a cookie value to remember the expanded/collapsed state of the sidenav bar. 

It tweaks the existing way sidenav state is stored in the *menu_controller.js*, and adds an extra style to styles.less to turn off animations when first showing the page. 

The *_sidebar.html* and *index.html* files are updated to use this style to turn off the animations when first loading.

### DIFF
--- a/seed/static/seed/js/controllers/menu_controller.js
+++ b/seed/static/seed/js/controllers/menu_controller.js
@@ -17,6 +17,7 @@ angular.module('BE.seed.controller.menu', [])
   'user_service',
   '$timeout',
   '$route',
+  '$cookies',
   function(
     $scope,
     $http,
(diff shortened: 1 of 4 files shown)


rebrownlbnl:
It looks like all the outstanding PRs are failing the Travis build. Is this the same problem that we had noticed before the v1.2 push? It's failing the test_get_columns test. Who knows that area of the code and can look into why it's failing?

mmclark:
I believe either Nick Serra or Nick Long were going to look at the 
failing test, but after the latest email exchange, I'm not sure which.  
I'd like to get this sorted out before merging any further PRs into develop.

RDmitchell:
I am thinking that the most logical person to look into this is Nick Long.

mmclark:
Daniel - can you submit this PR against develop, instead of master?  Thanks!

PR #319, opened 2015-08-12
Issue #318 - Save number per page in sessionStorage

### LINKED ISSUE #318: Displayed # of Buildings resets when switching pages
Extension of Issue (https://github.com/SEED-platform/seed/issues/12)

Change the display buildings to 50 buildings, when I switched to the building detail, the display buildings went back to 10. The expected behavior is the displayed building count stays at 50 until the user makes the change.

nicholasserra:
Save building number per page in session storage. 

Fixes #318


RDmitchell:
Tested and works well.

Issue #83, opened 2014-12-18
Uploading files with duplicate data


RDmitchell:
Almost all users are going to have the issue of uploading data files, such as Portfolio Manager, multiple time during their benchmarking process. SEED should only load the data that is not already associated with the building record, and should ask the user if they want to update changed data for records already in the master building list. 

The first step is to create a design / specification document.

RDmitchell:
Here is a link to the current specification. 
https://docs.google.com/a/lbl.gov/document/d/1GH8VdxhBOVjU7iObpryLu-IwxPp0KdxwIM0wZ5HGZCQ/edit?usp=sharing

For this first cut, it is a relatively simple implementation of not writing "copies" of existing data to a building snapshot. We are not going to do any UI changes for this first implementation, it will all happen behind the scenes.

danielmcquillen:
Hi Robin,

Just requested access to this doc.

I noticed you assigned Stephen to the issue. Is your thinking that he'll do
all the client-side Javascript and just turn to me for any UI that's
needed, if it turns out that a dialog or something is required? Or do you
want me to be involved in writing the client-side Angular code?

Daniel


  [image: photo]
*Daniel McQuillen*
McQuillen Interactive Pty. Ltd.
 p:+61 4 9980 2564 | e:daniel@mcquilleninteractive.com | w:
www.simplediagrams.com | a:Melbourne, Australia
 <http://facebook.com/simplediagrams>
<http://au.linkedin.com/in/mcquilleninteractive/>
<http://twitter.com/simplediagrams>
<http://plus.google.com/+Simplediagrams>
    Get a signature like this: Click here!
<http://ws-promos.appspot.com/r?rdata=eyJydXJsIjogImh0dHA6Ly93d3cud2lzZXN0YW1wLmNvbS9lbWFpbC1pbnN0YWxsP3dzX25jaWQ9NjcyMjk0MDA4JnV0bV9zb3VyY2U9ZXh0ZW5zaW9uJnV0bV9tZWRpdW09ZW1haWwmdXRtX2NhbXBhaWduPXByb21vXzU3MzI1Njg1NDg3Njk3OTIiLCAiZSI6ICI1NzMyNTY4NTQ4NzY5NzkyIn0=>

RDmitchell:
[...]

PR #327, opened 2015-08-18
Issue 83 uploading duplicate data

StephenCzarnecki:
Should have changes for both uploading duplicate data (Issue #83) and providing a changelog for merging buildings.

NOTE:  This involves a database migration so need to do
python manage.py makemigrations --settings=config.settings.dev
python manage.py migrate --settings=config.settings.dev

Added a 'duplicate' column to BuildingSnapshots table. Updated front end to display number of duplicates and tell user they will not be used. Added and modified unit tests. Added API endpoint to delete duplicate records for a given input file. NOTE: The raw import records are not being deleted in order to retain functionality in the review mapping screen.

Note on how rows are added to BuildingSnapshot table when a record is importing (prior to these changes):
1:  Pre-mapping:  A raw BuildingSnapshot record is created with references to the import file.
2:  Post-mapping:  A processed BuildingSnapshot record is created.
3:  If there is a match:  A new canonical building record that is the merge of the post-mapping and the previous canonical building.

Now with these changes #1 is still always created.  If it is a duplicate then the second is created and given the ID of the record it is a duplicate of.  I added a "delete duplicate records from file" api call and changed the javascript code to call that after the matching to remove the duplicate records.  If it is a duplicate there is no changes to the canonical building or new record created.

For the changelog behavior the main changes were to make save_snapshot_match also return a list of any changes to the existing canonical building.  Then the places that call save_snapshot_match (handle_results and handle_id_matches) add the changes to the AuditLog.  In order to get the display to look OK I had to wrap the cell in the table in building_audit_log.html with a pre tag.

### DIFF
--- a/seed/mappings/mapper.py
+++ b/seed/mappings/mapper.py
@@ -86,7 +86,6 @@ def merge_extra_data(b1, b2, default=None):
 
     return extra_data, extra_data_sources
 
-
 def merge_building(
     snapshot, b1, b2, can_attrs, conf, default=None, match_type=None
 ):
@@ -102,6 +101,7 @@ def merge_building(
     """
     default = default or b1
     match_type = match_type or models.SYSTEM_MATCH
+    changes = []
     for attr in can_attrs:
         # Do we have any differences between these fields?
         attr_values = list(set([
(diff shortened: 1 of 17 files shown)


StephenCzarnecki:
Forgot to mention before:  Added a bit to the string in the matching prompt that the user sees after the automated matching.  It says the number of duplicates that were found and informs the user that they are being removed from matching.  On the review matching screen only non-duplicates should appear under either "matched" or "non-matched'.  For example, if a file with 32 records is uploaded but only two are not duplicates there should only be 2 buildings shown in the review matching screen.

Issue #325, opened 2015-08-15
Fields in JSON Blob do not show data in matching screen


RDmitchell:
It appears that in the Matching view, the program only displays data for fields that are in the BuildingSnapshot (BS) table, not in the JSON blob. We need to be displaying data from all the fields, whether explicitly a field in the BS table or data from the JSON Blob extra_data field in the BS table.

Here is an example.

I imported the standard 2 example files, first the covered building data, then the PM data. I mapped the covered building list as follows:
![image](https://cloud.githubusercontent.com/assets/6314950/9286276/e2a3f3d0-42a4-11e5-9cd3-7909622af950.png)

The resulting Building List is shown below
![image](https://cloud.githubusercontent.com/assets/6314950/9286306/97f5a03a-42a5-11e5-92a0-c840a5efac5c.png)

I mapped the PM data as shown below (just showing the relevant mappings -- I let everything else default). "Year Ending" is an explicit field that is in the Building Snapshot table, while "Premises City" is not and will be stored in the JSON blob "extra_data" field in the Building Snapshot table.

![image](https://cloud.githubusercontent.com/assets/6314950/9286333/270b0a58-42a6-11e5-89b6-ef6af895ccdf.png)

When I go to the Matching view, I select the fields to view (from the "View / Hide Columns" button) to be Address Line 1 (explicit field in the BS table), City (an explicit field in the BS table, but only mapped for the covered building list not in the PM file), Premises City (not an explicit field in the BS table, so it is in the JSON extra_data field), and Year Ending (an explicit field in the BS table, only found in the PM file). 

On the left side of the Matching view, which is showing the data from the PM file, the City field is blank because City wasn't mapped (so that is correct even though from a UI standpoint it's a bad design), and the Premises City field is also blank. But Premises City is how the City field was mapped, and does have data, which is not being displayed. I assert that it is not being displayed because it is in the JSON blob field. 

![image](https://cloud.githubusercontent.com/assets/6314950/9286495/c35323ca-42a9-11e5-83cd-f19be6e33b96.png)

nicholasserra:
On initially uploading covered-buildings-sample and hitting the mapping screen, I can't get the system to give me a green check box in the "Premises City" field, as shown in your screenshot. I'm testing against develop.

RDmitchell:
The green checkbox has to do with whether it is in the BEDES mapping file or not. 

It is slightly odd that you are not seeing it, but I am not sure you need to worry about it.

[...]

PR #328, opened 2015-08-18
Issue #325 - Handle extra data fields on matching table

nicholasserra:
Extra data fields were not handled correctly in the DOM for the matching view table. I added checks to handle both string and numeric cases of the JSON data. 

Energy star & organization extra data fields shown:
screen shot 2015-08-18 at 5 12 03 pm

Zip extra data field shown:
screen shot 2015-08-18 at 5 12 12 pm

Refs #325


RDmitchell:
Great ! Thanks Nick

PR #333, opened 2015-08-20
Better coverage reporting settings

nicholasserra:
Made some changes to our coverage config to get a more accurate report. Basically don't check test files, migrations, or debug code. Looks like we're close to 70% coverage.


nllong:
:+1:

PR #335, opened 2015-08-20
Issue #325 - Handle right side matching extra_data fields.

nicholasserra:
Handle extra_data JSON fields on right side of matching view table.

Refs #325


RDmitchell:
Thanks Nick! I will test once Miles merges it in and puts it on seedtest

PR #356, opened 2015-09-01
Adding directive to resize table columns, feature to sort table columns

juliecov:


mmclark:
Julie - there are merge conflicts against develop.  Can you take a look?

juliecov:
Hi, Looks like the problem was caused when I added a javascript include while another (bootstrap) was updated.  I've pulled in the bootstrap version change, let me know if there is still a problem merging.

RDmitchell:
Thanks Julie !

Issue #90, opened 2015-01-14
Get rid of whoopra


RDmitchell:
The title says it all !

RDmitchell:
Nick -- take a look at this and see what you think. Rich thought it might break some dependencies. But we do need to get rid of it if it's not too big of a deal. thx Robin

nicholasserra:
Not sure what this is, did you mean `Woopra`? And if so, it looks like that also isn't really wired in, I only see a small reference to it in a settings file.

rebrownlbnl:
Yes, that's it.
The FE makes calls to a service for user tracking (look at your browser
console).
We aren't using it and it slows things down.

nicholasserra:
I don't see any reference to it on the FE. Are we sure it's still in there? Is there some other package alias it goes by?

RDmitchell:
This is probably a question for Miles. I will ping him about it.

mmclark:
Woopra is used by the django-analytical package.  There's a setting related to it in config/settings/main.py

nicholasserra:
Ah, gotcha. Should I just pull out the woopra stuff and leave django-analytics no-op, or want me to pull out django-analytics entirely? @RDmitchell

rebrownlbnl:
Let's leave the django-analytics in for now, but just remove the woopra.
(assuming there's not much overhead to having django-analytics in there but
not connected to any cloud-based tracking service).

nicholasserra:
PR merged, assigning back for testing.

RDmitchell:
Seems to be gone on seedtest .. THANK YOU !!!

PR #361, opened 2015-09-03
Issue #90 - Remove woopra

nicholasserra:
Remove the config for the Woopra domain. This stops django-analytical from putting the tracking code in the DOM. I don't see any dependency issues, as local dev never includes it anyway.

Refs #90


RDmitchell:
This can be merged into master and onto SEED with the September release.

Issue #378, opened 2015-09-14
Matching address with abbrev street type and direction


cycustodio:
This was discovered in issue (https://github.com/SEED-platform/seed/issues/315). The case is for ```Peach Ave. East``` and ```Peach Avenue E```, both should be normalized to ```peach avenue east``` so they match. 

![address_unmatched](https://cloud.githubusercontent.com/assets/8461331/9857811/b87dee7c-5ad1-11e5-979c-82f6e276a011.png)

nicholasserra:
Reassigning to Piper so he can get acquainted with the BE

pipermerriam:
This ticket says that the following should both map to `peach avenue east`

- `Peach Ave. East`
- `Peach Avenue E.`

The current address normalization behavior behaves as follows.

- Things like `Street, Avenue, Boulevard` are mapped to their shorthand versions, `st, ave, blvd`.
- Compass directions (`North, South, East West, North`) are mapped to their single letter shorthand, `N, S, E, W`.

Assuming we want to fix the issue in this ticket so that the two addresses above are normalized to the same value, as well as preserving the current behavior of the mapping, then I would like to suggest changing the acceptance criteria of this issue to the following.

The above addresses should both be mapped to `peach ave e` instead of `peach avenue east`.

@RDmitchell  Is this acceptable?

nicholasserra:
That should be correct ^

RDmitchell:
Yes that seems fine

cycustodio:
Thanks for catching that Piper. Both should match, ```peach ave e```.

pipermerriam:
The fix for this is ready for review in #380

RDmitchell:
Claudine -- can you please test this -- the PR for it was merged yesterday. Thx

cycustodio:
It's fixed. Thanks, @pipermerriam.

PR #380, opened 2015-09-14
Piper/issue 378 normalize street type

pipermerriam:
ref #378 

based on pr #379 (will rebase this PR if that pr gets merged first)

### What was wrong

The two addresses `Peach Ave. East` and `Peach Avenue E.` were not being normalized to the same values, and thus were being detected as different addresses when they should have been handled as the same address.

### How was it fixed.

New normalization function for the `StreetNamePostType` value of the deconstructed address so that the longhand street type `Avenue` is normalized to the desired shorthand `ave`.

#### Cute animal picture

baby-goats-09

### DIFF
--- a/seed/tasks.py
+++ b/seed/tasks.py
@@ -1039,6 +1039,17 @@ def _normalize_address_direction(direction):
         return direction_map[direction]
     return direction
 
+
+POST_TYPE_MAP = {
+    'avenue': 'ave',
+}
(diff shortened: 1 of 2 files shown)


nicholasserra:
Looks like this needs rebased/develop merged in now.

pipermerriam:
Rebased.

Issue #386, opened 2015-09-18
Matching and unmatching records orphans some records


RDmitchell:
I will add more content to this soon

RDmitchell:
Figuring out this issue is going to require some work to discover how matching works. During this discovery process, please document how the program is currently doing matching, and not just in the code. 

There are lots of issues with the matching screen but this issue is concerned with one particular aspect, which is that when you unmatch records, which should add one building to the canonical building list for every record that is unmatched. At least in the example I am giving. 

I have made a video which shows the following:
* load a file with 5 records, which creates 5 canonical buildings
* in the matching screen, none of the buildings will be matched to each other
* Match all the records together "by hand" -- see the video
* The Building list will now show 1 building
* The detailed view of that building will show the Master record with the 5 other records associated with it
* Go back to the Matching screen and unmatch the records. 
* You will end up with less than 5 records in the Building List, even though you should end up with the 5 original buildings that you started with -- the first time I did this I had 4 buildings, then I deleted the buildings (and the dataset), repeated the process, and ended up with only 3 records (this is what is shown in the video). 

here is a link to the directory with a video explaining the process to see the bug as well as the example file in the video

https://drive.google.com/a/lbl.gov/folderview?id=0B3fTKpZ9Dx7LbW9td24tc1F3dG8&usp=sharing

bravacheLBL:
I reproduced Robin's test: match 5 buildings, then unmatch them, and got back the 5 original building. In my opinion, this problem is solved, but let me know if you see a better way to test this issue further.

Thanks

RDmitchell:
This is working now with this simple dataset. I will test with a more complex data set and start another issue if any more problems are found. Thanks Piper

PR #389, opened 2015-09-23
Account for null values in the normalized connonical building address

pipermerriam:
I encountered this bug while doing investigatory work on #386 

### What was wrong.

When trying to execute the building matching phase of data import, I was getting an error because `lower()` was being called on `None` in the `_find_matches` function.

```
[2015-09-23 12:20:33,918: ERROR/MainProcess] Task seed.tasks._match_buildings[626978cc-7f2d-4c4d-941e-9bc5e854f937] raised unexpected: AttributeError("'NoneType' object has no attribute 'lower'",)
  File "/Users/piper/python-environments/seed/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/Users/piper/python-environments/seed/lib/python2.7/site-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/Users/piper/sites/seed/seed/decorators.py", line 50, in _wrapped
    response = fn(import_file_pk, *args, **kwargs)
  File "/Users/piper/sites/seed/seed/tasks.py", line 1208, in _match_buildings
    results = _find_matches(un_m_address, canonical_buildings_addresses)
  File "/Users/piper/sites/seed/seed/tasks.py", line 1101, in _find_matches
    if un_m_address.lower() == cb.lower():  # this second lower may be obsolete now
AttributeError: 'NoneType' object has no attribute 'lower'
```

When the [`_match_buildings`](https://github.com/SEED-platform/seed/blob/2070456678b07bbe73f99f94a414928a021c5a23/seed/tasks.py#L1108) task is run it creates a list of address pieces called [`canonical_buildings_addresses`](https://github.com/SEED-platform/seed/blob/2070456678b07bbe73f99f94a414928a021c5a23/seed/tasks.py#L1193) via the [`_normalize_address_str`](https://github.com/SEED-platform/seed/blob/2070456678b07bbe73f99f94a414928a021c5a23/seed/tasks.py#L1053) function.  Later, this list is passed into the [`_find_matches`](https://github.com/SEED-platform/seed/blob/2070456678b07bbe73f99f94a414928a021c5a23/seed/tasks.py#L1208) function, which assumes all of the list values are strings, and calls `lower` on them.

This blows up, because the `_normalize_address_str` function returns `None` on [line 1065](https://github.com/SEED-platform/seed/blob/2070456678b07bbe73f99f94a414928a021c5a23/seed/tasks.py#L1065) if the address value is falsy.

### How was it fixed.

I adjusted the `_find_matches` function to check whether the value is `None` and to just continue on with the next loop iteration when this is the case.

#### Cute animal picture.

goat-on-a-horse

### DIFF
--- a/seed/tasks.py
+++ b/seed/tasks.py
@@ -1098,6 +1098,8 @@ def _find_matches(un_m_address, canonical_buildings_addresses):
     if not un_m_address:
         return match_list
     for cb in canonical_buildings_addresses:
+        if cb is None:
+            continue
         if un_m_address.lower() == cb.lower():  # this second lower may be obsolete now
             match_list.append((un_m_address, 1))
     return match_list


pipermerriam:
An alternative fix for this would be to filter the `canonical_buildings_addresses` list and remove any `None` values from it, but this seemed more appropriate.

RDmitchell:
I think that checking for blanks address field data and then just continuing on is a good approach.

cycustodio:
Tested and ready for production.

PR #394, opened 2015-09-24
Piper/issue 386 fix bug when unmerging snapshots

pipermerriam:
PR for issue #386 

Is based off of PR #389 though the merge order should not matter.

### What was wrong?

When unmerging buildings, it was possible for the building to get left in an error state.

In the case where the building which is being removed from a set of other builds happens to be the one that has the active canonical building (within the tree of building snapshots), then when that building snapshot is unmerged from the tree, it's canonical building will end up with no canonical snapshot.  When this happens, it stops being returned in queries for buildings, and thus disappears from the frontend.

### How was it fixed.

The error state occurs when the orphan'd children get deleted, and one of those children happens to be a canonical snapshot.  When this happens, now it goes and finds a *sibling* of the child and sets the tip of that sibling's tree as the canonical snapshot prior to deletion of the orphaned child.

#### Cute animal picture.

518889810_c_294_220

### DIFF
--- a/seed/models.py
+++ b/seed/models.py
@@ -1274,7 +1274,7 @@ class BuildingSnapshot(TimeStampedModel):
     use_description_source = models.ForeignKey(
         'BuildingSnapshot', related_name='+', null=True, blank=True
     )
-    
+
     #Need a field to indicate that a record is a duplicate of another.  Mainly used for cleaning up.
     duplicate = models.ForeignKey(
         'BuildingSnapshot', related_name='+', null=True, blank=True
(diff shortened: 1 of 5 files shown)


pipermerriam:
https://travis-ci.org/SEED-platform/seed/builds/82063389

This test failure is also presenting itself in the `develop` branch.  Is fixing this being worked on by someone?  If not, I can work on it.  Please advise.

RDmitchell:
This will be fixed with this PR<https://github.com/SEED-platform/seed/pull/398>

PR #400, opened 2015-09-28
Issue178

danielmcquillen:
This is the first implementation of the reporting page, which shows two configurable graphs in the Building section, under the new "Reports" tab. Three new methods were created on the backend (in  views.py) to provide data to the charts on the client. 

Gathering the required data from the seed_buildingsnapshot table is somewhat involved as the charts need historical dat. We therefore can't rely just on the data captured in the last good canonical building, but instead must traverse the parents of the canonical building to gather historical data.

I'm working with Stephen and Nick Serra to document how those methods work and will update the wiki when finished.


RDmitchell:
reports look good.

PR #416, opened 2015-10-02
Fix for deleted buildings still showing up in reports.  When a user d…

StephenCzarnecki:
When a user deletes a building all that really happens is that the CanonicalBuilding record for that building has the 'active' property set to false.  Changed the reporting code to filter out those canonical buildings.


RDmitchell:
tested, looks good

PR #469, opened 2015-11-05
Merging branch Selenium tests into SEED develop

rHorsey:
Adding Selenium to the SEED environment.
Adding initial end-to-end testing through test_firefox.py selenium test.
Adding template in test_api.py for api 'unit-ish' tests.
Changed seed/tasks.py file to standardize on a single cache format.
Changed seed/tasks.py file to allow tasks to work either synchronously or asynchronously.
Changed travis build config.
Changed test and untracked_local settings files.
Added id tags to a number of partials to allow for selenium testing.


seed/tests/api/data/covered-buildings-sample.csv
```diff
@@ -1,513 +0,0 @@
-UBI,GBA,BLDGS,Address,Owner,City,State,Zip,Property Type,AYB_YearBuilt
```
nllong: @rHorsey: was this file removed on purpose?

seed/views/main.py
```diff
@@ -2017,7 +2020,7 @@ def progress(request):
     progress_key = json.loads(request.body).get('progress_key')
 
     if cache.get(progress_key):
-        result = cache.get(progress_key)
+        result = get_cache(progress_key)
```
nllong: Is this correct? Aren't sin't `cache.get` the new method for getting the cache data?

seed/tests/api/data/covered-buildings-sample.csv
rHorsey: Yes. We already have the file in seed/tests/data, and that's where all the tests were pointing to.

seed/views/main.py
rHorsey: Interesting. In tasks.py all the commands used on caches are get_cache and set cache. It should be cache.get and cache.set when we get all the progress stuff classified.

rHorsey: Alex confirmed that it should be get_cache.

Issue #459, opened 2015-10-28
Data Cleansing Report not working


RDmitchell:
Assigned to @nllong 

When I click on the Data Cleansing Report button in mapping, I get the error messages below. In the UI, nothing happens (ie, no error, and no display of the data cleansing dialog box).

This is on seedtest. I was using the Data Cleansing file that Claudine made, but I think it will happen with any file. Here is a link to the file
https://drive.google.com/open?id=0B3fTKpZ9Dx7LNnA4Y1hfcmVqZkE

Below is the error message set:

12angular.js:12477 TypeError: s.is_initial_state is not a function
    at fn0 (eval at <anonymous> (https://seedtest.s3.amazonaws.com/vendors/bower_components/angular/angular.js:13322:15), <anonymous>:4:483)
    at Object.expressionInputWatch [as get] (https://seedtest.s3.amazonaws.com/vendors/bower_components/angular/angular.js:14302:31)
    at Scope.$digest (https://seedtest.s3.amazonaws.com/vendors/bower_components/angular/angular.js:15818:40)
    at Scope.$apply (https://seedtest.s3.amazonaws.com/vendors/bower_components/angular/angular.js:16097:24)
    at HTMLButtonElement.<anonymous> (https://seedtest.s3.amazonaws.com/vendors/bower_components/angular/angular.js:23554:23)
    at HTMLButtonElement.p.event.dispatch (https://ajax.googleapis.com/ajax/libs/jquery/1.8.0/jquery.min.js:2:37944)
    at HTMLButtonElement.g.handle.h (https://ajax.googleapis.com/ajax/libs/jquery/1.8.0/jquery.min.js:2:33881)(anonymous function) @ angular.js:12477(anonymous function) @ angular.js:9246Scope.$digest @ angular.js:15844Scope.$apply @ angular.js:16097(anonymous function) @ angular.js:23554p.event.dispatch @ jquery.min.js:2g.handle.h @ jquery.min.js:2
[...]

PR #472, opened 2015-11-07
fix response with list object

nllong:
This fixes cleansing. It broke after we tried to standardize the return codes. Please review and make sure my change is appropriate.

### DIFF
--- a/seed/decorators.py
+++ b/seed/decorators.py
@@ -99,8 +99,12 @@ def ajax_request(func):
         response = func(request, *args, **kwargs)
 
         status_code = 200
-        if response.get('status') == 'error' or response.get('success') == False:
-            status_code = 400
+
+        # This is somewhat hackish, but the response is not always an HttpResponse. I'm explicitly
+        # excluding the list because it is a known type that is returned in the cleansing routines.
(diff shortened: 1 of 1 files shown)


rebrownlbnl:
Is it possible to add a test to detect the original problem?

seed/decorators.py
```diff
@@ -99,8 +99,12 @@ def wrapper(request, *args, **kwargs):
         response = func(request, *args, **kwargs)
 
         status_code = 200
-        if response.get('status') == 'error' or response.get('success') == False:
-            status_code = 400
+
+        # This is somewhat hackish, but the response is not always an HttpResponse. I'm explicitly
+        # excluding the list because it is a known type that is returned in the cleansing routines.
+        if not isinstance(response, list):
```
mmclark: Would it make more sense to test if response is an HttpResponse before calling get(), as opposed to testing if it's not a list?

nllong:
This will fix #459

seed/decorators.py
nllong: @mmclark -- i'm going to fix this today and add a test.

PR #476, opened 2015-11-11
Removed Less, added Sass and Gulp

axelstudios:
### Changes
This PR removes Less (Apache license), removes client-side compilation, and instead uses Sass for server-side compilation.  It also adds Gulp for workflow automation, with the new Gulp tasks described below.

Since the stylesheet must now be compiled before loading the page there is a Gulp task to watch for filesystem changes and automatically recompile on-the-fly.

### Developer instructions
Install the new developer dependencies: `./bin/install_javascript_dependencies.sh`

#### `gulp`
Default task – acts as an alias for `gulp watch`
#### `gulp jshint`
Reports on JavaScript code-correctness with errors & warnings
#### `gulp jshint:fix`
Automatically fixes the previously found issues
#### `gulp sass`
Compiles `style.scss` into `style.css`, adds browser prefixes for compatibility, compresses/optimizes the output, and creates a sourcemap for debugging purposes `style.css.map`
#### `gulp watch`
Runs `gulp sass` first, and then monitors for updates to `style.scss` to recompile when necessary


mmclark:
It's likely something particular about my environment, but I needed to install gulp-util manually before gulp would run succesffully:

```
$ npm install --save-dev gulp-util
```

rebrownlbnl:
Just coming back to this closed PR.
Do we need to update the developer and installation docs to reflect this change?

mmclark:
Yes, if the developer or installation docs don't discuss the use of gulp, then they need to be updated.

mmclark:
Alex - a question regarding the css and map file output by the sass compile.  Since style.css and style.css.map are generated by sass, it seems they should not be managed by git.  Would you agree?

axelstudios:
Typically you don't want compiled files committed.  However, compiled stylesheets have generally become an exception to that rule.  It makes it convenient for developers who don't commonly work with the stylesheets not to have to compile them before making changes.  It also helps to have it available if the repo is cloned on a sytem without installing developer dependencies, like the CI or a production environment -- that way, a command like `git pull` or a branch checkout would be enough to have the right compiled stylesheet.

mmclark:
Makes sense - thanks.

Issue #317, opened 2015-08-05
Address Matching -- deal with address ranges


RDmitchell:
The address parsing module has some issues with address ranges. If there is no whitespace in an address range it parses the 'house' appropriately (though not necessarily in a useful way - there is no indication of range start and range end). If there is whitespace then the parser is less predictable and the output is not so well formed.

Address ranges and matching datasets on their permutations have been a real struggle for me so this might add unwanted complexity to your normalization task but getting it right could improve matching.

example test addresses

322 S Green St #house:322
300-22 S Green St #house:300-22
300/322 S Green St #house:300/322
300 322 S Green St #house:300,street_name:'322 S green'

RDmitchell:
@nicholasserra -- can you see if this is still an issue, and if it is, propose a way to deal with the address ranges without whitespaces. I created the issue above, but I believe it was emailed to me by another developer. 

If it is useful, there is a set of data associated with this issue that has several records with address ranges. In general they are well formed, although one does not have white space between the address ranges. The addresses in the two files starting with "matched" all have a one-to-one matching on address, even though they have street number ranges, probably because they are formatted identically. These files could be used to make a test set where the addresses should match but are not so well aligned. 

If it is possible to delineate, without too much effort, as requested above, the range start and range end in the parser, that might be useful.

RDmitchell:
@nicholasserra -- can you investigate if this is still an issue, ie, dealing with white space in address ranges. I made a set of files that have address ranges, and do match 1:1, and the program matches them without issue on just address. They are formatted exactly the same, so maybe that is not a good test. You could potentially start with those files (in the link below, starting with "matched") and make them slightly more difficult to match, ie, one with white space and the other not. 

I believe this issue was emailed to me by another developer (even though I authored the issue, I didn't generate the content).

RDmitchell:
Link to sample files
https://drive.google.com/open?id=0B3fTKpZ9Dx7LUlQyUG9YZkQ2aDQ

The two files with the prefix "matched" have 1:1 matching on address line 1

nicholasserra:
Those are some nasty addresses. Anything is possible, just need to write regex to handle odd cases. Can you give me what you think the normalization output should be for those example addresses?

RDmitchell:
[...]

PR #483, opened 2015-11-18
Add normalization rules for address ranges

pipermerriam:
Fixes https://github.com/SEED-platform/seed/issues/317

### What was wrong.

Addresses that are declared as ranges were not being normalized or parsed correctly.

All of the following should be normalized to the same canonical form `310-310 green st`

* `300-310 green st`
* `300 - 310 green st`
* `300/310 green st`
* `300 310 green st`
* `300-10` green st`  (omits the leading digit on the second number)

### How was it fixed.

The numeric portion of the address is now passed through a normalization function `seed.tasks._normalize_address_number` which uses a regex to detect numeric ranges and normalize them.

#### Cute animal picture

2828e45400000578-0-image-a-74_1430678167323

### DIFF
--- a/seed/tasks.py
+++ b/seed/tasks.py
@@ -1129,6 +1129,42 @@ def _normalize_address_post_type(post_type):
     return POST_TYPE_MAP.get(value, value)
 
 
+ADDRESS_NUMBER_RE = re.compile((
+    r''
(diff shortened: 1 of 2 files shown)


nicholasserra:
NICE

PR #487, opened 2015-11-24
Split delineated tax id values into multiple buildings

### LINKED ISSUE #480: Handle multiple data elements in matching field
See the specification
https://docs.google.com/a/lbl.gov/document/d/1ecDCeNg7Yb3hPhPdVfuN5R9WxC2ZD7LCfwnmTUSIR3s/edit?usp=sharing

pipermerriam:
Fixes #480 

### What was wrong?

Some buildings in places like New York actually encompass more than one Tax Lot ID.  Because of this, data is being submitted to seed with multiple tax lot id's in the same field, with some separator like a semi-colon.

This caused data integrity problems since if another update came into the system with only a subset of the tax-lot-id's previously submitted it would not be matched as the same building.

### How was it fixed.

During the data import, after the fields have been mapped, the tax lot-id field is split into multiple id's based on either a semi-colon or comma separator, and then the values are normalized by removing all non alpha-numeric characters and upper-casing the alpha characters.  These normalization rules were derived from the spec document linked below.

https://docs.google.com/document/d/1ecDCeNg7Yb3hPhPdVfuN5R9WxC2ZD7LCfwnmTUSIR3s/edit

### How was it tested.

The tax-lot-id normalization has a thorough test suite that ensures that it complies with the specification outlined in the google document linked above.

In addition, manual testing was done against the provided sample data sets.  For example, this document (https://docs.google.com/spreadsheets/d/1gkVTLqfNgKRzqWXy2afHHcg0Vg26UmWMa4Hgz1KcJ50/edit#gid=254055333&vpid=A1) which contains 15 rows, but 25 unique tax-lot-ids parses correctly to 25 distinct buildings.


pipermerriam:
@RDmitchell This went quicker than expected.  Hurray for unexpected simplicity!

I was able to implement this in a pretty clean and simple way, but I'd sleep better at night if someone can verify that the results I'm seeing are inline with the desired results.

RDmitchell:
Piper -- thanks for getting this going so quickly. We will do testing on this with some real data from cities once Miles gets it merged into seedtest.

rebrownlbnl:
Can you add a test for a "compound" date type, with multiple types of separators and delimiters in the same string, like:
"1234-5678;9876 5432, 1048 5938"

I know that's pretty irregular, but these are hand-entered data so we can get all kinds of strange things.

rebrownlbnl:
sorry, that should be "compound data type"

pipermerriam:
Done in ba3649684f993916778b086b37f67edcde162e62

rebrownlbnl:
:+1: 
And a cute animal picture back to you:
img_2586

PR #496, opened 2015-11-30
Fix building of the documentation

pipermerriam:
### What was wrong

The documentation configuration hadn't been updated since the AppRegistry was introduced in django 1.7.  This was causing some automated imports to trigger the registry not ready exception.

### How was it fixed.

Manually imported and called `django.setup`

#### Cute animal picture

e8e62ec49afa5c1893096e80cdfb9ad0


RDmitchell:
Thanks Piper !

Issue #477, opened 2015-11-12
fix readthedocs 


nllong:

nllong:
edwin lee is looking into this

Myoldmopar:
@RDmitchell,  as per the [comment](https://github.com/SEED-platform/seed/issues/477#issuecomment-158129867) above, my esteemed coworker @nllong has asked me to work on generally improving the documentation and fixing the RTD support for this repo.  I had a few questions and thoughts and wondered if I should air them out here or if you'd prefer to chat via any other method.  Thanks!

RDmitchell:
You can send me an email to start the conversation, RDMitchell@lbl.gov

RDmitchell:
I believe this is now working, so I am closing this issue. We can reopen a new issue if needed for any followup.

Myoldmopar:
Thanks @RDmitchell, I opened #508 to wrap up a few polishing tasks.

RDmitchell:
perfect, thanks

nllong:
I'm reopening this and adding in a few tasks

[] enable spell check
[] remove BE attributions
[] update excludes

nllong:
This should be closed again... new docs are here: http://seed-platform.readthedocs.org/en/latest/

PR #499, opened 2015-12-01
Fix Automated ReadTheDocs Builds

Myoldmopar:
Automated builds with ReadTheDocs were closer with the merge of #496, but it was failing due to the DOMAIN_URLCONFS not being defined in the Settings dictionary.  This wasn't defined because on ReadTheDocs, there won't be a local_untracked.py file, and the dev.py file did not include an initialization of this variable.  For development/deployment, when local_untracked.py _is actually present_, the value will be overwritten with whatever the dev has defined in local_untracked.py.

To complete the RTD automated builds, the following steps are needed if they aren't already present:
 - [ ] This PR will need to be merged, or some other solution found for the uninitialized variable issue
 - [ ] In the Github repo settings: Webhooks -> Add Webhook -> ReadTheDocs -> Yes/Accept
 - [ ] In the ReadTheDocs project settings: Check the "Install your project inside a virtualenv" box, and add `requirements/local.txt` in the "Requirements file:" box.
 - [ ] Since develop is the default branch in Git, you shouldn't need to update the "Default branch:" entry unless you have some preference to use a different branch.

I believe that is it.  At that point, each commit to https://github.com/SEED-platform/seed will trigger a new build of the docs.  

Thanks

FYI @nllong

### DIFF
--- a/config/settings/dev.py
+++ b/config/settings/dev.py
@@ -116,6 +116,11 @@ if "COMPRESS_ENABLED" not in locals() or not COMPRESS_ENABLED:
 
 ALLOWED_HOSTS = ['*']
 
+# ReadTheDocs won't have a local_untracked, so even here in dev, there needs to be a few initializations
+DOMAIN_URLCONFS = {}
+DOMAIN_URLCONFS['default'] = 'config.urls'
+
+
 # use imp module to find the local_untracked file rather than a hard-coded path
 # TODO: There seems to be a bunch of loading of other files in these settings. First this loads the common, then this, then anything in the untracked file
 try:


Myoldmopar:
Fixes #477

Myoldmopar:
- [ ] There is also the addition of the badge from ReadTheDocs on the repo README.

rebrownlbnl:
I added ReadTheDocs as a "service integration" for the github repo. Will that work, or does it need to be a webhook? I just noticed this URL on RTD for what they call a "Post Commit Hook": curl -X POST https://readthedocs.org/build/seed-platform, which seems like what the target should be for the github webhook. Please advise if I should change it. (reading further on RTD, they say to use the GitHub service integration, so I think I've set it up correctly).

In the RTD settings, it already had the setting for virtualenv set and the requirements path.

Actually, I just tested it from github and the build failed with this message:
>Exception occurred:
> File "/home/docs/checkouts/readthedocs.org/user_builds/seedplatform/envs/stable/local/lib/python2.7/site-packages/django/apps/registry.py", line 124, in check_apps_ready
>raise AppRegistryNotReady("Apps aren't loaded yet.")
>AppRegistryNotReady: Apps aren't loaded yet.

One final thought: I'm thinking that RTD should only be updated when we update the Master branch, since those are our public releases. Does that sound right?

mmclark:
I updated the requirements setting on rtd to requirements/local.txt and manually set the branch to develop and the build passed.

Since we're actively working on RTD integration for this release, I'd suggest we keep RTD pinned to the develop branch for now.  When we make the December release, we should move it to master and leave it there so RTD mirrors the public release of SEED.  The only downside is that since develop is the default branch in the github repo, if anyone is comparing RTD with the source, it may be out of sync.

rebrownlbnl:
I think the version numbering will help people understand the synching
between github and RTD.

Rich Brown
Sent from my mobile phone

On Dec 3, 2015, at 7:43 AM, Miles Clark <notifications@github.com> wrote:

I updated the requirements setting on rtd to requirements/local.txt and
manually set the branch to develop and the build passed.

Since we're actively working on RTD integration for this release, I'd
suggest we keep RTD pinned to the develop branch for now. When we make the
December release, we should move it to master and leave it there so RTD
mirrors the public release of SEED. The only downside is that since develop
is the default branch in the github repo, if anyone is comparing RTD with
the source, it may be out of sync.

—
Reply to this email directly or view it on GitHub
<https://github.com/SEED-platform/seed/pull/499#issuecomment-161681379>.

PR #502, opened 2015-12-01
Piper/issue 434 labels

pipermerriam:
### What does this PR do?

It implements general scaffolding needed to use DRF as well as adjusting the current frontend implementation to use DRF for label administration.

* 63cd05d updates the `StatusLabel` model to include a `'gray'` option as to line up with the choices implemented on the frontend.  The migration in this commit won't modify the database in any way.
* bdd1a77 adds [underscore.js](http://underscorejs.org/) as a javascript dependency.  This contains some quite useful toolbox style functionality.  The subsequent 48f1b1d commit contains a few areas where it was used in the code.
* 4356e4d adds [Django Rest Framework](http://www.django-rest-framework.org/) as a dependency to the project.
* 48f1b1d is the initial work on converting the API endpoints to use DRF.  This contains changes to the frontend code sufficient to make the label administration function correctly.

This PR also contains the following two unrelated cleanup commits that could be extracted and PR'd directly into develop if including them into this branch isn't desirable.

* 98953d8 fixes some url confs that were missing a leading `^` in their regex.
* 47239aa contains whitespace error fixes that my editor automatically does for me.

### What this does not do.

This doesn't implement the searching/filtering logic, or the bulk labeling logic.  Those will be done in subsequent PR's

#### Cute animal picture

corgi-with-leaf


pipermerriam:
Worth noting that this is almost guaranteed to cause some small breakage on the building list page since the format of the response data has changed.  If you're ok with it, I'll address this first thing after this is merged.  Otherwise I can work on getting it into this PR.

PR #509, opened 2015-12-03
Add badges for stable and latest

Myoldmopar:
Here is the addition of the badges (in action [here](https://github.com/Myoldmopar/seed/tree/AddRTDBadges#standard-energy-efficiency-data-seed-platform)).  The stable badge is showing red because:
 - The master branch hasn't been updated with the fixes in develop, and that will happen at release time
 - ReadTheDocs may need some update to point to the master branch for stable


rebrownlbnl:
:+1:

PR #512, opened 2015-12-03
Formatted the time of upload in the Data Cleansing dialog

axelstudios:
This also includes the timezone offset:

image


RDmitchell:
so this is the offset from GMT, based on some information from the server, browser ?? so your display of -7 means it's Mountain Time, right?

Issue #513, opened 2015-12-04
Leading zeros on Matching on IDs - 


RDmitchell:
The program is now normalizing IDs in terms of ignoring (deleting) punctuation (which is great). But here is a case where the program doesn't match two IDs that it should, presumably because of the leading zero in the first case

In this case, the program normalized the Tax Lot ID field from the PM data (which had hyphens in it) but then should have matched these two records (each one from a different data upload), but it didn't, presumably because of the leading zero.
![image](https://cloud.githubusercontent.com/assets/6314950/11578646/39108632-99dc-11e5-82dc-a052612fab43.png)

I think we should ignore (delete?) leading zeros when matching. I will check with users to verify that this is the correct behavior. It could be in the long run that we let the users specify whether to ignore leading zeros but I think it is safe to say that they can be ignored. 

There are two files which can be used to test this. They are found in
https://drive.google.com/a/lbl.gov/folderview?id=0B3fTKpZ9Dx7LSjRzRGxMTnlmaUk&usp=sharing
and are called
* Building data_TLI_3440962.xlsx
* PM Data_TLI_3440962.xlsx

Here is the mapping for the Building data file -- matching field only for Tax Lot ID (not Address Line 1)
![image](https://cloud.githubusercontent.com/assets/6314950/11578712/c0c16f74-99dc-11e5-84a7-85c54a5fbc20.png)

Same mapping for the mapping for the PM data file -- matching only for Tax Lot ID (not Address Line 1)

@pipermerriam -- I am assigning this to you because you worked on the ID normalizing issue. Let me know your thoughts about this leading zero case. I am setting it to priority level = 2 behind the label feature, which is priority = 1. I consider this part of the 2 week round of testing and debugging (Dec 11-18) for our December release

RDmitchell:
Thanks @pipermerriam -- I am assigning this to myself for testing

RDmitchell:
I am closing this as we reverted this code

PR #519, opened 2015-12-07
Strip leading zeros off of tax ids

pipermerriam:
https://github.com/SEED-platform/seed/issues/513

### What was wrong.

Since tax id's are treated like strings, but sometimes represented numerically with leading zeros, this lead to equivalent tax ids formatted with and without leading zeros to not be matched.

### How was it fixed.

Added an `lstrip('0')` to remove leading zeros from the tax-id string during normalization.

#### Cute animal picture.

enhanced-buzz-14897-1354816372-0


mmclark:
Piper - can you look into why the Travis build is failing?

pipermerriam:
Lets see how it does now.

mmclark:
Looks good - thanks

Issue #273, opened 2015-06-18
null Addresses in the database break the matching function


cycustodio:
Origins of the issue: 
(https://github.com/SEED-platform/seed-lbnl/issues/54)

Files that have blank address_line_1, a matching field, break the current matching addresses function. The function takes each new address and compares the address to the items in the database. For consistency the addresses are converted into lower case. When the function encounters a null address from the database, the lower case conversion throws and error because it expects a string.

To recreate the issue use the sample files in [google drive](https://drive.google.com/a/lbl.gov/folderview?id=0B5fKMBregb8HUThOY0R0M3dxd2c&usp=sharing). Upload the covered-buildings-sample.csv first, since it has blank address lines (13 out of 99); map "Address" to "Address Line 1".
Next, upload the portfolio-manager-sample.csv (99 buildings); map "Address 1" to Address Line 1". In the GUI the progress bar for matching should be stuck at 0%. 

The celery log shows the error.
```
 R = retval = fun(*args, **kwargs)
  File "/home/vagrant/.virtualenvs/seed/local/lib/python2.7/site-packages/celery/task/trace.py", line 420, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/vagrant/seed/seed/decorators.py", line 60, in _wrapped
    response = fn(import_file_pk, *args, **kwargs)
  File "/home/vagrant/seed/seed/tasks.py", line 1049, in _match_buildings
    results =_findMatches(un_m_address,canonical_buildings_addresses)
  File "/home/vagrant/seed/seed/tasks.py", line 958, in _findMatches
    if un_m_address.lower()==cb.lower(): #this second lower may be obsolete now
AttributeError: 'NoneType' object has no attribute 'lower'
```

RDmitchell:
It seems like this only happens on the first file imported? If I have two files (see Montana sample data) if I import the first file that has address data in all records, then import a second file that has some missing address data, with both data sets having the address field set as a matching field, then the program will import and match the 2nd data set to the first.

RDmitchell:
@pipermerriam -- when you have finished your other priority = 1 and 2 tasks, can you take a look at this? Not sure it is still a problem. Thx

RDmitchell:
Piper -- thanks for fixing this -- I have assigned to myself for testing.

RDmitchell:
tested on seed v1.3, am closing

PR #520, opened 2015-12-07
Account for null addresses during matching

pipermerriam:
https://github.com/SEED-platform/seed/issues/273

### What was wrong?

Some of the code within `seed.tasks._match_buildings` didn't account for a possible null value, and was treating it as if it was a string.

### How was it fixed

Simple conditional statement.  The location of this code is pretty deep in a big method.  This is a pretty strait forward fix so I opted to just toss it in without the large overhead that would go into covering this with a test.  Alternatively, this code could be injected into the `_find_matches` function, but I was a little hesitant to change it's behavior.

#### Cute animal picture

cute-corgi-stuffed-squid-400x300


RDmitchell:
Thanks Piper !

Issue #521, opened 2015-12-08
Multiple IDs in Matching Field Fix


RDmitchell:
The Multiple IDs in Matching field feature ( #480 ) isn't working quite the way I was expecting. Below is a link to a doc explaining what I am seeing. 

https://docs.google.com/a/lbl.gov/document/d/1PBgcD6Irm145m8mkJbDU2zcQaa5f9jxo8xhvGqh1qjc/edit?usp=sharing

@pipermerriam -- I am assigning this to you, and depending on what you think once you read the issue, you, @StephenCzarnecki and I might want to have a call to discuss it.

pipermerriam:
Issue received.  Starting to look into it now.  I'll let you know what I find and whether I think I need a call to discuss.

RDmitchell:
sounds good. thanks !

pipermerriam:
Ok, I think I'm seeing where this is happening.  I'm going to take some notes here in the issue so that there in a public place for refererence.

The problem appears to be in the logic of the [`seed.tasks.get_canonical_id_matches`](https://github.com/SEED-platform/seed/blob/f924be35e874787f7fabd577958c4a8323bdb6d4/seed/tasks.py#L1020) function which is called in this call chain

* [seed.tasks.match_buildings`](https://github.com/SEED-platform/seed/blob/f924be35e874787f7fabd577958c4a8323bdb6d4/seed/tasks.py#L1002)
* [seed.tasks._match_buildings`](https://github.com/SEED-platform/seed/blob/f924be35e874787f7fabd577958c4a8323bdb6d4/seed/tasks.py#L1300)
* [seed.tasks.handle_id_matches](https://github.com/SEED-platform/seed/blob/f924be35e874787f7fabd577958c4a8323bdb6d4/seed/tasks.py#L1082)
* [`seed.tasks.get_canonical_id_matches`](https://github.com/SEED-platform/seed/blob/f924be35e874787f7fabd577958c4a8323bdb6d4/seed/tasks.py#L1020)

The `get_canonical_id_matches` function does a query where it returns `BuildingSnapshots` that match *any* of `pm_id, tax_id, custom_id` and even stranger, it will match records who's PM_ID matches the TAX_ID.  It then takes *all* of those matches and merges them into a single record.

[...]

PR #523, opened 2015-12-08
Fix overmatching of buildings

pipermerriam:
A possible fix for #521

DO_NOT_MERGE

### What was wrong?

The function `seed.tasks.get_canonical_id_matches` is responsible for looking up other building snapshots which are a match based on the three fields `pm_property_id`, `tax_lot_id`, and `custom_id_1`.  The current implementation is written such that it will match buildings that have *any* of these fields in common, or even snapshots which have a `pm_property_id` that is equal to another snapshot's `tax_lot_id`.

### How was it fixed?

I'm not actually sure that the logic is *wrong*, but I do know that it is wrong for the desired functionality of #487 since it will always result in buildings that were split based on their tax id being re-merged due to something like a common `pm_property_id` or `custom_id_1`.

Instead of doing a query that matches any of these, the query has been adjusted such that it matches records which have *all* of the fields in common *(or the record being matched has a null value)*

#### Cute animal picture

planking_corgi1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 cache:
   directories: 
-    - node_modules
-    - seed/static/vendors/bower_components
     - $HOME/.pip-cache/
 
 language: python
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-sudo: false
-
 cache:
   directories: 
     - node_modules
--- a/seed/tasks.py
+++ b/seed/tasks.py
@@ -1036,25 +1036,17 @@ def get_canonical_id_matches(org_id, pm_id, tax_id, custom_id):
     params = []
     can_snapshots = get_canonical_snapshots(org_id)
     if pm_id:
-        params.append(Q(pm_property_id=pm_id))
-        params.append(Q(tax_lot_id=pm_id))
-        params.append(Q(custom_id_1=pm_id))
+        params.append(Q(pm_property_id=pm_id) | Q(pm_property_id__isnull=True))
     if tax_id:
-        params.append(Q(pm_property_id=tax_id))
-        params.append(Q(tax_lot_id=tax_id))
-        params.append(Q(custom_id_1=tax_id))
+        params.append(Q(tax_lot_id=tax_id) | Q(tax_lot_id__isnull=True))
     if custom_id:
-        params.append(Q(pm_property_id=custom_id))
-        params.append(Q(tax_lot_id=custom_id))
-        params.append(Q(custom_id_1=custom_id))
+        params.append(Q(custom_id_1=custom_id) | Q(custom_id_1__isnull=True))
 
     if not params:
         # Return an empty QuerySet if we don't have any params.
         return can_snapshots.none()
 
-    canonical_matches = can_snapshots.filter(
-        reduce(operator.or_, params)
-    )
+    canonical_matches = can_snapshots.filter(*params)
 
     return canonical_matches
 
--- a/seed/tests/test_tasks.py
+++ b/seed/tests/test_tasks.py
@@ -275,10 +275,10 @@ class TestTasks(TestCase):
         self.assertEqual(
             mapped_bs.address_line_1, u'1600 Pennsylvania Ave. Someplace Nice'
         )
-        
+
     def test_is_same_snapshot(self):
         """Test to check if two snapshots are duplicates"""
-        
+
         bs_data = {
             'pm_property_id': 1243,
             'tax_lot_id': '435/422',
@@ -289,14 +289,14 @@ class TestTasks(TestCase):
             'city': 'Gotham City',
             'postal_code': 8999,
         }
-        
+
         s1 = util.make_fake_snapshot(
             self.import_file, bs_data, ASSESSED_BS, is_canon=True,
             org=self.fake_org
         )
-        
+
         self.assertTrue(tasks.is_same_snapshot(s1, s1), "Matching a snapshot to itself should return True")
-        
+
         #Making a different snapshot, now Garfield complex rather than Greenfield complex
         bs_data_2 = {
             'pm_property_id': 1243,
@@ -308,15 +308,15 @@ class TestTasks(TestCase):
             'city': 'Gotham City',
             'postal_code': 8999,
         }
-        
+
         s2 = util.make_fake_snapshot(
             self.import_file, bs_data_2, ASSESSED_BS, is_canon=True,
             org=self.fake_org
         )
-        
+
         self.assertFalse(tasks.is_same_snapshot(s1, s2), "Matching a snapshot to a different snapshot should return False")
-        
-        
+
+
 
     def test_match_buildings(self):
         """Good case for testing our matching system."""
@@ -330,7 +330,7 @@ class TestTasks(TestCase):
             'city': 'Gotham City',
             'postal_code': 8999,
         }
-        
+
         #Since the change to not match duplicates there needs to be a second record that isn't exactly the same
         #to run this test.  In this case address_line_2 now has a value of 'A' rather than ''
         bs_data_2 = {
@@ -378,7 +378,7 @@ class TestTasks(TestCase):
             AuditLog.objects.first().action_note,
             'System matched building ID.'
         )
-    
+
 
     def test_match_duplicate_buildings(self):
         """
@@ -394,12 +394,12 @@ class TestTasks(TestCase):
             'city': 'Gotham City',
             'postal_code': "8999",
         }
-        
+
         import_file = ImportFile.objects.create(
             import_record=self.import_record,
             mapping_done=True
         )
-        
+
         # Setup mapped PM snapshot.
         snapshot = util.make_fake_snapshot(
             import_file, bs_data, PORTFOLIO_BS, is_canon=True,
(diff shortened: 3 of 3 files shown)


pipermerriam:
I'm attempting some fixes for the weird selenium/firefox failures in this branch.  I've removed some of the caching directives in the travis yml and have cleared the cache on travis-ci.  We'll see if this fixes anything when I re-run the latest CI test run.

StephenCzarnecki:
@pipermerriam I got your issue-521-fix-id-splitting-overmatching branch from https://github.com/pipermerriam/seed.  I was going to both see how these changes interact with the multi-year graphing functionality as well as possibly try to improve test coverage.  But I'm running into a problem when I try to upload a file.  Details below, let me know if anything stands out to you or you've run into any of this before.

The file I am trying to upload is very simple, enough so that I can copy the contents here:
Tax Lot ID,Property Type,Year Built,Gross Floor Area,Site EUI,Weather Normalized Source EUI,ENERGY STAR Score,Total GHG Emissions (MtCO2e),Year Ending,some other field
00080806;00080808,Multifamily Housing,1966," 5 ",28.2,103.1,Not Available,1595.97,12/31/2013, 2_records_2013_row_1_some_other_field

So this file just has tax lot ID.  It does not have an address or a pm_property ID.  I can get through the intial mapping screen here:
image

But after that it hangs here:
image

If I look at the broswer console it looks like this:
image
And if I look at the celery output it just cycles through the same output over and over:

image

The most common lines are like
[2015-12-10 15:28:53,025: INFO/MainProcess] Received task: celery.chord_unlock[4f3e9833-a2c7-47ff-96cf-a763e44087f2] eta:[2015-12-10 15:28:54.020766-08:00]
[2015-12-10 15:28:53,029: INFO/MainProcess] Task celery.chord_unlock[4f3e9833-a2c7-47ff-96cf-a763e44087f2] retry: Retry in 1s: AttributeError("'DisabledBackend' object has no attribute '_get_task_meta_for'",)


But there is also the occasional like that looks like
[2015-12-10 15:29:19,271: INFO/MainProcess] Task seed.tasks._map_data[b8337771-eeaa-4554-9cff-57ecf9eb508a] succeeded in 0.0278465049341s: {'status': 'error', 'message': 'waiting for raw data save.', 'progress_key': u':1:SEED:_map_data:PROG:414'}

Let me know if there is any more information I can provide.  If this is a symptom of missing any installs I did pip install -r requirements.txt, node-and-npm-in-30s.sh, and install_javascript_dependencies.sh. 

Thanks

RDmitchell:
@pipermerriam -- it seems like you should work with @StephenCzarnecki tomorrow to get him to a point that he can write some more tests for this fix, so that we can determine if it is the right strategy. I think it is, but we were thinking it might be good to put a few tests together and have Stephen run through them on this branch. thx

pipermerriam:
@StephenCzarnecki I'm unable to reproduce this on my local machine.  I've tried with a freshly created database.  Can you send me more explicit steps on reproducing this?

The second thing you can do is open the developer tools in your browser and open the networking tab prior to loading the app.  Then, if any of the requests throw 4xx or 5xx level errors, get me the content of the response.  We can potentially do a google hangout or something as well so we can screen share which might be the fastest way for us to see what's going on.

pipermerriam:
Also, I'm about to rebase this branch off of the latest develop branch. I'm not sure how well you know git so if you run into any problems, then I would suggest first deleting your local branch and then checking the code out again after running a `git remote update`.

pipermerriam:
Test failure is the same issue ( #531 ) that's been plaguing the other recent work with the firefox selenium tests.  Everything is passing.

StephenCzarnecki:
@pipermerriam Here's what I did.

1:  Reverted my dev vm back to a point where none of the databases had any of the label migrations applied.
2: Since this xkcd rather embarrassingly describes my git workflow Image of xkcd
I renamed my existing seed folder and checked out your branch from https://github.com/pipermerriam/seed
3:  Copied over the local_untracked.py file from my original seed location.
4:  Started celery
5:  Started seed.
6:  Tried to log in, got 400 errors about bower stuff.
7:  Re-did the install steps (pip install -r requirements.txt, node-and-npm-in-30s.sh and install_javascript_dependencies.sh)
8:  Logged in with no issue.
9:  Deleted some existing buildings via seed.
10:  Tried to upload new building as csv.

I just tried it again with a file that only has one tax lot ID and I am seeing the same thing.  I have the Network developer console open and no 4xx or 5xx errors.

Also, and I may be just mistaken about this and/or it may be a red herring but I think the page that I am stuck on is the same page that will eventually haven the verify-mapping-table that is causing the selenium test to fail.

RDmitchell:
@pipermerriam -- Stephen is going to roll back his VM to see if he can figure out when it last worked. maybe this will shed some light on the CI test failures, or maybe it's a red herring. 

I am going to tell @mmclark to go ahead and merge this PR to develop and seedtest, so that I can test your changes to the matching feature.

StephenCzarnecki:
@pipermerriam Long story short I am eventually able to run the selenium tests locally by increasing some timeouts on the wait_for_visibility calls.  So all tests including the selenium tests pass yet when I try to manually upload a file I keep getting the same error from celery.  I'm documenting what I did below so far for posterity.  In the meantime I will try to do both some investigation to the different commits and also see if I can figure out what could be different between when a file is uploaded via Selenium and when I do it by hand.

When I try running the tests locally on this branch (after I install Selenium) I get the element not visible error in the firefox tests but for "review-mapping" instead of "verify-matching table".  I can watch as the selenium test happens and it is getting stuck here for me:
image

And this is the failure:
image

I changed the timeout from 60 to 300 and tried again.  The test got further to org-owned-tables.  I then added some more delays and got everything passing.

RDmitchell:
I believe the change that we made with this fix is not going to work. I just tried uploading and matching the sample data (covered buildings and PM data) and only one record from the PM data matched the covered buildings, when there should be a 100% 1:1 match between the records. 

So we can talk tomorrow about this -- maybe we need to completely revert the changes we made for the duplicate data in the matching field, and really sit down and figure out what the correct data model for it is, so that we can get the matching that we want under all circumstances. 

The program only matched one record but should have matched all 512 records

I will write up in more detail in a google doc and add the link to a comment in this PR

RDmitchell:
Here is a link to the google doc which documents what I did to test this. Basically importing the sample covered building list and then matching it to the sample PM data. The matching doesn't seem to work properly. We can talk about this tomorrow in our call to talk about this. 
https://docs.google.com/a/lbl.gov/document/d/1f4qrotIRKBlZwuIk9OmitvjVee8aIVTw5oSIhDKG5FE/edit?usp=sharing
@pipermerriam -- you might try reproducing my steps in your development environment -- import the sample covered building data, map it the way I have in the above referenced doc, then import the PM data, map it as in the doc, and then look at the matching results. Thx !

StephenCzarnecki:
Got things back working (wasn't setting the DJANGO_SETTINGS_MODULE when starting celery) and going through Robin's document here's what I see.

After initial import of covered buildings there are 1024 records in BuildingSnapshot.  This makes sense since we expect there to be one "raw" record and one "processed" record that is also the canonical at this point.

Once I import and match the second file I see the same results as Robin.  The only match is for 126421 NW Honeylocust Street

Looking at both files one thing I noticed was that address seemed like the only one in the covered buildings sample with a purely numeric value after the / in the UBI field:  10108/23890  The others have things like 10107/71db4

When I clear the database and change one other record in the covered buildings sample to be numeric in both places (10107/71db4 -> 10107/71004)  But that did not help, I still only got the one match with 126421 NW Honeylocust Street

pipermerriam:
Am I correct that there is an expectation for the buildings to match solely on "address line 1", and that they have zero overlap on the fields `pm_property_id`, `tax_lot_id`, `custom_id_1`?  Do we know definitively that prior to this change that buildings would match under these conditions?

RDmitchell:
We can talk about this on the call, but basically there are currently 4 fields that the program allows matching on, depending on how the user does the mapping. In the case of the sample files, for the covered buildings files, 2 fields can be mapped to a matching field, Tax Lot ID and Address Line 1. In the second file, the PM data, there are also two fields that can be mapped to a matching field, Address line 1 and PM Property ID. So in this particular case, where the only common matching field between the two files is Address Line 1, that is the field that the program can match the records on. And the data is set up so that there is a 1:1 match on Address Line 1 between the two files. So after importing both files, you should have 512 records, each one with one match to both the covered buildings and the PM records that matched on Address Line 1. When I did that import, mapping and matching, I ended up with 512 records, but that is because I believe 511 of them matched to one record.

StephenCzarnecki:
If I make a copy of the covered buildings file and pm property ID file and delete everything except the first record I get a match.  If I do the same but with the first two records I only get one match.  If I then look at the details of that matched building I see it matched both records:



The below is possibly irrelevant given the above, including it just in case.

If I change the UBI of that first record to 1234567 I get a match.  If I then apply that change to the larger file I again only get one match out of the 512 but now it is on 155830 SW Garfield Way.  If I try clicking on it in the building list to see the details it takes a while on
[15/Dec/2015 09:37:15] "GET /app/get_building/?building_id=14091&organization_id=1 HTTP/1.1" 200 1376355

And then takes longer on
[15/Dec/2015 09:39:32] "GET /static/seed/partials/building_audit_log.html HTTP/1.1" 200 2426

Finally I get an unresponsive script error and kill the page and it looks like this:

image

PR #524, opened 2015-12-08
434 develop labels

danielmcquillen:
**What was wrong:**
Label feature needed to be updated and enhanced.

**How was it fixed:**
This PR introduces some new functionality around labels and how they're applied in SEED. 
- Labels are now applied directly to buildings, not project-buildings. 
- Label management (CRUD) happens in a separate (not modal) view
- Labels are applied/removed from buildings in a modal window above the building list (with a convenience feature for users to add a new label in that window).
- Labels can be used to help filter the building list.
- Building details now show what labels have been applied to that building
- A request to remove the "instantaneous" filtering feature was incorporated into this branch, as it was somewhat related to adding the label filter. Now, in the mapping view, project view and building list view, the building table list should only update filters after the user clicks the "Update filters" button (not after a keystroke in one of the filter inputs).

In adding these features, a few FE libraries were included
- ng-tags-input, for a nice pill-based label filter
- angular-xeditable, for the Label CRUD page
- angular-ui-notification, for more immediate notifications after simple operations
- underscore, for more efficient FE code. See Piper's use of filters and map, no need to worry about ES5 support / transpiling

On the BE
- we are experimenting with using a REST framework as a way to make endpoints more efficient and effective.
- label methods were refactored into separate view
- data model was updated and migration was built to migrate existing labels
- 'default' labels migration was added to pre-populate some commonly used labels

A few things we punted on until the next iteration:
- Updating the style of the "Update Filters" button when a filter input is dirty
- Showing the color of the building label in the ngTagsInput field (see https://github.com/mbenford/ngTagsInput/issues/579)
- Allowing edits of applied labels at the building detail view
- Showing applied labels for each building row in the building list

IMPORTANT: We're still working on a couple issues with this branch:
- The Travis-CI build is failing on test_firefox : it's waiting for the verify-mapping-table element to appear and times out when it doesn't. Not sure if this is related to changes made in this branch.
- The "Add to an existing project" action on the pulldown in the building list is failing, due to a problem with the download cache key. Again, not sure if this is due to changes made in this branch.

**Cute animal pic**

image


danielmcquillen:
@rHorsey  and @axelstudios, do you think something is amiss with the get_cache() method on the server, or how I'm calling it from the FE? When the method get_adding_buildings_to_project_status_percentage (projects.py) is called, it'll fail if the download cache key hasn't been set. I added a try/except to make this issue more explicit.

danielmcquillen:
Piper found some issues in the BE cache methods. His PR is merged above. @rHorsey does this look ok to you?

PR #525, opened 2015-12-08
Piper/issue 434 weird add buildings to project bug

pipermerriam:
So this fixes *part* of the bug which involved the progress cache object not having a "status" key.


pipermerriam:
After looking into the commit which introduced this bug I think it is wrong to include these changes in this branch.  Thoughts?  We might want to revert this or rebase these commits out of your branch.

https://github.com/SEED-platform/seed/commit/a9b4ca4aa57b81b032408983d4b7e16643cbb19a#diff-ad7529b5f5cf3c633b7b84230df508a1R448

Issue #527, opened 2015-12-09
Bugs related to recently implemented caching utilities


pipermerriam:
cc @RDmitchell @rHorsey @danielmcquillen 

In our recent work on the new Labeling functionality in issue #494 daniel and I have uncovered some very strange bugs that seem related to the work that @rHorsey did back in November via PR #485, and more specifically in commit a9b4ca4aa57b81b032408983d4b7e16643cbb19a

One way this bug is manifesting is when you use the "Add to an existing project" functionality in the "Building Actions" menu while browsing the building list.  The `POST` to `/app/projects/add_buildings_to_project/` succeeds, but the next `/app/projects/get_adding_buildings_to_project_status_percentage/` raises a 500.

The source of this 500 error is from [`seed.utils.cache.get_cache#line-47`](https://github.com/SEED-platform/seed/blob/develop/seed/utils/cache.py#L47) where `data['status']` is accessed which throws a `KeyError` due to the key `'status'` not being present.

Looking at the api endpoint which initiates this task [`seed.views.projects.add_buildings_to_project`](https://github.com/SEED-platform/seed/blob/develop/seed/views/projects.py#L278) I see no place where the cache is populated with the appropriate cache key.  Looking at [`seed.tasks.add_buildings`](https://github.com/SEED-platform/seed/blob/develop/seed/tasks.py#L165) it appears that this cache key is first initialized on line 165 and it does not set a `'status'` key in the object that is written to the cache.

The bug we are seeing is caused by a race condition.  In the event that the request to `/app/projects/get_adding_buildings_to_project_status_percentage/` occurs before the `add_buildings` task has populated the cache, then the call succeeds because the `get_cache` function falls back onto the `default` value which contains a `'status'` key.  However, in the event that the task has already inserted a value into the cache, the call results in a 500 error.

This bug would not manifest itself in tests because tasks are run synchronously which guarantees that by the time the second call happens the entire process has run it's course.

pipermerriam:
I don't have any evidence to support this yet, but I suspect that the selenium/firefox errors we're seeing could be related to this as well.

rHorsey:
[...]

Issue #497, opened 2015-12-01
Problem with get_cache()


danielmcquillen:
I noticed an issue with the get_cache() method in utils/cache.py. 

To recreate on seedtest: select one or two building in the Building List and then select "Create a new project" from the pulldown.

The server will return a 500 error, as within the get_cache() method a Key Error occurs. It looks like on line 47 the code is trying to set data['status'] but that property doesn't exist at that point (at least in this scenario.)

RDmitchell:
@axelstudios -- I just assigned this to you, as you were working on the caching.

rHorsey:
This should be me. Robin, can you assign me to this? Thanks!

RDmitchell:
@rHorsey -- assigned to you ! thx

rHorsey:
Yep; this code is using the previous not-functional way of incrementing async tasks, which will no longer work. @RDmitchell are there progress bars other than mapping, matching, upload, and add-to-project? If so I'll make sure that they have the same updated code.

RDmitchell:
That list of progress bars seems correct.

RDmitchell:
@bravacheLBL -- if you can test this on the latest seedtest tomorrow (Fri, Dec 17) that would be great. thx

RDmitchell:
Just tested this, seems to be working. Am closing

PR #528, opened 2015-12-10
Updating caching methods for project control via api.

rHorsey:
### What was wrong?

API requests for creating projects, adding building to projects, and removing buildings from projects relied on a cache which could be initialized without a status using [`set_cache_raw`](https://github.com/SEED-platform/seed/blob/develop/seed/utils/cache.py#L9) but then would later be reset using [`set_cache`](https://github.com/SEED-platform/seed/blob/develop/seed/utils/cache.py#L17). This lead to an error when [line 30](https://github.com/SEED-platform/seed/blob/develop/seed/utils/cache.py#L30) was executed, assuming that the key had been created using `set_cache` and that it thus had a dictionary key for 'status'. This is noted in issue #527. The behavior of not being able to add projects is noted in issue #497. 

### How was it fixed?

Issue #527 is still outstanding. There are a number of ways of addressing this issue, and I'm not at all sure which is the optimal route. What this PR does address is the specific instance of this generalized bug in issue #497. The code for managing the caching has been updated in seed/tasks.py to follow the same standard used in the data-upload process, which should be immune to async / sync issues. Additionally, the projects view documentation has been updated, and the building list controller has been updated to access the renamed key progress, in favor of the previously used percentage_done. 

### Other issues uncovered:

Exporting is currently having a sync / async problem. I'll resolve this tomorrow.

### DIFF
--- a/seed/static/seed/js/controllers/building_list_controller.js
+++ b/seed/static/seed/js/controllers/building_list_controller.js
@@ -149,11 +149,11 @@ angular.module('BE.seed.controller.building_list', [])
         var stop = $timeout(function(){
             project_service.add_buildings_status(cache_key).then(function(data) {
                 // resolve promise
-                if (typeof data.progress_object !== "undefined" && data.progress_object !== null && typeof data.progress_object.percentage_done !== "undefined") {
-                    $scope.progress_percentage = data.progress_object.percentage_done;
+                if (typeof data.progress_object !== "undefined" && data.progress_object !== null && typeof data.progress_object.progress !== "undefined") {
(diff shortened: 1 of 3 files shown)


rHorsey:
@danielmcquillen @pipermerriam @RDmitchell

rHorsey:
And a cute animal picture:
Image of Komondor

pipermerriam:
:+1:  this looks good.

pipermerriam:
@rHorsey take a look at https://github.com/SEED-platform/seed/issues/527#issuecomment-163688577

pipermerriam:
This has been merged into #524

PR #571, opened 2015-12-15
Updates label service to include user's org id on label objects sent to API

danielmcquillen:


danielmcquillen:
Piper, I can see how it makes sense to include `organization_id` directly on label object for the create operation.

But for update and delete operations, should we even include `organization_id`  since the server can already find it, given we're sending the label ID? If including it is good as a check to make sure user can perform this operation, semantically does it make sense to include `organization_id` directly on label object or is it better to send as a separate parameter (e.g. I'm a user from org 1 and I'm trying to delete this label).

danielmcquillen:
Issue already fixed in separate pull request.

Issue #560, opened 2015-12-14
New label doesn't always show up in Filter list


RDmitchell:
I am getting inconsistent results when I make a new label and then try to use it from the Filter pulldown list. The first time I added a label it appeared, but now I added a 2nd label and it doesn't appear in the list, even though records have been labeled with it.

RDmitchell:
I will make a video for this issue.

RDmitchell:
Here is a video that shows the behavior I am seeing. 
https://drive.google.com/a/lbl.gov/file/d/0B3fTKpZ9Dx7LMzNrSkw0MHRfOFk/view?usp=sharing

@pipermerriam -- I am assigning this to you. Feel free to assign to @danielmcquillen as needed. thx !

danielmcquillen:
In testing via Robin's account, it looks like labels are being created for org ID 2, but the user is logged in as using org ID 6.

pipermerriam:
That makes complete sense.  All of the label backend code uses
`user.orgs.all()[0]`

RDmitchell:
Am assigning this to myself for testing

RDmitchell:
In my +1 acct with 2 orgs, filtered on !"" for Building Name (using file from #271), got 9 buidlings, made a label, clicked Done to apply it and got this error message
![image](https://cloud.githubusercontent.com/assets/6314950/11888699/7f0e9772-a4f5-11e5-864e-ed7c700d7829.png)
 The label was created, and if I don't use that filter and just apply the label to a record, it works fine. 

This doesn't seem to be a problem with an account with just one org.

@pipermerriam -- am assigning this to you, not sure if you will be able to reproduce it. And it is a lower priority because it seems to only apply to accounts with > 1 org, which is not the norm. So if you don't get to it by EOD on 12/18, it's ok.

RDmitchell:
@pipermerriam -- am giving this a Jan 2016 milestone -- we can do a hot fix in late Dec or early Jan

pipermerriam:
@RDmitchell This looks like it *could* be the same error that caused https://github.com/SEED-platform/seed/issues/627.  I'm going to spend a few more minutes attempting to reproduce this but would you mind testing this is still happening when you test #627 ?

RDmitchell:
@pipermerriam -- yes I will test this when I test 627. Thanks for all the fixes !

[...]

PR #572, opened 2015-12-15
Handle users with multiple organizations correctly with labels

pipermerriam:
https://github.com/SEED-platform/seed/issues/560

### What was wrong

The new labeling functionality didn't handle users with multiple organizations correctly.  It was always using `user.orgs.first()` when it should have really been using the `organization_id` that was currently active for the user.

### How was it fixed.

Fixed the organization selector to actually work, after which I modified the `label_service` to always send through `organization_id` as a query parameter.  Then within the two label api views, the organization used is fetched from that organization id.

#### Cuter animal picture

baby-pangolin-riding-mothers-tail


pipermerriam:
Just confirming that the current test failure is only the firefox issue.

Issue #553, opened 2015-12-14
Search_service not clearing selected_buildings


danielmcquillen:
The selected buildings are not being cleared correctly and in sync with the buildings "checked" property.

RDmitchell:
am closing.

PR #589, opened 2015-12-16
Fix 'unseen selected buildings' issue after filters are updated

danielmcquillen:
https://github.com/SEED-platform/seed/issues/553

### What was wrong

Selected buildings that fall outside of a new filter change were not being deselected.

For example, a user might filter once and see four buildings (rows). They might select all four, filter again, and then see two buildings (rows) that are selected. The user might then assume only two buildings were selected, when in fact all four were still selected.

### How was it fixed.

Added a method to `search_service` to explicitly deselect all selected buildings, in the process clearing out the `selected_buildings` array and setting `select_all_checkbox` to false.

This method is now called by controllers after the user updates filters and before the controller calls for a new search.

### To Do
It was agreed that, in a future iteration, we should revisit the concept of updating selections after filters are changed, since we might want to preserve selections that still fall within the new filtered subset. The fix in this PR just removes the potential for error and is probably not an optimum solution.

image

### DIFF
--- a/seed/static/seed/js/controllers/building_list_controller.js
+++ b/seed/static/seed/js/controllers/building_list_controller.js
@@ -65,16 +65,21 @@ angular.module('BE.seed.controller.building_list', [])
     * SEARCH CODE
     */
 
-     $scope.do_update_buildings_filters = function(){
-        refresh_search();
-    };
-
-    $scope.do_update_projects_filters = function(){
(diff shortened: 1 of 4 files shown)


RDmitchell:
Nice animal !! Can you put the label "ready" on this so that Miles knows to merge it into Develop tonight? thx !

Issue #593, opened 2015-12-17
"View Project" button doesn't work after adding buildings to existing project.


danielmcquillen:
**To recreate:**
- Select two buildings in the building list
- Select "Add to existing project" from the pulldown
- Select "View Project" in the success dialog

**Expected results:**
- Server shows project view

**Actual results:**
- Nothing happens

**Analysis:**

Looks like the click event is calling $scope.go_to_project() without providing a `project_slug` argument. The method in this case will defer to $scope.project.project_slug to get a slug, but in my test case $scope.project is itself null at this point, so the FE errors out with a 'can't read property of null' error.

RDmitchell:
@danielmcquillen -- Should this be assigned to @pipermerriam ?

danielmcquillen:
I'm looking into it now. If I can't figure it out perhaps Piper can help. But let me have at it a bit more. Somehow the project object is being remove from the $scope when that final success dialog is shown.

RDmitchell:
Sure, so I assigned it to you. Having it assigned to someone reassures me that someone is looking into it !! Thanks !

danielmcquillen:
The core issue is that the $scope.project object is getting set to null after buildings are successfully added to a project. I can't see exactly where this is happening, but I believe it's due to the fact that the `building_list_controller` calls init() after the buildings are added.

Within its init method, the `building_list_controller` will call project_service.get_projects() because it's currently in building list view state, not project list view state. (It figures this out by checking for the presence of a project_id, if none exists it must be in building list view state.)

When the server returns the list of projects, somehow the $scope.project object is getting set to null.

RDmitchell:
Should I test this on the production server to see if this is an issue that existed before we made the recent changes?

RDmitchell:
Looks like this works on the current production server -- I did the following
* checked 2 records in the building list and made a new project
* viewed the project,
* back to building list, checked 2 more buildings, and added them to the project i just made
* clicked View Project, program highlights the Project nav button and shows the building list in the project

RDmitchell:
It is not just View Project that doesn't work -- if I make a new project, the program makes the Project Name, but doesn't add the records to it that I had selected (after doing a filter on the data)

danielmcquillen:
For the add to existing, the difference is: on the production server after the buildings are added and init() is called, which calls get_projects(), the $scope.project object is not being set to null.
Somehow in the current code, it is getting set to null.

danielmcquillen:
[...]

PR #599, opened 2015-12-18
Update controller and view fix missing .project bug

danielmcquillen:
https://github.com/SEED-platform/seed/issues/593

### What went wrong
In the latest SEED build, at the end of the 'Add to existing project' process on the buildings list, the "View project" button wasn't showing the updated project view when clicked.

To the best of my knowledge, what is happening is that the pulldown shown in the 'select existing project' modal is nulling out the $scope.project property. It does this after the entire add process is complete and the controller has called the project_service to reload the projects. When that happens, the UI select control reloads and nulls out the $scope.project due via its binding.

This behavior is not seen in the current production build, but that could be because of changes to Angular and related libraries in the newer versions we are now using.

### How it was fixed
I created a model on the $scope whose sole purpose is to remember the selected existing project. This avoids clobbering the $scope.project property via an unintended change via binding.

I also updated the way both the "Create a new project" and "Add to existing project" actions are first started after a click event -- having each call a method specific to the action -- to help improve the code clarity a bit.

### To Do
The modals used in this view and the interaction between them and the controller need to be refactored, as the current setup isn't very clear.

### DIFF
--- a/seed/static/seed/js/controllers/building_list_controller.js
+++ b/seed/static/seed/js/controllers/building_list_controller.js
@@ -60,10 +60,7 @@ angular.module('BE.seed.controller.building_list', [])
     $scope.create_project_error = false;
     $scope.create_project_error_message = "";
     $scope.selected_existing_project = null;
-    
-    $scope.$watch('project', function(newValue, oldValue) {
-      $log.debug("project changed to ", newValue);
-    });
+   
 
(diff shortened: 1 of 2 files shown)


RDmitchell:
Great @danielmcquillen -- thanks for fixing this.

Issue #596, opened 2015-12-18
Default labels not available for new accounts


RDmitchell:
@danielmcquillen -- I am assigning this to you initially but it might be an issue for @pipermerriam 

I made a new account on seedtest for testing. I played around with testing filtering, then decided to test labels. 

When I filtered the building list and then clicked on Add/Remove labels, the default labels do not appear. I can make new label, and I see it in the Label Admin screen, but the default labels are not there. 
![image](https://cloud.githubusercontent.com/assets/6314950/11885902/cb9cac86-a4d8-11e5-8b5a-563645f64e1c.png)

The default label list doesn't appear on an existing account either
![image](https://cloud.githubusercontent.com/assets/6314950/11885942/1a392c02-a4d9-11e5-86b8-ed5632b45cd5.png)

But once I make a new filter, either from the Building list or from the Label Admin screen it seems to work fine in terms of being in the filter list, filtering properly, showing up on the building detail screen.
![image](https://cloud.githubusercontent.com/assets/6314950/11885970/75feef2c-a4d9-11e5-81d0-00df5d0ea26a.png)
Program filters properly on the new filter
![image](https://cloud.githubusercontent.com/assets/6314950/11885981/a1a1f462-a4d9-11e5-89f6-94fa3109fd34.png)

RDmitchell:
@pipermerriam -- I am putting this in your queue, since tomorrow is Daniel's Saturday. If you can fix this by EOD on Friday (12/18) that's great, but if not it's not the end of the world. The rest of the label functionality seems to be working well, so we don't have to have the default labels working if you don't have time or it's complicated.

pipermerriam:
@RDmitchell can you tell me how this new account was created?

RDmitchell:
Made a new account, default records are in the Label view. Closing

PR #604, opened 2015-12-18
Piper/issue 530 default labels for new orgs

pipermerriam:
https://github.com/SEED-platform/seed/issues/596

### What was wrong?

Organizations created as a result of either the management command `create_default_user` or the api endpoint `add_user` were not populating the default set of labels.

### How was it fixed?

Changed both of these to use `seed.utils.organizations.create_organization` which contains the code to populate the default label set.


pipermerriam:
Hurray for code linting! https://travis-ci.org/SEED-platform/seed/jobs/97752118

PR #606, opened 2015-12-18
fix non-deterministic test case

pipermerriam:
### What was wrong?

The test `seed/cleansing/tests.py::test_cleanse` was constructing a list using `[v for v in some_dict.values()]`.  Since dictionaries are not ordered, when this was later checked to be equal to another list of the expected values, the items didn't end up being in the same order and thus the equality check failed.

### How was it fixed

Instead of using `self.assertEqual` which requires true equality, the test now uses [`self.assertItemsEqual`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertItemsEqual) which ensures that the items in the two iterables are the same (independent of ordering).

#### Cute animal picture

4894-sleeping-puppy-1920x1080-animal-wallpaper


pipermerriam:
Lets see if travis agrees with my assessment of this test?

PR #624, opened 2015-12-22
Don't override extra data with blank data

pipermerriam:
### What was wrong?

Data found in the `extra_data` field(s) would be overridden if new data was uploaded that contained blank values.

### How was it fixed

Changed the logic in `seed.utils.mapper.merge_extra_data` to not clobber existing data with blank data.

#### Cute animal picture

412187-dogs-cute-sleeping-puppy-in-a-bathroom


pipermerriam:
Not ready for merge yet.  I have some testing and refactoring that I still need to do.

pipermerriam:
ok, ready to merge when tests are green.

Issue #611, opened 2015-12-19
Update Master record with good data if it was overwritten with blank data


RDmitchell:
Per #610, we need to figure out a way to correct the problem in an existing organization's data. The program has a record of all the merged data, so the original data that was overwritten by blank data is known. 

We will work on this the first week of Jan when we are all back from the holidays. I am assigning it for now to @StephenCzarnecki

RDmitchell:
@pipermerriam -- I am assigning this to you as your first priority for the next round of issues. This might be a bit tricky. If you can at least ponder it, and figure out how hard this would be to do, that would be a start. Thx !!

RDmitchell:
One question about this is whether this is something we do "behind the scenes" on individual orgs that request it, or we put the functionality into the UI so a user can trigger this action as needed. If we decide to put it in the UI, it could be a superuser function, or it could be something that individual users (probably "owners") of an organization, could trigger. We can figure this part out once we figure out how hard it will be to do.

pipermerriam:
@RDmitchell I think I'd prefer to do this as a data migration.  That way it's a one-off thing that runs on the next deploy and we don't end up with code that needs to be maintained long term since the application should now treat this data correctly.  Thoughts?

RDmitchell:
That does seem like the right approach.

RDmitchell:
I am assigning this to Stephen for testing.

StephenCzarnecki:
In short looks good to me.  

I don't know how to write an automated test to test a migration so I just did it by hand.  I made some files with the following columns: tax lot id, address line 1, foo.  For some of the later files to prevent duplicates I added a "q" column too.  The idea being that neither "foo" nor "q" should be recognized and therefore would be extra data.  

I came up with a handful (18) of test cases some of which I now realize are probably redundant.

I started on revision 04cb03d0ff871ba6bf73e9ff18af027796e6e9ef but had problems viewing the building details so I moved to c6eccd95d3d6ada6e6334a1f6d68ff75991931e6 (commit Bump version number to 1.3.0)

I uploaded file extra_data_file_1.csv in the attached.  On the mapping screen I unchecked the MAP checkbox for the "foo" column.  Found no matches as expected.

Then I uploaded extra_data_file_2.csv.  The MAP checkbox next to the foo column was already uncheacked.  Matching found 19 as expected.

I verified that I was seeing the issue on the cases I expected to see it (1,2,6,7,8,12,13) and the other cases behaved as expected.

[...]

PR #626, opened 2015-12-22
Piper/issue 611 populate extra data fields that were overwritten

pipermerriam:
https://github.com/SEED-platform/seed/issues/611

This PR is based off of #624 

### What was wrong?

Historically, if new data was imported with `extra_data` which contained blank values, those values would overwrite existing values that were not blank.  This has caused data losses.

### How was this fixed?

Since the data model saves all historical states all of the *lost* data is recoverable.  This PR includes a data migration that does the following.

* Finds all building snapshots which are:
    * A canonical snapshot.
    * Has parent snapshots (is the result of merging two snapshots)
    * Has no child snapshots (has not been merged with other snapshots)
* For each of these canonical snapshots:
    * Recursively perform this merge algorithm.
    * For each parent, merge the extra data, prioritizing the *latest* snapshot's data.

With the new extra data merging logic, this will result in any non-blank values that were overridden at some point propagating through the tree towards their canonical snapshot

#### Cute animal picture

msg-126083882108-3

### DIFF
--- a/seed/mappings/mapper.py
+++ b/seed/mappings/mapper.py
@@ -75,14 +75,20 @@ def merge_extra_data(b1, b2, default=None):
     extra_data_sources = {}
     default_extra_data = getattr(default, 'extra_data', {})
     non_default_extra_data = getattr(non_default, 'extra_data', {})
-    extra_data = copy.deepcopy(non_default_extra_data)
-    extra_data.update(default_extra_data)
+
+    all_keys = set(default_extra_data.keys() + non_default_extra_data.keys())
+    extra_data = {
+        k: default_extra_data.get(k) or non_default_extra_data.get(k)
+        for k in all_keys
+    }
 
     for item in extra_data:
-        if item in default_extra_data:
+        if item in default_extra_data and default_extra_data[item]:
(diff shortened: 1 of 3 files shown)


RDmitchell:
Stephen is going to do some testing just to make sure this is working properly before we put it onto seed production.

seed/mappings/mapper.py
```diff
@@ -1,7 +1,6 @@
 """
 :copyright: (c) 2014 Building Energy Inc
```
rebrownlbnl: I notice the old copyright notice is in the mapper.py file.
Looks like that file was missed when @pipermerriam made the pass to put in the new notice, or there's a versioning issue.

pipermerriam: It's correct in the upstream branch 

https://github.com/SEED-platform/seed/blob/develop/seed/mappings/mapper.py

This branch was just made prior to that change.

rebrownlbnl: Sorry, I got thrown off because the automated testing says "This branch has no conflicts with the base branch." I assumed the base is the current develop branch, but that must not be the case.

Issue #639, opened 2016-01-04
Matching view still has 'active' filter inputs.


danielmcquillen:
The filter input fields in the matching view table header are still dynamically updating the filter on each keypress. These need to be switched to only updating on 'Enter' key.

danielmcquillen:
The matching table list doesn't have the date filters shown in the building list. Is that intentional?

RDmitchell:
probably not. I have made an issue for it to be dealt with in the June 2016 release. #643 
We need to make a pretty significant pass through the matching screen after the June release to make it more useable, and we can revisit how to handle date filtering (there and everywhere else in the program) at that time.

RDmitchell:
This matching screen no longer exists

Issue #545, opened 2015-12-11
Activate Update Filter button with Enter key


RDmitchell:
I don't know if this is possible, but the change from the "immediate response" filtering that we have disabled to now having to click the Update Filter button every time you want to do a filter, seems like a bit of a regression. However, I think it is necessary to get rid of the flakey filter issues we were seeing before ( #471 ).

One thing that would help is if you could type in the filter criteria and then just hit the Enter button on your keyboard, rather than having to click the Update Filter button with your mouse. 

@danielmcquillen -- I am giving this to you -- you can tell me what the level of effort would be for doing this and then we can decide when to do it. Thx.

danielmcquillen:
This was included as part of the Labels work https://github.com/SEED-platform/seed/pull/524. 

(Sorry, should have been a separate PR to address just this issue. )

RDmitchell:
Daniel -- when you have this fixed, assign it to me and label it "test" and I will then test it and close it. Thx !

danielmcquillen:
There's a note in the code (ALECK) to at some point start using the search_service's filter_search() method rather than then one implemented in the matching_controller. 

I'm not going to tackle making that change in this issue. That refactoring should happen in a separate issue/PR.

rebrownlbnl:
@danielmcquillen can you create an issue for that refactoring, but make it a little more general to evaluate whether search_service is the right tool to use in this case?

RDmitchell:
@danielmcquillen & @rebrownlbnl -- I just created an issue #644 for the refactoring. Feel free to add whatever info you need to there, and change the title if desired. Robin

RDmitchell:
This is working on seed production. Closing

Issue #607, opened 2015-12-18
Add spinner when the program is taking a long time to do something


RDmitchell:
For example, adding a label 16,000 records -- when you click Done the program just sits there for a few seconds before that dialog box goes away as it applies the labels to the records.

@axelstudios had this idea, which seems good to me
For Angular web apps, I typically implement spin.js (http://fgnass.github.io/spin.js/) to appear in the middle of the window when the $rootScope detects a $stateChangeStart event, and to stop the spinner on $stateChangeSuccess and $stateChangeError.  That way, pages that take a long time to load will always show/hide the spinner during the transition.  You can always show/hide the spinner to indicate loading for other long events, too.

![image](https://cloud.githubusercontent.com/assets/6314950/11910821/bb3b6f34-a5b5-11e5-8a9e-443432b178f3.png)

Not too high priority (have put to priority = 2) but if it's not too much work would be good to add so that the user knows what's going on. Am setting the milestone to the June release

danielmcquillen:
Example usage: show progress indicator after user changes a filter in building list and hits return or clicks "Update Filters" button.

RDmitchell:
One example of where the program takes a long time to display results: 

* Matching
 * 5,000+ records, 100 records displayed at a time -- takes several seconds to display the 100 records the first time you set the "display" pull down to 100. Then moving to the next set of 100 records using the "Next" button at the bottom of the screen takes another few seconds. This would be a good place to implement the spinner.

RDmitchell:
Another place for the spinner is in the Building List when Show Matched and Show Unmatched are selected. The program takes a while to display the records, and you don't know when it is finished or not. So a spinner while the program is generating the list to display would be good.

RDmitchell:
Just deleted 14,000+ records from the building list and it took a few seconds, when nothing was being displayed on the screen. Might be another candidate for the spinner.

nicholasserra:
So do we want to go with a full screen version of this, or do we want to place targeted spinners where needed? Full screen would probably be quicker and easier. Not sure if we want to dim out the rest of the page. Could use an overlay.

RDmitchell:
I have been noticing the spinner on other web apps. It does seem to be fairly common that the spinner is full screen.
@danielmcquillen -- I would like to get your design input on this -- you were the one who suggested it and it would be great if you could provide some design guidance.

nicholasserra:
[...]

PR #640, opened 2016-01-04
545 return key submits filter

danielmcquillen:
https://github.com/SEED-platform/seed/issues/545, #639, https://github.com/SEED-platform/seed/issues/646

### What was wrong?

It's tedious for the user to click on the "update filters" button every time she changes a field in a buildings list table. 

Also, the filters on the 'Matching' table haven't been updated and still had the old called-on-every-keystroke behavior.

Finally, a small filter / results page bug was causing trouble when filtering on a results page other than 1 (#646).

### How was this fixed?

Updated the buildings table to call an enter-key-pressed function, which then decides how to update filters (project buildings or buildings). This included creating a new directive to handle an ENTER keypress in an input field. 

This new directives was also added to the filter fields in the matching list table and the related controller was updated in order to give the matching buildings list view the same behavior as the buildings list, project buildings list, and mapping buildings list.

Finally, the `building_list_controller` and `mapping_controller` were updated to use search_service's `filter_search` method rather than its `search_buildings` method.

## Notes

Currently we have some directives that have the old 'be' (Building Energy) prefix, and others which don't have a prefix at all.  We probably should be using a project-specific prefix for things like directives, so I started a convention of using 'sd' for SEED on the new directive `sdOnEnterKey`. Crazy? Good idea? Thanks for any feedback.

## Animal pic

image


danielmcquillen:
I noticed while implementing this that I missed the Mapping view when updating the filters to avoid submitting on every keypress.

rebrownlbnl:
I think your proposed naming convention for directives makes sense.
Should we make a pass through all the FE code to update to the new convention?

danielmcquillen:
Yes, Rich, that wouldn't take long. There are nine directives that would need updating. I can create an issue and separate pull request if you'd like.

RDmitchell:
Yes, go ahead and do this.

danielmcquillen:
Ok, I created a separate issue for refactoring directive names. https://github.com/SEED-platform/seed/issues/641

In the meantime, I did include the fix to the building matching table filters in *this* PR, as it's directly related to the work and uses the new directive created in this PR.

danielmcquillen:
Note that this function is probably a good example of where we need a progress indication (as we were discussing at the end of 2015). Right now there's no indication of what's happening after you hit Enter or click the Update Filters button.

danielmcquillen:
This is almost good to go. I just want to test it a bit more tomorrow to see if I can break it.

RDmitchell:
@danielmcquillen -- This is labeled DO NOT MERGE and In Progress. Can you remove those labels if it is ready to be merged, so that Miles will know the status of this PR? thx Robin

RDmitchell:
agreed. But we will have to put this off until after the Feb release. #607 (add spinner) is tagged with the June 2016 release milestone

danielmcquillen:
I missed the fact that a directive already existed for enter-key submits, so I updated the naming to sd prefix and changes all references to use that directive.

Issue #641, opened 2016-01-05
Refactor directive names


danielmcquillen:
Some directives has a "be" prefix, the rest don't have a prefix at all.
Standardize all directives to have an "sd" prefix (for SEED).

axelstudios:
No longer an issue - all directives currently begin with `sd`

PR #652, opened 2016-01-11
Refactored all directives except download to use 'sd' prefix

danielmcquillen:
#641 

### What was wrong?
Directives should use a project-specific prefix to avoid colliding with Angular directives current or future. Some of our directives did not have prefixes, and some had a legacy 'be' prefix ( for Building Energy).

### How was this fixed?
Refactored directives to use 'sd' prefix for SEED.


danielmcquillen:
I haven't yet refactored the 'dropdown' directive.

Issue #627, opened 2015-12-22
Can't add label from Filtered building list


RDmitchell:
When I do a filter on the building list, and then try to add a label to it (on seed.lbl.gov) I get the following error.
![image](https://cloud.githubusercontent.com/assets/6314950/11968635/33cb553a-a8c3-11e5-9c99-2726265b94e4.png)

RDmitchell:
The program also puts up an error message
![image](https://cloud.githubusercontent.com/assets/6314950/11968671/917f6428-a8c3-11e5-9e73-b0d55dcdfad6.png)

RDmitchell:
@pipermerriam -- I am assigning this to you, Priority = 1. If you can't replicate this in your development environment, let me know and I will add you to my seed account that shows this problem, in case that helps you debug.

RDmitchell:
Can replicate with my +3 account on production seed

RDmitchell:
@pipermerriam -- I am testing the current version of seedtest (12/23/2015 12:11 PM), and I can't replicate this problem. So for some reason this might only be happening on seed (production). I am going to do more testing on seedtest and make sure it is all working. Then if we move the code from develop to master and onto seed (production) I will test again.

pipermerriam:
Looking into it now.

pipermerriam:
Ok, I'm able to reproduce on seedtest but not locally.  Given that it's throwing a 500 it should be easy to diagnose if I can get access to the server logs (or the app can be toggled into debug mode).  Any ideas on how to accomplish this? @RDmitchell

RDmitchell:
I think @mmclark can put the server into debug mode. Not sure what his work schedule will be over the holiday. We can work on this the first week of Jan if not before.

pipermerriam:
I can no longer reproduce this on seed test.  @RDmitchell  can you see if this is still happening to you?

RDmitchell:
On seed.lbl.gov (production), which has all the latest merged PRs, I am getting the following error. The sequence was:
* filter on a field, add an existing label, that worked
* delete that filter, click update filters so that you see all the records
* filter on that same field, but with a different value (text string in this case), then click Update Filter button and get the problem below.
![image](https://cloud.githubusercontent.com/assets/6314950/12306873/f09bc148-b9f0-11e5-88e3-0273d6857198.png)

@pipermerriam -- I can add you to the organization (test org) on seed that has this issue, if you think that would help.

[...]

PR #663, opened 2016-01-13
Fix 500 error when fetching labels

pipermerriam:
Issue #627 

### What was wrong?

If you were trying to add a label while sorting on a field that was in the `extra_data` section of the building a 500 was thrown.  This was due to how sorting was implemented on extra-data fields which results in a `list` type being returned.

Later in the code, it was have `.filter()` called on it which resulted in an `AttributeError`

### How was it fixed.

Sidestepped the bug by adding a new parameter to the `orchestrate_search_filter_sort` helper function to allow skipping the sort portion of that function's logic.

#### Cute animal picture

2772f037b04aae5a1a394e9184e3dda8

### DIFF
--- a/seed/filters.py
+++ b/seed/filters.py
@@ -41,6 +41,7 @@ class BuildingFilterBackend(filters.BaseFilterBackend):
         buildings_queryset = search.orchestrate_search_filter_sort(
             params=params,
             user=request.user,
+            skip_sort=True,
         )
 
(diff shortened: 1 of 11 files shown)


RDmitchell:
aha ! the good old extra_data field ... Thanks Piper

pipermerriam:
Good to merge when tests are passing

Issue #595, opened 2015-12-17
"" or !"" in a numeric field gives server error


RDmitchell:
the expressions "" and !"" are not allowed for numeric fields -- null and !null are used instead. 

However, if the user (like me who forgot !) types "" or !"" into a numeric field, the program gives a server error

![image](https://cloud.githubusercontent.com/assets/6314950/11885369/327fce6e-a4d5-11e5-820d-4b8632b6ac25.png)

Could we put that error into the UI, ie, saying something like "That expression is not allowed for this field type" or whatever makes sense? 

@pipermerriam -- I am putting this in your queue for the Feb release.

nicholasserra:
I'm thinking a good option on this is to just catch the error silently on the BE, but on the FE have a set of inputs that we throw a visual warning on? How do you see this error showing up in the UI @RDmitchell? Are there some existing error popups or tooltips I could use?

RDmitchell:
@danielmcquillen -- any UI ideas for this?

nicholasserra:
Another simple option is to just fix the server error. We're allowing any text in that field, so there are a lot of inputs that won't filter correctly.

nicholasserra:
@RDmitchell can you give me quick background on why we're doing `null` on numeric and `!""` on non numeric?

RDmitchell:
@nicholasserra -- it was my understanding that is what you implemented.

nicholasserra:
No, maybe @pipermerriam put that in. Either way, just need some decisions on what route to go for now. If we show an error message for `""`, we probably need to show error messages for any invalid input.

RDmitchell:
I guess I need clarity on what expressions can be used for filtering. Can you write up a list of the expressions that are common across both numeric and character fields? Going back through the issues it was confusing to me what worked for different cases, and also it didn't seem like there was consistent functionality.

What would be the expression to use to filter for Null and Not Null values in both character and numeric fields ?

nicholasserra:
I don't think the expressions are consistent across numeric and character fields. Example: `""` for characters and `null` for numeric. Also, numeric seems to have a set of comparison operators wired in now. So we would have to write some regex on the FE to match possible expressions. 

I think our best option right now is to:
 - Fix this 500. 
 - Decide if we want to normalize these values across character and numeric fields.
 - Decide if we want to build a FE regex pattern to stop a user from entering invalid stuff.

We should probably split those 3 out (maybe use this ticket as the first one), as they all require some discussion and work.

[...]

PR #671, opened 2016-01-20
Issue #595 - Catch invalid operator errors and return nothing.

nicholasserra:
#### What's this PR do?
Incorrect null syntax in filter fields will throw an error. This is a first step patch to at least not give a 500 back from the server. Return no results on invalid syntax.

Refs #595


rebrownlbnl:
:+1:

PR #681, opened 2016-01-25
655 updating copyright in config

StephenCzarnecki:
Updated the copyright in seed/config and added it to any files without one.  Also changed the README.md and any other python files I saw that still had "copyright: (c) 2014 Building Energy Inc"


rebrownlbnl:
Yes, this looks like all these files were modified properly.
I can't vouch that other files aren't lurking somewhere that need editing, but it seems like we've rooted everything out.

One annoying thing is that the copyright date is now out of date, it should read "(c) 2014-2016..."
Should we just make a wholesale edit in all the files just before our next release?

RDmitchell:
Yes I think we should make a final pass through the copyright and get it to read 2016.

I will make a new issue and assign it to @nicholasserra

Issue #75, opened 2014-11-27
Date problems in XLSX files


RDmitchell:
There are several issues concerning dates in XLSX files that SEED is not intepreting correctly. 

See one example file in the gdrive folder for this issue, where the date in cell 3M is set to 1/3/1900, even though it should be 1/3/2000. 

The problem is the combination of (1) some dates in Marshall's spreadsheet, (2) The way that Excel handles dates, and (3) the way that mcm handles a date ambiguity.

The full details of the date issue can be read here: https://secure.simplistix.co.uk/svn/xlrd/trunk/xlrd/doc/xlrd.html?p=4966 under "Dates in Excel Spreadsheets", but the short version is that Excel for Windows mistakenly treats 1900 as a leap year, which causes any date before 2/29/1900 to be ambiguous. Excel for Mac starts it's date counting in 1904, which increases the confusion.

The xlrd library, which reads incoming excel files, throws an "XLDateAmbiguous" exception when encountering a date like January 3, 1900 and mcm doesn't catch and handle that exception.

Marshall's spreadsheet has a number of January 3, 1900 dates in it (seed cell 3M for an example).

A temporary work-around is to save the excel spreadsheet as a CSV file and upload that. A permanent fix will likely require re-implementing a function from the xlrd library that parses date values and resolves the ambiguity in a way we think appropriate.

There are several other exceptions that the date parsing function in xlrd can throw on bad input data. The mcm module doesn't appear to handle any of them.

(This was originally LBL issue # 6)

RDmitchell:
Baptiste -- if you could test to see if this is still a problem that would be great. thx.

bravacheLBL:
I tried to import the data in the GDrive from Claudine's API test and it failed at get_column_mapping_suggestions. In SEED, the progress bar stops at 45% and then return an Unhandled Error 3.0 with the following trace:

"Traceback (most recent call last):
File \"/home/ubuntu/prj/seed-lbnl-v1.2.0/seed/tasks.py\", line 774, in _save_raw_data
cache_first_rows(import_file, parser)
File \"/home/ubuntu/prj/seed-lbnl-v1.2.0/seed/tasks.py\", line 701, in cache_first_rows
row = rows.next()
File \"/home/ubuntu/prj/seed-lbnl-v1.2.0/seed/lib/mcm/reader.py\", line 142, in next
yield self.excelreader.next()
[...]

PR #683, opened 2016-01-27
Issue #75 - Handle odd date format in excel dates.

nicholasserra:
Excel is bad at dates. Mac and Windows formats have different date formats. 

#### What's this PR do?
Handle odd date formats by using a new method on the excel parser package. Add correct error messages for any edge cases where this error may still occur. Add regression tests.

Refs #75


RDmitchell:
thanks Nick !!!

rebrownlbnl:
:+1:

Issue #634, opened 2015-12-23
Renaming Data set doesn't save new name


RDmitchell:
**Problem**
Using the Edit icon to rename a dataset, and clicking the Save button, the new name is not saved

**To recreate**
* Make a dataset
* Click the pencil (edit icon)
![image](https://cloud.githubusercontent.com/assets/6314950/11986079/6011cb34-a980-11e5-8814-8893ba49c816.png)
* Rename the dataset
* Click the Save button
![image](https://cloud.githubusercontent.com/assets/6314950/11986093/7f24993e-a980-11e5-9e18-1493aa364973.png)
* Program momentarily flashes the screen with the new name but then refreshes (or something) and the old name is still there
![image](https://cloud.githubusercontent.com/assets/6314950/11986109/a2138f9a-a980-11e5-9aa5-e353bdd6188d.png)
* If I manually refresh the browser, the name is still the old name. 

**Importance**
Am setting to priority = 3 for the Feb release but if it moves to the June release (or a hot fix in between) that will be ok.

nicholasserra:
This is working for me. May be invalid. Retest and close?

RDmitchell:
Will do. Am assigning to myself for testing

RDmitchell:
I still see this issue on seedtest, which represents the latest develop branch. 
Here is a video
https://drive.google.com/a/lbl.gov/folderview?id=0B3fTKpZ9Dx7LeENpdUxxZUgwRWM&usp=sharing

nicholasserra:
This looks like a data issue. Something weird with permissions. Still hunting it down.

RDmitchell:
Let me know if you need a copy of the seedtest database. @mmclark could provide that.

nicholasserra:
Nope I was able to reproduce locally finally. Fix incoming. Thanks!

RDmitchell:
Tested on seedtest -- fixed !! yippeee !! Thx Nick. Am closing

PR #688, opened 2016-01-27
Issue #634 - Renaming data set fix

nicholasserra:
#### What's this PR do?
Renaming data sets was broken unless you were flagged as a Django superuser. The endpoint needs the `organization_id` passed along.

I don't have any tests for this, but it's a good candidate for a selenium test. Not sure if those are ready. 

Refs #634


RDmitchell:
thanks Nick !

Issue #684, opened 2016-01-27
Admin/Superuser Delete Buildings not working


RDmitchell:
If I log into seedtest with my superuser account, so that I can see all the organizations, when I click on the red DELETE BUILDINGS button, the program doesn't do anything. There don't appear to be any error messages.

I tried logging in with an incognito window and it didn't make any difference. Incognito + browser refresh = same result of no movement in deleting the buildings. 

Also, it took almost 2 minutes for the program to display the list of organizations at the bottom of the screen (!!)

In the example below, I clicked on the Delete Buildings button for LBNL 77, and the program turned the button into a gray progress bar, but then never actually made any progress. This can sometimes take a long time, but now it isn't even starting. 
![image](https://cloud.githubusercontent.com/assets/6314950/12624987/94d2c1bc-c4e6-11e5-82be-b1996467322e.png)

StephenCzarnecki:
Seems like this has something to do with changing celery to use json instead of pickle in https://github.com/SEED-platform/seed/commit/2631c66bdd7382e35c79dea887dc36833e31503f

In short at that commit (and currently it seems) celery gives an error about something not being JSON serializable.

Prior to that commit the admin deleting function seems to work.  At that commit I get the following errror in celery when I try to do an admin "Remove Buildings":

```
[2016-01-29 15:45:13,541: ERROR/MainProcess] Task seed.tasks._delete_organization_buildings[0d4d0f2a-b8ef-48fd-b837-fec9a8ba2ee4] raised unexpected: EncodeError(TypeError('[17557, 17558] is not JSON serializable',),)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/vagrant/seed/seed/decorators.py", line 63, in _wrapped
    response = fn(import_file_pk, *args, **kwargs)
  File "/home/vagrant/seed/seed/tasks.py", line 1526, in _delete_organization_buildings
    _delete_canonical_buildings.delay(can_ids)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/task.py", line 453, in delay
    return self.apply_async(args, kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/task.py", line 559, in apply_async
[...]

PR #699, opened 2016-02-01
typecast value_list to list when serializing to celery

nllong:
### What was wrong?

The JSON serializer for celery could not handle value_list.

### How was this fixed?

Typecase value_list to a list


nllong:
#684

PR #703, opened 2016-02-02
Trying to update the copyright again.  

StephenCzarnecki:
Apparently I failed to do a pull before doing the last pull request causing some merge conflicts.  At least one was caused by editing a file that had been deleted.  Since the actual fix was mostly a global find-and-replace I just did it again in a new branch. Thought that would be easier than dealing with the conflicts.


StephenCzarnecki:
Trying again, this time without conflicts hopefully.

PR #704, opened 2016-02-02
Stop-gap solution for django-pgjson bug

pipermerriam:
https://github.com/SEED-platform/seed/issues/627

### What was wrong?

A bug in django-pgjson was resulting in a query that *should* have worked failing.

https://github.com/djangonauts/django-pgjson/issues/35

### How was it fixed

Instead of letting the django orm do it's magic, I cast the subquery to a tuple to force evaluation and bypass the bad code.

It's worth noting that if we can get switched over to Django 1.8 we can drop this 3rd-party dependency entirely in favor of `django.contrib.postgres` based fields which are likely to be more reliable.

#### Cute animal picture

6d2d35001417b44df88ee66f908dfdc4

### DIFF
--- a/seed/views/labels.py
+++ b/seed/views/labels.py
@@ -112,7 +112,11 @@ class UpdateBuildingLabelsAPIView(generics.GenericAPIView):
         """
         building_snapshots = self.filter_queryset(self.get_queryset())
         queryset = CanonicalBuilding.objects.filter(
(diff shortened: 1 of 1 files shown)


pipermerriam:
Ready (assuming tests pass)

RDmitchell:
Made a new issue #705 for Piper's note about removing this workaround once we upgrade to Django 1.8

Issue #694, opened 2016-01-29
Change text in Mapping screen for Custom ID


RDmitchell:
In the Mapping screen "Step 1" tab description, change name of matching field from "Custom ID" to "Custom ID 1"

![image](https://cloud.githubusercontent.com/assets/6314950/12664868/09a77238-c5e8-11e5-963e-67f18411710c.png)

pipermerriam:
Assigning this to myself.

RDmitchell:
Fixed thanks ! Simple but was creating confusion with users, so great to get it changed !

PR #706, opened 2016-02-02
Fix text reference to Custom ID 1 field

pipermerriam:
https://github.com/SEED-platform/seed/issues/694

### What was wrong

Reference to "Custom ID 1" field was incorrectly written as "Custom ID"

### How was it fixed.

Updated the appropriate template.

#### Cute animal picture

b880a22820f81093cda4b7fc66eef588


pipermerriam:
Ready (assuming tests pass)

Issue #577, opened 2015-12-16
Add Show/Hide Columns and Reorder Columns to Building Detail


RDmitchell:
In the Building List, it is possible to control what fields are displayed in the list, and what order they are in, under the List Settings link. I am putting the main ideas about the new functionality directly into this issue, but am also starting a spec document where the developer who works on this can add more details. Here is a link to that doc:
https://docs.google.com/a/lbl.gov/document/d/1wCBzs5iJxNg0Hlg8Xhfzyjx2DXfhGf4GXCofMQdHNsU/edit?usp=sharing

This functionality would also be useful in the Building Detail view.

**Show/Hide Columns feature in Building List**
![image](https://cloud.githubusercontent.com/assets/6314950/11832321/05549d08-a36c-11e5-9fa5-34f7a830c6a9.png)

**Reorder Columns Feature in Building List**
![image](https://cloud.githubusercontent.com/assets/6314950/11832309/e907a820-a36b-11e5-98af-549bf8979f67.png)

**Building Detail View**
Add a List Settings link
* This link would go to the same type of screens that the Building List goes to
 * Show/Hide Columns
 * Reorder Columns
* These settings would need to be stored separately from the Building Detail settings so that the user could have different settings for the Building List and the Building Detail views.
* These settings would controls what fields (and associated data) are shown on the left side of the screen, and what order the fields are in.

![image](https://cloud.githubusercontent.com/assets/6314950/11832350/43fa3c5c-a36c-11e5-97fe-b8413fbd376b.png)

RDmitchell:
assigned to @danielmcquillen with help from Piper for BE issues as needed.

nicholasserra:
@RDmitchell Would these column preferences be different on each building, or apply to all buildings?

RDmitchell:
@nicholasserra -- if it is simpler to make the preferences apply to all buildings in the org, I would say we should start there. If, after users work with this, they decide they want to control it on a building by building basis, we can add that. I think in general, they just want to hide the fields that they don't use, so probably having the settings apply to all buildings is generally what people will want. thx.

RDmitchell:
@nicholasserra -- feel free to add to the spec if you think that is useful or necessary. Or just add comments directly into this issue. thx.

nicholasserra:
[...]

PR #708, opened 2016-02-02
Issue #577 - Building Detail show/hide/reorder columns

nicholasserra:
#### What's this PR do?
Adds column show/hide/reordering on the Building Details page. Implements this in the same way as the Building List page is done. A new "settings" tab is added to the detail page for this.

I was able to use the existing buildings_settings_controller to accomplish this, by making minor modifications to handle a single building.

On the BE, these preferences are stored in the same way as the buildings list page, in a new field on the user model. Building Detail column settings are global in this implementation. I used a new field so that we could easily make changes if we feel there's a need for a different solution.

Try it out on a Building Detail page.

Refs #577 

animated

screen shot 2016-02-02 at 3 19 19 pm

screen shot 2016-02-02 at 3 19 38 pm

### DIFF
--- a/seed/landing/migrations/0003_seeduser_default_building_detail_custom_columns.py
+++ b/seed/landing/migrations/0003_seeduser_default_building_detail_custom_columns.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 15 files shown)


RDmitchell:
@nicholasserra -- looks good, look forward to testing it.

Issue #635, opened 2015-12-24
Using View/Hide Columns in Mapping screen causes program to go back to Map your Data


RDmitchell:
**Problem**
When you have mapped your data and you are in the Mapping screen where you can see the actual data with it's new mapping (the list view), if you click the View/Hide Columns button, and select new or deselect existing fields, when you click Save Settings, you are put back into the very first mapping screen, and when you move from that screen to the final mapping screen, the list settings that you changed were not saved

**To reproduce**
* Import a data file and map it
* After clicking Map Your data, you will be put into the Mapping screen where you can see the actual data with the new mappings
* Click on the View/Hide Columns button
![image](https://cloud.githubusercontent.com/assets/6314950/11987032/c39ac348-a98c-11e5-8378-e15ebec0b211.png)
* You will see the Building List Settings dialog box
![image](https://cloud.githubusercontent.com/assets/6314950/11987106/41803134-a98e-11e5-9dba-158b77525e42.png)
* Deselect a checked field (or select a new field) and click Save Settings
![image](https://cloud.githubusercontent.com/assets/6314950/11987215/f48ffc18-a98f-11e5-963f-8739ad096abc.png)
* Program puts up the "Please wait while your data is loaded"
![image](https://cloud.githubusercontent.com/assets/6314950/11987224/096b2428-a990-11e5-9e8f-a7a4f59a2db8.png)
* And then it displays the first Mapping screen
![image](https://cloud.githubusercontent.com/assets/6314950/11987225/2e9db9b8-a990-11e5-85c3-68e7deef2a85.png)
* So you have to click the Map Your Data button to get back to where you started (which is where the program should have put you in the first place after Save Settings)
* And it didn't save the setting change -- in this example I deselected State Province, but it is still displayed in the list
![image](https://cloud.githubusercontent.com/assets/6314950/11987240/7072c842-a990-11e5-983d-8c1ea6770636.png)

pipermerriam:
assigning to myself

RDmitchell:
assigned back to @pipermerriam based on @nicholasserra comments in the PR #715

RDmitchell:
I am closing this issue because @nicholasserra discovered that the functionality of the View/Hide button in Mapping isn't working. For now we are going to remove the View/Hide button since it doesn't work properly, and make a new issue to add it back with proper functionality

PR #715, opened 2016-02-03
Don't reload the page after changing what columns are being viewed

pipermerriam:
https://github.com/SEED-platform/seed/issues/635

### What was wrong.

When using the Show/Hide columns feature during the 2nd step of mapping, the page refreshed upon closing the modal which reset the view back to step 1.

### How was it fixed.

Removed the call to `location.reload()` at the end of the code which executes when the modal closes.  As far as I can tell everything works fine without this statement.

#### Cute animal picture

5567107626_6e39ecd9c1_z

### DIFF
--- a/seed/static/seed/js/controllers/buildings_settings_controller.js
+++ b/seed/static/seed/js/controllers/buildings_settings_controller.js
@@ -104,7 +104,6 @@ angular.module('BE.seed.controller.buildings_settings', [])
             //resolve promise
             $scope.settings_updated = true;
             $uibModalInstance.close(columns);
-            location.reload();
         });
     };
 


pipermerriam:
ready when tests are passing

RDmitchell:
@pipermerriam -- thanks -- I will test once @mmclark has put it up on seedtest.

nicholasserra:
Nice. My other thought was making those tabs each have their own url route, so the reload would toss it back to the same page at least.

nicholasserra:
Ugh @RDmitchell @pipermerriam this actually may cause some issues, as that reload is what updates the columns on all of the pages that are not the mapping page.

nicholasserra:
I think we should leave this as a DO NOT MERGE until I get all of these new columns preferences features in, so we can test all at once.

RDmitchell:
Just added DO NOT MERGE tag for now.

nicholasserra:
Closed because show/hide columns on mapping page is being removed.

pipermerriam:
@nicholasserra thanks for finding this.

RDmitchell:
@pipermerriam -- nice animal picture !!

Issue #687, opened 2016-01-27
Owners can downgrade themselves even when only member


nicholasserra:
Noticed that, on the organizations screen, I can downgrade my role from owner to member, even though i'm the only member, leaving the organization without any owners.

RDmitchell:
This doesn't seem good. May not have time to fix this for the Feb release, but I am going to set it to that milestone for now. Might want to make it a post-Feb release Hot fix -- we can assess in a week or so. thx for finding and reporting this.

pipermerriam:
Assigning to myself.

pipermerriam:
@RDmitchell most obvious path is to disallow this.  In the case that you are the last member of an organization this change would make it so that it was impossible to leave that org..  Not really sure what the best way to go about this would be.

nicholasserra:
I was thinking that's its kind of odd that you're able to change your own membership at all, in the members list. I can't think of a situation where that would be useful. Nobody should be able to upgrade themselves, and nobody would normally downgrade themselves, so my thought was to just make the row read only if row.user == yourself.

nicholasserra:
Specifically on the organizations -> members tab

RDmitchell:
I would like to think about this a bit and will try to put in a comment soon.

rebrownlbnl:
I could imagine a situation where the owner is passing ownership to someone
else in the organization, she might promote the other user to owner and
then demote herself to a member (or delete their use altogether). But that
still only applies when there is more than one user in the organization. So
it seems safe to have a rule that there must be at least one owner in an
organization, and the last owner cannot demote themselves.

nicholasserra:
The new owner could also just demote the old one in that first situation.

RDmitchell:
I believe the way the program works right now, you can't delete yourself, if you are the only owner in the organization. So seems like we should apply the same logic to this case, where the owner wants to downgrade their Role. 

**So can we just put in an error condition that doesn't allow the last owner of an organization to downgrade themselves to a status other than owner?**

But I would like to leave the rest in place for now, ie, that if there are multiple members, that their status' can be changed. 

In March, one of the big tasks is going to be looking at the Parent/Sub Org functionality, which is largely untested, probably broken in many places, and probably needs reworking to function the way users want it to. So i want to save any major changes to the current functionality until then.

nicholasserra:
Seems I can delete myself and downgrade.

[...]

PR #717, opened 2016-02-04
Piper/issue 687 dont allow organizations to end up empty

pipermerriam:
https://github.com/SEED-platform/seed/issues/687

### What was wrong.

When managing organization members you could perform actions that would leave organizations having either no owners, or no users at all.

### How was it fixed.

Restrict the ability to edit your own role or remove yourself from a group if doing so would leave the group with no owners or empty.

#### Cute animal picture

kitten_cheezburger


RDmitchell:
thx !

Issue #686, opened 2016-01-27
Export appears to be finished, but it isn't -- make progress bar show actual status


RDmitchell:
Export progress bar needs some work. 

The two states that it seems to represent is 0% when it is probably actually doing something, and 100% when it has probably not finished. 

Could we at least have the progress bar show something besides 0/591 buildings if it has started, and also not show that it has finished before it actually has. 

When exporting a few records everything seems to be working properly, but exporting larger number of records, the progress bar doesn't seem to be showing the true state of things

I will add a video and some sample data for this issue in my next comment
@nicholasserra -- I am assigning this to you.

RDmitchell:
In the video recording below, I didn't experience the "stuck at 0/591" that I had seen right before I made the video but I did see the problem where the progress bar reported that it was finished, but it actually hadn't made the file and didn't appear to be creating it (although I am not completely sure about that).
Here is a link to the google drive folder that contains both the video and the dataset
https://drive.google.com/a/lbl.gov/folderview?id=0B3fTKpZ9Dx7LZ0lZTzJKMG5vM2s&usp=sharing

RDmitchell:
@nicholasserra -- a few more data points. With that data file loaded, if I filter on Use Description for Multi, which results in 176 records, export works without issue. If I filter on Use Description for Office, which results in 293 buildings, the progress bar indicates that it is finished, but in reality it doesn't seem to be doing anything
![image](https://cloud.githubusercontent.com/assets/6314950/12632307/ec4de122-c50b-11e5-9606-b93bcd3ac841.png)

RDmitchell:
(didn't mean to close it)

RDmitchell:
I'm still not convinced that the export is actually working. I tried exporting 5700+ records, and let it sit for 1.5 hours, and it never finished (the export file was never created, and the dialog box never got to the state where it had the "close" button on it -- although the progress bar indicated that it had all the buildings (as shown in the screen shot above).

nicholasserra:
Maybe there's an issue on seedtest. I'll investigate on monday.

pipermerriam:
I'm starting on this (assigned to myself)

pipermerriam:
@RDmitchell I tried and wasn't able to reproduce this using the provided data set.  This is at least the 3rd time that there's been an issue which *if* I was able to easily access the error logging from seedtest I probably could have diagnosed and fixed it faster.

[...]

PR #738, opened 2016-02-10
Cleanup building exporter class.

pipermerriam:
This work was done during the investigation of #686 

### What was wrong

The `seed.exporter.Exporter` class was complex and did not follow some best practices.

### How was it fixed?

* Extracted methods that could be turned into vanilla functions.
* Changed file handling to use files as context managers.  This prevents file descriptors from getting left open which is really bad.
* Reduced the memory footprint for iteration over large querysets (I suspect this might be part of why exporting large building lists fails)
* Minor cleanup to how file extension is validated.

#### Cute animal picture

cat-in-reindeer-christmas-sweater

### DIFF
--- a/seed/lib/exporter.py
+++ b/seed/lib/exporter.py
@@ -20,6 +20,110 @@ from django.db.models.fields.related import (
 )
 
 
+def batch_qs(qs, batch_size=1000):
+    """
+    From: https://djangosnippets.org/snippets/1170/
+
+    Returns a (start, end, total, queryset) tuple for each batch in the given
+    queryset.
+
+    Usage:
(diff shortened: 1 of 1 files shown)


RDmitchell:
Great ! Thanks @pipermerriam -- will test when it's merged to seedtest !

rebrownlbnl:
Do any unit tests need to change for this? (did it _have_ unit tests?)

pipermerriam:
> Do any unit tests need to change for this? (did it have unit tests?)

It had unit tests and they all still passed after the refactor.  A brief look at the actual tests showed they do a decent job testing the overall functionality.

Issue #730, opened 2016-02-09
Delete buildings not working when filtering using labels


RDmitchell:
Deleting buildings does not seem to be working when you apply a filter using the "Filter by Label" function. 

Here is an example:

1. Import the standard PM sample data file (512 records)
2. Accept the default mapping
3. Go to Building List
4. Filter on Postal Code = 10102; 53 records are returned
5. Select all of those records with the upper left checkbox
6. From the Building Actions pulldown, select Add/Remove labels
7. Create a new label called Zip 10102, color gray, click Add, and Done so that label gets applied to those 53 building records
8. Remove the filter so that the 512 buildings are again displayed.
9. In the Filter by Label box, pick the "Zip 10102" label, so that 53 records are displayed.
10. Select all of those records with the upper left checkbox
11. From the Building Actions pulldown, select Delete Buildings
12. The program asks if you are sure you want to delete the buildings, you say Yes, it appears from the dialog box that they are deleted.
13. But they are still displayed on the screen.
14. Remove the label filter, and click the update filters button so that all 512 records are displayed
15. The records with zip = 10102 are still there (that's why the program reports that there are 512 records) -- they did not get deleted.

RDmitchell:
@nicholasserra -- am assigning this to you, as I am not sure if @pipermerriam still has time to work on seed. Feel free to give it to him if he does. Thx !

pipermerriam:
@RDmitchell 

This should have been fixed with https://github.com/SEED-platform/seed/pull/720 which was merged 3 days ago.  Do you know whether you encountered this issue on a testing site that had this code deployed to it?

RDmitchell:
I believe I tested this yesterday on seedtest, but let me test again and get back to you.

RDmitchell:
Looking at the time I made this issue, 21 hours ago, I was testing the current code on seedtest, yesterday afternoon. So I am still seeing the problem. Do you want me to set up an account that shows the problem and invite you as a member?

RDmitchell:
Let me know if you cannot reproduce this problem.

pipermerriam:
No, not yet.  I'm still investigating.

On Tue, Feb 9, 2016 at 1:14 PM, RDmitchell <notifications@github.com> wrote:

> Looking at the time I made this issue, 21 hours ago, I was testing the
> current code on seedtest, yesterday afternoon. So I am still seeing the
> problem. Do you want me to set up an account that shows the problem and
> invite you as a member?
>
> —
[...]

PR #741, opened 2016-02-11
Fix building deletion while filtering by label

pipermerriam:
https://github.com/SEED-platform/seed/issues/730

### What was wrong.

The `seed.views.delete_buildings` view was still using the *old* search logic so it didn't work when filtering by labels.

### How was it fixed.

Updated it to use the new search functionality.

#### Cute animal picture

tumblr_np90u2dxgu1qbjrrso7_500

### DIFF
--- a/seed/tests/test_delete.py
+++ b/seed/tests/test_delete.py
@@ -218,50 +218,6 @@ class DeleteViewTests(TestCase):
             10
         )
 
-    def test_delete_buildings_deselected_buildings(self):
-        """tests delete_buildings for de-selected_buildings"""
-        # arrange: standard case, try to delete all but 10 unchecked buildings
-        self._set_role_level(ROLE_MEMBER)
-        self.assertEqual(BuildingSnapshot.objects.filter(
-            canonicalbuilding__active=True).count(),
-            self.NUMBER_ACTIVE
-        )
-        ids = BuildingSnapshot.objects.filter(
-            canonicalbuilding__active=True
-        ).values_list('pk', flat=True)[:10]
-
-        # act
-        resp = self.client.delete(
-            reverse_lazy("seed:delete_buildings"),
-            data=json.dumps(
-                {
-                    'organization_id': self.org.id,
-                    'search_payload': {
-                        'select_all_checkbox': True,
-                        'selected_buildings': list(ids),
-                    }
-
-                }
-            ),
-            content_type='application/json',
-        )
-
-        # assert
-        self.assertDictEqual(json.loads(resp.content), {'status': 'success'})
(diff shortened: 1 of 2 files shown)


pipermerriam:
I need some help understanding this code.

```python
    if not body.get('select_all_checkbox', False):
        # only get the manually selected buildings
        selected_buildings = get_search_query(request.user, {})
        selected_buildings = selected_buildings.filter(
            pk__in=selected_building_ids,
            super_organization__in=orgs
        )
    else:
        # get all buildings matching the search params minus the de-selected
        selected_buildings = get_search_query(request.user, body)
        selected_buildings = selected_buildings.exclude(
            pk__in=selected_building_ids,
        ).filter(super_organization=orgs)
```

This code translates into this logic.

* If the *select all* checkbox is **not** checked.
    * delete all of the buildings that were selected individually.
* If the *select all* checkbox is checked.
    * delete all of the buildings except for the ones that were individually checked.  (THIS IS THE WEIRD PART)

This functionality seems wrong.  If the select all checkbox has been selected why wouldn't you delete all of the buildings?  If the idea was that you could use the *select all checkbox* to select all of the buildings and then deselect a few that you didn't want to delete, then this query is actually wrong.

Please advise.  @mmclark it appears you were the original author of this code so maybe you can shed some light.

RDmitchell:
@pipermerriam -- I agree with you -- if the select all checkbox is checked, the program should delete all the checked buildings. This should basically overwrite the original records that were checked by hand. I wouldn't think that @mmclark wrote this code, I would have thought it was original code from BE, but I guess I could be wrong.

nicholasserra:
Yeah that seems like some weird UI thing, where maybe there was a "delete all EXCEPT checked" functionality. I'd rewrite it to what you think is sane.

RDmitchell:
agreed

pipermerriam:
Ok, removing this functionalitiy and the associated test.  Lets see what travis says.

rebrownlbnl:
Could that logic have been used for the use-case where you want to delete all but a few buildings, so you check the "select all" checkbox, then manually unselect the few buildings that you want to keep?

pipermerriam:
> Could that logic have been used for the use-case where you want to delete all but a few buildings, so you check the "select all" checkbox, then manually unselect the few buildings that you want to keep? 

I think that was the intention but it would *unintuitively* keep the buildings you left checked (the unchecked ones would still be deleted).  Very poor UX and major violation of [least astonishment principle](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

RDmitchell:
agreed -- hadn't ever heard about the LAP, so thanks for the link !!!

Issue #747, opened 2016-02-12
Optimize slow get_organizations endpoint


nicholasserra:
The get_organizations endpoint at `app/accounts/get_organizations/` takes 3 minutes if you're a superuser on seedtest.

nicholasserra:
Going to wait on a seedtest data dump to fix this locally. cc @mmclark @RDmitchell

mmclark:
Shared via gdrive

nicholasserra:
Seems "sub_orgs" from the get_organizations payload isnt used anywhere in the app.

nicholasserra:
I lied, it's used, but in a weird way.

RDmitchell:
I don't know how things are set up in the code, but Sub Orgs do exist as a concept and should have specific functionality, even if the program isn't doing it right now. 

Maybe in the software structure, there is no difference between an org and a sub org, and it is just a matter of what organizations are associated with each other (?) that would make them the Parent or the Sub Org (?)

I wrote this doc (in 2014) when I was trying to figure out how Organizations worked (based on the state of SEED at that time), including the Parent/Sub Orgs relationship, which I think doesn't have much content in it. worked in the version of SEED that existed then. 

https://drive.google.com/a/lbl.gov/file/d/0B3fTKpZ9Dx7LOHpfanV0VktyN0U/view?usp=sharing

I want to revisit the Parent/Sub Org feature in March or April to see what is currently implemented in the program, what we think the feature should really be and how to get it to that point.

RDmitchell:
Might be worth documenting, possibly in this issue, or in a new issue (?), how it is currently used?

nicholasserra:
For the purposes of this ticket, no sub org stuff was changed.

nicholasserra:
Reassigning for merge and testing.

RDmitchell:
took slightly over a minute to get the list of organizations.

PR #748, opened 2016-02-13
Issue #747 - Speedup get_organization endpoint

nicholasserra:
Deleting org buildings in the admin view feels very slow, this may be one reason. 

#### What's this PR do?
Speeds up the get_organizations endpoint. This was taking up to 3 minutes on seedtest. Locally, this speedup brought a 16 second response time down to 1 second. This mainly removes a bunch of count() queries that were able to be combined.

Refs #747 #684

### DIFF
--- a/seed/views/accounts.py
+++ b/seed/views/accounts.py
@@ -50,33 +50,33 @@ def _dict_org(request, organizations):
     orgs = []
     for o in organizations:
         # We don't wish to double count sub organization memberships.
-        owners = [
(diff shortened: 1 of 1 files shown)


RDmitchell:
Took slightly over a minute to display the organization list at the bottom of the superuser Admin screen. So maybe this is an improvement? Better than 3 minutes.

nicholasserra:
Yeah, it's a start.

PR #754, opened 2016-02-15
Switch to supported django-storanges pypi package

nicholasserra:
Switch to a pypi supported and current version of django-storages. The previous one was 3 years unmaintained and was installed via mercurial.


RDmitchell:
any way to test this ?

nicholasserra:
I think there was actually an issue with this that @mmclark resolved in #755

PR #755, opened 2016-02-16
update boto version

mmclark:
Updates in the previous pull requests required a new version of the python boto library (which was quite out of date).


nicholasserra:
Merged develop in.

Issue #752, opened 2016-02-15
Figure out Organization vs SuperOrganization usage


nicholasserra:
There seem to be two concepts of organization in the application. One is the traditional Organization model, which is in `seed.lib.superperms.orgs.models`. 

The second is from the package `django-organization` via `from organizations.models import Organization`. This package is included in INSTALLED_APPS, and is only used on the ImportRecord model as ImportRecord.organization. ImportRecord also has ImportRecord.super_organization.

My hunch is that ImportRecord.organization is unused and we can remove the entire package.

nicholasserra:
This is turning out to be a blocker for 1.8, so going to take this on now.

nicholasserra:
Any ideas from @nllong @mmclark @RDmitchell @StephenCzarnecki as to why django-organizations Organization model is wired to this ImportRecord model here?

https://github.com/SEED-platform/seed/blob/develop/seed/data_importer/models.py#L110

Almost finished removing it, but wondering what the deal was. Also would be interested to see if the db tables `data_importer_importrecord_organization` and `organizations_organization` are empty, as that's the for sure tell if it's unused.

rebrownlbnl:
Yes, agree we don't want two different treatments of organizations.

I don't know the history about why a different package is used (or at
least mentioned) for file importing.

Consistency is good; as always, please try to improve the docs with
whatever you learn.

nicholasserra:
So this will be a multi step migration patch. First I need to make a migration to remove the field in ImportRecord. Once that migration is deployed on all environments, I need to do a migration reset on the data_importer app to clean up history.

I'll need to figure out how these migrations will be deployed with @mmclark. The first migration needs to make it to production before the reset happens. So not sure if that requires it be in the feb release, that might be easiest.

RDmitchell:
So this is all going to happen after the Feb release, in terms of getting these changes into develop and master, right? 

It seems like we will want to do a bunch of monkey (and unit?) testing on seeddev or seedtest (after the Feb release) when this change is in place.

nicholasserra:
I need @mmclark's opinion on that. We can't reset migrations on the data_importer app until the new migration has been ran in all environments. He can probably do a hotfix to production.

RDmitchell:
OK. We can definitely move this to a hotfix after the Feb release (which we might be able to release before the end of the month, depending on results of testing)

StephenCzarnecki:
@nicholasserra 
[...]

PR #757, opened 2016-02-16
Issue #752 - Remove ImportRecord.organization

nicholasserra:
This is the first step in a series of patches to remove the unused ImportRecord.organization field and ultimately the entire django-organizations package. 

This step just removes the field. This migration needs to be ran on all environments before the next step, which is to reset migrations for the `data_importer` app.

Refs #752


nicholasserra:
@RDmitchell @mmclark I'm going to mark this as Feb release because this needs to get onto production as soon as possible to continue Django 1.8 work

RDmitchell:
sure.

Issue #723, opened 2016-02-05
Investigate refactoring search_service and building_service columns methods


nicholasserra:
`search_service.generate_columns` and `building_service.create_column_array` are very similar, aside from a check for columns being selected. Most likely these can be rolled into one method to avoid confusion. 

Currently only the matching controller calls the building_service method.

RDmitchell:
@nicholasserra -- any manual tests I should do for this?

nicholasserra:
Nah, just a small refactor.

PR #775, opened 2016-02-19
Issue #723 - Refactor FE columns methods into single method.

nicholasserra:
`search_service.generate_columns` and `building_service.create_column_array` were very similar. `building_service.create_column_array` was only used during matching.

#### What's this PR do?
Combine these methods into one, using the `search_service.generate_columns` method. Call the search_service method on the matching page instead. Removes duplicate code.

Refs #723


RDmitchell:
@nicholasserra -- any manual tests I should do for this?

nicholasserra:
Nah, just a small refactor.

Issue #751, opened 2016-02-15
Switch SESSION_ENGINE to cookie sessions


nicholasserra:
Right now we're using the cached db session engine. We should switch to signed cookie sessions to eliminate the need for the session table. It's also faster.

This will terminate active sessions when it is deployed, so I want to get @mmclark and @RDmitchell input.

`SESSION_ENGINE="django.contrib.sessions.backends.signed_cookies"`

Also turn on `SESSION_COOKIE_HTTPONLY`
https://docs.djangoproject.com/en/1.7/ref/settings/#std:setting-SESSION_COOKIE_HTTPONLY

RDmitchell:
@nicholasserra -- I need a bit more explanation about what the implications of this are, ie, "This will terminate active sessions when it is deployed". I don't really know what that means. 

Also, @rebrownlbnl -- could you weigh in on this? thx

nicholasserra:
When the code is released to the server, users who are currently logged in will be logged out.

nicholasserra:
They'll have to log in again.

RDmitchell:
ah, so just a one-time thing. I think that's fine. We can do the switch late in the day or in the evening, and I can tell users when we are going to do it.

rebrownlbnl:
Are there any other implication from changing the way we manage sessions?

-------------------------------------------------------
Rich Brown
Building Technology and Urban Systems Division
Lawrence Berkeley National Laboratory
Office: (510) 486-5896
Mobile: (510) 502-5968
REBrown@lbl.gov
www.lbl.gov/~rebrown
-------------------------------------------------------

nicholasserra:
No, I don't think so. A major positive is not having them stored in the database. The database table can pile up quickly if a cron job isnt clearing those sessions.

nicholasserra:
Just need to verify you can log in.

RDmitchell:
@nicholasserra -- Yes I can log in. I will close this issue

PR #777, opened 2016-02-20
Issue #751 - Django cookie sessions

nicholasserra:
Change the sessions to use signed cookies. Refs #751


RDmitchell:
I verified that I could log into seedtest both as a superuser and as a regular user.

Issue #492, opened 2015-11-30
Filtering tweak to filter out particular strings


RDmitchell:
From Baptiste #367 --  For the !"" it works for empty values, but it might be interesting to be able to filter out a given string also (for example filtering out a city or a use description by having "! Commercial").

nicholasserra:
Do you think this should be an exact or inexact match? aka `!"Commercial"` would show all results that do not have the exact text "Commercial" in the field. 

Compared to a query that shows all results that do not contain "Commercial" in some way.

RDmitchell:
could it be:

!commercial - not an exact match
!"commercial" -- exact match

Or does the thing being filtered on need to be enclosed in quotes?

nicholasserra:
Should be fine, that works.

nicholasserra:
Assigning to @RDmitchell for merge and testing.

RDmitchell:
tested, works, closing

PR #789, opened 2016-02-24
Issue #492 - Exclude filters

nicholasserra:
#### What's this PR do?
Adds ability to filter by excluding words.

Refs #492

### DIFF
--- a/seed/search.py
+++ b/seed/search.py
@@ -183,6 +183,8 @@ def filter_other_params(queryset, other_params, db_columns):
             case_insensitive_match = search_utils.is_case_insensitive_match(v)
             is_numeric_expression = search_utils.is_numeric_expression(v)
             is_string_expression = search_utils.is_string_expression(v)
+            exclude_filter = search_utils.is_exclude_filter(v)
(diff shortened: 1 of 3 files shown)


rebrownlbnl:
Is there any guidance for the users about how to use this functionality? (or any of the other filtering capabilities we've added in the last couple months, for that matter)

nicholasserra:
I don't know if there are docs around this at all. @RDmitchell probably has an idea.

RDmitchell:
I have a doc on the seed website about filtering, but it would be nice to incorporate that info into the program with help icons or something. This is something I will probably give to @danielmcquillen  to work on this Spring, ie, getting help information better incorporated into the program

RDmitchell:
this page has a link to the current "filtering" documentation
http://seedinfo.lbl.gov/detailed-support

RDmitchell:
@nicholasserra -- what did you end up implementing for this?

nicholasserra:
To exclude a piece of text, use an exclamation point at the begging with no quotes. aka `!exclude`

nicholasserra:
Or wrap in quotes for exact exclude `!"ExClUdE"`

Issue #809, opened 2016-03-01
Investigate necessity of Schema model


nicholasserra:
Investigate the necessity of the Schema model on the BE. On production, there is only one row in the DB. The `get_column_mapping_suggestions` endpoint does complex logic to get the Schema for each column, but ultimately, they're all the same.

Not sure if this can be removed and hardcoded, or if it's in there for future use.

cc @RDmitchell

RDmitchell:
What is the functionality of the schema? I am wondering if it was put in place for a need that we know about but having seen the need to implement yet?

nicholasserra:
I'm not sure yet. All i've gathered so far is that there's only 1 row in the production database for it, and that the codebase doesn't create them at all. From what I can see, I can hardcode its value right now without any issue. I need to spend a few hours digging into it.

rebrownlbnl:
This may have been a way to store field mappings in the database, rather than in text files, which is what we do now. Is that possible?
Where is this in the code?

nicholasserra:
`seed_schema` table. Only one row, and the only content is a column called `name` with content `BEDES`.

nicholasserra:
Basically, it seems that every Column object is supposed to have a schema. But every column just points to this one Schema, BEDES.

rebrownlbnl:
okay, I know what this is doing. This is how SEED originally know whether the little green check mark should show up in the mapping screen, to indicate if a field came from "BEDES" or not. (BEDES is a standard data spec for building-related data).
At that time, BEDES was the only schema we had, which is why all the fields say BEDES.
I think we deprecated that approach, right?
If we are no longer showing the check mark we should stop doing that query of the schema table.

We've reconsidered our whole use of BEDES in SEED, so if and when we re-implement this it needs to be redesigned from the ground up.

RDmitchell:
Good sleuthing !!! 

And I agree that if we need to reimplement it (which we might at some point) we would redesign it from the ground up. 

So, it can go !!!

nicholasserra:
Cool, so i'll remove the model and any code around it. Should speed some things up too.

RDmitchell:
yup, go for it !

RDmitchell:
don't know if there is any monkey testing I can do regarding this. I haven't hit any issues in mapping that I know of. So am closing.

Issue #803, opened 2016-02-29
Mapping screen is very slow to display


RDmitchell:
I know a few people have looked into this here and there, but we need to figure out, and possibly fix, the extremely slow display of the mapping screen. 

In a very non-robust timing test, it took the program 30 seconds to display the mapping fields from a data file (imported as a spreadsheet) with 34 fields and approx 980 records. I know it's not the end of the world, but it leads to the general impression (which is generally true) that the program is very slow. 

@nicholasserra -- I am assigning this to you to investigate what the issues are that are making this display slow, and then we can decide whether to fix it. Thx

nicholasserra:
Which exact step is this? Is this while the page says "Please wait while your data is loaded" or similar?

RDmitchell:
Sorry, I meant to add a screen shot.

It is the generation of the first mapping screen
First you see this
![image](https://cloud.githubusercontent.com/assets/6314950/13408740/46de013e-dee4-11e5-842d-940bb99da86e.png)
Then it takes several seconds for this screen to appear
![image](https://cloud.githubusercontent.com/assets/6314950/13408852/c32d54ce-dee4-11e5-8de6-d1947f762349.png)

nicholasserra:
Locally with the covered buildings sample file this endpoint is going 15,000 queries hah. So i'm going to spend some time figuring out why.

RDmitchell:
Yes, I am not surprised. @StephenCzarnecki -- do you have some insight into this? I can't remember if you were looking into this or not?

RDmitchell:
It is very fast now to show the mapping screen -- WONDERFUL !!!

PR #810, opened 2016-03-01
Issue #803 - Fix slow get_column_mapping_suggestions

nicholasserra:
`get_column_mapping_suggestions` endpoint during mapping was taking upwards of 30 seconds on seedtest.

#### What's this PR do?
Optimize the queries for this endpoint. Currently this endpoint does around 15,000 queries with a small input file. This PR brings that number down to around 40. Response time for this endpoint is current around 25 seconds locally. This PR brings it down to around 500ms. 

Also added some organization membership checking for security.

Added issue #809 to investigate whether all this logic is necessary for getting the Schema info.

Refs #803

### DIFF
--- a/seed/views/main.py
+++ b/seed/views/main.py
@@ -5,6 +5,7 @@
 :author
 """
 
+import copy
 import json
 import logging
 import datetime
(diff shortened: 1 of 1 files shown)


RDmitchell:
@nicholasserra -- just curious why it was doing 15,000 queries before and now it only has to do 40 !!!
I suspect there are a lot of places in the code where this is happening, which affects performance. Thanks -- can't wait to test it !

nicholasserra:
There was some code that was making queries in an easy and intuitive way, but also in a way that was terrible for performance. Basically for every column available, it was doing a couple queries to get schema for that column. I changed it to grab all schemas in one query and then match with python.

Issue #804, opened 2016-03-01
Sorting Building List affected by blank data in JSON extra_data field


RDmitchell:
With the Show All setting active, if you sort on a field that is a real field in the Building Snapshot table (not in extra_data), the program reports and displays the correct total number of records. 

However, sorting by a field stored in the extra_data JSON blob field, the program only shows records with data in those fields, although it should be showing all the records, whether there is data in the extra_data field or not. 

Use the data from #799 and you can see the issue (you can probably see if by just loading the building list, but I loaded and matched the two data files):

Import Building List -- 10 records
Map a few of the fields to make sure they will end up in the Extra Data field.
![image](https://cloud.githubusercontent.com/assets/6314950/13414637/0abccb66-df09-11e5-94df-e7a798edcff0.png)

Upload the PM data -- 3 records all of which match to a building
again make sure that a few of the fields will end up the Extra Data field
![image](https://cloud.githubusercontent.com/assets/6314950/13414699/7166bd18-df09-11e5-9583-3b5c0477a2f1.png)

Go to the Building List
Sort by PM Property ID -- the sorting will result in 10 records being displayed, including the 7 records from the building data that do not have matching PM records, and so have PM Property ID blank
![image](https://cloud.githubusercontent.com/assets/6314950/13414765/ef0f1030-df09-11e5-98b4-6bea96cc67f9.png)

Sort by one of the "extra data" fields defined in the PM file, and the program reports and displays only 3 records (even though "Show All" is selected). It should show all 10 records with the 7 building records not matched to a PM file showing that field as blank
![image](https://cloud.githubusercontent.com/assets/6314950/13414807/3f826742-df0a-11e5-861a-20faed5e7782.png)

We need to get this fixed and put up as a hot fix as soon as possible. I thought that we had fixed this already, but when I look at old issues about this it isn't clear that it was ever fixed. If you filter on "sorting" for closed issue you will see lots of references to this problem.

nicholasserra:
I'm able to reproduce locally. On it. Another good candidate for selenium tests.

RDmitchell:
I was thinking the same thing.

nicholasserra:
This is an edge case that only happens when you are sorting by an extra data field that is only present on one of the files. I'm going to try to tweak the json query that handles this.

RDmitchell:
Right !!! I think that is what was discovered in one of the other many issues about this problem. If you could tweak the query to handle this edge case, that would be great. Then I can close all those other issues !

PR #812, opened 2016-03-01
Issue #804 - Fix for extra data order by results

nicholasserra:
#### What's this PR do?
Put in a fix for how a json query is handled when doing an order_by. This changes the behavior of the json_query, but ordering is only done in select situations, so i'm hoping the limited scope of this ensures we don't have any regressions.

Refs #804


RDmitchell:
I just labeled this as a March hotfix, so if it's not too much trouble, can it get merged to the 1.4.0 branch? thx !

nicholasserra:
Yes I'll have to close and reopen this. Any future hotfixes we should label right away.

PR #819, opened 2016-03-03
WIP selenium travis

nicholasserra:
WIP


.travis.yml
```diff
@@ -39,3 +41,7 @@ script:
 
 after_script:
   - coveralls
+
+addons:
+  # Sauce credentials in Travis env vars
+  sauce_connect: true
```
pipermerriam: Here's the problem

tox.ini
```diff
@@ -5,17 +5,15 @@ max-line-length = 100
...
+    functional
 skipsdist = True
 
-[testenv]
-basepython=
-  py27: python2.7
+[testenv:python]
 commands=
-  - cp -n {toxinidir}/config/settings/local_untracked.py.dist {toxinidir}/config/settings/local_untracked.py
+  cp -n {toxinidir}/config/settings/travis.py {toxinidir}/config/settings/local_untracked.py
```
pipermerriam: I would prefer that the `tox.ini` file not have travis specific stuff if possible so that it can be used as a local test harness.

.travis.yml
nicholasserra: :clap:

tox.ini
nicholasserra: Fixed

Issue #818, opened 2016-03-03
Mapping pulldown list is not displaying correct choices


RDmitchell:
When displaying the choices in the pulldown list (by typing a letter or word) the program is not displaying the correct choices.

Importing and try to map the standard sample covered buildings file, if I type "Tax" I should see a choiced called "Tax Lot ID"; instead I am seeing tax_lot_id
This holds for all the other choices that might be coming from the building snap shot fields (?) -- I don't know how the list of choices is generated. But something that was changed has broken it. 

This happens even with a "virgin" account where there is no history of previous mappings.

Here is what I see now
![image](https://cloud.githubusercontent.com/assets/6314950/13509138/d7f85d66-e13e-11e5-9e76-92172e1bdae5.png)

Here is what I should see (what I see on seed production)
![image](https://cloud.githubusercontent.com/assets/6314950/13509410/1f4011ae-e140-11e5-929e-2b48237aacfe.png)

RDmitchell:
@nicholasserra -- am assigning this to you. Might have something to do with the sped-up mapping ? Not sure

RDmitchell:
This only needs to be fixed if we put the possibly related sped up mapping fix on production. We could decide that none of the hot fixes are really that urgent.

nicholasserra:
Confirmed my speedup broke this. Working on a fix.

nicholasserra:
The issue here is simply the order of the columns in the payload. Some undesirable code is assuming the first 37 columns should be changed to be more user readable. Just need to fix this output.

RDmitchell:
Great, thanks.

RDmitchell:
Seems to be fixed, Thanks @nicholasserra

PR #822, opened 2016-03-04
Issue #818 - Better logic for mapping column typeahead & regression fix

nicholasserra:
#### What's this PR do?
Fix a regression dealing with the mapping page column typeahead.

Refs #818


RDmitchell:
Tested, seems fixed.

RDmitchell:
tested on seedtest

Issue #797, opened 2016-02-26
Building Detail Page UI Refinements


danielmcquillen:

danielmcquillen:
Building detail:
- Field column should have min size
- Max width with truncate on data columns
	- Include some kind of convention to show whole title
	- Some kind of combination of rollover and click to show modal
	- Do some different versions in prototype
- Better scrollbars
- We want frozen headers

Building list:
- Store width settings of columns

RDmitchell:
@danielmcquillen -- see #793 -- we could fold that into this issue, or keep them both. Up to you.

RDmitchell:
@danielmcquillen -- the Building List has the same issue of needing fixed field headers when you set the Display to > 10 records. When you scroll down through the records, you lose the field headers.

danielmcquillen:
Robin, the more recent version of angular-bootstrap, a library we use for a lot of our interactivity, is on version 1.2 and we're still using 0.14. We need the more recent version to get an efficient way of hiding the popover (we're planning on table cells) when the user clicks off anywhere else on the screen.

However, when I update to the latest angular-bootstrap, other elements in the site break, such as the Building Actions pulldown on the Building List page.

I'll look into it a bit more, but we might want to postpone popovers in this issue, finish the other parts, then create a new issue to update the relevant bower dependencies, since we want to keep these up-to-date anyways.

RDmitchell:
sure, we can put this together in any order that makes sense to you. Feel free to add whatever issues you need to, and indicate an order (using Priority or whatever else makes sense to you).

danielmcquillen:
@RDmitchell aha. Actually figured it out. As you'd expect, moving to angular-bootstrap 1.2 from 0.14 has a few required update steps...in this case it was adding 'uib-dropdown-menu' class to dropdowns. 

Shall I go forward with the introduction of angular-bootstrap 1.2 and bootstrap 3.3.6, and then we'll just have to really test the UI when considering a PR for this issue? It might be good to get this library up to the latest if we have time and scope within this issue.

RDmitchell:
I think we should update bootstrap as part of this issue if you think that makes sense. 

You might just start a Slack channel called GUI and say that's what you plan to do, reference this issue, and see if anyone has anything to say about it.

danielmcquillen:
Done.

RDmitchell:
Assigned to myself for testing

danielmcquillen:
[...]

PR #825, opened 2016-03-06
797 building detail page UI

danielmcquillen:
#### What's this PR do?
Updates the building detail page to have resizable columns. Also updates table cells to show ellipsis for text values, and a popover when clicked (so that text can be selected and copied).

To get the popover-trigger directive -- as well as to just keep us up-to-date in our dependencies --this PR also updates angular-bootstrap to 1.2.1. Therefore, developers will need to run a bower update command once this PR is merged. 

A side-effect of the angular-bootstrap update is that existing pulldown break, so this PR also adds in missing uib-dropdown-menu directive as a remedy. (Still looking for other issues caused by update.)

#797


danielmcquillen:
Build is failing with a "DoesNotExist: OrganizationUser matching query does not exist" ... I see this in another PR as well. Might be an upstream issue in develop.

RDmitchell:
@danielmcquillen -- do you think you can resolve this, or should I give it to someone else?

danielmcquillen:
Closing this PR and redoing in new PR.

PR #833, opened 2016-03-08
Timeseries Updates

wanghp18:
- Trim user input at front end
- Show energy type instead of cost in monthly energy data list
- Finer and monthly timeseries data, GreenButton request info are only requested when user goes to building's energy tab.


nllong:
looks nice. thanks!

PR #843, opened 2016-03-10
Timeseries Improvement

wanghp18:
Updates:
- Record very first and very last finer timeseries data's timestamp of each building in KairosDB, put the 'meta data' into unused timestamp range. Very first timestamp is not used for now, maybe used in future. Very last timestamp is used to narrow down the query time interval for a shorter response time.
- To check if KairosDB insertion data is saved, do insert-query loop until query returns at least same amount of data as inserted, a unique tag is created in each batch insertion for this purpose.


nllong:
:+1

Issue #861, opened 2016-03-19
Program can't load labels when semicolon is used for filter


RDmitchell:
Doing a test on a large dataset, filtering on semicolon in Tax Lot ID, returned 265 records, tried to add a label to those records, and the program wasn't able to load the labels. 
![image](https://cloud.githubusercontent.com/assets/6314950/13895965/34723b24-ed39-11e5-8ede-80458f422f8d.png)

I will make a video with the steps and the data.

@nicholasserra -- am assigning this to you and can make you an owner of the org on seed where I am seeing this issue.

RDmitchell:
Here is a link to the video
https://drive.google.com/a/lbl.gov/file/d/0B3fTKpZ9Dx7La0gxalprOTRaZ0E/view?usp=sharing

nicholasserra:
@RDmitchell thanks for the video! Would be helpful to be added to this org too so I can check it out on the server. I'll try to reproduce locally too.

nicholasserra:
@RDmitchell which files were all imported for this? I only imported the new york file and didnt get any semicolons in tax lot ids, so I think I need to do some specific matching to get this scenario.

RDmitchell:
@nicholasserra -- I will add you to the org, and send you documentation about how to upload, map and match the files on your development environment.

nicholasserra:
Perfect thank you

nicholasserra:
https://app.getsentry.com/lawrence-berkeley-national-lab/seed-platform-django/issues/120160889/

filter on semicolon and add label

RDmitchell:
I am still seeing this issue on seedtest with the large dataset

In this particular case, the total # of filtered records is 867, and I am displaying them 100 records at a time, selecting several (by hand by checking the checkbox on the left). I was displaying the 201-300 block of data when this happened. 
![image](https://cloud.githubusercontent.com/assets/6314950/14653611/9de80b20-062e-11e6-8c2c-5a60e148c75b.png)

RDmitchell:
I get the error message right away after clicking the Building Actions, Add/Remove Labels option

RDmitchell:
This is working to generating labels for those records, but it seems that when I then filter by the label that labeled the records with semicolons in the Tax Lot ID field, the program will not export those records -- seems to be hanging on 0/265 buildings. 

I will make a new issue for this, and test other imports as well to see if it is a problem with exports in general or just this particular case. 

[...]

PR #867, opened 2016-03-22
Issue #861 - Use a custom $http serializer to encode semicolons.

nicholasserra:
Semicolons inside of JSON inside of GET params were not being urlencoded by angular. Angular is doing this on purpose for some situations.

#### What's this PR do?
Adds a custom params encoder to use where necessary. This may be an issue in many places, so it's reusable. This is also temporary, as I plan on opening a PR on Angular to make this behavior less strict.

Refs #861 

#### Definition of Done:


nicholasserra:
Refs https://github.com/angular/angular.js/issues/6140

mmclark:
Loaded on seedtest

PR #870, opened 2016-03-24
Celery logging for sentry

nicholasserra:
There are several issues that are probably throwing errors in celery that we can't see.

#### What's this PR do?
Get Sentry working with celery.


mmclark:
Loaded on seedtest

Issue #567, opened 2015-12-15
Add git $Id:$ keyword to display the sha of the current version on the About screen


axelstudios:
This will help with identifying which git commit seedtest is currently serving.

axelstudios:
This is fixed with #875 - the solution is a little different than I originally intended, but the outcome is the same (version is in package.json, About page shows version and git sha)

axelstudios:
@RDmitchell I think this is ready for testing and closure.

RDmitchell:
so what type of testing would be useful?

RDmitchell:
look at about screen and it should show the git commit #

RDmitchell:
This may not be testable on seedtest because it doesn't have git -- @mmclark ??

mmclark:
seedtest does have git.  That's how I pull down updates from github.  I see the /about/version/ call is returning a 500.  I'll check into it.

RDmitchell:
Great, thanks !

axelstudios:
[The command](https://github.com/SEED-platform/seed/blob/f7caf8cb4236f6ad77495decc32e1b4c16269ed8/seed/views/main.py#L162) it's attempting to run is `git rev-parse --short HEAD` from the working directory.

mmclark:
The issue is resolved.  It was a problem with the way I was setting the PATH environment variable in the upstart script.

axelstudios:
Looks good!

RDmitchell:
OK, I'm closing this issue. Thanks all !

PR #875, opened 2016-03-28
Issue #567 - Tons of improvements

axelstudios:
More details coming..

This pull request adds support for ng-pluralize (smarter plural text [e.g. "1 Project"]), new dependency locations, easier (and portable) development environment configuration, dependency updates, tons of js lint cleanup (with added eslint), fixed extra scrollbars, and more.

#### Workflow changes
- The SEED version number (e.g. 1.4.0) is now located in package.json, not buried and hardcoded in about.html

#### Developer changes
- Node must be installed system-wide (node isn't used by seed, but it's required for installation of some npm dependencies)
```bash
curl -sL https://deb.nodesource.com/setup_5.x | sudo -E bash -
sudo apt-get install -y nodejs
```
- A custom (legacy) node.js build is no longer necessary, and node_modules is now located within the seed directory
- Dependencies are correctly managed with package.json and bower.json
- ~/local/ can be deleted
- The following lines can be removed from ~/.bashrc
```bash
export PATH=$HOME/local/bin:$PATH
export NODE_PATH=$HOME/local/lib/node_modules
```
- Install the dependencies (if global dependencies are missing, or if this is the first time running, the script will request root access to install bower, grunt-cli, and gulp-cli globally)
```bash
cd ~/seed
```

#### Related Issue
#567
#### Screenshot
image


nllong:
tested with docker and local (OSX). No issues. All the tests passed. Monkey testing seemed successful. 

Running install_javascript_dependencies.sh on OSX required the root password. I think we should not have that happen if at all possible.

axelstudios:
That's easily avoidable - I added it to simplify installation of the three dependencies that need to be available from the command line:
https://github.com/SEED-platform/seed/blob/workflow-improvements/bin/install_javascript_dependencies.sh#L5-L14

All we need to do is add some notes to the developer instructions to install those three dependencies before running this script, and the script can be changed to report on whether or not it found them (and not attempt installation)

nllong:
Also, I just hit rebuild on the PR tests and it seemed to pass. not sure why it failed the first time around.

rebrownlbnl:
yes, that sounds good.
probably also good to put a comment in that shell script about the dependencies assumed to be installed.

danielmcquillen:
Just ran the above steps and all worked well on my dev VM. Now testing the app locally...

danielmcquillen:
There are still 14 tests failing, but this could be due to upstream changes in develop.

nllong:
@danielmcquillen  are you running the js tests locally -- if so, are we not running these on travis?

nllong:
@danielmcquillen -- can you enable the js tests on travis?

RDmitchell:
@nllong -- if we implement the javascript tests in the CI world, what are we testing? Would it help us find some of the UI problems we might be having? 

I am trying to figure out what priority this should have. I will find out from Daniel what the level of effort would be to implement the tests, and figure out if this is something he should be working on now, or whether we can put it off til later. In my mind, the highest priority right now is to get a handle on the data model refactoring, and Daniel is pondering some of the UI implications of that.

danielmcquillen:
@nllong yes I always just used the url to view these Angular-based tests locally. I'm not sure how to hook this into Travis but could investigate. I'll wait to hear from @RDmitchell

danielmcquillen:
@axelstudios I think moving to Lodash from Underscore has caused some things to break in the FE, e.g. _.pluck is not implemented in Lodash.

axelstudios:
Underscore was actually a fairly recent addition (which shouldn't have been since lodash was already there) - the issue is that _.pluck was deprecated with the update to the latest version of lodash (_.map is the replacement since it takes the same arguments)

danielmcquillen:
I think @pipermerriam added underscore originally, and then I started using it after I found out how awesome it was (didn't realize Lodash was already there). Piper, was there any function in Underscore that you were using that might not translate well to Lodash? Or should all incompatibilities be trivial, i.e. _.pluck to _.map?

pipermerriam:
@danielmcquillen I think migrating to Lodash should be just fine.  You may need to make some minor changes given that the latest lodash doesn't have the same API as underscore anymore.

mmclark:
Loaded on seedtest

danielmcquillen:
@mmclark When this is loaded onto seedtest, does it automatically update the FE dependencies as described by bower.json?

Issue #813, opened 2016-03-01
Matching shows 0% progress, but is actually matching (on a large dataset)


RDmitchell:
Using a large dataset (12,000 - 13,000 records in both the Building List and the Portfolio Manager data), the program is slowly matching the records, but not showing any progress in the matching screen. 

Here is what the matching screen shows:
![image](https://cloud.githubusercontent.com/assets/6314950/13446023/fe8e0794-dfc4-11e5-88bd-a9e6f44d751f.png)
The Chrome Developer tool shows now errors, with the Progress post being returned happily
![image](https://cloud.githubusercontent.com/assets/6314950/13446041/20386682-dfc5-11e5-9768-8b9c0f273d2b.png)

In order to check whether records were indeed being matched, even though the program wasn't showing any progress, I logged into the same account with another browser window (just happened to be Mozilla but could have been another Chrome window).

Looking at the building list, and setting the program to Show Matched, the program initially showed that approximately 5,000 records had been matched, and they showed that the PM Property ID was there, etc, so the PM data had actually been merged. Refreshing the browser now and again, the program is still working away, matching the records. 

Refreshed the browser a bit ago and the program showed 9400+ records matching
![image](https://cloud.githubusercontent.com/assets/6314950/13446115/802dac3c-dfc5-11e5-9ee1-ee350e454045.png)

Then refreshed again, and the program is reporting 10,700 records matched. So it is definitely chunking through the data.
![image](https://cloud.githubusercontent.com/assets/6314950/13446136/a5c8da98-dfc5-11e5-8eb8-6e484e962125.png)

And still reporting 0% progress in the original window that is doing the matching
![image](https://cloud.githubusercontent.com/assets/6314950/13446150/ce94a01a-dfc5-11e5-8a21-d6b5301a8a89.png)

I will put a link to the data to replicate this.

RDmitchell:
In the end, it finished matching (still reporting 0%) and found 11,718 matches, out of 12,752 records, which is actually really great. So we just need to figure out how to get the program to report the progress better to the user. 
![image](https://cloud.githubusercontent.com/assets/6314950/13446214/6a679880-dfc6-11e5-87f1-8398bc8c523b.png)

RDmitchell:
It actually doesn't have to be that big of a dataset -- it happened on data where there were only a few thousand records being matched.

RDmitchell:
[...]

PR #877, opened 2016-03-29
Issue #813 - Progress bar fixes

nicholasserra:
Progress bar for uploading and mapping wasn't displaying correct values, was causing false page changes when jobs weren't completely done.

#### What's this PR do?
Tweak the way progress cache is incremented and simplify the logic for chunking out progress. 

Refs #813 

#### Definition of Done:


mmclark:
loaded on seedtest

Issue #841, opened 2016-03-10
Clean up / remove sessionStorage search filter stuff from cleansing_controller


nicholasserra:
For some reason there's a bunch of code around filtering and saving filters in the `cleansing_controller` file, even though this file is only used to show the `data cleansing results` modal. 

For example: https://github.com/SEED-platform/seed/blob/develop/seed/static/seed/js/controllers/cleansing_controller.js#L74-L101

... looks like it was pulled right from the `search_service`.

It doesn't seem like this is used, but I may be wrong. Hoping maybe @axelstudios can shed some light on this.

axelstudios:
Yep, it's used.  As I recall, the issue with directly using the search_service was that any changes overrode filtering for the other tables, and the cleansing dialog needed to be handled independently.  It copies the search_service, but then changes a few things - for instance, all other tables filter and sort by making HTTP requests, whereas the cleansing dialog does everything client-side.

Using sessionStorage doesn't matter to me, that was simply an artifact of the search_service.

nicholasserra:
I guess I was confused because I don't see any searching or filtering in the cleansing dialog? 

via Hiptop

axelstudios:
Here's a screenshot with both sorting and filtering active:

![image](https://cloud.githubusercontent.com/assets/411466/13709646/4fd3a9c4-e772-11e5-8e32-120b0dcd072a.png)

nicholasserra:
@axelstudios Got it. Is the filtering done on the FE? I don't see any pagination, number per page, or ordering options? Is the intention to save the filters when the modal opens and closes?

nicholasserra:
Mainly asking because I had to do some sessionStorage filter stuff, saw this, and was a bit confused. So i'd like to add some comments and possibly slim it down.

axelstudios:
Yep, filtering and sorting is all done on the front end -- there's no pagination in this dialog.  As for saving the current filter/sort settings it doesn't matter to me.  The only reason I even included it was to match the existing code in search_service that applies to all of the other tables using sessionStorage.

nicholasserra:
Awesome, thank you. In the interest of less complexity i'd like to clean up as much as I can while I know what's up with it. @RDmitchell Do you have any preference as to whether filters are saved when you close and reopen the data cleansing modal?

RDmitchell:
Ideally, the filtering on data cleansing would work the same way through out the program, with the model being how it works in the Building list. 

[...]

PR #879, opened 2016-03-29
Issue #841 - Cleansing modal doesn't use number per page.

nicholasserra:
#### What's this PR do?
Removes an unused call to session storage in cleansing modal. Mainly removing this for cleanup and to make it less confusing.

Refs #841


mmclark:
loaded on seedtest

Issue #883, opened 2016-03-31
Export not working from Project


RDmitchell:
On seedtest, with NYC data (15,000 records), tried to export the 300 records from a project, and the program appears to hang. Progress bar stays at 0%, no errors. There is a sentry error which may not be releated but here it is:

New Issue on Lawrence Berkeley National Laboratory SEED Django (staging)

AttributeError: type object 'ProjectBuilding' has no attribute 'status_label'
ID: 019d48501b254ce187b2ac666ded7c91
March 31, 2016, 3:35 p.m. UTC
Exception

AttributeError: type object 'ProjectBuilding' has no attribute 'status_label'
(2 additional frame(s) were not displayed)
...
  File "seed/tasks.py", line 112, in export_buildings
    exporter.export(selected_buildings, selected_fields, _row_cb)
  File "seed/lib/exporter.py", line 166, in export
    export_method(buildings, fields, row_cb)
  File "seed/lib/exporter.py", line 249, in export_csv
    for field in fields
  File "seed/lib/exporter.py", line 249, in <genexpr>
    for field in fields
  File "seed/lib/exporter.py", line 61, in get_field_name_from_model
    par = getattr(par, component)
Tags

level = error  server_name = ip-172-31-30-126

nicholasserra:
I think that error is related.

nicholasserra:
https://app.getsentry.com/lawrence-berkeley-national-lab/seed-platform-django/issues/120298616/

nicholasserra:
So it's trying to export the label as a field. I'll add a check to stop that.

RDmitchell:
that's very interesting, because that is actually something we need to add to the program, ie, being able to export the labels associated with each record. I started to write it up as an issue, and realized that we needed to think about it a bit, because there may be different numbers of labels associated with each record. So some design may be needed. 

For now, let's fix this so that export works again. 

And very soon I will start an issue about adding labels by record to exports and we can explore the options about how that could manifest itself.

RDmitchell:
I am trying to same type of thing on seedtest, after the merge, this time from the Building list, rather than Project. The program reports 0/2616 buildings are exported, progress keeping chunking along under the network debugging tab, and then there is the following sentry error
![image](https://cloud.githubusercontent.com/assets/6314950/14401151/f7da948e-fdbb-11e5-8855-77b85c34a398.png)
[...]

PR #888, opened 2016-04-04
Issue #883 - Project export fix

nicholasserra:
"status" column on project buildings list was causing export to fail.

#### What's this PR do?
Remove status column on project buildings list since it is no longer used.

Export via project buildings list.

Refs #883 

#### Definition of Done:


mmclark:
loaded on seedtest

Issue #889, opened 2016-04-05
Fix migrations from clean db


nicholasserra:
Migrations are not working from a fresh database. Hoping this is something we can fix without bumping to Django 1.9.

nicholasserra:
@danielmcquillen can you confirm I fixed this? :P

danielmcquillen:
Sorry @nicholasserra , only seeing this now. Yes I believe the migrations worked correctly.

RDmitchell:
@danielmcquillen -- if you think everything is working properly, you can close this issue.

danielmcquillen:
Closing. This issue is long gone.

Issue #860, opened 2016-03-19
Superuser admin -- Add Creation Date to Organization list


RDmitchell:
In the list of orgs at the bottom of the admin screen that you see in superuser mode, if it is possible, add the date the organization was created -- maybe another column? For the existing orgs, if there is not a way to determine that, just leave it blank or whatever. But going forward it would be nice to keep track of when orgs are created, so that if they are inactive, we will know how long they have been there and that it's ok to delete them. 

![image](https://cloud.githubusercontent.com/assets/6314950/13895767/80979bf6-ed35-11e5-9f5a-9e92c12b71b3.png)

nicholasserra:
Orgs don't have a created date so i'll be adding a new field for this.

nicholasserra:
For some reason the orgs app does not have migrations, so I need to fix that for us to be able to add anything.

nicholasserra:
Unfortunately it seems there isn't a clean way I can accomplish this because of issues with migrations. There is a new flag in Django 1.9 that will allow me to implement this, assuming 1.9 is on our roadmap. This may have to be on hold until then.

https://github.com/django/django/commit/db97a8849519a3933bf4abd2184efd68ebc21965

RDmitchell:
Ok, that's fine. We can put this off until after the June release -- I will change the Milestone. 

Thanks for investigating it.

RDmitchell:
Tested ! Looks Great !! Thanks !

PR #900, opened 2016-04-11
Issue #889 -  Fix migrations

nicholasserra:
Migrations were broken when starting fresh, and missing migrations for orgs app.

#### What's this PR do?
Adds migrations for org app and updates some migrations.

@mmclark when these migrations are ran on staging and production, you'll need to pass `--fake-initial` to the migrate command. You'll know you didn't pass it if you see a `relation "orgs_" already exists` error. You'll only need to pass that flag once per environment. 

Refs #889 #860 

#### Definition of Done:


mmclark:
loaded on seedtest

Issue #897, opened 2016-04-06
Audit pip packages


nicholasserra:
We have a lot of stuff being installed via pip, some that is not even used. I found a few via the Django 1.9 upgrade, but i'd like to go over each one and figure out if it's unnecessary.

RDmitchell:
sure

PR #901, opened 2016-04-11
Issue #897 - Remove unused pip packages

nicholasserra:
#### What's this PR do?
Removes over 20 unused python packages. Removes a couple `reconcile` files that were not used. Removes references to New Relic.

Refs #897 

#### Definition of Done:


nicholasserra:
@mmclark I ended up just going with a simple `aws.txt` for server requirements. Replaces production.txt

mmclark:
Looks like the spelling job if failing in Travis.  Likely not a big issue, but it would be good to have it passing.  What's the fix for this?

nicholasserra:
Hoping @nllong can tell me how to rebuild those sphinx docs, as they need updated to clear that error.

nllong:
well it looks like the spelling is doing its job. The issue is this line: https://travis-ci.org/SEED-platform/seed/jobs/122371686#L519

regex isn't a real word I guess. Add it to the list here: https://github.com/SEED-platform/seed/blob/develop/docs/source/spelling_wordlist.txt

It is also warning us about some other things that I was hoping would resolve automagically when we upgrade to django 1.8. Regardless, if you fix regex, it should pass.

nicholasserra:
What's up with all the import errors? I removed some python files, and the docs are trying to import them? Do you know how to regenerate the autodoc or should I manually edit it?

nicholasserra:
Also the word "regex" has been in the source all along, it wasn't added in this PR.

nicholasserra:
@mmclark this is green now, should be fine to merge.

@nllong thanks for the look. I think there's still some weird stuff going on with spelling step, but we can revisit later.

mmclark:
loaded on seedtest

Issue #914, opened 2016-04-19
Number of Buildings not reporting properly 


RDmitchell:
On the building list, for some reason the number of building is not reporting properly. I will give an example once this issue is created.

RDmitchell:
Now I am having trouble recreating the issue. Now the program is reporting the correct number of buildings in both the left nav bar and the building list header, where before it was only reporting "1" building in the building list header, no matter what I did (Clear Filter, Show All)

RDmitchell:
Now, using user's real data, I see the same issue -- "1" building is reported in the Building list header but 1,166 records are actually there.
![image](https://cloud.githubusercontent.com/assets/6314950/14624275/1a15d1c4-058e-11e6-8cf3-089607e3df67.png)

RDmitchell:
I can reproduce this issue by importing 2 years of public disclosure data from the DC benchmarking website. Here is a screen shot -- I have uploaded 2012 and 2013 data, and even though there will be some matching, the Building list is only reporting 1 building in the header, when it should be reporting 1,1218.

![image](https://cloud.githubusercontent.com/assets/6314950/14624544/87d0e968-0590-11e6-8d64-43df0344e1df.png)

I can share those files if they are useful in debugging this issue. I didn't see the problem when I imported the first file -- the number of buildings was reported correctly. But when I imported the 2nd file, mapped and matched it, the program then reported only 1 building in the Building list header.

danielmcquillen:
Dang. I can't recreate this issue locally. Trouble is, we're using Angular 1.4.10 on develop right now (as well as Django 1.9 and some other changes), so it's going to be even harder to recreate.
@axelstudios I think this has to do with ng-pluralize's binding to scope.number_of_buildings. Do you have any thoughts on what might be going wrong? 
@RDmitchell should we postpone working on this issue until the latest on develop gets pushed to staging, as it might be resolved by the upgrades. Or is this version going out to users soon?

RDmitchell:
@danielmcquillen -- I don't quite understand your comment "until the latest on develop gets pushed to staging".... what is in develop that is not in staging (ie, seedtest) ?

axelstudios:
I'm not able to reproduce this on seedtest or locally (using Angular 1.4.10, Django 1.8.9).

But I also don't think it's ng-pluralize - `search.number_matching_search` is exactly what is was bound to before that change.  I have a feeling it's an issue with the data coming from the backend (search results).  To test this, try the following:

[...]

PR #919, opened 2016-04-20
Separate quantity and pluralized label

danielmcquillen:
We introduced ng-pluralize as part of a bigger FE update, but this directive didn't respond well to the existing code's use of the  "number" formatter for the relevant value. 

I tried fixing via a documented strategy of formatting the value within the "when" attribute, but that didn't work in this situation (perhaps because of the scope.object.value nature of the value). So in the interest of time I just pulled out the number from the ng-pluralize, which now just updates the noun in the label.

#### What's this PR do?
Separates noun being pluralized from formatting of actual number.

Make sure "Buildings" header properly shows number of buildings, including when buildings > 1000.

#914 

image

#### Definition of Done:

### DIFF
--- a/seed/static/seed/partials/buildings.html
+++ b/seed/static/seed/partials/buildings.html
@@ -4,8 +4,11 @@
            
         </div>
         <div class="page_title">
-            <h1><i class="fa fa-building-o"></i>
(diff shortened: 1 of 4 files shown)


mmclark:
loaded on seedtest

RDmitchell:
Thanks Miles, I will test !!

PR #926, opened 2016-04-27
Remove DOMAIN_URLCONFS setting.

nicholasserra:
DOMAIN_URLCONFS is defined in all the settings files, but I can't find a reason for it to exist.

#### What's this PR do?
Remove the setting.

#### Definition of Done:

### DIFF
--- a/config/settings/dev.py
+++ b/config/settings/dev.py
@@ -90,10 +90,6 @@ if "COMPRESS_ENABLED" not in locals() or not COMPRESS_ENABLED:
 
 ALLOWED_HOSTS = ['*']
 
-# ReadTheDocs won't have a local_untracked, so even here in dev, there needs to be a few initializations
-DOMAIN_URLCONFS = {}
-DOMAIN_URLCONFS['default'] = 'config.urls'
-
 
 # use imp module to find the local_untracked file rather than a hard-coded path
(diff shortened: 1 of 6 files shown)


nicholasserra:
cc @nllong @mmclark Maybe you two have an idea of what this was used for? I don't see any reason for it to be around anymore.

nllong:
No idea on my end. 

- nick

rebrownlbnl:
One of the comments talked about readthedocs, could it have been in there for the docs build on readthedocs?

nllong:
@Myoldmopar -- was this there for readthedocs?

Myoldmopar:
I assumed it was related to pointing to the proper urls file for a given setup. I don't recall atm, but i can look later/in the morning.

nicholasserra:
I don't think read the docs needs this. Only reference I can find is to a django-simpledomains package that is not used. Wondering if maybe it was used a long time ago.

rebrownlbnl:
yes, it's legacy.
I just went back and confirmed that these lines were in the original commit of the BE code back in 2014, so it was something they had put in there.
This could have something to do with wanting to run SEED as one "app" on a big, shared infrastructure, so there might have been multiple domains being served from the same server infrastructure.
Would that explain it?

I can't reasonably envision a scenario for running SEED that way now, so fine to delete if that's its purpose.

mmclark:
loaded on seedtest

PR #934, opened 2016-05-05
797 building detail ui

danielmcquillen:
Users were having difficulty viewing data in the building detail view when fields contained a lot of text (e.g. address).

#### What's this PR do?
Updates the building detail page to have resizable columns. Also updates table cells to show ellipsis for text values when they overflow the cell, and a popover when clicked (so that text can be selected and copied).

To get the popover-trigger directive -- as well as to just keep us up-to-date in our dependencies --this PR updates bower.json to use angular-bootstrap to 1.3.2. Therefore, developers will need to run a bower update command once this PR is merged.

Also, page title and table header styles were updated so they show ellipsis when value extends beyond parent container.

A side-effect of the angular-bootstrap update is that existing pulldowns break, so a big part of this PR is adding in missing uib-dropdown-menu directives and updating class styles. (Still looking for other issues caused by update.)

Testers must run **bower update** to get latest angular-bootstrap
Make sure to run **gulp sass** to compile updates to scss file

To Test:
All pulldowns should be tested.
Building detail page should be tested with data that has long strings, e.g. address 1.
  - Make sure columns can be dragged to new widths
  - Make sure ellipsis is shown when cell contents are longer than cell width
  - Make sure cells show a popup with full value of cell when clicked
  - Make sure you can select and copy text in popup
  - Make sure popup goes away when clicking anywhere on screen

#797

image

#### Definition of Done:


danielmcquillen:
@RDmitchell I think this can be merged, as the pulldowns that were broken by the UI Bootstrap upgrade are working correctly. It might need one pass of testing on seedtest to make sure the ui is behaving the same in the test environment.

RDmitchell:
OK, I will label it ready and Miles can merge it in with others. Thanks !

mmclark:
Loaded on seedtest

Issue #916, opened 2016-04-19
Error loading Building List page


RDmitchell:
I keep running into the issue that the program errors out trying to display the Building List. I am testing a large dataset with 14,000+ records, and it happens fairly often. The only solution seems to be to log out, close the browser, start another browser. I tend to start an incognito browser, which I don't think users should have to do.

In the example below, I had filtered on two labels, and then clicked on List Settings, reordered the columns, then clicked on Building List to get back to the Building list view.

![image](https://cloud.githubusercontent.com/assets/6314950/14657257/9ac65542-0641-11e6-832a-39ec5d86f5c7.png)

Then after clicking on "Clear Filters" the program gives an error when trying to display the building list
![image](https://cloud.githubusercontent.com/assets/6314950/14658581/bc97394a-0649-11e6-804e-c0f37b89b62c.png)

RDmitchell:
Here is a link to a video that shows the problem
https://drive.google.com/open?id=0B3fTKpZ9Dx7LdExveEptN2FCa0U

RDmitchell:
This keeps happening to me, and the only way I know that the program is not going to display the list is to look at the error via the developer tools. Is there a way that this error can make it's way to the surface so that the user knows what's going on, and can log off and start over?

nicholasserra:
Initial testing leads me to believe this is just a slow query across those buildings. I can look into optimizing that endpoint if you'd like. We might just have to up the timeout on nginx cc @mmclark

mmclark:
The uwsgi timeouts on nginx are probably ok - they're set at 300s. This morning the /app/label/ call was taking 18-20 seconds.  It sounds like it was slower for @RDmitchell, but not several minutes slower.  Are there additional timeout settings that you think would effect this call?

nicholasserra:
There is an nginx default timeout of 60 seconds that is probably being hit. I think `proxy_read_timeout` is the setting. I want to try to speed up this endpoint first though.

http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout

nicholasserra:
There are some pretty ugly queries around building search that i'm fixing up. We should see a performance improvement for all filtering.

RDmitchell:
wonderful !!!

RDmitchell:
Testing in the App Contest org on seedtest with 22,214 records, displaying the building list is working much better. Will also do a bit of testing with city data files that have less records (12K) but more fields.

[...]

PR #935, opened 2016-05-06
Issue #916 - Slow building filtering

nicholasserra:
Building list page is slow with a ton of records. 

#### What's this PR do?
This PR is a performance tweak for searching. The building list page calls `search_buildings`. The mapping/matching/other pages call `search_building_snapshots`. 

Both of these endpoints hit the same search methods that return children, parents, and coparent info. Building list does not need this information. So this PR passes along some info so that these related queries are not ran, speeding up the building list search.

@RDmitchell This needs solid testing, as it involves filtering in general. I ran through the normal flow with sample files and it looks good. But you may want to run your full test once this is merged.

Refs #916

#### Definition of Done:

### DIFF
--- a/docs/source/spelling_wordlist.txt
+++ b/docs/source/spelling_wordlist.txt
@@ -34,6 +34,7 @@ concat
 cond
 config
 coparents
+coparent
 csv
 dataset
 datasets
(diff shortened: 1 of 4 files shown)


mmclark:
loaded on seedtest

Issue #924, opened 2016-04-26
Error when filtering records on matching


RDmitchell:
The following error was generated, I believe, in the matching screen, after matching was finished, and I went back to it from the Data nav bar, and entered a filter (street name) on one of the fields, and then hit Enter to initiate the filter.

![image](https://cloud.githubusercontent.com/assets/6314950/14804612/e19c66d2-0b19-11e6-8abf-6af9f195abbe.png)

Seems like the issue might be the apostrophe in the field name "Assessor's Database Address"

![image](https://cloud.githubusercontent.com/assets/6314950/14804631/fffc05f6-0b19-11e6-8ab2-607cb8194357.png)

nicholasserra:
You're correct, that field name is invalid with the apostrophe. Do you have data files that caused this?

RDmitchell:
Here is a link to a file with this case. There are a few field names that have apostrophe's
https://drive.google.com/a/lbl.gov/folderview?id=0B3fTKpZ9Dx7Lc2wxaWV5Rkx5cG8&usp=sharing

![image](https://cloud.githubusercontent.com/assets/6314950/15085320/717e408c-138e-11e6-9c11-002dcba3cd08.png)

![image](https://cloud.githubusercontent.com/assets/6314950/15085338/80abdd58-138e-11e6-8ed6-f9a3a7e4be88.png)

![image](https://cloud.githubusercontent.com/assets/6314950/15085346/8c4a7836-138e-11e6-9286-c5ecac6464cd.png)

nicholasserra:
I think the best bet is to sanitize these extra_data fields on import. I'm not having much luck escaping them when querying, as the json query stuff is complex. I'm going to spend a little time to see if I can simplify it. I think Django shipped some changes that we can use.

RDmitchell:
Seems to be working on the sample data.

PR #938, opened 2016-05-10
Issue #924 - Update json queries

nicholasserra:
#924 is hard to debug because of the custom json query stuff. 

#### What's this PR do?
This PR removes a lot of the custom json query stuff, as it was no longer necessary. The JsonField package we're using supports most of it.

Refs #924

#### Definition of Done:


mmclark:
Loaded on seedtest

Issue #942, opened 2016-05-12
Investigate Sentry errors


nicholasserra:

RDmitchell:
@nicholasserra -- I noticed there was a new sentry error this morning from philadelphia, having to do with projects. Maybe it is something we have already fixed, but if you could look into it that would be great. thx

RDmitchell:
@nicholasserra -- just got this Sentry error when I was trying to filter on Year Ending, which is a date field, so you have to enter min and max values -- it's actually very funky. 

ValidationError: [u"'2010-01-01T08:00:00.000Z' value has an invalid date format. It must be in YYYY-MM-DD format."]
(15 additional frame(s) were not displayed)
...
  File "seed/utils/api.py", line 97, in _wrapped
    return fn(request, *args, **kwargs)
  File "seed/decorators.py", line 103, in wrapper
    response = func(request, *args, **kwargs)
  File "seed/views/main.py", line 648, in search_buildings
    user=request.user,
  File "seed/search.py", line 621, in orchestrate_search_filter_sort
    buildings_queryset, other_search_params, db_columns
  File "seed/search.py", line 221, in filter_other_params
    queryset = queryset.filter(query_filters)

this was with my rdmitchell+22 account.

RDmitchell:
And now I can't even get the building list to appear -- gets an error on search buildings. sigh ...

RDmitchell:
I just loaded sample data from issue 85 (multi year data) in order to do some testing on date filtering. I imported the data onto seed production in my +25 account, which didn't have any data in it (not sure whether it used to or not). Imported, mapped and matched the covered building list, the program reported 10 records imported, I clicked on the Buildings nav bar and got the search buildings error (same as shown above)

nicholasserra:
Looks like the datepicker is sending the wrong values.

nicholasserra:
Okay so just realized the datepicker is a chrome only thing. Is that what was used when the error happened? I don't think I can make a fix for that on the frontend.

RDmitchell:
I was using Chrome, but I am puzzled that this is an issue. I am pretty sure it worked before on Chrome, although it is true that I haven't tested this in a very long time. 

I am wondering if we should abandon the current filtering UI for dates -- it is very hard to use, and doesn't seem very effective. 

[...]

PR #945, opened 2016-05-12
Issue #942 - Be strict about requiring organization_id

nicholasserra:
Buncha errors in sentry around org id missing. I'd like to surface these as informative messages to the browser.

#### What's this PR do?
Make a decorator for views requiring organization_id passed via GET. Throw a bad request with an error message if not. Unsure how these are happening but this is a start.

Refs #942 

#### Definition of Done:


mmclark:
Loaded on seedtest

PR #978, opened 2016-06-08
Adding ansible provisioning within AWS

andremichi:
A few weeks a go I've helped a friend of mine to setup the SEED platform at AWS. After some issues the result was great. Running this playbook you can get an entire infrastructure running the latest version of the platform.

This PR does not change any code within the application. It is just an addendum to help who wants to automate things.

To run the playbook you will need to have a working setup of ansible with dynamic inventory. 

Just to remind you again, it is just a tentative of contribution.

### DIFF
--- a/provisioning/README.md
+++ b/provisioning/README.md
@@ -10,3 +10,10 @@ The configuration is separated in two steps. The first one will take care about 
 
 The main configurations that you need to change are inside the file vars/seed.yml, there you can find the a lot of variables that you need to adequate to your environment. All the important variables are between ##.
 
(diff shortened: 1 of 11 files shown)


nllong:
I'm all for a provisioning mechanism for SEED. If we go down this route, then we should probably pick a couple to use, (e.g. ansible, chef, and docker). I am planning on making docker (or docker-compose) the default deployment path.

nllong:
not sure we are going down the ansible route. If this is something that is needed, then we need to restart a discussion. Thanks!

Issue #965, opened 2016-05-26
Bar chart section of report not working


RDmitchell:
I am getting the following error on seedtest for the right side graph, which is the bar chart graph. @danielmcquillen I am assigning it to you for now, but we could give it to @axelstudios also, if you don't have time to work on this. We probably can't do it for the June 1.5 release, so not as much time pressure.

![image](https://cloud.githubusercontent.com/assets/6314950/15586145/a6a4d0d8-2337-11e6-80e8-53408f32f11d.png)

tallus:
I believe this is caused by missing quote marks on line 32 of buildings_reports_controller.js

RDmitchell:
Bar chart seems to be working now ! Thanks Paul !

PR #981, opened 2016-06-14
Fixes js error in getAggChartData

tallus:
Fixes js error in getAggChartData
    - yAxisType: Category, missing quotes causing Undefined error
    - added quotes : produces bar chart

Chart wasn't loading due to js error
#### What's this PR do?
Adds quotes to fix js error
Go to reports page and generate charts
#965
selection_130

#### Definition of Done:
Existing test at seed/static/seed/tests/buildings_reports_controller.spec.js was failing, now passes


RDmitchell:
Miles -- I set the milestone for this as a 1.5 Hot Fix -- it might be nice to get this into the V1.5 production version, but probably not the end of the world if we don't.

PR #995, opened 2016-06-24
Added column menus, grid menus, and saving/restoring state

axelstudios:
(See title) More UI Grid additions to the bluesky property and taxlot grids.  The individual grid settings are saved separately, and there's a custom option under the grid menu at the top right to completely reset any saved grid changes (filters, sorting, column order, column width, column visibility).

Please deploy to seeddev at your convenience.


RDmitchell:
@mmclark, @maronnax  -- still not quite sure what the protocol is here, but I definitely would like to get this PR merged into the bluesky branch on the public repo (maybe that's what Miles does?) and also onto seeddev (maybe that's what Nathan does?) so I can check it out. 

I am hoping that you two have figured out the protocol for this?

PR #1001, opened 2016-06-30
Small UI Grid Improvements

axelstudios:
*I would suggest re-running `bower install` after this merge.*

- Fetches all tax lot data the first time
- All `Related` columns (including extra data) get aggregated by default now
- Added customized ui-grid library with larger auto-column-size estimate (we may improve this later with an angular decorator)
- Fixed the grid menu scrolling to the top after hiding/showing a column


RDmitchell:
and did you also put it on seeddev so that I can test?

Issue #984, opened 2016-06-21
In superuser organization list (under Admin) show the Organization ID from the database in a column


RDmitchell:
In the Admin view of a superuser account, at the bottom where the list of organizations is shown, add a column to the left of the Organization Name, to display the Organization ID. 

![image](https://cloud.githubusercontent.com/assets/6314950/16241839/64c650ac-37a4-11e6-82d0-40669e957b19.png)

RDmitchell:
Done

PR #1002, opened 2016-07-06
In superuser organization list (under Admin) show the Organization ID…

tallus:
… from the database in a column
    - added column to template in static/seed/partials/admin.html
    - altered width of last column to fit

#### What's this PR do?
Adds an ID column to the Remove buildings and organizations of the admin screen
Visit /app/#/profile/admin as a superuser
#984 
selection_133
#### Definition of Done:
- [ ] Passes manual testing is different browsers


RDmitchell:
Paul -- sorry that I wasn't clearer about where the ID comes from -- it needs to be the official organization ID that is in the database. So for example, I believe that New York City is ID # 49, City of Cambridge is ID # 69, etc. Thanks !

tallus:
Would that be parent_organization_id? At the moment  it's the the id taken from the organization_org table.

RDmitchell:
Yes I believe that is correct.

tallus:
New Screenshot:
selection_134

N.B. parent_org_id can be a null field

RDmitchell:
Can you show a screen shot that shows the Parent Org ID?

@StephenCzarnecki -- is parent org ID the right ID?

tallus:
Here you go:
selection_135

According to the Organization model (or at least its get_parent method) an org is its own parent if its not a sub-organization, so I have followed the same principal here. i.e. it will show its own id if its a parent, its parents id if it is a sub-organization

RDmitchell:
do you think it is worth having two columns, one for Parent and one for child, with the appropriate IDs or blanks ? Then it is completely clear what IDs are what ?? Not sure.

tallus:
I wondered about that. There's an is_parent field we could display, or we could use id & parent_id -- it shouid be clear what's going on if they match.

tallus:
Some like:
Parent Id | Has Children
23&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;  | &#x2714;

or 
Parent Id | ID
23&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;  | 23
23&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;  | 24

RDmitchell:
Let's go with the second option

Thanks !

tallus:
selection_136

Issue #1005, opened 2016-07-11
Error in number of pages calculation 


nllong:
The number of pages calculation is not reading from the local storage upon initialization. This causes the buildings page to load indefinitely because the program always thinks that the total_number_of_pages is equal to total_buildings divided by 10, and not total_buildings divided by number_per_page. 

This error had the largest number of events in sentry (#991)

RDmitchell:
I believe this broke #842 but we can close this for now, and reopen 842 to be fixed

PR #842, opened 2016-03-10
Issue #836 - Remember page number when navigating around building list.

nicholasserra:
#### What's this PR do?
Save page number when navigating between building list and building detail views.

#### What are the relevant tickets?
refs #836 

#### Screenshots (if appropriate)
gif

PR #1006, opened 2016-07-11
1005 number of pages

nllong:
There were many errors in Sentry where the user tried to access a page in the building list that wasn't allowed. This addresses the issue by making the number of pages a method that is called.

Also, the building view of project buildings did not have the number of items, filters, etc in the local storage. This was updated.

Note that the filters for a project and all buildings are now the same. If you go from one to the other, the filters/settings persist.

#### What's this PR do?
Fixes the most prevalent sentry error.

List all the buildings and change from 10 -> 100. Go out, then come back to the view and click `last record`. It should return the results immediately (before it just hung).

Open up a project (list the buildings), and make sure the number of items and filters stick when clicking around.

#1005 

#### Definition of Done:
- [X] Is there appropriate test coverage? (e.g. ChefSpec, Mocha/Chai, Python, etc.) None.
- [X] Does this PR require a Selenium test? (e.g. Browser-specific bugs or complicated UI bugs). It should, but I don't think this is setup yet.
- [X] Does this add new dependencies? If so, does PIP, npm, bower requirements need to be updated? No

### DIFF
--- a/config/settings/local_untracked.py.dist
+++ b/config/settings/local_untracked.py.dist
@@ -44,3 +44,20 @@ CACHES = {
 DEFAULT_FILE_STORAGE = 'django.core.files.storage.FileSystemStorage'
 
 INTERNAL_IPS = ('127.0.0.1',)
+
+LOGGING = {
+    'version': 1,
+    'disable_existing_loggers': False,
+    'handlers': {
+        'console': {
+            'level': 'ERROR',
+            'class': 'logging.StreamHandler'
+        }
+    },
+    'loggers': {
+        'django': {
+            'handlers': ['console'],
+            'level': os.getenv('DJANGO_LOG_LEVEL', 'DEBUG'),
+        },
+    },
+}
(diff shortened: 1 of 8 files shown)


nllong:
I was trying to work with @tallus today to figure out why these functional tests are failing. I don't think we made much progress.

RDmitchell:
@mmclark -- is this fix on seedtest? I wanted to test and then get these on production. Thanks !

mmclark:
Loaded on seedtest

RDmitchell:
Seems like the problem of the list hanging is fixed. But I am not seeing the fix that was theoretically merged into the 1.5 release for staying on the same page when coming back from Building Detail. Don't know if that issue just didn't get put into the right branch or if this fix somehow affected that fix? https://github.com/SEED-platform/seed/pull/842

RDmitchell:
@nllong -- I am having @mmclark verify this, but it looks like #842 was part of the V 1.5 release, so it seems that maybe your fix to this issue broke #842 ? Can you check into it (or should we have someone else look at this, possibly @tallus ?) since you are deep into the new data model world?

mmclark:
loaded on seedprod

RDmitchell:
Tested on seed production working great. Thanks Nick !!

Issue #1008, opened 2016-07-12
catch invalid organization id


nllong:
When accessing several of the API endpoints, there is a chance that the org id is not valid. The code is currently trying to catch the error, but the error is not propagating back to the user. 

For example this call: http://localhost:8000/app/get_datasets_count/?organization_id=

will return an invalid type cast error because '' can't be coerced into an integer.

#967

RDmitchell:
any monkey testing for this?

nllong:
just go to  http://localhost:8000/app/get_datasets_count/?organization_id=

on the seed site and see if it returns a nice message instead of a 500/server error.

RDmitchell:
When I use this URL:

https://seedtest.lbl.gov/app/get_datasets_count/?organization_id=

I get this message

it returns the message "There was a problem with our server!" 

Is that the correct behavior ?

RDmitchell:
@nllong -- when I enter an invalid org id, like 9999, I get the following message

{"status": "error", "message": "Organization does not exist"}

It seems like the "blank" org ID is the same case, ie, the organization does not exist. And the message above is a much better message than 

"There was a problem with out server" 

?

nllong:
it should have handled both cases. I'll have to test locally again.

nllong:
this is what my machine looks like on the release-1.5.0 branch

<img width="635" alt="screen shot 2016-07-25 at 3 36 45 pm" src="https://cloud.githubusercontent.com/assets/1907354/17118398/ac418584-527d-11e6-9bf3-ffa42eeed83c.png">

RDmitchell:
ok, so I think your code is not on seedtest, which is where I would test it. I will ping Miles about this.

RDmitchell:
looks like @mmclark loaded this code onto seedtest 9 days ago. So I wonder what the difference is between your local environment and seedtest. 

Can you try the URL on seedtest and see if you get the right behavior (that I am not seeing for some reason)?

mmclark:
seed prod is updated with the hotfixes and behaves like Nick's example above

nllong:
i'm confused why this isn't looking right on seedtest. Could it be something related to DEBUG? Can i log into seedtest and check the code base?

seedtest
![image](https://cloud.githubusercontent.com/assets/1907354/17141965/85f1ee6c-530b-11e6-914d-cb5a10219c87.png)

release-1.5.0 branch


[...]

PR #1009, opened 2016-07-12
return json when invalid organization is passed, catch the error nicely

nllong:
See #1008.

There were many errors on sentry related to this bug. I have not determined what action on the front end is causing this, but this PR will allow for the app to continue.

#### What's this PR do?
Makes the error of org id not existing explicit and a JSON. Before it was a string result.

Normal monkey testing.

#1008 

#### Definition of Done:
- [X] Is there appropriate test coverage? (e.g. ChefSpec, Mocha/Chai, Python, etc.). Tests were added
- [X] Does this PR require a Selenium test? (e.g. Browser-specific bugs or complicated UI bugs). No
- [X] Does this add new dependencies? If so, does PIP, npm, bower requirements need to be updated? Yeah, I updated django-nose

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -1,7 +1,7 @@
 # Test dependencies go here. The local and test are the same at the moment.
 -r base.txt
 coverage==3.7.1
-django-nose==1.4.1
+django-nose==1.4.4
 flake8==2.5.2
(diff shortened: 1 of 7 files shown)


mmclark:
loaded on seedtest

PR #1015, opened 2016-07-19
Added Selenium Test Infrastructure

tallus:
- added Page base class
    - added page specific page object classes
    - added Table, TableRow and TableColumn classes
    - rewrote existing tests to use new infrastructure
    - updated documentation
   - unit tests for page.py
 
#### What's this PR do?
Adds Selenuim test infrastructure using the Page Object design pattern recommended by Selenium

#### Definition of Done:

  [x]  Is there appropriate test coverage? (e.g. ChefSpec, Mocha/Chai, Python, etc.)
  [x] Manual testing not need if functional tests pass


nllong:
I'm wondering if we need to revisit this and get the tests running deterministically. Both @axelstudios and I have seen this fail randomly.

tallus:
If you can let me know which tests I'd be grateful. The failures are random, annoyingly so if you are trying to reproduce them, but the ones I have seen are due to Firefox being flaky. In newer version of the tests I either skip the problematic tests with Firefox (Firefox sometimes ignore its own config and puts out the first run pop up which covers some links) or added some extra tests so we can skip parts that would fail if we hit a browser issue that would cause them to do so. 

This lessens the value of the tests somewhat, but actual failures should still get flagged.

axelstudios:
@tallus: Here's an example of a failed build, but that may not have enough detail about the specific test failure: https://travis-ci.org/SEED-platform/seed/jobs/148403661

tallus:
That's not happening in the actual tests themselves, though. It's occurring  in the tearDown method when it tries to quit the browser.  I think because the remote web driver is refusing connections. Its hard to debug without direct access to Sauce Labs but  I think its related to an issue I've occasionally seen where Chrome will close prematurely. Once this happens it cascades down to subsequent tests.

I could try wrapping it in a try..except block though it might just push the errors elsewhere. 
It should go at the end of the tear down, at least,  so the user etc always get deleted.

tallus:
merged in to #1046

PR #1019, opened 2016-07-20
Bluesky Changes

axelstudios:
- Organization dropdown now has a loading indicator and scrolls within the confines of the screen
- Fixed a couple column types (for filtering)
- Fixed taxlot initial load (> 10)


maronnax:
ok

RDmitchell:
I don't believe this is on seeddev (?). Also, right now on seeddev, when I click on View by Tax Lot, I don't see any records at all, even though at the top the program reports that there are 1,000+ records. See PT https://www.pivotaltracker.com/story/show/127043357

RDmitchell:
* Tested Org scroll bar and loading indicator -- very nice !
* Don't know what to test for the changed field types
* Can't check the > 10 record loading because DC BlueSky Test only has 9 records (!) -- **will test Monday when the migration is done on seeddev** -- unless you want to migrate a few more records (maybe 25 (?)) today, then I will test today.

axelstudios:
@RDmitchell Check now, I migrated 150 records.
The field type changes just change the filtering/sorting behavior:
- `Property / Building ID` is now a string
- `PM Parent Property ID` is now a numberStr

RDmitchell:
* > 10 records loads and displays on both Tax Lot and Property tabs -- I discovered that I can only get this to work when I am in "incognito" mode in Chrome -- all my problems with viewing the correct records (this past week) seem to stem from that. Is this something we need to deal with? Should I make a new bug in PT for that?

axelstudios:
Well, this >10 change was only just deployed today, which may explain why you haven't previously seen it.  Do you still only see 10 tax lot records when you *don't* use incognito mode?

If so there's a couple of things you should try:
- In Chrome (not incognito), hit F12.  That will open the developer tools.  Go to the `Network` tab and make sure that `Disable cache` is checked.  Refresh the page, or try a force-refresh (Ctrl+F5).
- The other option is more involved and would clear your localStorage.  Let's try the previous option first.

RDmitchell:
I think this is fine for now and can be merged into develop. The problem with my browser, and cache, I am going to put in a new issue here and I guess in PT.

PR #1020, opened 2016-07-20
Bluesky Cycle Saving

axelstudios:
Saves the chosen cycle between property/taxlot views for each organization.  The selected cycle will persist for the duration of the session, after which point it will default back to the first choice.


maronnax:
ok

axelstudios:
Refs [#124315651](https://www.pivotaltracker.com/story/show/124315651)

RDmitchell:
I have tested this on seeddev as of 7/25/2016 at 6:36 pm and the cycle sticks moving from View by Property and View by Tax Lot

Issue #1025, opened 2016-07-25
Breakup Bluesky Model


nllong:
Restructure and remove the use of bluesky.

PR #1026, opened 2016-07-25
1025 flatten bluesky

nllong:
This removes the bluesky terminology from the backend. The API still support the app/bluesky/properties.json syntax, but should be deprecated for the app/properties.json (which should also be deprecated once we have api/v2/properties hooked up @Myoldmopar). Also, the frontend is still using the old endpoints and need to be updated. 

#### What's this PR do?
When starting from a clean database (production or pre-bluesky), this will migrate the database and add the tables for seed_* instead of bluesky_*. 

Start from a fresh db, or restore a copy of the prod database and run

`./manage.py migrate`

If you loaded the production database, then you can also run the following after giving yourself access to org 184 (DB Blue Sky Test).

`./manage.py manage_bluesky_data --org 184 --limit 30`

#1025 

#### Definition of Done:
- [X] Is there appropriate test coverage? (e.g. ChefSpec, Mocha/Chai, Python, etc.). Not really, as there are no tests in the new data model yet (tsk tsk tsk)
- [X] Does this PR require a Selenium test? (e.g. Browser-specific bugs or complicated UI bugs). Not really
- [X] Does this add new dependencies? If so, does PIP, npm, bower requirements need to be updated? Yes. The dependencies for running the migration steps were added to local.txt.


maronnax:
:thumbsup:

nllong:
merge this down after #1029  is merged to develop and back into this PR.

Issue #1031, opened 2016-07-26
Investigate Sentry error


RDmitchell:
Paul -- can you investigate this Sentry error? Miles is going to invite you to our Sentry account tonight, which should help you debug it. 

https://app.getsentry.com/lawrence-berkeley-national-lab/seed-django-production/issues/145481270/

I will forward you the email I got for this error, but really you will need to log into Sentry to figure it out.

RDmitchell:
This error was generated through the UI, not through an API call.

tallus:
It looks like it is generated during  the data mapping stage once Step 1 has been completed (when the user can 'Save Mappings'). On this page clicking on any of the table headings sends a request to search_building_snapshots with order_by set to that column name.

However since this gets translated directly into a DB call it only works for columns that correspond to fields in the database, not on extra data as orderby_by with JSONFields is not currently supported by Django:

https://stackoverflow.com/questions/36641759/django-1-9-jsonfield-order-by

It can be down with Raw SQL (though that's not ideal), I will take a look as to what can be done with that.

tallus:
Pull request is in
https://github.com/SEED-platform/seed/pull/1034

tallus:
Pull request against release 1.5.0: https://github.com/SEED-platform/seed/issues/1041

RDmitchell:
closing this issue as V 1.5 errors are no longer relevant

PR #1034, opened 2016-07-27
Fix for #1031 Investigate Sentry error

tallus:
- error was caused when an extra data field was used in an order by query, as this is not supported by Django
- added a raw sql query for order_by when order_by is an extra data field.

#### What's this PR do?
Add an alternate means of resolving order_by query in search_building_snapshots when the field is an extra data field 
Upload data containing columns that will need to be stored as extra data. At the end of Step 1 - Map Your Data, click on an extra data column and verify the page loads correctly
# 1031
#### Definition of Done:

### DIFF
--- a/seed/tests/test_views.py
+++ b/seed/tests/test_views.py
@@ -1554,7 +1554,7 @@ class SearchBuildingSnapshotsViewTests(TestCase):
             data=json.dumps(post_data)
         )
         self.assertEqual(1, json.loads(response.content)['number_returned'])
-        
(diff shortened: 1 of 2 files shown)


RDmitchell:
sorry I didn't make this clear in the issue, but this PR should be against the 1.5 release (as well as develop once it is tested in the 1.5 release).

tallus:
New pull request: https://github.com/SEED-platform/seed/pull/1038

Issue #1033, opened 2016-07-27
Add link to Swagger on front end


Myoldmopar:
I _think_ it might be nice to put a link to the Swagger page right on the front end.  Maybe something like:

![image](https://cloud.githubusercontent.com/assets/849824/17176035/04bb6716-53c8-11e6-911d-3ad1c304907d.png)

only.... :sunny: pretty ☀️ .

RDmitchell:
What do others think? I can see it is nice to be able to get to it easily through the FE, but most users using the FE will not know what this is or use it. So I can see it both ways, ie, with and without it. 

How else would someone get to the swagger pages if not through the FE?

Myoldmopar:
Maybe a link in the docs or on the GitHub readme, but it seems like a good
idea to have a clean simple link right on the FE.

RDmitchell:
agreed !

Myoldmopar:
Closable now, right?

RDmitchell:
Am closing, this has been implemented, rather nicely I must add.

PR #1036, opened 2016-07-28
Add link to Swagger on front end

Myoldmopar:
The Swagger page wasn't accessible from the front end; this fixes it.

#### What's this PR do?
This adds an icon in the sidebar to access the Swagger API documentation page.

The link should go to the API doc page, and the sidebar should look nice, and all other sidebar links should work.

#1033 

Sidebar closed:
image

Sidebar open:
image

#### Definition of Done:
I'm not sure if the Selenium test, etc., is necessary here or not.


RDmitchell:
I like the icon !

Myoldmopar:
OK, we can go with that icon then.  In that case, if @axelstudios might look over the changes, this could potentially go right in.  The main concern I have is the href and url reversing I put in, which looks slightly different from the others because there is no `#` in the Swagger url.

Myoldmopar:
I think I should probably add a link on the Github README too; I'll do that now.  

OK I added a description of where to find the swagger docs on a running server to the README.  I also updated the ReadTheDocs badges since they weren't being rendered properly.

Before:
image

After:
image

RDmitchell:
will this be deployed to seeddev at some point soon ?

axelstudios:
@RDmitchell Yep, I'm working on that now.  Nathan and I are currently looking into an issue where git isn't able to pull new changes on seeddev

RDmitchell:
great, thanks !

axelstudios:
@RDmitchell Try testing now

RDmitchell:
it's showing up on seeddev and looks great ! I haven't actually tested an API call yet, but just being able to see the documentation is very cool. Thanks Edwin and Alex !

RDmitchell:
I guess the only slightly funky thing is getting back to SEED. I just used the back arrow on my browser. Is there a better way to do it? Would it make sense to open the swagger window as a separate browser tab? Maybe not because of security issues that I don't have a clue about. Anyway, just my initial reaction. But I do think the actual swagger pages look great and I think will be incredibly useful.

Myoldmopar:
That's a good point.  I don't see an obvious way to add a link back to SEED from the Swagger page, although I am sure there is _some_ way to do it.  Given that it is such a different portal/view, it seems reasonable to me to open it in another tab.  The browser will remain logged in, so it should be seamless.  I can add this if others find it reasonable too.  @nllong @axelstudios ?

Myoldmopar:
Also, I'm seeing an error when using Swagger to actually access the data that I'm not seeing locally.   Hmm wait.  It's when I'm not logged in.  OK.  I think if you aren't logged in, it tries to hit a different branch of code to throw the exception which wasn't updated with the new v2 style DRF endpoints.  I can reproduce this locally though.  If I find something small, should I lump that in here or open a new issue/PR?

axelstudios:
@Myoldmopar It might be worth trying a tiny angular page that just uses `ng-include` to pull in the swagger page without leaving the main SEED interface..

Myoldmopar:
Thanks @axelstudios.  So it would open in the SEED page itself in its own container?  

`if so, then:`
I won't claim designer status, but it feels like it could be clumsy to a developer sorta working in the API and the GUI simultaneously (like adding datasets and seeing the representation in Swagger to develop their app, etc.).  
`endif`

Honestly though I don't care.  Whatever works best from a design/workflow standpoint.

axelstudios:
How's this?  Used an iframe:
image

Myoldmopar:
Looks superb!  Does the content persist if you browse to a different page and come back to it?

RDmitchell:
That looks fine to me (for whatever that is worth). And then you just click on another nav button to get to the other views, right? So it is just like the other nav buttons. Seems nice because it's consistent with the way the rest of the program works.

axelstudios:
Yep, it's just like another nav button.  @Myoldmopar nope, state isn't persisted.  As soon as you leave the page, like all pages, the angular scope is destroyed.

Myoldmopar:
Hmm.  In that case I'm not sure.  It seems like if someone is working on the Swagger page, they are most likely playing around trying to figure out how to get their app to talk to SEED, and possibly adding/deleting data which they may want to verify is coming through right from the GUI, and then they switch back to the Swagger page, and the nice request they put together is gone, and this is a very long sentence.

RDmitchell:
actually, on some of the existing pages (at least in the old world) the state of the page would persist. For example, filters are kept in the Building List, even if you navigate away and come back to it. But I suspect, at least as a first cut, you wouldn't need to persist the state when the user leaves the page.

axelstudios:
The things that are currently persisted, like filters on old tables, get saved to the database per user.  I'll see if I can prevent the `$destroy` event on this one.

RDmitchell:
Edwin -- per your comment, I guess it's a matter of trying to figure out who the audience is and how they will use it. If you really think an App developer would spend a lot of time on this page, and want to save the state of it, that is different than someone just experimenting with the API calls. But maybe first cut is that the state isn't saved, and then you can keep moving forward with the 2.0 migration and come back to this after talking to some developers about how they would want to use this.

axelstudios:
@RDmitchell @Myoldmopar My change is on seeddev now and ready for testing.

 It looks like if we want to persist the state the best option is to replace the default Angular $routeProvider with UI router which has a stickyState option.  That's a wishlist change, though.

RDmitchell:
I would say let's put the persist state option on the list for after the Sept release (unless we end up twiddling our thumbs in Sept). @Myoldmopar -- maybe you should make a new issue for it, either here, or in PT, just so we don't lose track of it. 

@axelstudios -- your change looks great to me

PR #1038, opened 2016-07-28
Fix for #10131 cut against master

tallus:
Cut against master branch.
- error was caused when an extra data field was used in an order by query, as this is not supported by Django
- added a raw sql query for order_by when order_by is an extra data field.

#### What's this PR do?
Add an alternate means of resolving order_by query in search_building_snapshots when the field is an extra data field 
Upload data containing columns that will need to be stored as extra data. At the end of Step 1 - Map Your Data, click on an extra data column and verify the page loads correctly
#1031
#### Screenshots (if appropriate)#### Definition of Done:

### DIFF
--- a/seed/tests/test_views.py
+++ b/seed/tests/test_views.py
@@ -1555,6 +1555,49 @@ class SearchBuildingSnapshotsViewTests(TestCase):
         )
         self.assertEqual(1, json.loads(response.content)['number_returned'])
 
+def test_search_building_snapshots_extra_data_order_by(self):
+        """
+        Test to ensure search_building_snapshot correctly handles
+        order_by when it is an extra data field.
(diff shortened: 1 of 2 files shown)


RDmitchell:
I think this should go against the 1-5-0 release branch, rather than master, since it's a 1.5 Hot fix? 

@mmclark -- can you verify?

tallus:
pull request against release 1.5.0: https://github.com/SEED-platform/seed/pull/1041

mmclark:
Yes, this needs to be issued against the release-1.5.0 branch

tallus:
Closing see: https://github.com/SEED-platform/seed/pull/1041 for pull request against 1.5.0.

Issue #1039, opened 2016-07-29
Swagger is inelegant when user isn't logged in


Myoldmopar:
The `RequiredLoginMixin` that is inherited by the RestFramework viewsets is causing troubles.  If the user isn't logged in, it follows this path through the Mixin code:
- [it returns the response of `handle_no_permission`](https://github.com/django/django/blob/master/django/contrib/auth/mixins.py#L55), which:
- if it isn't triggered to throw an exception, [tries to `redirect_to_login`](https://github.com/django/django/blob/master/django/contrib/auth/mixins.py#L46), which:
- tries to access a `request` attribute on the class instance, which in our case, is, for example, `DataViewSet`

I don't know why exactly it assumes `request` is a member; maybe that's how it used to be in the Django world, but it isn't now.  I see a couple paths forward:
- Maybe we should do whatever setting is needed to not try to redirect to the login page.  It would throw an exception, but I think it would be a handled exception and be better behaved.
- Or, maybe we should protect the Swagger page to only during a logged in session.  
- Or...

Myoldmopar:
Gyah...OK I think I already solved it.  I'll comment on the PR in a moment.

PR #1040, opened 2016-07-29
Override exception flag from access mixin in our derived class

Myoldmopar:
The Swagger page wasn't behaving nicely when the user wasn't logged in.  It was trying to access a class member that didn't exist in our structure, so the exception was a Python exception traceback and not a nice response.  

#### What's this PR do?
This PR simply adds a class member variable that overrides a variable on the inherited class.  This variable tells the logged-out code to throw an exception that can be caught nicely and return a 403 instead of trying to access an invalid member.

Well, the existing response when you aren't logged in looks like this:

image

Not great.  With this fix, it looks like this:

image

Maybe it could be polished a little, but it's way better, and clear to the user that they don't have permission to access the underlying views, and should clue them in to logging in.

#1039 

#### Definition of Done:
… doesnt have a request member

### DIFF
--- a/seed/views/datasets.py
+++ b/seed/views/datasets.py
@@ -25,6 +25,8 @@ _log = logging.getLogger(__name__)
 
 
 class DatasetViewSet(LoginRequiredMixin, viewsets.ViewSet):
+    raise_exception = True
+
     @require_organization_id_class
     @api_endpoint_class
     @ajax_request_class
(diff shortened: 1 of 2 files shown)


nllong:
@Myoldmopar can this get merged down?

Myoldmopar:
For sure, if it still needs to... Lemme check all the ViewSet derived classes real quick.

Myoldmopar:
It looks like all of our derived classes already got the flag put on, this should just be closed without any action.

Issue #1027, opened 2016-07-25
Can't delete data in extra_data fields in Building Detail


RDmitchell:
It seems that in the old data model (V1.5) does not allow deleting existing text in extra_data fields in Building Detail. However, it IS possible to delete data in structured fields in the building snapshot table.

RDmitchell:
Here is a link to a sample data file that has both real fields and extra_data fields. 
https://drive.google.com/open?id=0B3fTKpZ9Dx7Lano3OXp3Z2syX2M

Steps to replicate:
* Import the file into SEED, mapping the Extra Data fields so that they stay in extra_data
![image](https://cloud.githubusercontent.com/assets/6314950/17119851/7609c5e4-527d-11e6-8bd4-c3c019060302.png)

* Delete "triangle town" which is an extra data field using "Edit" in the Detailed view and save it. It looks like it is deleted when you are still in Detailed view
![image](https://cloud.githubusercontent.com/assets/6314950/17119971/3483397e-527e-11e6-9f1e-8ee4db2874b6.png)

* Go back to the Building List view, and you will see that the data is still displaying in that field. 
![image](https://cloud.githubusercontent.com/assets/6314950/17120339/641c1866-5280-11e6-8c28-7e9cc14e2630.png)

* Do the exact same thing with the field **City**, which is NOT in extra_data, but is in the structured field called "city", and the program will actually delete the value. 

So it appears that this is not working in the extra_data fields.

RDmitchell:
@tallus -- can you take a look at this and determine the level of effort to fix it? Then we can decide whether to fix it now or just wait for the September release. thx

tallus:
updating extra data  uses merge_extra_data  from [seed/mappings/mapper.py](https://github.com/SEED-platform/seed/blob/develop/seed/mappings/mapper.py#L60)
This *merges* extra data so it won't overwrite keys with blank values:
changing 
   k: default_extra_data.get(k) or non_default_extra_data.get(k)
to
 k: default_extra_data[k] if k in default_extra_data else  non_default_  extra_data.get(k)
on line 85 "fixes" the problem.

This won't take that long since I've done the hard work already.
 
Note however there is a unit test to ensure it doesn't do this (i.e. override with blank data).

So we also need ensure this won't have impact elsewhere (it shouldn't as the function is only called 
in one place) and that this is really what we want and won't have unintentional side-effects*, given that the unit tests say it shouldn't do this. This might take more time.

*like deleting data unintentionally.

[...]

PR #1043, opened 2016-08-01
Fix can't delete data in extra_data fields in Building Detail #1027

tallus:
-  adds allow_delete parameter to seed.mappings.mapper.merge_extra_data        
        if allow_delete is True extra_data fields can be set to null.
 -  adds allow_delete_extra_data to seed.models.update_building
        value is passed to merge_extra_data
 -  sets allow_delete_extra_data=True in the call to update_buildings in
        seed.views.main.update_building view

#### What's this PR do?
Allows users to "delete" (set to null) extra data fields when updating buildings

Edit a building with extra data fields in the Building Details view, by blanking the relevant box and saving. Visit Buildings List view to ensure previous value is no longer there.

#1027 
#### Definition of Done:

Note tests won't run if you have migrated to develop.

### DIFF
--- a/seed/mappings/mapper.py
+++ b/seed/mappings/mapper.py
@@ -77,7 +77,7 @@ def merge_extra_data(b1, b2, default=None, allow_delete=False):
     ..Note:
 
         If default is neither b1 or b2 b1 becomes the default and b2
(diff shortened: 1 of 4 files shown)


nllong:
closing without merging as v2 updates make this irrelevant.

PR #1050, opened 2016-08-26
APIv2

Myoldmopar:
Not nearly ready for merge yet, but starting to push the v2 API to a more final state.  Need to discuss deployment because this is a breaking API change.

For each endpoint here, I am updating the unit tests and the javascript services to interact with the new endpoints and any changed parameters so that the front end works seamlessly.  

Things to still resolve:
 - [ ] Complicated, hand-built view return types don't play nice with Swagger
 - [ ] Reversing dynamically built URLs in Angular land

I'll update this description and relevant tickets as this clears up even further.


api/urls.py
```diff
@@ -49,11 +49,11 @@
     url(r'^taxlot-columns/$', get_taxlot_columns, name='taxlot-columns'),
     url(r'^properties/(?P<property_pk>\d+)/cycles/(?P<cycle_pk>\d+)/$',
```
tallus: This can be rewritten like this:
property_detail = Property.asView({
    'put': 'put',
    'get': 'get_property',
})
url(r'^properties/(?P<property_pk>\d+)/cycles/(?P<cycle_pk>\d+)/$', property_view, name='property')

which might be a little cleaner  if you don't mind shared names.

nllong:
ouch time...

PR #1051, opened 2016-08-26
Add search_properties and search_taxlots as equivalents to search_buildings

tallus:
#### What's this PR do?
Add search_properties and search_taxlots as equivalents to search_buildings
https://www.pivotaltracker.com/n/projects/1488476/stories/128697981


seed/search.py
```diff
@@ -64,6 +65,46 @@ def search_buildings(q, fieldnames=None, queryset=None):
...
+    return _search(q, fieldnames, queryset)
+
+
+def search_taxlots(q, fieldnames=None, queryset=None):
+    if queryset is None:
+        return TaxLotState.objects.none()
+    if fieldnames is None:
+        fieldnames = [
+            'jurisdiction_taxlot_identifier',
+            'address'
```
nllong: I already merged this down, but the disconnect between the DB fields for tax lot and property is concerning. We call address_line_1 in one and address in the other. should we fix this? @RDmitchell

RDmitchell:
@nllong -- I do agree that it is a problem that the names for similar fields have different names in the two tables. The addresses may or may not be the same in the two files, but it does seem like the fields should have the same names, with a prefix of the table distinguishing them. And then in matching, if the user wants to match on address, they will automatically have the same name. 

I assume that @maronnax had a good reason for making them different, but now it seems like that might not make sense.

nllong:
so if we are to change the names, we should do that now-ish before we get more 'new data' into the system.

RDmitchell:
Right .. and we would have to change the migrations, etc. It would be nice to talk this over with @maronnax but he isn't back until next Tuesday. Not sure how to make the decision.

nllong:
yeah the migration would need to updated... let's just add it to the list.

RDmitchell:
OK, I will put in a PT issue for it, which we can hopefully discuss first thing when Nathan gets back next week.

rebrownlbnl:
I think the different names were an artifact of the field names that were in the existing database, where Nathan was trying to pull apart the PM data from the tax lot data.

I generally agree that we should use the same name for the same concept, to be consistent, maybe with a prefix to indicate which table it's in. My vote would be something more human readable like tax_lot_address and property_address.

The reason we had a field called "address_line_1" is because PM has the concept of a multi-line address. Do we need to worry about keeping multiple address lines? Or do we just concatenate it all into one field?

PR #1052, opened 2016-08-26
Major UI Refactor

axelstudios:
This is the current state of all the merged UI branches currently deployed to seeddev, which includes inventory list, details, settings, labels, cycles, reporting, etc.


RDmitchell:
Tested in seeddev, seems good to me from monkey testing.

tallus:
I'm re-opening this as it broke the functional tests. It appears that javascript files necessary for some views were removed.

If the building views are no longer needed then they needs to be cleaned up properly:
i.e. the urls (and attached views) need to be removed so the code/page is no longer reachable and tests need to be re-worked to reflect this.  Removing the UI elements to reach them is not sufficient.

PR #1062, opened 2016-09-09
Demo data with audit logs

StephenCzarnecki:
This should fix the issues with audit log data in the command for creating sample data.  Had a problem merging develop into the branch I had been working on so since it is only two files I just made a new branch.  This PR does not address issues with using the generated data as import files.
#### What's this PR do?
Adds a management command called create_sample_data that takes the following command-line params:
--A number of A cases to create
--B number of B cases to create
--C number of C cases to create
--D number of D cases to create
--Y years to be created in a csv string.  E.G. "2014,2015,2016"
--audit-depth the number of additional records to be created for each case for each year.  E.G. "--A 1 --B 0 --C 0 --D 0 --Y "2014,2015,2016" --audit-depth 5" will create 6 PropertyStates and 6 TaxlotStates (remember case A is one property to one taxlot) for each of 2014, 2015, and 2016.  Each year creates an additional 5 states to be used in the audit logs for a total of 6.  But there is only one Property and Taxlot view record created for each year.
Visually using the new UI changes.  Make sure all of the cases seem to agree in the building list and the audit log information looks reasonable in the detail view.
#### Definition of Done:
NA I think.  This is generating sample data for use in development, it is not actually used in production.

### DIFF
--- a/seed/management/commands/create_sample_data.py
+++ b/seed/management/commands/create_sample_data.py
@@ -1,0 +1,801 @@
+from __future__ import unicode_literals
+
+from django.core.management.base import BaseCommand
+from seed.lib.superperms.orgs.models import Organization
+import datetime
+import logging
+import itertools
+from random import randint, random
+import seed.models
+
+from seed.test_helpers.fake import FakePropertyStateFactory, FakeTaxLotStateFactory, BaseFake
+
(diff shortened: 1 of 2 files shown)


nllong:
👍

PR #1070, opened 2016-09-21
Clean up the Swagger docs

Myoldmopar:
The auto-generated Swagger docs needed some cleanup.

#### What's this PR do?
Focused on docstring updates.  But in the case of complex body parameters and complex response types, I created small Serializers to allow Swagger to render nicely.  These serializers could be replaced later...possibly.

Just check the Swagger page endpoints and make sure the request/response types, and endpoint descriptions, all look nice.

None that I know of.

No.

#### Definition of Done:

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -22,7 +22,7 @@ api_v2_router.register(r'data_files', DataFileViewSet, base_name="data_files")
 api_v2_router.register(r'projects', ProjectsViewSet, base_name="projects")
 api_v2_router.register(r'users', UserViewSet, base_name="users")
 # api_v2_router.register(r'labels', LabelViewSet, base_name="labels")
-api_v2_router.register(r'reverse_test', TestReverseViewSet, base_name="reverse_test")
+api_v2_router.register(r'reverse_and_test', TestReverseViewSet, base_name="reverse_and_test")
 
 urlpatterns = [
     # v2 api
--- a/api/urls.py
+++ b/api/urls.py
@@ -21,8 +21,7 @@ api_v2_router.register(r'organizations', OrganizationViewSet, base_name="organiz
 api_v2_router.register(r'data_files', DataFileViewSet, base_name="data_files")
 api_v2_router.register(r'projects', ProjectsViewSet, base_name="projects")
 api_v2_router.register(r'users', UserViewSet, base_name="users")
-# api_v2_router.register(r'labels', LabelViewSet, base_name="labels")
-api_v2_router.register(r'reverse_and_test', TestReverseViewSet, base_name="reverse_and_test")
+# api_v2_router.register(r'reverse_and_test', TestReverseViewSet, base_name="reverse_and_test")
 
 urlpatterns = [
     # v2 api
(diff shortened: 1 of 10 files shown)


api/urls.py
```diff
@@ -21,8 +21,7 @@
 api_v2_router.register(r'data_files', DataFileViewSet, base_name="data_files")
 api_v2_router.register(r'projects', ProjectsViewSet, base_name="projects")
 api_v2_router.register(r'users', UserViewSet, base_name="users")
-# api_v2_router.register(r'labels', LabelViewSet, base_name="labels")
-api_v2_router.register(r'reverse_test', TestReverseViewSet, base_name="reverse_test")
```
Myoldmopar: Just removed the test endpoints from the Swagger page.

api/views.py
```diff
@@ -80,6 +80,21 @@ def one_arg_reverse(self, request):
         reversed_url = reverse(reverse_string, args = [argument])
         return HttpResponse(json.dumps({reverse_string: reversed_url}))
 
+    @list_route(methods=['GET'])
```
Myoldmopar: This example demonstrates how we'll eventually use the multipart upload....if Swagger works with it...

seed/views/datasets.py
```diff
@@ -36,48 +36,23 @@ class DatasetViewSet(viewsets.ViewSet):
     def list(self, request):
         """
         Retrieves all datasets for the user's organization.
-
```
Myoldmopar: Lots of changes in this PR like this where I change the old docstring to a form that Swagger can interpret and render.

seed/views/datasets.py
```diff
@@ -165,38 +140,28 @@ def update(self, request, pk=None):
...
+                    description: A dictionary of a full dataset structure, including
+                                 keys ''name'', ''number_of_buildings'', ''id'', ''updated_at'',
+                                 ''last_modified_by'', ''importfiles'', ...
             parameter_strategy: replace
             parameters:
                 - name: pk
                   description: "Primary Key"
                   required: true
                   paramType: path
+                - name: organization_id
```
Myoldmopar: This was simply missing.

seed/views/organizations.py
```diff
@@ -195,6 +195,79 @@ def _save_fields(org, new_fields, old_fields, is_public=False):
 _log = logging.getLogger(__name__)
 
 
+class RulesSubSerializer(serializers.Serializer):
```
Myoldmopar: So here's a bunch of serializers.  These are never actually used to serialize data for the views.  They are simply there to render complex types in Swagger.  Maybe...we could...use these sometime to actually get data in/out of the views?  But it seems like overkill.

tallus: FYI, the existing serializers all live under seed/serializers, rather than views. (I've been using them for all the API endpoint I've been writing)

Myoldmopar: Thanks @tallus.  I had thought about putting the serializers in a separate file, but to avoid complications with Swagger I left them in the same file as the views.  I'll make the change and test it out and if Swagger still understands how to find the serializers I'll push it up.

PR #1083, opened 2016-10-06
save tax lot id to lot_number

nllong:
For pairing it helps to have the tax lot ids in the  property.lot_number.

#### What's this PR do?
Adds tax lot ids to the lot number.

Run the tests.

N/A

N/A

#### Definition of Done:
- [X] Is there appropriate test coverage? (e.g. ChefSpec, Mocha/Chai, Python, etc.) yes
- [X] Does this PR require a Selenium test? (e.g. Browser-specific bugs or complicated UI bugs) N/A
- [X] Does this PR require a regression test? All fixes require a regression test. N/A
- [X] Does this add new dependencies? If so, does PIP, npm, bower requirements need to be updated? No


nllong:
merging knowing that we have errors to fix.

RDmitchell:
hate to be a wet blanket at this stage in the game, but why call it "lot_number" in property, if the table name is prepended to the field name, ie, property.lot_number? 

Seems like it would be much more straightforward if the two fields had the same name, but with the table pre-pended, ie, 

taxlot.tax_lot_id 
& 
property.tax_lot_id

at least in the database. Still not sure how all this should manifest itself on the front end.

Issue #1090, opened 2016-10-10
Program automatically creates duplicate mapping even though the original fields names are unique


RDmitchell:
The program is doing some "automatic" mapping, and creating erroneous duplicate field names, when the original field names are actually different.

When importing the example file called 
example-data-properties-V2-original-noID.xlsx
here is a link to that file
https://drive.google.com/open?id=0B3fTKpZ9Dx7LRW0ydm5fdktpUGs (from issue #1089)

the program maps the two extra data fields to the same new SEED field name -- it should be able to detect that they actually have different original field names and should make the newly mapped names unique, so that the user doesn't have to unduplicate the field names, when they should never have gotten into that state in the first place. 

Here is what the program does 
![image](https://cloud.githubusercontent.com/assets/6314950/19252553/4bc4e1d8-8efa-11e6-9e4e-b302fdfb8271.png)

Because it is possible to work around this bug, it doesn't have a super high priority, but it is definitely an unnecessary annoyance to the user, and we should fix it sometime in FY 2017

nllong:
two ways to fix. easy way would be to append a uniq id when repeated fields show up. Or we can enforce the fuzzy logic to not look for a field that has already been chosen. The latter is more work than the former.

nllong:
other option would be to just use the field name if the confidence of the mapping match is low.

rebrownlbnl:
I think the 3rd option is originally how the mapping logic was intended to work, but it always seemed to have some quirks. @RDmitchell can you clarify what the desired behavior is in this case? Should it just keep the original header names?

RDmitchell:
We (NREL / LBNL) decided this morning on a call about this to do the third option that Nick suggests.

Really what this was supposed to catch were duplicate field names in the ORIGINAL file being imported, not to CREATE bogus duplicates with the crazy field suggestions that the program is currently making.

rebrownlbnl:
Good. My memory about how mapping was using the confidence metric wasn't too far off :-)

RDmitchell:
Definitely have to fix the duplicate field issue. Here is what happens when I import a Portfolio Manager file -- the program makes about 1/4 of the 250 fields into duplicates. This is another reason to also fix issue #1097 because using the default PM mapping file will eliminate this issue. 

The screen shot below is only some of the fields that are declared duplicates by the program.

this is on seeddemostaging with latest develop as of 10/17/2016 at 8:59 am

[...]

Issue #1085, opened 2016-10-10
Duplicate Fields in Mapping -- can't assign them to a Tax Lot or Property table


RDmitchell:
When the program finds a duplicate field, it puts them at the top of the Mapping screen. 

In V2.0, we need to allow the user to resolve the duplicates by changing the field names, and then also have a way for the user to assign those fields (presumably after resolving the duplicates?) to the appropriate table, either Tax Lot or Property. 

Right now, you can't do that (code as of 10/10/2016)

![image](https://cloud.githubusercontent.com/assets/6314950/19246284/62fc91d6-8eda-11e6-8586-83859d372f97.png)

nllong:
I think that this will disappear once we have unique names. If we enforce uniqueness, then we could remove the need for this "duplicates" section in the table.

RDmitchell:
For now, it seems that no duplicates are being generated in mapping. However, if there ever are, this will still be an issue. Am closing with the idea that if it comes up for a specific case, we'll open up another issu.e

RDmitchell:
Just imported Kansas City building list and see duplicates. This is on the pm import branch from Nick. Maybe it doesn't have these fixes in it? Am going to reopen this just to make sure it gets retested when all the code is merged. 

SHA: 1d26e16
instance: seeddev
org: LBNL 202
browser: chrome, incognito

![image](https://cloud.githubusercontent.com/assets/6314950/20845971/520cc118-b87c-11e6-8fc7-3e48b8278561.png)

RDmitchell:
see files here
https://drive.google.com/open?id=0B3fTKpZ9Dx7LNkxXX0x1b3p4LUU

PR #1106, opened 2016-10-28
1090 duplicate field names

nllong:

Mapping would sometimes return duplicate mapping results and the suggestions were sometimes way off.
#### What's this PR do?

This PR allows for only one unique mapping result per set of fields. If there are duplicates, then it will look at all the mapping suggestions and decide the next best one to pick based on the confidence of the match.

In addition, if a column has no suggestions or all the suggestions are removed, then the mapping suggestion will be set to the raw column name and end up as extra data.

Go through the mapping process with the covered building sample and the new property/taxlot examples.
#1085 #1090

N/A
#### Definition of Done:
- [X] Does this add new dependencies? If so, does PIP, npm, bower requirements need to be updated?

### DIFF
--- a/seed/data_importer/tasks.py
+++ b/seed/data_importer/tasks.py
@@ -1003,6 +1003,7 @@ def filter_duplicated_states(unmatched_states):
 
 
 class EquivalencePartitioner(object):
+
     @classmethod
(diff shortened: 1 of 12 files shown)


RDmitchell:
Thanks @axelstudios . I should have time to do some testing tomorrow.

PR #1116, opened 2016-11-08
Rewire column mapping suggestions

Myoldmopar:
@nllong added a small worker function to get column mapping suggestions to allow it to work during the backend changes.  A new, viewset-based function was added and also called this worker function.  The original function was left in there and the front end was still hooked to it.

#### What's this PR do?
This PR moves the front end to call the new class-based "v2" function, also updating the docs mention of the function.

Verify that column mapping suggestions is still working as expected.

https://www.pivotaltracker.com/story/show/133876497

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -41,7 +41,7 @@ class TestReverseViewSet(viewsets.ViewSet):
         i = dict()
 
         # try one from the main SEED app, which appears in the url as "/app/whatever", but it is namespaced to "seed"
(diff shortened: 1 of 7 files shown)


nllong:
need to fix this tests: test_get_column_mapping_suggestions_with_columns (seed.tests.test_views.TestMCMViews)

Myoldmopar:
@nllong Problem should be solved now.

Myoldmopar:
Like I was saying.  Problems should be fixed _now_.

PR #1119, opened 2016-11-11
1090 exclude fields

nllong:
Subsequent fix for 1090

#### What's this PR do?
Don't match on data_state field. Makes front end look at both inventory type and field name for duplicates. Check and flag duplicate rows across those two items and disable the map data button.

import data and manually create duplicates to see if behaviour is as expected.

#1090 

N/A

#### Definition of Done:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .project
 .pydevproject
 .coverage
+.noseids
 *~
 *.pyc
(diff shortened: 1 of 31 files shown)


RDmitchell:
Any idea when this can get merged to develop? When it is we will deploy to staging and I can test.

PR #1122, opened 2016-11-11
Import Files Endpoint and Associated Changes

Myoldmopar:
This will ultimately update the import_files related endpoints into API v2 including Swagger.

#### Definition of Done:


Myoldmopar:
- [ ] /app/get_first_five_rows/ - > /api/v2/import_files/1/first_five_rows/
- [ ] /app/get_import_file/ -> /api/v2/import_files/1
- [ ] /app/get_raw_column_names/ -> /api/v2/import_files/1/raw_column_names
- [ ] /app/save_raw_data/ -> /api/v2/import_files/1/ POST  (BIG CHANGE)
- [ ] /app/search_mapping_results/ -> /api/v2/import_files/1/filtered_mapping_results/ POST
- [ ] /app/start_mapping/ -> /api/v2/import_files/1/perform_mapping POST
- [ ] /app/start_system_matching/ -> /api/v2/import_files/1/perform_matching POST
- [ ] /cleansing/download/ -> /api/v2/import_files/1/cleansing_results.csv GET
- [ ] /cleansing/results/ -> /api/v2/import_files/1/cleansing_results.json GET

Myoldmopar:
So...the unit test for the `/app/search_mapping_results/' endpoint was `@skip`ped because it focuses on the canonical building stuff.  I don't know the path to change it to use the new backend, but that definitely needs some attention to make sure I got this change made right.

- [ ] Fix unit test to cover the `/api/v2/import_files/1/filtered_mapping_results/` endpoint

Myoldmopar:
So...the start_mapping endpoint is executing a background task using delay.  No problem.  But when I make a call to this endpoint on Swagger, it is returning null.  Presumably the background task is still running, but it isn't an interactive thing. I probably need to consult with @nllong or @axelstudios to see how to best handle this.  Also, I couldn't find a unit test for this one.

- [ ] Fix return from `/api/v2/import_files/1/perform_mapping/` endpoint, figure out test

Myoldmopar:
So...`start_system_matching` has the same issues as `start_mapping`.  Need to figure out how to handle it.  Here's the same comment, copied in.

> So...the start_mapping endpoint is executing a background task using delay. No problem. But when I make a call to this endpoint on Swagger, it is returning null. Presumably the background task is still running, but it isn't an interactive thing. I probably need to consult with @nllong or @axelstudios to see how to best handle this. Also, I couldn't find a unit test for this one.

- [ ] Fix return from `/api/v2/import_files/1/perform_mapping/` endpoint, figure out test

Myoldmopar:
- [ ] Add protection in views that hit _map_data for invalid pk.

Myoldmopar:
The cleansing views have all been moved into import_files.  Thus the views file and the urls file is empty.  We should decide at some point whether we remove the entire cleansing folder or not.  At this point there are only tasks and models that would need to be moved.

As for the actual move, there are unit tests in place for them, and they are succeeding, and I don't think I really changed anything structural, so it should be good.

Myoldmopar:
Need to update `mapping.rst` to reflect new endpoint names.

Myoldmopar:
So...I was going to update `/app/save_raw_data` to become `/api/v2/import_files/1 (POST)` as a de facto "create" method.  However, that's not really what this function does.  It doesn't take the information needed to create a new import file, it just takes an ID of an already existing import file and saves it.  So I left it as a standalone verb.

`/app/save_raw_data/` to `/api/v2/import_files/1/save_raw_data`

This completes the bulk of the work here, the only remaining items being the things I'd like some help with from @nllong or maybe @axelstudios as I mentioned in the comments above.

rebrownlbnl:
Looks like a lot of progress on the v2 APIs.
Your notes sound like you had to change some of the endpoints on the fly as you got into the implementation, and I'm just wondering how this compares to Nick's original endpoint spreadsheet:
https://docs.google.com/spreadsheets/d/1F0fDjvGccc8Tw_jiY3-BmJJ5IONRYPgpOoSooRavIfs/edit#gid=1820060935.

Does that document need to be updated to reflect the final endpoints, or is the Swagger documentation all we need for the v2 APIs?

Myoldmopar:
It may not hurt to update it; for speedy-purposes we started a new spreadsheet to just track the remaining pieces.  I'll pull in those changes to the original one so we have one centralized location.

Myoldmopar:
OK, this one is done enough.  If tests pass, I'd love someone to monkey test the front end since the unit testing isn't completely covering the views.

seed/data_importer/tasks.py
```diff
@@ -504,6 +504,8 @@ def _cleanse_data(file_pk, record_type='property'):
 def map_data(file_pk, *args, **kwargs):
     """Small wrapper to ensure we isolate our mapping process from requests."""
     _map_data.delay(file_pk)
+    prog_key = get_prog_key('map_data', file_pk)
+    return {'status': 'in process...', 'progress_key': prog_key}
```
nllong: need to verify the 'status' message here.

Myoldmopar: Did you want me to take out the status key?  Or change it to progress?  Or nothing?  The other conflicts are now taken care of, so if your answer is _nothing_ then this is ready to drop.

PR #1130, opened 2016-11-19
Root API things

Myoldmopar:
Some endpoints in the root API needed to be moved into the /api/v2 namespace.  Or otherwise cleaned up.

#### What's this PR do?
- [x] `/app/schema`
- [x] `/app/version`
- [x] `/app/progress`
- [x] s3 uploader stuff

The endpoints are covered in the core python API tests, so the dashboard should be sufficient.

133876237


nllong:
looks like spelling and python tests are failing

nllong:
@Myoldmopar -- can you merge up develop into this?

Myoldmopar:
Resolved conflicts; testing now and I'll push when the tests pass.

Myoldmopar:
Two unit tests are failing.  `test_get_property_columns` is failing [here](https://github.com/SEED-platform/seed/blob/133876237-RootAPIThings/seed/tests/test_views.py#L3628) because the `related` key is coming back as `True`, but the [expected value](https://github.com/SEED-platform/seed/blob/133876237-RootAPIThings/seed/tests/test_views.py#L3625) is `False`.  The taxlot test is probably the same problem.  Any thoughts why @nllong?

Myoldmopar:
Yeah, in [this line on Travis](https://travis-ci.org/SEED-platform/seed/jobs/182030574#L1339) it shows the same issue.  The hash that should match is the last one on the line, but it doesn't because `related` doesn't match.  I don't think that issue is on my end, so just let me know if I can help.

nllong:
i fixed that test on my branch. i'll merge it into mine and test it out.

Issue #1103, opened 2016-10-19
Mapping Review: only show fields that were mapped


RDmitchell:
Just tested latest develop (as of 01/19/2016 6:43 am) on staging

The Mapping review screen is better, but not quite right yet. 
- Import the file called "example-data-taxlots.xlsx"
  https://drive.google.com/open?id=0B3fTKpZ9Dx7LcXFQWXI0QlVkMDA

![image](https://cloud.githubusercontent.com/assets/6314950/19530829/4bf0de04-95ea-11e6-804b-3e7ae62a1116.png)
- Map the fields as shown below -- mapping "number_buildings" to Tax Lot should make that a field that gets put into extra data
  ![image](https://cloud.githubusercontent.com/assets/6314950/19530898/9a6f5d3a-95ea-11e6-8520-528600959e46.png)
- Click the Map Your Data button
- Program displays all the fields in the Tax Lot table, even if the user didn't map to them -- extra data field is displaying so that is good !!
  ![image](https://cloud.githubusercontent.com/assets/6314950/19531158/7f2b4f74-95eb-11e6-8a2a-0e12d60616fd.png)
- In V 1.5, the program only displayed the fields that were actually mapped. I would argue that is what we should do in V 2.0 also, so the Mapping review screen would look like this instead

![image](https://cloud.githubusercontent.com/assets/6314950/19531091/3d471dd6-95eb-11e6-81d6-55dcac02640f.png)

This was in issue #1084 but I deleted it from there and made a separate issue for it.

RDmitchell:
Same file as shown above, imported, mapped as shown below, has the following issues

1 - Fields that were not mapped are being displayed, which shouldn't be
2 - Associated Building Tax Lot ID is being displayed, which should not be
3 - Number of Buildings (the one field mapped to the Propety table) is not being displayed in the View by Property tab (it should be the only field that is displayed under that tab based on the mapping of this file. 

Mapping
![image](https://cloud.githubusercontent.com/assets/6314950/20855546/08b0ff3e-b8b3-11e6-94af-cfa7449f1bf4.png)

Mapping Review screen 
![image](https://cloud.githubusercontent.com/assets/6314950/20855522/c3b7f4aa-b8b2-11e6-9321-7840ec64aa38.png)

SHA: 1d26e16
instance: seeddev (Nick's branch)
org: LBNL 20
file: example-data-taxlots
browser: chrome, incognito

RDmitchell:
SHA: 341f613
[...]

Issue #1100, opened 2016-10-18
Mapping: extra data fields shown in Mapping Review screen with field name not display name


RDmitchell:
When importing to sample Tax Lot file, the number of buildings field is mapped so that it becomes extra data. In the final mapping screen, that field is shown in the column with the field name not the "display name" given in the Mapping screen. 

on seeddemostaging with latest develop as of 10/17/2016 at 8:59 am

Steps to reproduce
- Import sample tax lot file (link from #1099):    https://drive.google.com/open?id=0B3fTKpZ9Dx7LZkpJcG4taGtsaFk
- Map it as follows -- setting the "number_buildings" field to the Tax Lot table in the pulldown will make it a field that is added to the Tax Lot table extra data field
  ![image](https://cloud.githubusercontent.com/assets/6314950/19461582/a05ee428-949a-11e6-85f1-bb01e37aea42.png)
- Click the Map Your Data button
- See this mapping in the Mapping Review screen -- notice that the field that should have a column named "Number of Buildings" instead has a column name of "number_buildings", ie, it is showing the actual field name, rather than the "display name". 
  ![image](https://cloud.githubusercontent.com/assets/6314950/19461617/eef247b0-949a-11e6-9a20-c1bccf89a364.png)

RDmitchell:
This might be related 1094

RDmitchell:
SHA # 27603f6 on staging
I still see this same issue -- extra data shows up as "number_buildings" instead of "Number of Buildings"

nllong:
This will be address in this ticket #1146.

RDmitchell:
SHA: 341f613
instance: seeddemostaging
browser: chrome, incognito
files: covered-buildings-sample.csv & portfolio-manager-sample.csv (standard sample files)
org; LBNL 300 (created new for this test)

In the mapping review screen, this seems to no longer be an issue

PR #1154, opened 2016-12-09
1103 show only mapped fields

nllong:
In the mapping review screen all the columns were being shown instead of just the columns that have been mapped. The extradata fields were using the old field names, not the new ones.

#### What's this PR do?
1) shows only mapped columns in the mapping review screen
2) debug logger now outputs file and line number of log messages
3) added property_type on the PropertyState model as a native field.
4) fix extra_data display names.

Unit tests and monkey testing

#1100 
#1103 

#### Definition of Done:
- [X] Does this add new dependencies? If so, does PIP, npm, bower requirements need to be updated? **YES, run ./manage.py migrate**


nllong:
@axelstudios -- when you get to this, notice that the extra data fields are in the JSON request, but not always showing up in the grid.

image

PR #1181, opened 2017-01-04
135787537 performance updates

danielbressan:
SEED performance updates. reduce loops and postgres calls in django by pulling full tables of needed data and letting postgres do more work. taxlot and property calls only return viewable rows (this is still a GET though and should changed to a POST).


RDmitchell:
Will these changes result in a substantial increase in performance for organizations with large data sets, such as NYC with 13K+ records? If so, how big of a performance increase is expected? I would like to show NYC the performance improvements if they are significant, but otherwise, probably won't. Please advise. Thx

axelstudios:
@RDmitchell This should be a decent speedup for displaying the NYC data because now the backend only has to send the data that's visible.  So instead of ~800 columns of data for every record, now it only sends what the columns that are selected by the user, so ~50 by default

RDmitchell:
@axelstudios -- great. If we can get this into develop today or Monday, I will test it out -- the webinar with NYC is on Tuesday. Think we can get it merged into develop by EOD on Monday? Thx !

axelstudios:
@RDmitchell Merged!  And deployed to seeddev currently.

RDmitchell:
Doesn't look like there is data for NYC on seeddev so will put this code on staging for testing.

Issue #1182, opened 2017-01-05
Can't create new data set for new organization


RDmitchell:
Instance: seeddemostaging
SHA: 3a6051b
browser: chrome, incognito
priority: 1 -- can't do anything if you can't make a dataset to import a file
* This may not seem important because users all have existing organizations but it will make my testing almost impossible if I can't make new organizations to test from scratch
* Also this is a new bug, because I have been making new organizations when I test for months

Steps to reproduce
* Created a new organization and logged into it
* Clicked the Data nav button, then clicked Create Dataset
* Got these errors

angular.js:12261 GET https://seeddemostaging.lbl.gov/api/v2/cycles/?organization_id=256 403 (Forbidden)
(anonymous) @ angular.js:12261
sendReq @ angular.js:12023
serverRequest @ angular.js:11818
processQueue @ angular.js:16692
(anonymous) @ angular.js:16708
$eval @ angular.js:17990
$digest @ angular.js:17804
$apply @ angular.js:18098
(anonymous) @ angular.js:26585
dispatch @ jquery.js:4435
r.handle @ jquery.js:4121

![image](https://cloud.githubusercontent.com/assets/6314950/21664762/773e2068-d29d-11e6-82f6-f8201ea634fb.png)

RDmitchell:
ah, right, stupid me, I forgot to add a Cycle first.

RDmitchell:
Fixed SHA: 362adf5

rebrownlbnl:
Probably should add a warning to the user if they don't specify a cycle.

PR #1190, opened 2017-01-05
upon org creation, create a default cycle

nllong:
There was an error when new organizations were created that data sets could not be added. This was because there was not a cycle available. 

#### What's this PR do?
When an organization gets created a Default {year-1} Calendar Year cycle will be generated automatically.

Create org, then create a data set. Make sure that no error is thrown.

#1182 

#### Definition of Done:

### DIFF
--- a/config/management/commands/create_default_user.py
+++ b/config/management/commands/create_default_user.py
@@ -4,7 +4,6 @@
 :author
 """
 # stdlib
-from datetime import date, datetime, timedelta
 from optparse import make_option
 
 # Django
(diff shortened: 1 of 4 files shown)


RDmitchell:
ah, right, stupid me, I forgot to add a Cycle first. Another option would be to put up a message telling the user that they needed to create a cycle first. But this is nicer. Thanks.

nllong:
yeah, but it shouldn't have crashed... it should have just had an empty cycle drop down. This will prevent it from ending up in the weird state.

RDmitchell:
yes, thanks !!

PR #1202, opened 2017-01-11
Fix some failing tests that snuck in

Myoldmopar:


Myoldmopar:
@nllong What would cause the `related` term in the response to be missing keys?  That seems to be the issue on all the tests at this point.

nllong:
not sure to be honest. I thought the related key was always present in the property and taxlot endpoint.

Myoldmopar:
OK, so @nllong @axelstudios this last commit appears to fix all the unit tests.  Could you verify that the front end still plays nicely with it?

Issue #1111, opened 2016-11-01
Server times out mapping large data set


RDmitchell:
In working with NYC data, I have the following problem.

I have loaded a Tax Lot table with 13,678 records with approx 25 fields, most of which are in extra data.
This works without issue
I then try to map a Portfolio Manager table as a "property" table, which has 13,052 records and 259 fields.
The program is able to display the first mapping screen (where all the fields are displayed with their SEED mapping), but when I click on the "Map Your Data" button, the server times out while the program is trying to display the grid showing the mapped fields with the actual data. 

The timeout on the server is set to 300 seconds (5 minutes). This seems like it should be a long enough timeout setting??

RDmitchell:
Miles changed the server timeout to 10 minutes, and this step still timed out.

RDmitchell:
SHA: 3e9c017
instance: seeddemostaging
browser: chrome, incognito
Folder with description and files
https://drive.google.com/open?id=0B3fTKpZ9Dx7LN3BUejJ3ZU1tY0E

This is still an issue -- I let the program run overnight trying to map the 2nd file, the Portfolio Manager file, and when I came in this morning, this was the state:

![image](https://cloud.githubusercontent.com/assets/6314950/21364398/87fe45b2-c6a5-11e6-83d2-a9c580e593f0.png)

axelstudios:
It turns out this isn't a timeout, but a data issue.  It's a corner case that just needs to be handled, this will be fixed soon

RDmitchell:
Aha !!! Great, thanks for looking into this. Glad it's actually a data issue !

RDmitchell:
SHA: f59b579
instance: seeddev
browser: chrome, incognito
org: LBNL 20 (my test org)

Fixed with PR #1205 
I was able to import both the NYC Covered Building List and their Portfolio Manager files, both which are 13K+ records. 
Haven't checked all the data, but looks like the program was able to figure out the tax lot / property associations also.

THANKS @nllong  !!!! It is so great that you fixed this !!!!

PR #1205, opened 2017-01-12
increase length of lot_number

nllong:
The lot number was a charfield at 255 and some of the delimited lot_number fields were > 255.

#### What's this PR do?
Convert lot_number to TextField. Catch error when trying to save a model so that it doesn't fail the entire import.

Import large data set.

#1111 

#### Definition of Done:
**run migrations!**


RDmitchell:
Thanks @nllong -- I'll test this whenever you get it onto seeddev !

PR #1266, opened 2017-02-22
Hand matching

axelstudios:
Functional hand matching branch, with several bug fixes outside of matching


seed/tests/test_views.py
```diff
@@ -2008,70 +2001,6 @@ def test_save_match_audit_log(self):
         self.assertEqual(audit_log.action_note, 'Matched building.')
         self.assertEqual(audit_log.audit_type, LOG)
 
-    @skip('Fix for new data model')
```
nllong: we really need to clean up any old @skips

nllong:
On pairing page. If I import the covered-buildings-sample file, map UBI to jurisdiction_tax_lot_id, then drag over one of the properties onto the taxlots, then it does this:

image

nllong:
query to see matching results seems quite slow. we can address the performance later, but I wanted to make sure that we are aware of it.

Issue #1213, opened 2017-01-14
Some data getting mapped is not displaying in detail and list view


RDmitchell:
SHA: bd59264
Instance: seeddev

See this doc
1213 - Some data getting mapped is not displaying in detail and list view
https://drive.google.com/open?id=1trcdpJ7zGNEN832YJFyv21z597oL8bG5m8VQaDBQZy8

in this folder (folder also has the sample data file)
https://drive.google.com/open?id=0B3fTKpZ9Dx7LZmtYNkExNjEyMms

RDmitchell:
I think the issues shown in the doc may need to be resolved before releasing V 2.0

RDmitchell:
This is still an issue -- see the above doc, updated from testing on seeddev with the V2.0 release SHA. Just a few fields that are not either available in the Settings view to select, or if they are there, there is no data displayed when there should be data.

RDmitchell:
SHA: 9e6e90e
instance: seeddev
browser: chrome incognito
org: LBNL 400
files: Issue 1213 Data file-01.xlsx & Issue 1213 Data file-02.xlsx
Link to folder:
https://drive.google.com/open?id=0B3fTKpZ9Dx7LZmtYNkExNjEyMms

I thought I figured out a way to make the program properly import/display the data with problem fields by renaming the fields but if you read to the very end, you will see my conclusion was actually not correct. Just goes to show what happens when you try to understand / debug based on empirical data. Hopefully someone can look at the behavior I am seeing, and then the code, and make some sense out of it and fix the problem. 

Steps to reproduce:
* Import  Issue 1213 Data file-01.xlsx
![image](https://cloud.githubusercontent.com/assets/6314950/22229826/fab7c200-e18e-11e6-85ac-cface5978a9d.png)
* Map fields "Square Footage OPA" and "Gross Floor Area" to the same names as in the file.
![image](https://cloud.githubusercontent.com/assets/6314950/22230881/0b70d9dc-e195-11e6-842e-5d6fc1f87b83.png)
* Matching Screen -- these fields do not show up in the matching screen
![image](https://cloud.githubusercontent.com/assets/6314950/22229982/fb8929a2-e18f-11e6-91f9-67f92b4a0733.png)
* Inventory -- the field names are available in List Settings but no data is displayed.
![image](https://cloud.githubusercontent.com/assets/6314950/22230069/725fc842-e190-11e6-9e47-9da75dd8a207.png)
* Delete inventory, data set, start again
[...]

Issue #1191, opened 2017-01-05
Return display names from the backend


nllong:
The conversion of database fields to user displays exists in several places in the code base. Need to standardize on one solution and implement in the backend.

This will most likely result in adding a display name field to the columns table in the database.

RDmitchell:
Yes, it seems like adding a display name to the columns table makes sense. 

How does this interact with what the user sees in the Mapping pulldown list? Can the display name and the pulldown list name be the same? I guess this gives the option of not having them same when needed, but the default would be that they would be the same.

nllong:
yeah that would be the best approach... users would map to display names, not database field names.

RDmitchell:
Yup -- mapping and display names are one and the same. That should be what we do ... but what does that mean for 2.0 vs 2.1? We don't want to have to do remapping of data (if we can help it) after 2.0 (although maybe it's unavoidable given our time constraints).

RDmitchell:
@nllong -- the one field that is really weird in mapping is when you have a Tax Lot ID in a property file, and you want to set it to be a matching field in the Tax Lot table (ie, matching it to Jurisdiction Tax Lot ID), you map the field to "Lot Number", and I think apply it to the Property table (would have to I think). 

Seems like this field name (at least the display name / mapping name?) should be "Jurisdiction Tax Lot ID", but the program knows it is in the Property table and thus it will match on it. 

We don't have to do this now, I guess, because it would probably mean changing the matching code, and we don't have time to do that. And maybe it doesn't matter from the BE, because the field name would continue to be "Lot number", only the FE mapping name would change to "Jurisdiction Tax Lot ID". 

Anyway, something to ponder for the future I guess.

nllong:
#1230

RDmitchell:
Comment from #1151 which I have closed
*****
SHA: 3e9c017
instance: seeddemostaging
browser: chrome, incognito
file: covered-buildings-sample-small.csv, from #1126
https://drive.google.com/open?id=0B3fTKpZ9Dx7LMHgwYTloc0lGOTQ

Looking at the pm-mapping.json file, seems like the display names are not getting picked up for some reason.

![image](https://cloud.githubusercontent.com/assets/6314950/21337120/56995862-c61f-11e6-98d3-b09782df93fc.png)

![image](https://cloud.githubusercontent.com/assets/6314950/21337141/8a261d14-c61f-11e6-946d-2e9ce4e60b2b.png)

[...]

PR #1270, opened 2017-03-02
1213 migrate columns

nllong:
Legacy databases have lots of columns, most of which are not used. Some of the columns that are used are out of state. Some of the column mappings point to the same raw columns and mapped column which is no longer possible because the mapped columns need to specify which table they are being mapped to.

#### What's this PR do?
This PR includes 4 migration files and they are run independently to handle the transactions correctly, that is the result of the first migration is needed for the second migration and that won't happen unless the database transaction is committed.
1. Break up column mappings that point to the same raw and mapped column. Creates a new column and assigns that column to the raw column.
2. Deletes any columns that are not referenced in any column mapping.
3. Split column mappings that have multiple raw columns or mapped columns. Note that this will only split the mapped columns if they are extra_data. If the data are database fields then it is okay to have raw columns pointing to the same database field.
4. Clean up the columns. If the column is a raw column, then remove the table_name and makes sure that the extra_data flag is false. If it is a mapped column and there is no table_name, then set it to PropertyState by default.

Import fresh copy of database and run migrations. Monkey test

#1213 #1191 

#### Definition of Done:


RDmitchell:
will this have any impact on the old building snapshot table that we are currently using to migrate data? At some point that table will no longer be needed, but at this point, I think we may need to keep it for a few more months.

nllong:
This will not impact the BuildingSnapshot table, only the columns. The old mappings will stay intact, just broken apart if the `from` and `to` column were the same.

PR #1331, opened 2017-04-21
Old data quality

danielbressan:
#### Definition of Done:


RDmitchell:
seems like this might need a bit of monkey testing?

PR #1371, opened 2017-06-01
Release 2.1

nllong:
Merge down code for most release.


RDmitchell:
@nllong -- should we be putting all these changes into the 2.1 release? I was hoping that nothing would be changing from what was already done, as I don't have much more time to test and want to get it onto the production server over the weekend.

PR #1379, opened 2017-06-08
GBR merge 02 api helpers

fablet:
#### What's this PR do?
Adds helper classes and functions to simplify the creation of viewsets via an OrganizationUser based permissions class, a JSON rendered tailored to SEED response style, and a mixin to simplify filtering viewsets by Organization
#### Definition of Done:


seed/authentication.py
```diff
@@ -16,17 +18,28 @@ class SEEDAuthentication(authentication.BaseAuthentication):
...
 
         try:
-            from seed.landing.models import SEEDUser as User
+            if not auth_header.startswith('Basic'):
+                raise exceptions.AuthenticationFailed(
+                    "Only Basic HTTP_AUTHORIZATION is supported"
+                )
+
+            auth_header = auth_header.split()[1]
+            auth_header = base64.urlsafe_b64decode(auth_header)
```
nllong: this was a previous pull request and we never approved it because this broke the site. I can't recall why at the moment, but it seemed that base64'ing the auth header was not needed because we are running the site over https. If we base64 the auth header, then many of the tests broke.

seed/lib/superperms/orgs/permissions.py
```diff
@@ -0,0 +1,166 @@
+# !/usr/bin/env python
+# encoding: utf-8
+"""
+:copyright (c) 2014 - 2017, The Regents of the University of California,
+through Lawrence Berkeley National Laboratory (subject to receipt of any
+required approvals from the U.S. Department of Energy) and contributors.
+All rights reserved.
+:author Paul Munday<paul@paulmunday.net>
+"""
+from django import VERSION as DJANGO_VERSION
```
nllong: can you add a header to this file describing what its intent is?

seed/lib/superperms/orgs/permissions.py
```diff
@@ -0,0 +1,166 @@
...
+            request.user and
+            (
+                is_authenticated(request.user)
+                or not self.authenticated_users_only
+            )
+            and self.has_perm(request)
+        )
+
+
+class SEEDPublicPermissions(SEEDOrgPermissions):
```
nllong: did this get moved from another part of the code?

seed/tests/test_api.py
```diff
@@ -57,7 +59,10 @@ def setUp(self):
             'building_id': cb.pk,
             'organization_id': self.org.pk,
         }
-        self.auth_string = '%s:%s' % (self.user.username, self.user.api_key)
+        auth_string = base64.urlsafe_b64encode(
```
nllong: i don't think this is necessary, especially if we are forcing HTTPS

seed/authentication.py
fablet: I have a reply regarding this from Paul indicating that this was added because base64 encoding is part of the spec for using basic auth.  https://tools.ietf.org/html/rfc7617

seed/lib/superperms/orgs/permissions.py
fablet: To the best of my recollection and comparisons against develop, this is new code.  However, you may be thinking of the superperms/orgs/decorators which should be carrying the same or similar functionality just, obviously, as decorators rather than class level permissions.

seed/utils/viewsets.py
```diff
@@ -0,0 +1,64 @@
...
+class SEEDOrgCreateUpdateModelViewSet(OrgCreateUpdateMixin,
+                                      SEEDOrgModelViewSet):
+    """Extends SEEDModelViewset to add perform_create method to attach org.
+
+    For use on models that have a foreign key relationship to Organization via
+    fieldname 'organization'. Models using 'super_organization' or having
+    additional foreign key relationships, such as user, should extend
+    SEEDOrgModelViewset and override perform_create/perform_update in the view.
+    """
+    pass
```
fablet: actually no.  The perform_create and perform_update methods are implemented in this base class (see seed.utils.api.OrgCreateMixin and OrgUpdateMixin).  I have updated the docstring in an attempt to make the usage a little clearer

seed/authentication.py
nllong: Hmm, it is part of the basic auth standard, but this seems to break quite a few things in the testing world right now. Are we able to get all the tests working again on travis if this is enabled?

nllong:
looks lie the tests are passing. I am going to rely on the tests heavily because we have updated them quite a bit over the last couple months to cover the major functionality. 

https://travis-ci.org/SEED-platform/seed/jobs/241387090#L310

django_filter - issue.

Also, I'm going to merge these all into a new `gbr` branch in order to test the functionality together before the final merge into develop. Can you change the base to `gbr` and feel free to pull up that branch into this one.

Thanks!

fablet:
I assume you meant to say tests are _not_ passing based on the results from the travis link you posted.  
When I run the tests locally per Testing Notes (python manage.py test --settings=config.settings.test) everything passes, but I have django_filter installed in my local virtualenv.  However, I can see from the Travis job log that it is not installing any of the updated requirements.

Myoldmopar:
Travis cache issue @nllong ?

seed/authentication.py
fablet: Since these tests pass on my local machine, I am reasonably certain that the tests will work on travis once the issue of travis not using the updated requirements for this branch (specifically getting travis to install django-filter) is resolved.

PR #1381, opened 2017-06-08
GBR merge 03 residential cert endpoints

fablet:
This branch depends on the files added in the first 2 GBR feature branches.
#### What's this PR do?
Adds api endpoints for GreenAssessments
#### Definition of Done:


seed/serializers/certification.py
```diff
@@ -0,0 +1,360 @@
...
+    class Meta:
+        model = GreenAssessmentURL
+
+
+class GreenAssessmentPropertyReadOnlySerializer(serializers.BaseSerializer):
+    """Simple read only Serializer describing Green Assessment attached to
+    property. Use with prefetch_related to avoid extra database calls.
+    """
+
+    def to_representation(self, obj):
```
nllong: looks like this is to_dict, right? Should we just call it to_json?

fablet: Using to_representation is the correct method for customizing the serialization of a serializer class 
http://www.django-rest-framework.org/api-guide/serializers/#overriding-serialization-and-deserialization-behavior

PR #1382, opened 2017-06-08
GBR merge 04 a org mixin cycle

fablet:
This branch depends on the files added in the first 2 GBR feature branches.
#### What's this PR do?
Updated Cycle endpoint using new helper classes, and added filterset to allow filtering on start (start_lte), end (end_gte) or year (calendar year).
#### Definition of Done:


seed/filtersets.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
+# encoding: utf-8
+"""
+:copyright (c) 2014 - 2016, The Regents of the University of California,
```
nllong: 2017

seed/lib/superperms/orgs/models.py
```diff
@@ -140,14 +139,16 @@ def save(self, *args, **kwargs):
         # Create a default cycle for the organization if there isn't one already
         from seed.models import Cycle
         year = date.today().year - 1
-        cycle_name = 'Default ' + str(year) + ' Calendar Year'
+        cycle_name = '{} Calendar Year'.format(year)
         if not Cycle.objects.filter(name=cycle_name, organization=self).exists():
```
nllong: I know that I updated this in the bricr-models branch. Shouldn't conflict too much though. Any reason to remove the "Default"? You just want it to be cleaner?

seed/models/properties.py
```diff
@@ -245,9 +245,10 @@ def save(self, *args, **kwargs):
 class PropertyView(models.Model):
     """Similar to the old world of canonical building."""
     # different property views can be associated with each other (2012, 2013)
-    property = models.ForeignKey(Property, related_name='views')
-    cycle = models.ForeignKey(Cycle)
-    state = models.ForeignKey(PropertyState)
+    property = models.ForeignKey(Property, related_name='views',
+                                 on_delete=models.CASCADE)
+    cycle = models.ForeignKey(Cycle, on_delete=models.PROTECT)
```
nllong: nice!

nllong:
need to get tests passing. I'm fine merging these all into one branch (i.e. `gbr`) and fixing tests. Let me know what you would like to do

seed/lib/superperms/orgs/models.py
fablet: I was noticing that I was ending up with 2 default cycles anytime I was created an new organization, one from this point in the code and one from seed/management/commands/migrate_organization.py
Both places are checking for an existing cycle for the calendar year based on cycle name, but one was using "{year} Calendar Year" and one was using "Default {year} Calendar Year".  Since one of the two cycle names needed to be changed to insure only one calendar year is created per org creation, removing the "Default" from this calendar year name seemed the way to go since it does also make it cleaner, and brings it in line with an expedited filtering by calendar year through filtersets.

fablet:
I changed the pull request base on this to gbr, and also rebased this branch off of my 02 branch to make it easier for testing.  Once the issue with Travis not installing from the updated requirements in that 02 branch is resolved, these tests should pass

PR #1386, opened 2017-06-12
Filter matching

danielbressan:
#### What's this PR do?
1) Creates merged coverage from python and functional tests. Increases run time of functional tests by 5-7 mins. Could reduce that if we went to a different coverage tool than coveralls (like codecov). 
2) Updates to filter and sorting for hand matching detail and pairing pages. Also updates to css for pairing page.
3) Functional tests for (2)
See coverall reports for combined coverage. See matching detail and pairing pages for filter and sort updates. 
#### Definition of Done:


tox.ini
```diff
@@ -15,13 +15,11 @@ skipsdist=True
 [testenv:python]
 commands=
   coverage run manage.py test seed
-  coveralls
```
nllong: so coverage is only reported when running the functional tests now?

danielbressan: yes because coveralls can't take more than one report per commit. Other coverage tools like codecov can and merge automagically if we switch off coveralls.

nllong: 👍

PR #1389, opened 2017-06-15
GBR merge 05 properties

fablet:
    There is a significant difference between the PropertyViewSet and the one created by us, most notably, the inclusion of the create and update endpoints in our branch, vs the additional endpoints (pair, unpair, etc) in yours.  For the moment we made a change to the name of our viewset and added an extra endpoint, so that these could be evaluated to determine whether we wanted to merge functionalities into a single PropertyViewSet, or maintain two separate viewsets.  If the later, suggestions for a more generic naming choice (vs gbr_properties) would be welcomed.
#### What's this PR do?
    Adds full ModelViewSet endpoints for Property, PropertyState and PropertyView 
#### Definition of Done:


nllong:
@fablet thanks for submitting all these. I have confirmed that the tests for PR's 1-4 are passing. I'm looking into the issue travis at the moment. Thanks for your patience.

Issue #1418, opened 2017-07-20
Date sometimes does not display correctly in the Detail view


RDmitchell:
instance: seeddemostaging
SHA: f836a69

See the files in this folder for more details
https://drive.google.com/open?id=0B3fTKpZ9Dx7LNTgzRWIweTVPMTQ

RDmitchell:
instance: seeddemostaging
SHA: 1b9d15a
org: LBNL 403
from the folder above

This is fixed for new files imported and existing files.

PR #1421, opened 2017-07-21
Added a date filter for date objects on the inventory detail page

axelstudios:
Date fields (currently just `Recent Sale Date`, `PM Generation Date`, and `PM Release Date`) are now processed with a date filter on the front end to convert to local time on the inventory detail page.

`2016-08-27T03:41:49Z` becomes `2016-08-26 9:41 PM`, and you can still click on the field to see the raw timestamp value:


RDmitchell:
I think this seems fine.

RDmitchell:
see issue #1418 - tested and closed

Issue #1416, opened 2017-07-18
Can't filter on dates using operators


RDmitchell:
Right now you can filter on dates (if you know the right format to use in the filter) but you can't use operators.

In the example below, there are records with PM Release Date > 03-04-2016.
![image](https://user-images.githubusercontent.com/6314950/28296619-23663452-6b1e-11e7-914c-e08135a3cfd1.png)

Just filtering on 03-04-2016 (entered as 2016-03-04), the program returns the correct number of records
![image](https://user-images.githubusercontent.com/6314950/28296634-3f41c8f8-6b1e-11e7-9448-73fe6ef0f3cc.png)

Adding an operator to the date filter, to see records later than 03-04-2016, returns no results
![image](https://user-images.githubusercontent.com/6314950/28296659-6bfe14a0-6b1e-11e7-8fea-72df32d8b318.png)

RDmitchell:
Instance: dev1
SHA: 2e13a70

Seems like straight Equality filtering works, ie, 2016, 2016-10, etc
But nothing using an operator (!, >, etc) seems to work

See this doc for test results.
https://drive.google.com/open?id=1GziBu2CKAfRSX9IcWn7aDaS_IbGcOEux6GyleA4wMDU

axelstudios:
Sorry, the code is properly deployed but you won't be able to actually test it until the cloudflare cache is purged, and I lost most of my passwords the last time I formatted my computer.  I'll let you know when dev1 is up to date

RDmitchell:
No worries ! Happy to test when it is ready. Seems to be working well so far.

axelstudios:
This has been merged into develop so I think we can close

RDmitchell:
Yes, tested on dev1 and it works, so it can be closed

PR #1425, opened 2017-07-28
Date filter

axelstudios:
On the inventory list page, the 3 datetime fields (`Recent Sale Date`, `PM Generation Date`, `PM Release Date`) can now be filtered similar to the number filter.

The filter supports multiple filters separated by commas, and the date string can be either a year (`2016`), a year and a month (`2016-05`), or a full date with no time (`2016-05-31`).

The filter syntax is as follows:

#1416

### DIFF
--- a/seed/static/seed/js/controllers/inventory_list_controller.js
+++ b/seed/static/seed/js/controllers/inventory_list_controller.js
@@ -185,6 +185,7 @@ angular.module('BE.seed.controller.inventory_list', [])
       _.map($scope.columns, function (col) {
         var options = {};
         if (col.type === 'number') options.filter = inventory_service.numFilter();
+        else if (col.type === 'date') options.filter = inventory_service.dateFilter();
         else options.filter = inventory_service.textFilter();
         if (col.type == 'text' || col.type == 'numberStr') options.sortingAlgorithm = naturalSort;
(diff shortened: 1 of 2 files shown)


RDmitchell:
See my latest comment in issue #1416 -- operators don't seem to be working

RDmitchell:
instance: dev1
SHA: 
org: ID = 20

Greater than filter is probably working as designed, but it's a bit odd, because if I say > 2016, the program returns everything that starts with 2016, I think because the days are all greater than 2016-00-00 (?).

Intuitively I would think it would return the 2017 dates. 

The same thing is true with, for example > 2016-03 -- the program shows the 2016-03 records as well as everything greater than that (2016-04, etc). Again, a bit non-intuitive that it would include the 2016-03 records although I can potentially see why it would be if the program is saying that 2016-03 == 2016-03-00 (?). I would have expected it NOT to show the 2016-03 records and only show 2016-04 and greater.

image

axelstudios:
Right, this is as designed.  Any date is parsed as 00:00:00 on that date, any month is parsed as the first of the month, and any year is 1/1.  So in this case, all of those dates are greater than 2016-03-01 at midnight.

As you mentioned, this doesn't affect `=`, `!=`, `<`, `<=`, or `>=`, but I'll check and see how much work it would be to parse the date differently for the `>` operator

Issue #1410, opened 2017-07-01
Filtering in PM Property ID field doesn't do "contains" filter


RDmitchell:
instance: seeddemostaging
SHA: fc1d2b0
org: LBNL 201

Filtering on PM Property ID doesn't seem to do a "contains in" filter -- it is only doing an exact filter.

See the doc in this folder
https://drive.google.com/open?id=0B3fTKpZ9Dx7LOXJ4VjFxRkQ5ZDQ

RDmitchell:
instance: dev1
SHA: 5861591
Org: ID = 20

Working now, thanks @axelstudios !

PR #1427, opened 2017-08-02
PM Property ID now uses text filtering

axelstudios:
The PM Property ID column on the inventory list page can now be filtered using "contains" comparisons.

Also fixed natural sorting for all non numeric/date fields and removed the `numberStr` column type since natural sorting is used by default

#1410


RDmitchell:
Let me know when I can test on dev1. thx!

axelstudios:
@RDmitchell It's on dev1 now

Issue #1428, opened 2017-08-02
Add Pagination to Matching Detail View


RDmitchell:
Pagination was added to the Matching List View, so that not all the records are displayed at once. This was a major speed improvement to displaying the list. #1376 

Now that same technique needs to be applied to the Matching Detail view, which is still VERY SLOW to display.

RDmitchell:
instance: dev1
SHA: 639f1da

Added and tested.

PR #1433, opened 2017-08-08
Matching Detail Pagination & Performance

axelstudios:
#### What's this PR do?
- Adds client-side pagination to the matching to the matching detail page
- More performance improvements for the matching list page
#1428


nllong:
looks good.

PR #1436, opened 2017-08-11
Fix Cycle Filters and API Cleanup

nllong:
API was a bit disorganized. Cycle filters weren't working for time data because of timezones.

#### What's this PR do?

- Reorganize URLs
- Move APIs folder into seed folder
- Add another layer to the API namespace
- Add api_view objects to v1 endpoints
- Add JsonResponse to v1 endpoints
- Update DRF to 3.4.7

Lots of monkeys.

N/A

#### Definition of Done:


config/urls.py
```diff
@@ -1,49 +1,45 @@
-
 # !/usr/bin/env python
 # encoding: utf-8
 """
 :copyright (c) 2014 - 2017, The Regents of the University of California, through Lawrence Berkeley National Laboratory (subject to receipt of any required approvals from the U.S. Department of Energy) and contributors. All rights reserved.  # NOQA
 :author
 """
-from django.conf.urls import include, url
 from django.conf import settings
-from django.contrib import admin
```
Myoldmopar: Goodbye, admin!

config/urls.py
```diff
@@ -1,49 +1,45 @@
...
-    # i18n setlang
     url(r'^i18n/', include('django.conf.urls.i18n')),
-
     url(r'^robots\.txt', robots_txt, name='robots_txt'),
 
-    url(r'^api/v2/', include('api.urls', namespace="apiv2")),
-    url(r'^app/', include('seed.urls.properties', namespace="app", app_name='properties')),
+    # API
+    url(r'^api/swagger/', include('rest_framework_swagger.urls'), name='swagger'),
+    url(r'^api/', include(api, namespace='api')),
```
Myoldmopar: Much cleaner here, :+1:

seed/api/base/views.py
```diff
@@ -43,16 +43,16 @@ def try_coded_reverse(self, request):
         # i['seed:get_column_mapping_suggestions'] = reverse('seed:get_column_mapping_suggestions')
 
         # now try the same column mapping thing from API v2
-        i['apiv2:data_files-mapping-suggestions'] = reverse('apiv2:data_files-mapping-suggestions', args=[1])
+        i['apiv2:data_files-mapping-suggestions'] = reverse('api:v2:data_files-mapping-suggestions', args=[1])
```
Myoldmopar: Such a subtle, but important, improvement here.

seed/filtersets.py
```diff
@@ -88,7 +91,7 @@ def year_filter(self, queryset, name, value):
         name = "{} Calendar Year".format(value)
         cycles = queryset.filter(name__contains=name)
         if not cycles:
-            start = datetime(int(value), 1, 1)
+            start = make_aware(datetime(int(value), 1, 1), pytz.UTC)
```
Myoldmopar: Is this changing a default then?  Or is this just being intentional about what it was defaulting to before?

seed/tests/test_project_views.py
```diff
@@ -822,7 +822,7 @@ def test_move_inventory(self):
         )['project']
         url = "{}?organization_id={}&inventory_type=property".format(
             reverse_lazy(
-                'apiv2:projects-move', args=[project['slug']]
+                'api:v2:projects-move', args=[project['slug']]
```
Myoldmopar: It would be really nice if Github's diff view would accept a query parameter so I could filter out certain diffs.  I can already filter whitespace by appending `?w=1`, it could be great if I could apply `&filter=apiv2|api:v2` in some form like I can with my native diff tools.

seed/urls.py
```diff
@@ -4,3 +4,13 @@
 :copyright (c) 2014 - 2017, The Regents of the University of California, through Lawrence Berkeley National Laboratory (subject to receipt of any required approvals from the U.S. Department of Energy) and contributors. All rights reserved.  # NOQA
 :author
 """
+
```
Myoldmopar: This file was moved from `seed/urls/__init__.py`.  Does that mean there is no longer an __init__.py file there?  Maybe I'll find out as I read along.

seed/views/main.py
```diff
@@ -244,19 +245,20 @@ def export_buildings(request):
         selected_fields,
     )
 
-    return {
+    return JsonResponse({
```
Myoldmopar: I'm not sure this should work.  This particular line is fine, but if you are calling this as a DRF method, you shouldn't be accessing `request.body` up on line 172 anymore.  You should access `request.data`, which is already a `dict`, so you don't have to `json.loads` it.  I just wouldn't be surprised if this function fails.

seed/views/main.py
```diff
@@ -284,17 +286,18 @@ def export_buildings_progress(request):
     percent_done = progress_data['progress']
     total_buildings = progress_data['total_buildings']
 
-    return {
+    return JsonResponse({
```
Myoldmopar: Same thing here, I think `request.body` will fail to work here, right?  Is it working and I'm just bonkers??

Myoldmopar:
Got them changed to request.data; Travis _seems_ to be happy so far.  If he stays happy I'll merge this.

seed/filtersets.py
nllong: it wouldn't work before because we have timezones enabled. So this was failing. We have to be intentional with timezones all the time now.

seed/urls.py
nllong: file was moved. There used to be a seed/urls/*.py folder. That is gone now. It is just seed/urls.py because, well, the majority of the app is an API.

Myoldmopar:
OK all good.  Merging.

nllong:
oh man... that was a quick merge.

PR #1448, opened 2017-08-29
Filter By Label By Cycle

axelstudios:
#### What's this PR do?
Filter By Label on the inventory list view previously showed all labels that had been applied to any record from any cycle.  Now, the only labels that appear in the dropdown are ones that can be found on records for the currently selected cycle.

Have records in more than one cycle.  Apply a label to records in one cycle that has not been used for the other cycles.  Change cycles to see that the label becomes available or removed in the label filter.

### DIFF
--- a/seed/static/seed/js/controllers/inventory_list_controller.js
+++ b/seed/static/seed/js/controllers/inventory_list_controller.js
@@ -44,7 +44,9 @@ angular.module('BE.seed.controller.inventory_list', [])
       $scope.number_per_page = 999999999;
       $scope.restoring = false;
 
-      $scope.labels = labels;
+      // Reduce labels to only records found in the current cycle
+      $scope.selected_labels = [];
+      updateApplicableLabels();
 
       var localStorageKey = 'grid.' + $scope.inventory_type;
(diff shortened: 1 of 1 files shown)


nllong:
looks good. tested per your instructions.

RDmitchell:
@axelstudios - I don't think there is a Github issue for this, or a PT story. Seems like we should make one so that it is on my list of things to test .. ??

axelstudios:
Yeah, this was just a bug I stumbled across - I'll add a pivotal ticket

RDmitchell:
@axelstudios - is this deployed to the dev1 server for testing?

RDmitchell:
Actually, looks like it is, because develop and dev1 are both SHA: 8a68345

RDmitchell:
instance: dev1
SHA: c94bc89
org: ID = 69

Added label to Cycle = 2017, and it is the only label that shows up in the Filter by Label box

Going to Cycle = 2016, there are lots of labels applied, and the one I added for 2017 does not appear in that list, and the labels in 2016 do not appear in 2017. 

So seems to be working

PR #1451, opened 2017-09-01
Filter out bazillions of markdown package import debug log statements

Myoldmopar:
The Swagger page imports the markdown package, and upon import, it issues many, many debug lines to the console.  This only occurs in dev mode when the log is set to debug.  But it makes it hard to develop when all those lines mask actual messages.

#### What's this PR do?
Adds a log filter that will simply filter out the markdown import messages.

On the develop branch, run SEED in dev mode and load the swagger page.  The console should be littered with messages:

image

Then on this branch, run SEED in dev mode and those particular messages should be gone, leaving only relevant messages.

N/A

Above.

#### Definition of Done:
- [ ] Is there appropriate test coverage? N/A - I think.
- [ ] Does this PR require a Selenium test? No
- [ ] Does this PR require a regression test? No
- [ ] Does this add new dependencies? No

### DIFF
--- a/config/settings/dev.py
+++ b/config/settings/dev.py
@@ -52,12 +52,18 @@ LOGGING = {
             'format': "%(pathname)s:%(lineno)d - %(message)s"
         }
     },
+    'filters': {
+        'mute_markdown_import': {
+            '()': 'seed.utils.generic.MarkdownPackageDebugFilter'
+        }
+    },
     # set up some log message handlers to choose from
     'handlers': {
         'console': {
             'level': 'DEBUG',
             'class': 'logging.StreamHandler',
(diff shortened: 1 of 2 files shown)


config/settings/dev.py
```diff
@@ -52,12 +52,18 @@
             'format': "%(pathname)s:%(lineno)d - %(message)s"
         }
     },
+    'filters': {
+        'mute_markdown_import': {
+            '()': 'seed.utils.generic.MarkdownPackageDebugFilter'
```
nllong: bitmoji

Myoldmopar:
bitmoji

PR #1453, opened 2017-09-08
Fix typo, remove spurious 301 redirect.

rgm:
Seems djangorestframework will fill in a 301 redirect to add a trailing slash. No need for two AJAX requests here.


nllong:
appended slashes, ugh, thanks for catching this!

Issue #1257, opened 2017-02-20
Field name with apostrophe capitalized the next character


RDmitchell:
instance: seeddev
SHA: 514e883
Org: LBNL 405; deleted inventory and mapping as superuser before importing data
Chrome, incognito
Files: Google Drive\SEED Shared Folder\Software Development\Issues\Issue 1128
Import: MixedPropTaxLotData.xlsx

In the process of testing #1213, I did the following:

* Import MixedPropTaxLotData.xlsx
* Mapped as follows
![image](https://cloud.githubusercontent.com/assets/6314950/23143840/68074296-f778-11e6-8c95-efa229fefc3d.png)
* In mapping review screen, the program capitalized the "S" after the apostrophe in **Assessor's Database Address**
![image](https://cloud.githubusercontent.com/assets/6314950/23139216/2a765a06-f760-11e6-9ad0-4eda025259cf.png)
* It would be nice if the program didn't do that -- that is now the name of the field forever more. 
![image](https://cloud.githubusercontent.com/assets/6314950/23143759/fa9db0c8-f777-11e6-9226-4adc3e9c2549.png)

RDmitchell:
instance: dev1
SHA: a675da9

This is fixed

PR #1463, opened 2017-09-18
fix titlecase method with apostrophes

nllong:
Strings that were being title cased were getting a capital letter after the apostrophe. This is known issue in python's title() method.

#### What's this PR do?
use python's suggestion to fix the title() issue.  Note that this fix is code side only and no data in the database is stored with the weird title case.

Import a file that has an apostrophe in the column title and make sure that the mapping results look right.

#1257 

#### Definition of Done:


nllong:
@axelstudios can you test this quickly. Sanity test should suffice.

PR #1468, opened 2017-09-21
Add PUT function for PropertyState via BuildingSync

Myoldmopar:
For now just creating this PR to show what I'm adding when comparing to the original bricr branch, not develop.  I'll reset the base branch if needed later.

#### Definition of Done:


seed/api/v2_1/views.py
```diff
@@ -122,3 +125,89 @@ def building_sync(self, request, pk):
...
+            cycle = Cycle.objects.get(pk=cycle)
+
+        building_file = BuildingFile.objects.create(
+            file=the_file,
+            filename=the_file.name,
+            file_type=file_type,
+        )
+
+        try:
+            # do I need to pass cycle ID to get a specific cycle time?
```
Myoldmopar: @nllong What all do you need to pass in to ensure you get a unique PropertyView besides pk?  Cycle and Org?  Just Cycle?

seed/api/v2_1/views.py
```diff
@@ -122,3 +125,89 @@ def building_sync(self, request, pk):
...
+            return JsonResponse({
+                'success': False,
+                'message': 'Cannot match a PropertyView with pk=%s; cycle_id=%s' % (pk, cycle)
+            })
+
+        # here, instead of relying on BuildingFile.process to create a PropertyView,
+        # I'd like to either:
+        #   pass in the existing property_view and have it to the assignment inside process, or
+        #   just have the process() function only do the processing, and not also the PV creation, then I can do
+        #   the assignment here
```
Myoldmopar: Most of this function was pulled from the `BuildingFileViewSet.create()` method.  We could certainly pull out some of the functionality into a single worker to maximize DRYness.  But for the sake of time, I dropped it over here and modified it.

Here is the core change from the create method.  The process() function used here does (at least) two things: 
 - processes the building sync XML file into a structure
 - creates the propertyview to accompany this buildingsync file

For the call in this view I'm adding, I only want the first step.  Would you prefer that I:
 - Modify that function so that the processing and creating are split into two functions, so that I can call the process method from here easier?
 - Modify that function so that I pass in an optional property view ID, and it will skip creating a new one?
 - Something else?

seed/api/v2_1/views.py
```diff
@@ -122,3 +125,89 @@ def building_sync(self, request, pk):
         """
         return JsonResponse(
             {"status": "error", "message": "Not yet implemented. PK was {}".format(pk)})
+
+    @detail_route(methods=['PUT'])
+    def update_with_building_sync(self, request, pk):
```
Myoldmopar: I just put this name here to avoid conflicts with the GET method that we still need to wrap up.  When both methods are working well, it is simple to put them together into a single detail route.

nllong: sounds good.

seed/api/v2_1/views.py
nllong: pk, cycle, and org.

seed/api/v2_1/views.py
nllong: look here for processing BuildingSync. https://github.com/SEED-platform/seed/blob/bricr/seed/models/building_file.py#L89

I think we would want to break up the processing/creation into two functions. Processing should be done'ish with the link above. Creating and updating the propertyview is the challenge as we need to keep all the relationships on the propertystate intact (measures, scenarios, etc).

Myoldmopar:
OK, made some updates.  Instead of splitting the function, as a first step I tried just allowing an existing PropertyView to be passed in.  Does this look reasonable @nllong?  I'm thinking of this kinda like the function naming: once we have it working properly, it should be straightforward to rename/organize things.

seed/api/v2_1/views.py
```diff
@@ -122,3 +125,82 @@ def building_sync(self, request, pk):
...
+            cycle = Cycle.objects.get(pk=cycle_id)
+
+        building_file = BuildingFile.objects.create(
+            file=the_file,
+            filename=the_file.name,
+            file_type=file_type,
+        )
+
+        try:
+            property_view = PropertyView.objects.get(pk=pk, cycle=cycle)
```
Myoldmopar: I couldn't find anywhere that propertyviews were downselected to organization.  I could add it here, but it looks like maybe org is actually on the PropertyState?  View only joins the state, property, and cycle together, so I just wanted some confirmation about adding them.

seed/models/building_file.py
```diff
@@ -77,154 +77,161 @@ def str_to_file_type(cls, file_type):
...
-                self.save()
-
-                # add in the measures
-                for m in data['measures']:
-                    # Find the measure in the database
-                    measure = Measure.objects.get(
-                        category=m['category'], name=m['name'], organization_id=organization_id,
-                    )
+        if self.file_type != self.BUILDINGSYNC:
+            return False, None, "File format was not set to BuildingSync"
```
Myoldmopar: Gross, this diff looks worse than it is.  I tried to make it easier to chop up the main pieces of the function by replacing the if / else with some early return calls.  The only real change is [here](https://github.com/SEED-platform/seed/pull/1468/files#diff-74afe00e18809e72419ed1db0d6d2009R125) where I check if property_view is passed in, and if so I use that for further assignments.  If it's not, we go down the original path of creating a new state and view.

nllong: Eek. yeah, I was about to flip out.

seed/building_sync/building_sync.py
```diff
@@ -168,12 +176,67 @@ def export(self, property_state, process_struct=ADDRESS_STRUCT):
...
+        :return: string, as XML
         """
 
-        return "TODO: Populate this object!"
+        # if not property state is defined, then just return the BuildingSync unparsed
+        if not property_state:
+            return xmltodict.unparse(self.raw_data, pretty=True).replace('\t', '  ')
+
+        # parse the property_state and merge it with the raw data
+        new_dict = copy.deepcopy(self.raw_data)
```
Myoldmopar: I assumed I'd see deepcopy in here now after the discussion yesterday.

seed/api/v2_1/views.py
```diff
@@ -120,11 +123,54 @@ def get_queryset(self):
...
+        bs_file = property_view.state.building_files.last().file.path
+        if os.path.exists(bs_file):
+            bs.import_file(bs_file)
+            xml = bs.export(property_view.state, BuildingSync.BRICR_STRUCT)
+            return HttpResponse(xml, content_type='application/xml')
+        else:
+            return JsonResponse({
+                'success': False,
+                'message': "Could not find valid BuildingSync file attached to PropertyState"
+            })
```
Myoldmopar: Walking myself through it; so you:
- get the existing property view by pk
- fetch the last building file for it
- get the path of it
- if it exists in the uploaded files section you 
  - first import the file using the BuildingSync class import_file method
  - get an xml version of it using the BuildingSync export method
  - return that
- if not, just return failure

Now actual comment, what status should we return on failure here? 404?

seed/building_sync/tests/test_building_sync.py
```diff
@@ -54,10 +71,48 @@ def test_export(self):
...
+        self.assertTrue(result)
+        self.assertEqual(data["a"], 1999)
+
+        result = self.bs._set_node('c.d', data, "string_me")
+        self.assertTrue(result)
+        self.assertEqual(data["c"]["d"], "string_me")
+
+        # result = self.bs._set_node('d.e.f.g.h', data, 1.234)
+        # self.assertTrue(result)
+        # self.assertEqual(data["d"]["e"][""], )
```
Myoldmopar: Any reason this test is commented?  It seems like it should work...

seed/api/v2_1/views.py
```diff
@@ -117,8 +123,131 @@ def get_queryset(self):
...
+        except PropertyView.DoesNotExist:
+            return JsonResponse({
+                'success': False,
+                'message': 'Cannot match a PropertyView with pk=%s; cycle_id=%s' % (pk, cycle_id)
+            })
+
+        # passing in the property view pk should allow it to process the buildingsync file but not create a new PV
+        p_status, property_view, messages = building_file.process(organization_id, cycle,
+                                                                  property_view=property_view)
+
```
Myoldmopar: So at this point, we have a property_view that we got from the process() method.  But if I'm understanding correctly, that doesn't actually point to a valid property_state behind the scenes.  It's in a weird state.  Let me know if my path forward sounds correct:
 - I should still look up the current property_view.
 - Then get the current property_state from the view
 - Then do a deepcopy of the property_state
 - Call the process method, but pass in a property_state instance (instead of a property_view), and have it process the file and apply it to the passed in property_state
 - Once back in this method, do some data persistence:
   - PropertyState.extra_data should be persisted because of the deepcopy I think
   - BuildingFile.property_state - is this BuildingFile instance the original file that imported the building, or the new BuildingSync file being added?  Still a little fuzzy here
   - Scenario.property_state - keep the scenario ID from the original property view/state, and update that scenario with the newly created property state
   - Simulation.property_state - same
 - Then promote the newly created property state to a property view
 - Done?

seed/building_sync/tests/test_building_sync.py
nllong: no. it shouldn't be there. working on this a bit more today.

seed/api/v2_1/views.py
nllong: if failure, yeah, 404. I'm going to extend this a bit more today. There are some edge cases that I need to address. Namely:

* Create a new BuildingSync file if an existing one does not exist
* Create new nodes if node is not found
* Handle complex elements (e.g. compound lookups, if FloorAreaType = 'Footprint' then set area to value').

seed/api/v2_1/views.py
nllong: talk this morning?

Myoldmopar:
I added a bit here, fetched and merged, and it got conflicted.  I'll take a peek at why.

Myoldmopar:
Alright, the issue was because of some confusion on the base branch.  I used the bricr branch as the base branch, and I had forgotten that when doing some merging.  All is well, back to work.

nllong:
+1

Myoldmopar:
Pushed a few fixes up, I haven't written the tests, but this could probably be merged down into bricr for testing/fixes before eventually going into develop.

Myoldmopar:
@nllong 

So in the before time, the process method would:
- create a PropertyState
- create a PropertyView by calling PropertyState.promote() (which would internally create a Property also)
- it would then operate on the join tables for Measure/Scenario using the PropertyState.id field

In my new world, I now:
- added a property_state argument to the process method
- if this is passed in, no new state is created, only the original
- but the state is never promoted to a new view, and this seems to be a problem

So to remedy this:
- I modified the promote method to accept a property_id argument
- I will then modify the process method:
  - if the property_state already exists and was passed in, it will look up the underlying Property id and pass that into promote.
  
Remaining questions:
- if an existing property id is passed into promote(), should it still create a new property view, pointing at this already existing property?

nllong:
Yes, yes, yes!

Remaining questions: yes, if an existing property is passed into the promote() method, then it should promote the propertystate and property to a new property view.

Myoldmopar:
Alright, those changes have been pushed.  I am second guessing the arguments I pass into `process` now though.  I am passing in the existing property state and the existing underlying property id.  I could alternatively pass in the entire existing property view, but that may be overkill.  What do you think @nllong?

Myoldmopar:
@nllong I'm getting this error now:

```
       743: unexpected token at 'AssertionError at /api/v2/building_file/
       (u"Creating a ModelSerializer without either the 'fields' attribute or the 'exclude' attribute has been deprecated since 3.3.0, and is now disallowed. Add an explicit fields = '__all__' to the ScenarioSerializer serializer.",)
```

Any reason not to follow the recommendation and add the fields = '__all__' to the serializer?

Myoldmopar:
OK, adding that in there allows the problematic bricr test to pass! @nllong we must be close.

However, the bricr tests don't pass completely because the xml file renaming happened and the test code wasn't updated.  Or I didn't have the local code updated.  Either way, the property update test seems to run.  

I have some changes to NREL/bricr in my working directory, some of which are from our interactive debugging session.  I'll look those over to see if they need to be pushed or not.

seed/api/v2_1/views.py
```diff
@@ -117,8 +123,131 @@ def get_queryset(self):
...
+        if not cycle_id:
+            return JsonResponse({
+                'success': False,
+                'message': "Cycle ID is not defined"
+            })
+        else:
+            cycle = Cycle.objects.get(pk=cycle_id)
+
+        try:
+            # TODO: not checking organization? Is that right?
```
nllong: this is inheriting `SEEDOrgReadOnlyModelViewSet` which verifies the organization. So yes, this is okay.

seed/api/v2_1/views.py
nllong: org is on property state, yes.

Issue #1464, opened 2017-09-19
Field mapping -- Letter after number gets capitalized


RDmitchell:
instance: dev1
SHA: a675da9
Org: LBNL 405
https://drive.google.com/open?id=0B3fTKpZ9Dx7LOWZaX1VoNDllcHM

Mapped as such
![image](https://user-images.githubusercontent.com/6314950/30609325-f8cecd96-9d2f-11e7-9824-9b40a7ecdd57.png)
Mapping Review -- program capitalized the letter after the number
![image](https://user-images.githubusercontent.com/6314950/30609419-39d032f8-9d30-11e7-88df-88cd761e5dfe.png)
Inventory
Same field name as in the Mapping Review
![image](https://user-images.githubusercontent.com/6314950/30609492-612fe334-9d30-11e7-9e09-9df994222f6d.png)

RDmitchell:
Instance: staging (LBNL)
SHA: 36cc59d (2.3.1 release)

Fixed !

![image](https://user-images.githubusercontent.com/6314950/37237389-5a99530c-23c7-11e8-9a9b-0b993798677e.png)

Issue #1450, opened 2017-08-31
Need spinner when changing cycles in Inventory list for large datasets


RDmitchell:
instance: Dev1
SHA: 
Org: ID = 49

This org has lots of records in the inventory (16K+ for one of the cycles). 

On the Inventory List screen, when changing from one cycle to another, it takes a while for the program to display the data for the newly changed cycle, but there is not a spinner, so the user doesn't know what's going on. The cycle pulldown still works while the program is preparing to display the data, so it seems like the program has finished getting the data for the new cycle, only the data doesn't change to the data from the new cycle. Eventually it updates the display but in the meantime the user (me!) has been switching between cycles, just making the whole process take even longer, because I didn't know whether the program was still working on getting the new data. 

This could go in the Dec 2017 list of performance enhancements

RDmitchell:
Instance: staging (LBNL)
SHA: 36cc59d (2.3.1 release)

Spinner spins when changing between cycles.

PR #1478, opened 2017-10-03
Multiple Fixes

axelstudios:
###### Testable changes
- Data Quality modal column sort now uses natural sort order
- The spinner is shown when changing cycles on the inventory list view
- Column title casing is improved, so you'll see `3rd Role` instead of `3Rd Role`
###### Backend changes
- Switched to JS strict type checking
- More one-time binding improvements
#1450
#1464

### DIFF
--- a/seed/data_importer/models.py
+++ b/seed/data_importer/models.py
@@ -51,7 +51,7 @@ STATUS_UNKNOWN = 10
 STATUS_MATCHING = 11
 
 # TODO: use these instead of the others defined in models.py
-IMPORT_STATII = [
+IMPORT_STATUSES = [
     (STATUS_UPLOADING, "Uploading"),
     (STATUS_MACHINE_MAPPING, "Machine Mapping"),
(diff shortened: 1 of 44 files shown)


seed/data_importer/models.py
```diff
@@ -113,7 +113,7 @@ class ImportRecord(NotDeletableModel):
     matching_done = models.BooleanField(default=False)
     is_imported_live = models.BooleanField(default=False)
     keep_missing_buildings = models.BooleanField(default=True)
-    status = models.IntegerField(default=0, choices=IMPORT_STATII)
+    status = models.IntegerField(default=0, choices=IMPORT_STATUSES)
```
nllong: Thank you for fixing this!

nllong: If you want to go the next level, then put the IMPORT_STATUSES inside the class that is referencing it.

axelstudios: Boom.  Scoped.  55b8435

PR #1457, opened 2017-09-13
Lay foundations for `pint` Quantity objects for EUI, Areas on PropertyStates

rgm:
#### Any background context you want to provide?

The EUIs and areas in the PropertyState are stored as simple floats, and are implicitly kBtu/ft2/year / square feet. Incoming data from Canadian sources is going to be various flavours of metric.

#### What's this PR do?

Introduces 7 new PropertyState columns for unit-tagged EUIs and areas. Lays a UI foundation—hidden behind a dark-launch flag until the QA work around working with metric units is completed—to allow a user to specify mappings and data quality rules with a selection of units. This will enable working with heterogeneous data files (eg. some columns metric, others imperial), and to understand data quality rules in a user's preferred units without having to do mental translation (eg. 15 kWh/m2/year for Passive House).

#### How should this be manually tested?

Because this is foundational, with more non-database work to come, leave the `release:use_pint` flipper off in Angular and Python and confirm that there are no user-visible changes and all tests continue to pass.

#### What are the relevant tickets?

n/a

#### Definition of Done:

- [x] Is there appropriate test coverage? (e.g. ChefSpec, Mocha/Chai, Python, etc.)
- [x] Does this PR require a Selenium test? (e.g. Browser-specific bugs or complicated UI bugs)
- [x] Does this PR require a regression test? All fixes require a regression test.
- [x] Does this add new dependencies? If so, does PIP, npm, bower requirements need to be updated?

PR #1481, opened 2017-10-16
Hides `*_pint` columns in Property List Settings view

rgm:
Follow-on work to #1457. Hides the `*_pint` columns until the main body
of the UI work for differentiating metric and imperial units is ready.

There are now database columns ready to receive tagged Quantity units for the EUIs and Areas. The UI is not yet ready, so these columns should not be displayed to users. They have a `(Pint)` suffix in the UI, just in case they display, to avoid mistakenly stuffing production data into them.

#### What's this PR do?

Fixes one missed case where the Quantity-related columns (eg. `Site EUI (pint)`) show in the Property List Settings.

See "Definition of Done" below.

#1457 

#### Definition of Done:

Given a migrated 2.2 database, the page `/app/#/properties/settings` should not show any columns with the suffix `(pint)` available for toggling display.

### DIFF
--- a/seed/static/seed/js/controllers/inventory_settings_controller.js
+++ b/seed/static/seed/js/controllers/inventory_settings_controller.js
@@ -12,7 +12,8 @@ angular.module('BE.seed.controller.inventory_settings', [])
     'user_service',
     'all_columns',
     'shared_fields_payload',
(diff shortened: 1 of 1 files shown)


RDmitchell:
@nllong , @axelstudios -- I can test this when it gets onto develop. Just let me know. Thx @rgm !

nllong:
@RDmitchell  -- on dev1

Issue #1363, opened 2017-05-25
DQ Check - Add export to Inventory DQ Results modal


RDmitchell:
instance: dev1
SHA: 3750bb2

There is an export button in the DQ results modal resulting from the Mapping DQ process, but there is not one for the DQ results model resulting from the DQ process in Inventory.

We should add one to that modal also

![image](https://cloud.githubusercontent.com/assets/6314950/26468546/b46a73f2-414b-11e7-884a-efcf30280aac.png)

RDmitchell:
Instance: dev1
Org: LBNL 24
Cycle: 1238 Test

Program seems to be hanging when I do the DQ check from the Inventory

See this doc for more details
https://drive.google.com/open?id=1LKCAedXLWLbv7T6j3N1o77pi5J72TmuQ

RDmitchell:
Instance: dev1
Org: LBNL 24
Cycle: 1238 Test

File downloads properly and contains the correct data

PR #1490, opened 2017-10-25
Export DQ from Inventory

nllong:
Data quality results that were presented in the modal were only exportable from the mapping results.

#### What's this PR do?
Add the export capability to the DQ results from the inventory list

- Go to inventory page
- Select a few records (that will have DQ issues)
- Run DQ
- From modal, make sure export exists
- Try to download results and inspect file.

#1363 

#### Definition of Done:


seed/static/seed/js/controllers/data_quality_modal_controller.js
```diff
@@ -3,172 +3,171 @@
  * :author
```
nllong: this was entirely code cleanup/autoformatting on this file

PR #1493, opened 2017-11-01
Removes node_modules from fine-uploader after building

axelstudios:
Running Django's `collectstatic` command has to copy and validate about 15,259 files and directories.  Cleaning up after fine-uploader reduces this total to 2,741

#### What's this PR do?
Deletes the `node_modules` directory from the fine-uploader directory after building

From the seed directory:
```bash
rm -rf seed/static/vendors/bower_components/fine-uploader
```


nllong:
looks good!

Issue #1501, opened 2017-11-10
migrate django password reset to class based views


nllong:
Django 1.11 prefers using class based viewsets for account management.

PR #1502, opened 2017-11-10
fix password reset

nllong:
Password reset wasn't working after django 1.11 upgrade.

#### What's this PR do?
This is a temporary fix and simply removed the unused argument.

The correct way to fix this is to convert from function based views to classed based views per the [django documentation](https://docs.djangoproject.com/en/1.11/topics/auth/default/#django.contrib.auth.views.PasswordResetView)

Test if you can reset your password.

#1503 
Follow on ticket to address class based views - #1501

#### Definition of Done:


RDmitchell:
Aha !!! Berkeley was reporting a problem with this, and I hadn't gotten around to talking to you about it, so in the interim, I just made a new user for them (because they couldn't reset the password on an existing account). Glad you fixed this !

PR #1507, opened 2017-11-15
removing instruction to python setup.py install

nmerket:
It caused problems with finding the settings for celery in on mac. Also those instructions weren't there in the Ubuntu instructions.

@nllong @axelstudios


nllong:
great find!

Issue #1298, opened 2017-03-20
Add Import Date as a viewable field in List and Detail view


RDmitchell:
It would be nice to have "Import Date" as a field that could be viewed in the List and Detail views. It was added to migrated data and so seems to be available as a field that can be selected from List Settings, but there is not a value in it for newly imported data.

RDmitchell:
@nllong -- so is this date only going to appear for newly imported records, not records that are already imported?

nllong:
existing records will get a date, but the date will be when migration is
run.

RDmitchell:
instance: dev1
SHA: 94df60a
Org: LBNL 28

@nllong -- so I guess the fields are called "Updated (Property)" and "Updated (Tax Lot)"?

I imported covered-building-sample.csv as a spreadsheet, mapped fields to both Tax Lot and Property, here are the resulting "updated" field dates
![image](https://user-images.githubusercontent.com/6314950/33909247-d7c1f216-df3f-11e7-8a12-6424a221076d.png)

I then imported the file "portfolio-manager-sample.csv as a Portfolio manager file, all fields mapped to the Property table.
![image](https://user-images.githubusercontent.com/6314950/33909388-5d1eb6ce-df40-11e7-9469-eaccccf59af0.png)

In the Detail view, the program is showing the created and updated dates for the master record but not the individual files that were imported. This seems ok, but would it also make sense to put those dates in the "Created" field for each file that was imported? ie, the Updated date in the master record in this case would be the Created date for the Portfolio Manager record that was imported.
![image](https://user-images.githubusercontent.com/6314950/33909760-87e646fa-df41-11e7-9bfb-9c0e68123d55.png)

RDmitchell:
@nllong -- I'm not sure what you mean by migration -- are you saying that the migration will figure out when all the different children were imported, and then put those dates into these date fields? Or will the migration just put the date of the migration? If it's the second option, I'm not sure adding a date that a migration done is very useful for existing data. It actually might be confusing. Maybe we should leave it blank for existing data if we can't actually determine when the records were created, depending on what the migration is actually doing.

RDmitchell:
I think one of the original reasons I wanted an "import" date (which is now an updated or created date) was to keep track of when data was updated over time, which means putting that information in the individual file import columns, as well as the master record column. 

don't know how hard that is to do, but it seems like it would be very useful to the user to see that information over time for the data they have imported.

nllong:
[...]

PR #1515, opened 2017-12-04
1298 import date field

nllong:
Datetimes were not stored on when properties and taxlots were imported and/or modified. 

#### What's this PR do?
Add datetime stamps to the Property and TaxLot objects upon creation and modification. The tax lot data are stored on the Property/TaxLot tables and not the State tables. When a user updates the propertystate directly via the inventory detail page, then the database has a callback to update the Property modified time.

This PR also deprecated more API v1 around columns.

Import some data. Import another file with some modifications. Make sure that the created and updated fields change correctly in the inventory list. 

Also, try editing a detail property and taxlot and make sure that the updated datetime changes correctly.

- [x] @Myoldmopar -- in /api/v2/properties/{id}/, the id is the Property.pk, right? (not PropertyView). 

#1298 

The roll-ups look a little funny right now.
screen shot 2017-12-04 at 1 30 48 pm

#### Definition of Done:


Myoldmopar:
> @Myoldmopar -- in /api/v2/properties/{id}/, the id is the Property.pk, right? (not PropertyView).

Right, `/api/v2/properties/{id}/` will lead to the [`PropertyViewSet`](https://github.com/SEED-platform/seed/blob/e3b25cae56afeadece76cc6bdbe116c966783dbf/seed/views/properties.py#L184), and on the `retrieve` function there, the `pk` is passed to the [`_get_property_view`](https://github.com/SEED-platform/seed/blob/e3b25cae56afeadece76cc6bdbe116c966783dbf/seed/views/properties.py#L408) method, which selects the `PropertyView`s that have a matching underlying `property_id` with that `pk`.

So yes.

nllong:
thanks for double checking the `property_id`. Yeah, it would be good if you could run some manual tests. Thanks!

Myoldmopar:
Submitted `covered_building_sample.csv`, got the following timestamps:

image

Made one change and it found the difference:

image

And now that file has an updated timestamp:

image

Manually edited the single property and also got an updated timestamp:

image

Then I took a view by tax lot:

image

And manually edited it to find a new timestamp there also:

image

This is good to go.  @nllong shall I merge?

nllong:
merge away... just note that the roll-ups are a bit weird for now.

PR #1517, opened 2017-12-12
Bricr buildingsync upload

nllong:
When updating a propertystate with a new buildingsync record, the relationships were not moving over to the new state correctly. Also, when calling update on a propertystate the data were not being persisted.

#### What's this PR do?
Fixes the update and update_with_buildingsync method. update_with_buildingsync now uses the merge_state method to merge the previous state with the new state. 

The merge_state method hase a merge_relationships method. This method is marginally tests and could benefit from a clear understanding on how we want to merge/update/move scenarios, building_files, measures (attached to scenarios), and simulation. Presently a measure will be copied to the new state if, and only if, the measure has not already been attached. 

Try import a buildingsync file, then import a new one with small changes (especially in the measures section).  Importing is only via the API at the moment.

N/A

N/A

#### Definition of Done:


seed/api/v2_1/views.py
```diff
@@ -241,220 +235,46 @@ def update_with_building_sync(self, request, pk):
 
         the_file = request.data['file']
         file_type = BuildingFile.str_to_file_type(request.data.get('file_type', 'Unknown'))
-        organization_id = request.data['organization_id']
-        cycle_id = request.data.get('cycle_id', None)
+        organization_id = request.query_params.get('organization_id', None)
```
Myoldmopar: Cool, not sure why this was coming in on the POST anyway (if it was).

seed/api/v2_1/views.py
```diff
@@ -241,220 +235,46 @@ def update_with_building_sync(self, request, pk):
...
-        try:
-            property_view = PropertyView.objects.get(pk=pk, cycle=cycle)
-        except PropertyView.DoesNotExist:
-            return JsonResponse({
-                'success': False,
-                'message': 'Cannot match a PropertyView with pk=%s; cycle_id=%s' % (pk, cycle_id)
-            })
+            property_view = result.pop('property_view')
+
+            # passing in the existing property state allows it to process the buildingsync without creating a new state
```
Myoldmopar: OK, so the `process` method appears to now take the old `property_view`.  This is fine, but the comment should be changed.

seed/api/v2_1/views.py
```diff
@@ -241,220 +235,46 @@ def update_with_building_sync(self, request, pk):
...
+            p_status, new_pv_state, new_pv_view, messages = building_file.process(
+                organization_id, cycle, property_view=property_view
+            )
```
Myoldmopar: And with all the operations below removed I assume they are simply handled inside the `process()` method.

seed/lib/merging/merging.py
```diff
@@ -83,7 +92,90 @@ def get_state_attrs(state_list):
         return get_taxlotstate_attrs(state_list)
 
 
-def merge_extra_data(b1, b2, default=None):
+def _merge_relationships(merged_state, state1, state2):
```
Myoldmopar: Is it obvious this method is for PropertyState objects?  It's not in the function name or the docstring.  Obviously we know what it does, but it would be really vague for others.  It feels like this should belong on the PropertyState class instead of just hanging out here in a separate lib file, but I'm sure you have your reasons for putting it here.

seed/lib/merging/merging.py
```diff
@@ -83,7 +92,90 @@ def get_state_attrs(state_list):
...
+    Merge together the old relationships with the new.
+    """
+    # get some items off of this property view
+    no_measure_scenarios = [x for x in state2.scenarios.filter(measures__isnull=True)] + \
+        [x for x in state1.scenarios.filter(measures__isnull=True)]
+    building_files = [x for x in state2.building_files.all()] + [x for x in state1.building_files.all()]
+    simulations = [x for x in Simulation.objects.filter(property_state__in=[state1, state2])]
+    measures = [x for x in PropertyMeasure.objects.filter(property_state__in=[state1, state2])]
+
+    # TODO: dedup the relationships, if they are exact then don't add
```
Myoldmopar: Is that just adding a line in each of the following loops to check some attributes for duplication?

seed/lib/merging/merging.py
```diff
@@ -177,8 +269,11 @@ def merge_state(merged_state, state1, state2, can_attrs, default=None):
...
 
-    merged_extra_data, merged_extra_data_sources = merge_extra_data(state1, state2, default=default)
-
+    merged_extra_data, merged_extra_data_sources = _merge_extra_data(state1, state2, default=default)
     merged_state.extra_data = merged_extra_data
 
+    # merge measures, scenarios,
+    if isinstance(merged_state, PropertyState):
+        _merge_relationships(merged_state, state1, state2)
+
```
Myoldmopar: Hmm, maybe this line clears up my earlier comments in this file.  Does this merge function work on classes other than PropertyState?

seed/models/building_file.py
```diff
@@ -123,17 +124,20 @@ def process(self, organization_id, cycle, property_state=None, property_id=None)
...
-            property_state = PropertyState.objects.create(**create_data)
-            property_state.extra_data = extra_data
-            property_state.promote(cycle)
-            property_state.save()
-        else:
-            # invalid arguments, must pass both or neither
-            return False, None, "Invalid arguments passed to BuildingFile.process()"
+        # always create the new object, then decide if we need to merge it.
+        # create a new property_state for the object and promote to a new property_view
+        property_state = PropertyState.objects.create(**create_data)
```
Myoldmopar: Interesting.  I thought we were trying to avoid creating a new property state.

seed/models/building_file.py
```diff
@@ -227,13 +229,43 @@ def process(self, organization_id, cycle, property_state=None, property_id=None)
...
+                description='Automatic Merge',
+                import_filename=None,
+                record_type=AUDIT_IMPORT
+            )
+
+            property_view.state = merged_state
+            property_view.save()
+
+            merged_state.merge_state = MERGE_STATE_MERGED
+            merged_state.save()
```
Myoldmopar: Wow, the different uses of the word state here had me thrown.  I get it though.  You merge two property states together, and store the newly merged state onto the property view's state member, and on the merged state's merge_state member, you store the status merge.

seed/static/seed/js/seed.js
```diff
@@ -168,7 +168,7 @@ SEED_app.run([
     make('release:use_pint', 'boolean', false);
 
     var make2 = _.partial(flippers.make_flipper, 'nicholas.long@nrel.gov', '2018-01-01T00:00:00Z');
-    make2('release:bricr', 'boolean', false);
+    make2('release:bricr', 'boolean', true);
```
Myoldmopar: Ooooh adding another flipper?

seed/api/v2_1/views.py
nllong: we went back and forth on what to pass... in the end I think this was the best. I'll update the comment.

seed/api/v2_1/views.py
nllong: yes

seed/lib/merging/merging.py
nllong: i actually implemented the deduping in the measures section below.

seed/lib/merging/merging.py
nllong: no, that is the reason for the PropertyState catch.

seed/models/building_file.py
nllong: we were. I changed my mind because i wanted to leverage the merge functionality instead of doing 'magic' on the propertystate itself. The merge requires 2 states to be passed.

seed/static/seed/js/seed.js
nllong: i am just enabling the flipper on the bricr branch. when we merge down, we will need to disable again.

seed/lib/merging/merging.py
nllong: moving to PropertyState class method.

PR #1518, opened 2017-12-12
Tightens up check for whether celery's running in dev

rgm:
The old implementation (count the # of times "celery" is greppable from
the process list) is vulnerable to skipping a needed start if I'm
tailing `/tmp/celeryd.log` in another terminal window. Here we let the
celery process write a pid file and check for the presence of that
instead.

1. Before merge, use `./bin/start-seed.sh` to start the dev environment
2. Open a second window and then tail `/tmp/celeryd.log`
3. Kill django and celery in the first window
4. Restart with `./bin/start-seed.sh`

Expected: celery should restart.
Observed: celery doesn't restart; we're told it's already running because the tail command is spoofing it.

After merge, repeat steps. Expected and observed should now match.

n/a

### DIFF
--- a/bin/start-seed.sh
+++ b/bin/start-seed.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 # Start SEED in developer mode
 
+CELERY_PIDFILE="/tmp/celery-seed-dev.pid"
+
 # check if "$prog" is running
 running () {
     prog="$1"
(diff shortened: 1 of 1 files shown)


nllong:
looks good. I'll merge this once I can figure out what is going on with the functional test.

PR #1522, opened 2017-12-20
Fixes HTML error in sub-org submit button

rgm:
#### What's this PR do?

Fixes the "Create sub-org" submit button. Lost the close tag on the line above while adding the translate tags in 1931c42f1.

Go to `/app/#/accounts/<org_id>/sub_org` and hit "Create a new sub-organization".

Expected: bootstrap primary button should now show correctly (won't without this patch).


nllong:
thanks!

Issue #1532, opened 2018-01-09
Auto ESPM import doesn't add data to correct Cycle


RDmitchell:
instance: dev1
SHA: 2.3.0.c2dab8d
Org: LBNL 20

When I use the new Portfolio Manager Import functionality, the program does not put the data into the specified cycle. It seems to default to the first cycle in the list of cycles.

Here are the steps:
* Create new data set
![image](https://user-images.githubusercontent.com/6314950/34744525-59234216-f542-11e7-9e23-c71be322b9c6.png)

* Select the cycle called "Automatic ESPM Import" from the cycle list and click the Import Portfolio Manager button
![image](https://user-images.githubusercontent.com/6314950/34744608-a7ab0400-f542-11e7-80ac-1b0818c6d7fc.png)
* Enter the credentials for the test ESPM account and click Get Report Templates
![image](https://user-images.githubusercontent.com/6314950/34744689-ee8b2378-f542-11e7-9dcc-226799314759.png)
* Select SEED City Test Report template and click Submit
![image](https://user-images.githubusercontent.com/6314950/34744738-184f1f0c-f543-11e7-823e-f55836a2fc95.png)
* Changed the field mapping for 2 fields
![image](https://user-images.githubusercontent.com/6314950/34744785-387b1a1a-f543-11e7-9167-d06869efeb6c.png)
* Mapping review shows the 3 records. Click Save Mappings
![image](https://user-images.githubusercontent.com/6314950/34744829-5a1bfce8-f543-11e7-8d63-c4ab0306f724.png)
* SEED adds three new properties
![image](https://user-images.githubusercontent.com/6314950/34744928-bbe5fba4-f543-11e7-8f43-01635b26a6ef.png)
* Go to the Inventory List, select cycle = Automatic ESPM Import -- it is empty -- the 3 records should be here
![image](https://user-images.githubusercontent.com/6314950/34744980-ea3337ba-f543-11e7-9dcc-7ec8c7269778.png)
* Go to the first cycle in the list, called 1101 KC Test, go to the end of the list of data, and the records are there. So they were imported into the wrong cycle.
![image](https://user-images.githubusercontent.com/6314950/34745792-8d8c3220-f546-11e7-9bf1-1e9eb19c0665.png)

nllong:
@RDmitchell -- can you test and close this if it works?

RDmitchell:
Instance: production (LBNL)
SHA: 68bdd74 (2.3.0)

Fixed.

PR #1533, opened 2018-01-10
Fixed the selected cycle for PM imports

axelstudios:
1. Go to add a new data file
1. Change the cycle away from the default
1. Import a PM report
1. Check that the selected cycle was used

#1532


RDmitchell:
Thanks @axelstudios

PR #1535, opened 2018-01-12
display analysis state string instead of int

nllong:
The inventory list page did not show the analysis state as a string. 

#### What's this PR do?
Converts the analysis_state field to the enumeration string.

Load data, check inventory page for analysis_state as 'Not Started' instead of 0.

N/A

image

#### Definition of Done:

### DIFF
--- a/seed/models/tax_lot_properties.py
+++ b/seed/models/tax_lot_properties.py
@@ -117,6 +117,8 @@ class TaxLotProperty(models.Model):
                 # Do not make these timestamps naive. They persist correctly.
                 related_dict['updated'] = related_view.property.updated
                 related_dict['created'] = related_view.property.created
+                # Replace the enumerations
+                related_dict['analysis_state'] = related_view.state.get_analysis_state_display()
             elif lookups['obj_class'] == 'PropertyView':
                 # Do not make these timestamps naive. They persist correctly.
                 related_dict['updated'] = related_view.taxlot.updated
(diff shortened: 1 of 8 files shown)


seed/models/tax_lot_properties.py
```diff
@@ -117,6 +117,8 @@ def get_related(cls, object_list, columns):
                 # Do not make these timestamps naive. They persist correctly.
                 related_dict['updated'] = related_view.property.updated
                 related_dict['created'] = related_view.property.created
+                # Replace the enumerations
+                related_dict['analysis_state'] = related_view.state.get_analysis_state_display()
```
Myoldmopar: So was `analysis_state` already a key here somewhere and you are now overriding it with this new `_display` method?

seed/models/tax_lot_properties.py
```diff
@@ -238,6 +236,7 @@ def get_related(cls, object_list, columns):
                 # Do not make these timestamps naive. They persist correctly.
                 obj_dict['created'] = obj.property.created
                 obj_dict['updated'] = obj.property.updated
+                obj_dict['analysis_state'] = obj.state.get_analysis_state_display()
```
Myoldmopar: And then similarly here.

seed/models/tax_lot_properties.py
nllong: it exists when calling model_to_dict. I originally tried hooking in the PropertyStateSerializer but it would require a bit of customization to work right. Ideally that is what we will need to do sooner rather than later.

Myoldmopar: I figured there was some serializer action automatically transforming the members; makes sense.

PR #1536, opened 2018-01-15
Quantity (aka. Pint) fields to accommodate metric and US Customary energy/area data

rgm:

Use of SEED by Canadian organizations requires importing energy/area data as either imperial or metric (or, in the case of the City of Toronto, in imperial energy units but metric area units (!)). Different organizations tend to culturally work with different units as well: eg Passive House-influenced organizations will tend to use kWh/m²/year, where federal agencies favour GJ/m²/year.

The database groundwork for accommodating this was laid in #1457; this completes that work with minor database work and more user-interface work. Recognizing that there is quite a lot of existing production data this touches, we are favouring a conservative approach has of adding new EUI and area database columns to accommodate this, rather than changing existing ones.

#### What's this PR do?

1. Renames the original float-type fields for EUIs and areas to add the suffix `_orig`.
1. Renames the Quantity fields with suffix `_pint` that were introduced in #1457. It remove the suffix and these become the canonical columns for this data.
1. Migrates existing PropertyStates to bring forward old float data as new Quantity data, preserving old float data in case of catastrophe.
1. Migrates any existing data quality rules to have Pint-parseable units (eg. 'ft**2' instead of 'sq. ft'). 
1. Adds an organization-level preference for the units used in display of EUI and area data, including persistence and user interface.
1. Adds server-side code needed to "collapse" Quantity data so that as far as the front-end is concerned, it is still getting floats and these are converted to reflect the organization's preference for units.
1. Adds server-side code to handle and report unit-aware data quality checks.
1. Removes release flipper `release:use_pint` (from #1457). 
1. Adds new release flipper `release:orig_columns` to hide the display of `_orig` float data in the user interface while letting it remain in the database. (This should go away in the 2.4 cycle, with the scheduled deletion of the `_orig` columns).

(cosmetic changes):
1. Alphabetically sorts existing English, French translation file to make future changes more readily apparent.

Needs QA against a fresh database, and an existing production database clone. Procedure is similar for both:

1. Create a new data set
1. Add data quality rule to label anything over 3 GJ/m2/year as "High EUI" (this is equivalent to 264 kBtu/ft2/year and should catch some reports in the sample data file).
1. Import sample Portfolio Manager file, proceed to mapping
1. Confirm "Site Eui" and "Source Eui" are showing `kBtu/sq ft/year` in unit selection
1. Map "Weather Normalized Site EUI (kBtu/ft2)" as "Site Eui Weather Normalized" and confirm that units dropdown has appeared, is showing "kBtu/ft2/year". (Repeat with the weather normalized source EUI).
1. Map file. Confirm that data quality results contain entries for "High EUI" (ie. that units have converted correctly between the metric rule and the US Customary data)
1. View inventory. EUIs should be showing in kBtu/m2/year.
1. Go to Organizations > (Name of Logged in Org) > Settings. Change EUI and Area units
1. Go back to inventory, confirm that inventory is now showing and exporting in new units, and that column headers are displaying correct units.
1. Go to the detail of a property, confirm that property states are showing correct unit data and that header is showing correct units.
1. Go to inventory report, confirm that EUI graphs are showing correct units in axis labels and that data is converted to correct units.

#1457, (not sure otherwisen; don't have access to Pivotal tickets)

org-prefs
data-quality
units-mapping
data-quality-result
units-inventory
units-reports

#### Definition of Done:

No, something is definitely breaking within the test suite while manual testing in the browser seems OK. I'm having trouble telling where the problem is (maybe fixture data, assumptions that have become embedded within the test code, or genuine application-code bugs). I'd appreciate some assistance in diagnosing this.

May require a minor add to the API test to ensure that unit selection is correctly added to the mapping step

Don't believe so.

No.


seed/serializers/properties.py
```diff
@@ -177,6 +190,43 @@ def to_representation(self, data):
 
         return result
 
+    def _collapse_pint_value(self, property_state, field_name):
```
nllong: I was just messing with this today in order to fix a bug on a production system. I think we'll want to converge the two together. In short instead of creating a SerializerMethodField, I created a new serializer.Field.

nllong: @rgm -- see this PR https://github.com/SEED-platform/seed/pull/1542/files. I was going to merge this into master after some more testing, then eventually pull it up to develop and all dev-branches.

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -130,7 +130,7 @@
     "to_field": "gross_floor_area",
     "to_table_name": "PropertyState",
     "from_field": "propGrossFloorArea",
-    "units": "",
+    "units": "ft**2",
```
nllong: I like that we can take what were once just cosmetic units and actually make them useful!

seed/lib/mcm/mapper.py
```diff
@@ -99,11 +99,21 @@ def apply_column_value(raw_column_name, column_value, model, mapping, is_extra_d
     :rtype: model inst
     """
 
-    def is_pint_column(column_name):
+    def is_quantity_type_column(column_name):
         """Test if the column_name is a QuantityField"""
-        # TODO rgm change this to a test against a set (the 7 db column names) after feature release
         # TODO rgm re-locate to someplace else, eg. Column
-        return re.search(r'_pint$', column_name)
+        quantity_column_names = [
```
nllong: this is good for now. I think we should eventually put this in the mapping jsons as a field and then do a simple lookup on that.

seed/lib/superperms/orgs/models.py
```diff
@@ -39,6 +39,18 @@
...
+    ('ft**2', 'square feet'),
+    ('m**2', 'square metres'),
+)
+
+MEASUREMENT_CHOICES_EUI = (
+    ('kBtu/ft**2/year', 'kBtu/sq. ft./year'),
+    ('kWh/m**2/year', 'kWh/m²/year'),
+    ('GJ/m**2/year', 'GJ/m²/year'),
+    ('MJ/m**2/year', 'MJ/m²/year'),
+    ('kBtu/m**2/year', 'kBtu/m²/year'), # really, Toronto?
```
nllong: =)

seed/lib/superperms/orgs/models.py
```diff
@@ -39,6 +39,18 @@
     (STATUS_REJECTED, 'Rejected'),
 )
 
+MEASUREMENT_CHOICES_AREA = (
+    ('ft**2', 'square feet'),
+    ('m**2', 'square metres'),
+)
+
+MEASUREMENT_CHOICES_EUI = (
```
nllong: can you scope this to be part of the OrganizationUser so it isn't just hanging out there on its own? Something like:

```
class OrganizationUser:
    MEASUREMENT_CHOICES_EUI = (...)
```

seed/migrations/0084_copy_propertystate_orig_pint.py
```diff
@@ -0,0 +1,49 @@
...
+#
+# Because the `*_pint` columns are new, the forward migration to _start_ using
+# them is straightforward: just copy `*_orig` -> `*_pint` for every
+# PropertyState. This works as a straight float copy because the Pint base
+# units (ft² and kBtu/ft²/year) match the implicit units of the original
+# EUI/area columns.
+
+
+def copy_orig_fields_to_pint_fields(apps, schema_editor):
+    PropertyStateModel = apps.get_model('seed', 'PropertyState')
```
nllong: I will need to run this locally with a copy of the production db and make sure everything works as expected. It looks right though.

seed/migrations/0085_rename_propertystate_from_pint.py
```diff
@@ -0,0 +1,60 @@
...
+            new_name='site_eui_modeled',
+        ),
+        migrations.RenameField(
+            model_name='propertystate',
+            old_name='site_eui_weather_normalized_pint',
+            new_name='site_eui_weather_normalized',
+        ),
+        migrations.RenameField(
+            model_name='propertystate',
+            old_name='source_eui_modeled_pint',
```
nllong: need to add this one to the PintQuantitySerializerField

seed/migrations/0085_rename_propertystate_from_pint.py
```diff
@@ -0,0 +1,60 @@
...
+            new_name='occupied_floor_area',
+        ),
+        migrations.RenameField(
+            model_name='propertystate',
+            old_name='site_eui_pint',
+            new_name='site_eui',
+        ),
+        migrations.RenameField(
+            model_name='propertystate',
+            old_name='site_eui_modeled_pint',
```
nllong: need to add this one to the PintQuantitySerializerField

seed/migrations/0086_auto_20180111_1531.py
```diff
@@ -0,0 +1,33 @@
...
+    human-readable strings instead of Pint specs. This converts them so
+    that any custom rules can be maintained instead of just dumped and
+    re-created.
+    """
+    Rule = apps.get_model('seed', 'Rule')
+    for rule in Rule.objects.all():
+        if rule.units == 'square feet':
+            rule.units = 'ft**2'
+        elif rule.units == 'kBtu/sq. ft./year':
+            rule.units = 'kBtu/ft**2/year'
```
nllong: hmm. we should really just update the old units with the pint version of the units. save it for another day.

seed/serializers/pint.py
```diff
@@ -1,11 +1,109 @@
...
 from django.core.serializers.json import DjangoJSONEncoder
 from quantityfield import ureg
 
+AREA_DIMENSIONALITY = '[length] ** 2'
+EUI_DIMENSIONALITY = '[mass] / [time] ** 3'
+
+AREA_DEFAULT_UNITS = 'ft**2'
+EUI_DEFAULT_UNITS = 'kBtu/ft**2/year'
+
+SIGNIFICANT_FIGURES = 2  # this should be... someplace else?
```
nllong: probably want to make this a user-configurable setting

rgm: cool, see 756cbefa

axelstudios:
I was following the testing steps and got a 500 while trying to go the detail view for an inventory record:

```
Got AttributeError when attempting to get a value for field `gross_floor_area_pint` on serializer `PropertyStateSerializer`.
The serializer field might be named incorrectly and not match any attribute or key on the `PropertyState` instance.
Original exception text was: 'PropertyState' object has no attribute 'gross_floor_area_pint'.
```

seed/tests/test_properties_serializers.py
```diff
@@ -331,10 +331,13 @@ def test_get_history(self):
         """Test get_history"""
         obj = mock.MagicMock()
         obj.state = self.property_state
+
+        data = self.serializer.get_history(obj)
```
nllong: Hi @fablet -- This PR adds quantities to various fields in the seed database (basically adding units to certain fields). I hacked these tests to get them passing and I realized that this may cause issues. The results of the get_history is a serializer object but since it hasn't been passed to JSON yet, the encoding still includes Quantity fields. You can see my hack by forcing the quantity field to a magnitude. Anyway, I was wondering if you think this will cause issues? If so, any ideas on how to address this more elegantly?

Thanks!

seed/lib/mcm/cleaners.py
```diff
@@ -39,6 +39,9 @@ def default_cleaner(value, *args):
     if isinstance(value, unicode):
         if fuzzy_in_set(value.lower(), NONE_SYNONYMS):
             return None
+        # guard against `u''` coming in from an Excel empty cell
+        if (value == u''):
+            return None
```
nllong: not optimal, but I think we go with it for now.

PR #1537, opened 2018-01-15
Re-sorts angular localization files

rgm:

The translations somehow came unsorted. When unsorted, it makes for messy
diffs/commits when we do the automatic download from lokalise.co.

#### What's this PR do?

Just alphabetically sorts the locale files so that when we use `script/get_angular_translations` it lays down new phrases in a easy-to-see way, instead of a bit noisy re-order diff that buries the actual change.

Load up and assure via clicking around that nothing appears different in the UI.

#### Definition of Done:

All answers no.


nllong:
this looks good. I'll merge it down once the tests pass. thanks!

Issue #1543, opened 2018-01-18
Export not returning all data when exporting from taxlot view


nllong:
If exporting from taxlot, all the data from the property state is not showing up.

![image](https://user-images.githubusercontent.com/1907354/35109597-12605492-fc34-11e7-8287-c2a7e0792a46.png)

Issue #1512, opened 2017-11-22
Export has (primarily) internal field names not SEED user mapped field names


RDmitchell:
*****
instance: production (V 2.2.0)
SHA: 0f3774d

In V 2.2.0, the export (from Actions/Export) has the user-mapped SEED field names in the export file
![image](https://user-images.githubusercontent.com/6314950/33153616-aff773ee-cf97-11e7-9c76-05f5df62de54.png)
*****
instance: staging (V 2.2.1)
SHA: 2c52627
covered-buildings-sample.csv (imported in the previous version, not in 2.2.1)

In V 2.2.1, the export (from Actions/Export) has the internal (?) field names rather than the user-mapped SEED field names. This is a file that was imported from a previous version and already existed in the Inventory database
![image](https://user-images.githubusercontent.com/6314950/33153754-906adbfa-cf98-11e7-9218-1d94d11ed379.png)

Importing from scratch in V 2.2.1:
Mapping
![image](https://user-images.githubusercontent.com/6314950/33153800-eac6f43a-cf98-11e7-8fb8-cc9266b683ce.png)

![image](https://user-images.githubusercontent.com/6314950/33153886-7a515b2c-cf99-11e7-8efc-ca2b7a0f78d7.png)

nllong:
#1543

RDmitchell:
instance: dev1
SHA: d266f3a
Org: LBNL 20

This appears to be fixed

PR #1544, opened 2018-01-18
CSV Export Header

nllong:
CSV column header was not showing display names

#### What's this PR do?
* Show display names in CSV header
* Fix exporting on tax lot page

* Import some data
* Select some items, then go to export
* Verify columns that are selected are exported with the correct headers
* Do the same on tax lots

#1543 #1512 

screen shot 2018-01-18 at 9 46 50 am
screen shot 2018-01-18 at 9 36 07 am

#### Definition of Done:


Myoldmopar:
Pulled branch, installed deps, migrated, etc.  Imported new dataset and exported from both the property and tax lot views.  Column headers look great on both.  Approved.  Two comments however:

- Can we change the flipper date for the Pint work?  The below console log is incredibly intrusive with how many times the warning gets thrown.
- Why do we require the user to enter a name for the export?  The next step in the workflow is the "save as..." dialog, so what does having this modal step gain?

image

I will update the branch and if it comes back happy from CI I'll merge unless there is a reason to hold on it (another branch about to go in, for example).

nllong:
The need to enter the filename must be browser/platform specific. On my machine, it just starts downloading once I say 'export'.

Issue #1548, opened 2018-01-23
ESPM Auto Import -- Import file name 


RDmitchell:
Instance: dev1
Org: LBNL 20
Cycle: Automatic ESPM Import

This is working in terms of importing the records from ESPM by logging into an ESPM account from SEED.

There are a few issues which we may want to refine over time, but the main one that I see now is the Filename that shows up each time an ESPM custom report template is imported into SEED. The name is always the same, ie, PM_Import.csv.

Is it possible to add a sequential number or even better a download date (2018_01_22, 2018_01_22a, etc) to that name, just so that the user can distinguish between the different records in the Inventory Detail view when a record is updated?

See this doc for an example of when this might be useful:
https://drive.google.com/open?id=1VqPrK20Lj_IjvV4iO4SXHgEnSdmmW9o56wVrla-pCWM

nllong:
@RDmitchell - Can you check and close?

nllong:
@RDmitchell can we close this too?

RDmitchell:
instance: LBNL staging
SHA: 1a8cd10 (2.3.3 release)

It doesn't seem like the name includes the date. The files are still called "PortfolioManagerImport". See the file imported today (7/10/2018) in the screen shot below.

![image](https://user-images.githubusercontent.com/6314950/42541903-3c1eb7d0-8459-11e8-988a-541143fe3a82.png)

RDmitchell:
instance: dev1
SHA: get an error on the page when trying to show the SHA -- whatever the state of the code is on 7/10/2018 

Same on dev server - filename doesn't include date, always called PortfolioManagerImport
![image](https://user-images.githubusercontent.com/6314950/42542025-c13f41a0-8459-11e8-8a87-b79224735048.png)
![image](https://user-images.githubusercontent.com/6314950/42542046-df092ec6-8459-11e8-95f9-bd447b831d87.png)
![image](https://user-images.githubusercontent.com/6314950/42542102-2539b55a-845a-11e8-8985-0e22b4f32c51.png)

nllong:
on dev1 (again)

RDmitchell:
instance: dev1
SHA: 
Org: LBNL 20
Cycle: 

Import file name now includes the date. Thanks !!!

![image](https://user-images.githubusercontent.com/6314950/44065320-db2474d4-9f1e-11e8-94c1-89a85f2a67ad.png)
[...]

PR #1556, opened 2018-01-24
Add date to PM import filename

Myoldmopar:
@RDmitchell noted that the PM import user experience would be nicer if the filenames had a timestamp embedded in them.

#### What's this PR do?
This saves the data file as `pm_import_20YY-MM-DD.csv`, and will only append a random string to ensure uniqueness in case the user imports multiple in one day.  Which should be rare.

Use the PM auto import workflow and check the filename that is automatically generated on the property detail view.

Issue #1548 

image

#### Definition of Done:


seed/data_importer/views.py
```diff
@@ -4,15 +4,16 @@
...
+import csv
+import datetime
 import hashlib
 import hmac
 import json
 import logging
 import os
-import pint
 
+import pint
```
Myoldmopar: Import changes are just reordering them and then adding the datetime import to allow us to generate a date stamp

seed/data_importer/views.py
```diff
@@ -289,8 +290,12 @@ def create_from_pm_import(self, request):
                 'message': "Must pass pm_data in the request body."
             }, status=status.HTTP_400_BAD_REQUEST)
 
+        # base file name (will be appended with a random string to ensure uniqueness anyway, but this helps for display)
+        today_date = datetime.datetime.today().strftime('%Y-%m-%d')
+        file_name = "pm_import_%s.csv" % today_date
+
         # create a folder to keep pm_import files
-        path = os.path.join(settings.MEDIA_ROOT, "uploads", "pm_imports", "pm_import.csv")
+        path = os.path.join(settings.MEDIA_ROOT, "uploads", "pm_imports", file_name)
```
Myoldmopar: Instead of using `pm_import.csv`, we use `pm_import_20XX-MM-DD.csv`.

The next block below then uses the FileSystemStorage class to make this unique if it is needed (if the user does two imports in the same day, for example).  It will become `pm_import_20XX-MM-DD_abcd1234.csv`.

RDmitchell:
This change did not make it into 2.3.1-- see my update to #1548 

So maybe we should take it out of the Release Notes for 2.3.1 ??

PR #1597, opened 2018-03-13
Allow the mapping columns to be resizable

### LINKED ISSUE #1337: SEED Header input box in Mapping -- need to be able to make column wide enough to see long field names
V 2.0.2

When importing a Portfolio Manager file, there are many fields that are very long. 
Right now in SEED, it is not possible to see the entire field name that SEED is defaulting to without going into the SEED Header field and moving your cursor to the right in that field to discover what the mapping really is. 

It seems like letting the user expand that column to the size needed to see the longest field would be ideal 

Options could include
* Ability of the user to expande the field to whatever width they desire by pulling on the right hand side of the header of the column (like in other lists).
* Double right click on the right hand side of a column in Excel which causes the program to expand that column to the size that fits the largest data for that column. But even without that fan

axelstudios:
#### What's this PR do?
Allows the mapping columns to be resizable with drag handles
Upload a file, go to the mapping page, drag the column dividers to resize
fixes #1337

### DIFF
--- a/seed/static/seed/partials/mapping.html
+++ b/seed/static/seed/partials/mapping.html
@@ -137,16 +137,16 @@
                             <th class="sub_head is_aligned_center hide">Map
                                 <a uib-popover="{$ MAP_copy $}" popover-placement="right"><i class="fa fa-info-circle" style="cursor: pointer;"></i></a>
                             </th>
-                            <th class="mapping_field" translate>SEED Header</th>
+                            <th class="mapping_field ellipsis-resizable" sd-resizable><translate>SEED Header</translate></th>
                             <!--th class="sub_head is_aligned_center">Duplicate <a uib-popover="{$VALIDATE_copy$}" popover-placement="right"><i class="fa fa-info-circle" style="cursor: pointer;"></i></a></th-->
-                            <th class="import_column_name" translate>Data File Header</th>
-                            <th translate>Inventory Type</th>
-                            <th translate>Measurement Units</th>
-                            <th translate>Row 1</th>
-                            <th translate>Row 2</th>
-                            <th translate>Row 3</th>
-                            <th translate>Row 4</th>
-                            <th translate>Row 5</th>
+                            <th class="import_column_name ellipsis-resizable" sd-resizable><translate>Data File Header</translate></th>
+                            <th class="ellipsis-resizable" sd-resizable><translate>Inventory Type</translate></th>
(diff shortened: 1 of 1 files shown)


nllong:
This doesn't appear to be working. Disabled cache, hard refresh. class looks right on the column.

image

axelstudios:
@nllong Here's what I see when I drag on the column handles:
video_2018-03-13_172203

nllong:
That’s what I was expecting. I can show you tomorrow

nllong:
Works on mac with an external display. There may be an issue using the laptop display. Approving.

RDmitchell:
@nllong / @axelstudios -- let me know when these various PRs are on one of your dev servers and I will test.

axelstudios:
@RDmitchell Just deployed `develop` to dev1

nllong:
FYI @RDmitchell  -- we will move items into the Test bucket on the [github project page](https://github.com/SEED-platform/seed/projects/3)

RDmitchell:
Perfect. I'll test and close if they are fixed.

RDmitchell:
instance: dev1
SHA: 51231c0
Files:
* Issue 1597-long field names (ESPM).xlsx
* Issue 1597-short field names.csv
Located in this folder
https://drive.google.com/open?id=1FjHCwMcRkrmbQMEJfcSvfYYIm-kgkKlQ

I think that this feature isn't working well enough for the case that it was designed to help, ie, an ESPM file with lots of very long fields. See the 2nd example, and try importing the file above called 
Issue 1597-long field names (ESPM).xlsx

I'm not sure what the solution is -- maybe allow truncating of the ROW data, or add a horizontal scroll bar (?)

The first time I tried this feature, the right hand side of the SEED Header field wouldn't move at all, even though I could see the double horizontal arrow. So I ran developer tools on with disabled cache and it worked. 

Then I tried a file that had short field names (and only 1 record) and I could stretch the SEED header column to a very wide width. Then I tried it with an ESPM file, and I couldn't expand the column very much because the file had long field names in both the SEED Header column and the Data File Header column as well as in the ROW columns. See the examples below.

Here is what I can do for the sample file that has very short file names -- And I think since there was only 1 record in the file, I could expand the field very far. But this is not really a valid test since it only has 1 record.
image
But for the ESPM file, which is really the problem case, since all users are going to be importing and mapping these files, the field names can be quite long, and the data in the ROW columns is also long, so I was not able to expand the SEED Header column very far at all.
image

Issue #1297, opened 2017-03-20
Data quality errors when typecasting data (TypeError Str -> Float)


RDmitchell:
instance: dev1
SHA: d261eb8
See this document for files and details
https://drive.google.com/open?id=1NJsdPueCJSbbxvD0TJKOH-8ratQB4n21QIvD9g97Sgw

(this is also a problem on current production, SHA: 1d98c72)

RDmitchell:
this initially happened on v2 production server, where there were many cycles with 12K records in them. It doesn't appear to have occured on dev1 when there is only 1 cycle of data imported. I will do more testing to see if something on dev1 fixed this, or if enough data is loaded, it will also happen on dev1. 

So am assigning this to myself.

RDmitchell:
instance: staging (LBNL)
SHA: 36cc59d (2.3.1 release)

Now get unicode error when trying to import this file

![image](https://user-images.githubusercontent.com/6314950/37185890-6af50604-22f7-11e8-9fb5-8e45e854ac61.png)

RDmitchell:
might be related to #1565 ?

nllong:
This rule is specifying a number for a field that is a string. I am going to catch this error and report it in data_quality.

![image](https://user-images.githubusercontent.com/1907354/37604750-501d77a4-2b57-11e8-96dd-47e69cbb055e.png)

RDmitchell:
Yes, this is on some list of mine, for the Data Quality setup to check to make sure the rules make sense. We should probably display (?) what the field type is -- in a lot of cases something that looks like a number is stored in extra data as a string (right?) so that might be a problem. I have thought for a long time that the user should be able to specify in mapping what type of a field each is. For the Portfolio Manager files, we should be able to set the field type for most of the fields (at least the ones we know about) that will get put into extra data -- maybe the program is already doing that?

I think there needs to be more transparency to and control by the user for the field types that get put into extra data.

RDmitchell:
How will you report it in Data Quality when the user can't even get to the mapping screen, where the DQ report is available?

RDmitchell:
Instance: dev1
SHA: 6077134

See this doc (in the issue folder referenced at the beginning of this issue)
https://drive.google.com/open?id=1NJsdPueCJSbbxvD0TJKOH-8ratQB4n21QIvD9g97Sgw

Problem file imports, but this error is displayed (doesn't seem to affect anything functionally)
[...]

PR #1604, opened 2018-03-19
1297 data quality typeerror

nllong:
If there is a data quality rule that checks for numbers and the data is a string, then the data quality checking would error out and stop the mapping progress.

#### What's this PR do?
Check if the data are convertible to a number, if not, then log then error as a data quality rule error.

* Go to organization settings
* Create a data quality rule on a normally string field but set the type to number (e.g. city)
* Import some data that and verify that the data quality rules are now reporting that it is unable to convert city to number (or whatever field you are testing).

#1297 

#### Definition of Done:


RDmitchell:
I see that you are trying to convert strings to numbers so my comment on #1297 is probably less relevant since you are already doing this level of checking and converting.

Issue #1583, opened 2018-02-28
Unable to log into ESPM account from SEED


RDmitchell:
### Expected Behavior
Should be able to log into ESPM account from SEED
### Actual Behavior 
Get an error when trying to do that
### Steps to Reproduce
See doc in this folder
https://drive.google.com/open?id=1ytYeN0XSRdTpIYxZVU5JUPMYBIxtRWhG
I think someone from NREL is going to have to do a screen share with the user to try to figure out what the problem is. 
### Instance Information

Server Instance: LBNL production
Server SHA: 68bdd74

Myoldmopar:
@RDmitchell I just requested access to the Google Doc there so I can start investigating.

RDmitchell:
@Myoldmopar -- you should have access to it now. Let me know if you don't. Thanks for working on this.

RDmitchell:
SHA: 2c23bd5
Instance: Dev1

DC just tested this for their ESPM account on Dev1 and they could get the list of Report Templates, but when they picked one, they got another 500 internal server error.

I think we will need to have a screen share with NREL and the user (and me if possible) to see what all the issues are -- they have a pretty complex set of report templates. 

See this doc for more info
https://drive.google.com/open?id=1pb9Twge57ZIifYVuj1X5cf78GS7ArfIxs80nMNLdk_s

RDmitchell:
@Myoldmopar -- should we close this one, and just use the latest one I have generated with the specific issue of which report templates to pull, ie, #1621 ?

RDmitchell:
I believe this is resolved, so I'm closing this, and when I get users to test for #1621 I can make a new issue if needed.

PR #1607, opened 2018-03-27
Extra error handing for PM template list

Myoldmopar:
Issue #1583 describes an error where the user got 500 errors when trying to connect to PM.  Looking at Sentry, it looks like the connection to PM was successful, but something bad was coming through in the response.  The templates weren't fully formed.  It's possible the report template isn't well formed on PM or something.

#### What's this PR do?
This PR adds extra error handling to the template response.  Instead of blindly assuming that each template row will have id and name keys, it checks for them explicitly.  If they aren't there, a debug message is thrown, and that template is not included in the list offered to the user to select from.

For well-formed responses, like those I receive on my own machine with our test PM account, this won't change anything.  For responses that fail because of a bad template response, log messages should be thrown, but the user should simply see the list of valid templates.

#1583

#### Definition of Done:

### DIFF
--- a/seed/static/seed/js/controllers/data_upload_modal_controller.js
+++ b/seed/static/seed/js/controllers/data_upload_modal_controller.js
@@ -64,6 +64,7 @@ angular.module('BE.seed.controller.data_upload_modal', [])
       };
       $scope.pm_buttons_enabled = true;
       $scope.pm_error_alert = false;
+      $scope.pm_warning_message = false;
       /**
        * dataset: holds the state of the data set
        * name: string - the data set name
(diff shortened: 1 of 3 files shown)


Myoldmopar:
@nllong does Sentry include the logging output?

nllong:
I don't think Sentry includes the logging output (as in the server logs).
AKAIK it is just the stack trace.

RDmitchell:
@Myoldmopar / @nllong -- once this is settled and on dev1, I can have the user log in to a test account on dev1 and see if this fixes the problem

RDmitchell:
Let me know when this is on dev1 so it can be tested ... thanks !!

Myoldmopar:
So I can think of two options, and maybe both of them need to be done:
1. Change this so that it reports an error that Sentry traps (just raise an exception?)
2. Change the response to the front-end so that it includes an error that the user can address.

I will feel better about deciding whether to include those once I know if the user's issue is addressed by the code changes currently in place.

Myoldmopar:
OK, I added some more error handling, the server side should be extra robust now.

For the front end, I added a warning message, but I'm looking for advice on the message to be shown:
image

I'll push my changes as of now, and we can refine that and @axelstudios can tell me if I already did something wrong on the JS side.

RDmitchell:
thx @Myoldmopar 

what is the "I will keep trying" part? What does that tell the user?

Myoldmopar:
This is really just the difference between errors and warnings.  In this case, the code encounters "errors" in the data coming back from PM, but they may not impact the user's workflow.  So I still want to warn the user about the issues, but that the code is going to continue to run.

RDmitchell:
I'm not sure the text "I will keep trying" conveys to the user what is going on. 

Maybe something like, "I have encountered issues, but you should be able to import your data" or something?

nllong:
@Myoldmopar can we delete this PR?

PR #1609, opened 2018-04-02
added oauth, add_columns endpoint, and adjusted pint handling

fablet:
#### What's this PR do?
1.  adds an 'add_column_names' endpoint that allows Columns to be added based on an existing record. We found this to be necessary to make column selections available when records are upload through api endpoint rather than the front end.
2. adds OAuth2 functionality, with a JWT Grant option.  See https://github.com/GreenBuildingRegistry/jwt-oauth2 for additional details and JWT Grant base client side setup.  SessionAuthentication and SEEDAuthentication were left in place.  The 'authentication_classes' attribute was removed from the views that had set these two classes explicitly so that the default athentication classes, which includes all three would be use.
3.  Removed pint handling from SEEDJSONRenderer and updated PintQuantitySerializerField to handle the remaining cases requiring collapse_unit.  Manually testing endpoints that depend on PropertyStateSerializer produced multiple failures (for example, 'history', where there were further nested 'state' serializations, or where the data_name was "properties" but the result was a single instance).  PintQuantitySerializerField should handle all instances of Quantity type fields

#### Definition of Done:
- [x ] Does this add new dependencies? If so, does PIP, npm, bower requirements need to be updated?
no

### DIFF
--- a/seed/serializers/pint.py
+++ b/seed/serializers/pint.py
@@ -97,11 +97,11 @@ def add_pint_unit_suffix(organization, column):
 
     try:
         if column['dataType'] == "area":
-            column['displayName'] = \
-                format_column_name(column['displayName'], organization.display_units_area)
+            column['displayName'] = format_column_name(
+                column['displayName'], organization.display_units_area)
         elif column['dataType'] == "eui":
-            column['displayName'] = \
-                format_column_name(column['displayName'], organization.display_units_eui)
+            column['displayName'] = format_column_name(
+                column['displayName'], organization.display_units_eui)
     except KeyError:
         pass  # no transform needed if we can't detect dataType, nbd
     return column
@@ -125,9 +125,15 @@ class PintQuantitySerializerField(serializers.Field):
     """
 
     def to_representation(self, obj):
-        if isinstance(self.root.instance, PropertyView):
-            org_id = self.root.instance.state.organization_id
-            org = Organization.objects.get(pk=org_id)
+        if isinstance(obj, ureg.Quantity):
+            if isinstance(self.root.instance, list):
+                state = self.root.instance[0] if self.root.instance else None
+            else:
+                state = self.root.instance
+            try:
+                org = state.organization
+            except AttributeError:
+                org = state.state.organization
             value = collapse_unit(org, obj)
             return value
         else:
--- a/seed/serializers/pint.py
+++ b/seed/serializers/pint.py
@@ -11,9 +11,6 @@ from django.core.serializers.json import DjangoJSONEncoder
 from quantityfield import ureg
 from rest_framework import serializers
 
-from seed.lib.superperms.orgs.models import Organization
-from seed.models import PropertyView
-
 AREA_DIMENSIONALITY = '[length] ** 2'
 EUI_DIMENSIONALITY = '[mass] / [time] ** 3'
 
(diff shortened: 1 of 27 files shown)


fablet:
The travis tests are failing with the following error:
ImportError: No module named oauth2_provider
I checked the 'python installed' list and can see that none of the new requirements I added were installed.  I had this same issue last time I submitted a pull request.  Is there an extra step I am missing to get travis to find and install any new packages?

nllong:
Thanks for the PR! The module error is due to Travis's caching. I've purged the cache. I'll review this soon and get back to you if there are any issues.

Did this address the Pint serializer email that you sent?

fablet:
I did, and a few other issues with pint and the PropertyState serializer dependent endpoints while I was debugging the original issue.
Thank you.
Fable

nllong:
Thanks for fixing it! And sorry for not catching it sooner.

seed/views/organizations.py
```diff
@@ -205,7 +203,6 @@ class OrganizationUsersSerializer(serializers.Serializer):
 
 class OrganizationViewSet(viewsets.ViewSet):
     raise_exception = True
-    authentication_classes = (SessionAuthentication, SEEDAuthentication)
```
nllong: seems like we should get rid of all of these (e.g. https://github.com/SEED-platform/seed/blob/a27db02f2f837ffa3592f2baf5474b73ec4e6584/seed/api/v2/views.py#L13).

This will allow for the new oauth2 to work as expected because of this:
https://github.com/SEED-platform/seed/blob/986e7bdf5adc81477a1bd14683a3f9d12bb242c1/config/settings/common.py#L266

seed/views/certification.py
```diff
@@ -306,180 +314,37 @@ class GreenAssessmentPropertyViewSet(SEEDOrgModelViewSet):
         Return an assessment property instance by pk if its associated
         assessment is within the specified organization.
 
-        :GET: Expects organization_id in query string.
-        :Parameters:
```
nllong: thanks for cleaning these up!

seed/serializers/pint.py
```diff
@@ -125,9 +122,15 @@ class PintQuantitySerializerField(serializers.Field):
     """
 
     def to_representation(self, obj):
-        if isinstance(self.root.instance, PropertyView):
-            org_id = self.root.instance.state.organization_id
-            org = Organization.objects.get(pk=org_id)
+        if isinstance(obj, ureg.Quantity):
```
nllong: much better, thank you.

seed/views/organizations.py
fablet: agreed.  I thought I caught them all already, but I guess a couple slipped past me.

nllong:
we can grab them in another PR someday. Not a big deal.

nllong:
@fablet -- it appears that something in here (most likely) is now causing a permissions error when accessing the API via the browser. It seems like the sessionauthentication is not longer working as expected.

image

When I tested this pull request it seems to be fine but I vaguely remember that I didn't migrate the DB when I tested.

Any ideas?

fablet:
@nllong -- I'm afraid I don't have any ideas.  When I try the same action in my local environment I only get the permissions error if I am logged out of the front end.  Are you able to get a result through the swagger interface?

nllong:
okay. Let me mess around a bit more. This seems strange.

Issue #1621, opened 2018-04-25
PM Report Template list -- not all the reports are listed in SEED


RDmitchell:
### Expected Behavior
See all the report templates in a PM account
### Actual Behavior 
Not all are being displayed -- some users create "Nested Templates" and only the main templates, not the nested templates, are being displayed in SEED
### Steps to Reproduce
See the doc called "Issue 1621 - PM Report Template list -- not all reports are listed"
in this folder
https://drive.google.com/open?id=1AVFXVzT9CpfVz1Xjz5NMjetBybDbLThQ
### Instance Information
Instance: dev1 (NREL test instance)
SHA: 5744ecc (not completely sure -- user tested this last week, SHA could have been updated since then)

RDmitchell:
@Myoldmopar --  In addition to the info in the doc referenced above, I believe I have recreated the Data Request nested Report Template in the sample PM account we have been using internally for this testing. 

So if you log into that test account, you should see this
![image](https://user-images.githubusercontent.com/6314950/39220429-df36fc58-47e5-11e8-9d73-d30a57d2e93d.png)

So the "Date Request: SEED City Test Report April 24, 2018" is I believe the type of "Report Template" that the cities need to see in the report template pulldown list in SEED -- as suggested from DC in the google doc above, it could be a "2 stage" pulldown list, ie, see the "parent" template, and then the nested "Data Request" associated with the parent.

RDmitchell:
@Myoldmopar -- see page one of the doc referenced in the main description of this issue. I have added instructions about how to get to a sample Data Request in our test account, which now has data in it. So hopefully you can use that to test, and I believe it matches how the users are getting their data.

RDmitchell:
Instance: dev1
SHA: c18e0ca
Org: LBNL 61

See the notes for issue #

All the fields were imported properly (I compared them field by field to an Excel file downloaded by hand from ESPM). 

There were a few field names in the import (and mapping) that didn’t match the ESPM Excel file, partly due to the internal ESPM field names, and partly due to default SEED mapping. The field names can be changed once in the SEED mapping and then the program will remember them the next time the file is imported.

Examples include
* **Year Ending** in ESPM Excel file == **period_ending_date** in the auto ESPM import
* **Address 1** in ESPM Excel file == **Address** in the auto ESPM import
* **National Median ENERGY STAR Score** in ESPM Excel file == **national_median_scor**e in the auto ESPM import

So this seems good to go. 

[...]

PR #1625, opened 2018-05-14
Fix some portfolio manager child data request stuff

Myoldmopar:
The portfolio manager workflow was designed around using a single template, and having SEED generate and then download that report.  A new workflow was encountered, where the templates could have child data requests that behave a bit differently.

#### What's this PR do?
With this PR, we now checks to see if the current row in PM has any child rows such as these:
image
If those are encountered, we keep track of those rows as slightly different entities.  If the user selects a regular template row to report on, the code behaves as it has, and generates a report then downloads it and imports the data, exactly as it currently does.  So we haven't broken anything.  But if a child row is selected, there is no generation step, instead we just directly call the server with a query string and it gives us back the XML data.  So we now should be able to support the new workflow.

- Test template list request with invalid authentication, it should fail
- Test template list request with valid authentication, it should succeed
  - if there are child data requests, they should show up under the parent row:
image
  - (I thought about indenting these, but that might introduce a bit of unnecessary complexity, so I didn't attempt it right now)
- Test report generation for the parent object, if it has properties, it should import them, if it is empty, it should gracefully fail to import
- Test report generation for any child objects, again their success should be dependent on whether the request is properly formed on PM

#1621 

Above

#### Definition of Done:

### DIFF
--- a/seed/views/portfoliomanager.py
+++ b/seed/views/portfoliomanager.py
@@ -16,9 +16,9 @@ import urllib
 import requests
 import xmltodict
 from django.http import JsonResponse
+from rest_framework import serializers, status
 from rest_framework.decorators import list_route
 from rest_framework.viewsets import GenericViewSet
-from rest_framework import serializers, status
 
 _log = logging.getLogger(__name__)
 
@@ -82,7 +82,15 @@ class PortfolioManagerViewSet(GenericViewSet):
         template = request.data['template']
         pm = PortfolioManagerImport(username, password)
         try:
-            content = pm.generate_and_download_template_report(template)
+            if 'z_seed_child_row' not in template:
+                return JsonResponse(
+                    {'status': 'error', 'message': 'Invalid template formulation during portfolio manager data import'},
+                    status=status.HTTP_400_BAD_REQUEST
+                )
+            if template['z_seed_child_row']:
+                content = pm.generate_and_download_child_data_request_report(template)
+            else:
+                content = pm.generate_and_download_template_report(template)
         except PMExcept as pme:
             return JsonResponse(
                 {'status': 'error', 'message': pme.message},
@@ -96,8 +104,11 @@ class PortfolioManagerViewSet(GenericViewSet):
             properties = content_object['report']['informationAndMetrics']['row']
         except KeyError:
             return JsonResponse(
-                {'status': 'error', 'message': 'Template XML was properly formatted but missing expected keys.'},
-                status=500
+                {
+                    'status': 'error', 'message':
+                    'Processed template successfully, but missing keys -- is the template empty on Portfolio Manager?'
+                },
+                status=status.HTTP_400_BAD_REQUEST
             )
 
         return JsonResponse({'status': 'success', 'properties': properties})
@@ -152,7 +163,7 @@ class PortfolioManagerImport(object):
             response = requests.get(url, headers=self.authenticated_headers)
         except requests.exceptions.SSLError:
             raise PMExcept("SSL Error in Portfolio Manager Query; check VPN/Network/Proxy.")
-        if not response.status_code == 200:
+        if not response.status_code == status.HTTP_200_OK:
             raise PMExcept("Unsuccessful response from report template rows query; aborting.")
         try:
             template_object = json.loads(response.text)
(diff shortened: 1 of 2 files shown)


Myoldmopar:
@nllong @axelstudios @RDmitchell  I think this could be ready for a test deployment on whatever server you like.

seed/views/portfoliomanager.py
```diff
@@ -82,7 +82,15 @@ def report(self, request):
...
-            content = pm.generate_and_download_template_report(template)
+            if 'z_seed_child_row' not in template:
+                return JsonResponse(
+                    {'status': 'error', 'message': 'Invalid template formulation during portfolio manager data import'},
+                    status=status.HTTP_400_BAD_REQUEST
+                )
+            if template['z_seed_child_row']:
+                content = pm.generate_and_download_child_data_request_report(template)
+            else:
+                content = pm.generate_and_download_template_report(template)
```
Myoldmopar: "Child" data request rows, and "parent" template rows behave significantly different, so we have a switch here on which worker function to call to actually get the XML data response.

seed/views/portfoliomanager.py
```diff
@@ -96,8 +104,11 @@ def report(self, request):
             properties = content_object['report']['informationAndMetrics']['row']
         except KeyError:
             return JsonResponse(
-                {'status': 'error', 'message': 'Template XML was properly formatted but missing expected keys.'},
-                status=500
+                {
+                    'status': 'error', 'message':
+                    'Processed template successfully, but missing keys -- is the template empty on Portfolio Manager?'
+                },
+                status=status.HTTP_400_BAD_REQUEST
```
Myoldmopar: Cleaned up the verbiage and the response status code (400 is more appropriate here than 500...) to better indicate the response status from PM.

seed/views/portfoliomanager.py
```diff
@@ -164,10 +175,30 @@ def get_list_of_report_templates(self):
...
             raise PMExcept("Could not find rows key in template response; aborting.")
         templates = template_object["rows"]
+        template_response = []
         for t in templates:
+            t['z_seed_child_row'] = False
+            template_response.append(t)
             _log.debug("Found template,\n id=" + str(t["id"]) + "\n name=" + str(t["name"]))
-
-        return templates
+            if 'hasChildrenRows' in t and t['hasChildrenRows']:
```
Myoldmopar: This section of code is where I check for the child data request rows and add them to the array being returned.  I tag them with a key so that we can check later to determine which type we are operating upon.

seed/views/portfoliomanager.py
```diff
@@ -234,6 +265,33 @@ def generate_and_download_template_report(self, matched_template):
             response = requests.get(download_url, headers=self.authenticated_headers)
         except requests.exceptions.SSLError:
             raise PMExcept("SSL Error in Portfolio Manager Query; check VPN/Network/Proxy.")
-        if not response.status_code == 200:
+        if not response.status_code == status.HTTP_200_OK:
+            raise PMExcept("Unsuccessful response from GET trying to download generated report; aborting.")
+        return response.content
+
+    def generate_and_download_child_data_request_report(self, matched_data_request):
```
Myoldmopar: This new worker function simply makes one URL request to get the XML data response.

seed/views/portfoliomanager.py
```diff
@@ -96,8 +104,11 @@ def report(self, request):
             properties = content_object['report']['informationAndMetrics']['row']
```
Myoldmopar: It was great that the child data request XML objects came back in the same form once we get to this point, so there wasn't any major import code changes to make.

RDmitchell:
instance: dev1
SHA: f5fb0b3
Org: LBNL 200

@Myoldmopar -- when I test this, I don't see the child reports. I only see the "Parent"

image

Maybe you code isn't on develop yet?

Myoldmopar:
Definitely not on develop yet.  I just pushed the branch up.  I've got to clean up at least one failing test (only a string change afaik), then it could/should be pushed to a dev machine for you to test on.  I also have an idea for how to unit test this whole module better, so I'll be pushing up additional unit tests tonight/tomorrow, but the test deployment does not need to wait on that.

RDmitchell:
Ah, ok. I jumped the gun! Thx

Myoldmopar:
I'm getting 100% pass now with the last commit, hopefully Travis is happy with it as well.

RDmitchell:
This seems to be working well. I can see and select the "sub" reports as described in #1621

PR #1630, opened 2018-05-17
Superuser improvements

axelstudios:
#### What's this PR do?
- Superusers can always get to the admin page now, even if they don't belong to the current org they have selected
- Orgs you don't belong to have a light pink background in the org dropdown (this should only apply to superusers)
- The admin link does not appear for non-superusers now
- Improved data resolutions for the notes page to happen before page load

- As a superuser, remove yourself from an org, switch to that org, and try going to the admin page.  It should allow you now.  Also check that the background is pink in the org dropdown
- As a non-superuser, the `Admin` link should no longer be visible


nllong:
Looks good. I tested as superuser and as not-superuser. Performed as expected. I like the pink background for orgs that I don't belong to when running as superuser.

@axelstudios -- the jasmine tests are failing, can you look at them before merging?

RDmitchell:
@axelstudios -- I don't quite understand the test instructions. 

When I log in as superuser, and go to the admin page, I do see that all the orgs I am not a member of (the superuser email address) are pink. 

But I don't know what I am supposed to do after that. I don't think I understand what this is fixing, probably because I don't use the superuser functions in the same way that you do. 

If you are convinced that this is well-enough tested by you, I would say you can close it.

nllong:
I think it showing up pink says that it works as expected.

RDmitchell:
OK, cool. I will move it to Done !

PR #1631, opened 2018-05-18
Fix `Only Show Populated` to use the current cycle

axelstudios:
The `Only Show Populated Columns` feature wasn't always requesting the correct inventory records if multiple cycles were present, and in that case could select the wrong columns.
#### What's this PR do?
Fixes the cycle id that was passed to the inventory data request.
- Create two cycles, e.g. 2016 and 2017
- Import different data into one or both cycles
- Run the  `Only Show Populated Columns` feature for both cycles
- Check that the correct columns are shown for data belonging to the current cycle

### DIFF
--- a/seed/static/seed/js/controllers/inventory_list_controller.js
+++ b/seed/static/seed/js/controllers/inventory_list_controller.js
@@ -165,8 +165,8 @@ angular.module('BE.seed.controller.inventory_list', [])
             currentProfile: function () {
               return $scope.currentProfile;
             },
-            cycle_id: function () {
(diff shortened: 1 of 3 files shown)


RDmitchell:
is this up on dev1 for testing?

RDmitchell:
instance: dev1
SHA: c18e0ca
Org: LBNL 61

Made 2 cycles, imported the same file into each but with different mapping. For each cycle, used the Show Only Populated Columns option to make a profile, which then reflected the changed fields for each cycle. It all seemed to hang together.

Issue #1632, opened 2018-05-19
Units are off for Energy Use and EUI when using the ESPM Login from SEED to import data


RDmitchell:
### Expected Behavior
When importing data from ESPM using the ESPM log in function in SEED, I expect to get exactly the same data that I would get if I exported that data by hand from ESPM as an Excel File
### Actual Behavior 
The units for the ESPM data being imported into SEED via the SEED ESPM login are off by 1,000 for all fields of Energy Use and EUI. This may be something to do with how units changed in SEED, or it could be that ESPM does some units math before exporting the final values to the Excel files that are downloaded "by hand" from ESPM. 
### Steps to Reproduce
See the doc in this Issue folder with information about how to reproduce this
https://drive.google.com/open?id=1WgGnUzk8Fsvk8RSYjxwyuqO15beVhmKz 
### Instance Information
Instance: dev1
SHA: c18e0ca
Org: LBNL 61

RDmitchell:
instance: dev1
SHA:  8132824
org: LBNL 61

Using the instructions posted above, reimported and checked all the data. Looks good !!

PR #1637, opened 2018-05-21
Use kBtu during imports

Myoldmopar:
During PM imports, the value was converted to BTU and assigned a proper pint unit quantity.  Unfortunately, this unit did not persist with the data throughout the entire import process.  Because the data gets written to an import_file locally and then processed from there like a normal import, the unit did not get carried along.  I mistakenly thought Pint didn't have kBTU as it isn't listed in the definition file.  However, the btu unit can be tagged with a kilo prefix, so that's what I changed this to do.

#### What's this PR do?
When we import from PM, we now keep the numeric value as it was from PM, and tag it with kBTU (even though that isn't necessarily persisted).  This should allow the normal import process to grab the data in what it thinks is kBTU and everything will be good.

Test a pm import file process and a pm import process and make sure both come in with the same units.

#1632 

#### Definition of Done:

### DIFF
--- a/seed/data_importer/views.py
+++ b/seed/data_importer/views.py
@@ -247,11 +247,9 @@ class LocalUploaderViewSet(viewsets.ViewSet):
                 return {'success': False, 'message': 'Could not cast value to float: \"%s\"' % string_value}
             original_unit_string = pm_value['@uom']
             if original_unit_string == u'kBtu':
-                new_val = float_value * 1000  # convert to btu manually
-                pint_val = new_val * units.BTU
+                pint_val = float_value * units.kBTU
             elif original_unit_string == u'kBtu/ft²':
-                new_val = float_value * 1000  # convert to btu manually
-                pint_val = new_val * units.BTU / units.sq_ft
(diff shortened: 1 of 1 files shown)


nllong:
The test_search_identifier is, unfortunately, a non-deterministic test. Sorry. Just hit restart. We need to fix this test some day.

seed/data_importer/views.py
```diff
@@ -247,11 +247,9 @@ def _get_pint_var_from_pm_value_object(pm_value):
                 return {'success': False, 'message': 'Could not cast value to float: \"%s\"' % string_value}
             original_unit_string = pm_value['@uom']
             if original_unit_string == u'kBtu':
-                new_val = float_value * 1000  # convert to btu manually
-                pint_val = new_val * units.BTU
+                pint_val = float_value * units.kBTU
             elif original_unit_string == u'kBtu/ft²':
-                new_val = float_value * 1000  # convert to btu manually
-                pint_val = new_val * units.BTU / units.sq_ft
+                pint_val = float_value * units.kBTU / units.sq_ft
```
nllong: 👍

Issue #1587, opened 2018-03-07
Mapping: If columns are mapped to both Property and Taxlot tables, ensure that a matching field is present for each table


axelstudios:
### Expected Behavior
Import a file and map columns to both tables.  You should not be allowed to proceed if both a Property matching columns AND a Taxlot matching column are not mapped.

### Actual Behavior 
Mapping a single matching column for either table allows you to proceed, creating unmatchable records on the other table

### Steps to Reproduce
Import a file, map all columns to one table with a matching column, and then map a single non-matching column (like `City`) to the other table.

RDmitchell:
Ah, good. I thought the current behavior was not a good idea. Thanks !

RDmitchell:
Another example -- a bit of a weird edge case, but another example of why we need to prevent users from creating a mapping that doesn't have a matching field for both tables when they are both in one mapping.

https://drive.google.com/drive/folders/10zqJk2pF9h7_LgyY2me17R9RNiUTCIIK?usp=sharing

RDmitchell:
instance: LBNL staging
SHA: 9e003b3
Org: LBNL 551
Cycle: Test 1639

This seems to be working

* If there are no matching fields for either table the program puts up an error message about each table
![image](https://user-images.githubusercontent.com/6314950/41205814-489dd912-6caf-11e8-90a5-f409c8fae476.png)

* If only one of the tables (in the case where both tables are being mapped) has a matching field the program displays the error message about the table missing a matching field.
In this case Property has a matching field but Tax Lot does not and the program reports that fact
![image](https://user-images.githubusercontent.com/6314950/41205846-b2261bb0-6caf-11e8-9711-cb27ad1e4594.png)
In this case Tax Lot has a matching field but Property does not
![image](https://user-images.githubusercontent.com/6314950/41205867-03bffde2-6cb0-11e8-9f0f-b7c9524c7784.png)
* If both tables have a matching field set, no errors are reported.
![image](https://user-images.githubusercontent.com/6314950/41205872-275b0eae-6cb0-11e8-9933-329d9f4291d7.png)

PR #1651, opened 2018-05-25
Strict table mappings

axelstudios:
#### What's this PR do?
During mapping, if columns are mapped to both the Property and Tax Lot tables then require at least one matching field from each table
- Upload a file
- Set all columns to `Property`.  There should be an error only if at least one of the 4 required Property fields is not set (Address Line 1, Custom Id 1, Pm Property Id, Ubid)
- Set all columns to `Tax Lot`.  There should be an error only if at least one of the 3 required Tax Lot fields is not set (Address Line 1, Custom Id 1, Jurisdiction Tax Lot Id)
- Set both `Property` and `Tax Lot` columns.  There should be an error if a mapping field for each table is not set.
#1587


RDmitchell:
@axelstudios -- should I put 1587 in the Test column in the Project Tracker?

axelstudios:
@RDmitchell No, it's not deployed yet, pending review and merging

RDmitchell:
Right, I'm always jumping the gun. I realized it hadn't been merged after I made that comment. I'll let you put it in the project when you think it's ready !! :)

RDmitchell:
@nllong -- So I guess this didn't make it into the 2.3.2 release?

axelstudios:
@RDmitchell - It did, it's in v2.3.2

PR #1653, opened 2018-05-31
Generates PM Import file name based on date

Myoldmopar:
Multiple ESPM imports could not be distinguished because they were given the same file name.

#### What's this PR do?
Generates a filename based on the current date to allow the user to distinguish between different imports

- Baseline
  - import from ESPM on develop branch
  - look at a property detail for newly imported property
  - file name should just be pm_import.csv
- Fix
  - import from ESPM on this branch
  - look at property detail for newly imported property
  - file name should be pm_import_2018_05_31.csv or something like that
- Additional verification
  - I guess you could test it again the next day and it would make it clear which import the property details were imported

#1548 

Baseline images shown in the ticket.

Fixed like this:
image


Myoldmopar:
@RDmitchell @nllong I have no idea how this wasn't already in trunk.  Anyway, this will fix #1548 directly.

RDmitchell:
@nllong -- do we want to add this to the 2.3.2 release?

nllong:
@RDmitchell -- This will be post 2.3.2. It is more of a cosmetic fix.

RDmitchell:
@nllong -- it's actually a bit more than cosmetic, as it is what appears in the column heading in the detail view for each record, so it will help users keep track of what has been uploaded and when. But post 2.3.2 is ok.

Issue #1654, opened 2018-06-01
500 Error when logging into ESPM from SEED


RDmitchell:
### Expected Behavior
Log into ESPM account from SEED
### Actual Behavior 
Got an error when trying to do this
### Steps to Reproduce
See this doc
https://drive.google.com/open?id=12FvCxHNs8ZXpQuP_2vs8vMn8Xg7mmTOn
### Instance Information
Server: dev1
SHA: 90e4a7d

RDmitchell:
@Myoldmopar -- SF is having the same problem they originally did logging into ESPM from SEED. The details are in the doc referenced above. Thx

RDmitchell:
Server: dev1
SHA: 90e4a7d

I just tested logging into our (simple) ESPM test account and it worked.

RDmitchell:
We will probably have to work directly with SF to figure out the subtleties of what is going on.

RDmitchell:
Both SF and Berkeley are having this problem.

nllong:
test on dev1

RDmitchell:
instance: dev1
SHA: 6639ad8
Org: LBNL 20

From the LBNL ESPM test account that I set up, I imported data from the "Data Request:SEED City Test Report April 24 - Test 2" under "SEED City Test Report".
![image](https://user-images.githubusercontent.com/6314950/45389430-c3f15c00-b5d0-11e8-8bd5-8032558aeb66.png)
The 4 records imported properly into SEED
So I am not getting this error anymore.

However, when I compare, field by field, the data imported into SEED vs the data in the file downloaded from ESPM, two fields are missing from the SEED import -- 
* Site EUI (kBtu/ft²)
* Source EUI (kBtu/ft²)

I will make a separate issue for that, since this issue is about a 500 error which doesn't seem to be happening anymore. 

The real test for this issue, however, is if San Francisco can get their large data set imported directly from ESPM, which it seems they can.

PR #1661, opened 2018-06-11
Address PM Import Error

Myoldmopar:
Users are seeing 500 errors when trying to connect to portfolio manager.

#### What's this PR do?
As of right now it just adds extra error handling, because I can't reproduce the problem to investigate.  We'll need to get a screen share session to allow investigating further.
* Sort the user email addresses when creating a new organization

This should be pushed to a dev server, and a screen share should be set up to allow me to see a little more of the problem so I can figure out what we need to do to increase our robustness.

#1654


nllong:
should we deploy this branch on dev1 for test, or merge down and then test?

Myoldmopar:
I guess your call.  My preference would be to deploy on a branch to dev, and have a client check it out first, but I'm not too concerned.
 
I've gathered a little more info from the Sentry responses, so I feel like I understand what is happening on our end to cause the 500: the templates coming from ESPM are missing the "id" field.  I can continue trapping for this "erroneous and unexpected" condition, but I would love to also figure out why ESPM is providing templates without an ID for these cities.

nllong:
I just deployed this branch on dev1.

Myoldmopar:
@RDmitchell could you have some folks test out on dev1.  The 500 server error should be gone, but I would really like the client to look closely at what templates are available in the list.  If there are any that are missing from the list then that will help clue us in to why some of the templates come from ESPM without an ID.  Thanks!

RDmitchell:
OK, I will have them test logging in. 

If they still can't log in, should I facilitate a screen share?

Myoldmopar:
@nllong I added a bunch more error handling to try to diagnose this.  I think it's ready for a deployment and quick check from the client to get some informative info during the reporting step.

nllong:
deployed as of an hour ago with the latest develop merged into this as well.

nllong:
@Myoldmopar -- we are planning on releasing this change on friday cob. Do you think this will be warpped up b then?

Myoldmopar:
Found a few things thanks to my webmeeting today.
1. The filename of the xml coming from ESPM needs to be sanitized for slashes and the like.
2. The data request children need to be indented under the parent to clearly replicate the formulation on ESPM
3. There is some funny shenanigans with a template report if it is a "Generate Preview" based report -- it only comes with a single property and the xml is different.  Yet another special case to deal with.

@nllong @axelstudios Tomorrow morning I should have my local branch tidied up (from some quick hacking earlier today) to a point where we can deploy and have this tested out.

nllong:
:+1

seed/static/seed/partials/data_upload_modal.html
```diff
@@ -225,7 +225,7 @@ <h4 class="modal-title" ng-switch-when="13" translate>Portfolio Manager Import</
         <div class="row" style="padding-bottom:10px">
             <label class="control-label col-lg-4 col-sm-4" for="inputPMReportTemplateName">Report Template Name <small>{$ pm_templates.length ? '(' + pm_templates.length + ')' : '' $}</small></label>
             <div class="col-lg-8 col-sm-8">
-                <select ng-model="pm_template" ng-disabled="!pm_buttons_enabled" ng-options="x.name for x in pm_templates"></select>
+                <select ng-model="pm_template" ng-disabled="!pm_buttons_enabled" ng-options="x.display_name for x in pm_templates"></select>
```
Myoldmopar: To show the "child" template rows better, I added a display name key to the template dictionaries.  I needed to keep the "name" key unaltered to use it in operations, but this display_name key allows me to indent children objects.  It looks really good.

seed/views/portfoliomanager.py
```diff
@@ -57,6 +57,11 @@ def template_list(self, request):
                 {'status': 'error', 'message': pme.message},
                 status=status.HTTP_400_BAD_REQUEST
             )
+        except Exception as e:
```
Myoldmopar: This is just a catch-all during the template list retrieval, for anything not covered.

seed/views/portfoliomanager.py
```diff
@@ -82,36 +87,55 @@ def report(self, request):
...
+                if isinstance(possible_properties, list):
+                    properties = possible_properties
+                else:  # OrderedDict, anything else
+                    return JsonResponse(
+                        {
+                            'status': 'error',
+                            'message': 'Property list was not a list...was a preview report template used on accident?'
+                        },
+                        status=status.HTTP_400_BAD_REQUEST
+                    )
```
Myoldmopar: This little error block is an important addition.  A key issue being hit was that SF was (and even I was) accidentally selecting the wrong template in the list because of a sort-of duplicated name.  The one being selected was set on ESPM as a "Preview", which meant the data didn't come back as a list of properties at this report->informationAndMetrics->row key level.  Instead, it came back as a dict for a single property.  I'm using this type inference as a check for whether this request is based on a preview, and warning the user if so.

seed/views/portfoliomanager.py
```diff
@@ -176,9 +200,14 @@ def get_list_of_report_templates(self):
             raise PMExcept("Could not find rows key in template response; aborting.")
         templates = template_object["rows"]
         template_response = []
-        for t in templates:
+        sorted_templates = sorted(templates, key=lambda x: x['name'])
```
Myoldmopar: Sorting the templates here is also an important addition.  This sorts the "parent" templates, children stay as they are underneath the parents.  This makes the template list dropdown choices look exactly like they do over on ESPM when they are sorted by name there.  SF sorts them by name because they have one parent object with a prefix asterisk, as it is their primary one.  Now that one shows up at the top of the list both on ESPM *and* on this dropdown.  :+1: :+1:

seed/views/portfoliomanager.py
```diff
@@ -258,15 +288,20 @@ def generate_and_download_template_report(self, matched_template):
 
         # Finally we can download the generated report
         template_report_name = urllib.quote(matched_template["name"]) + ".xml"
-        download_url = "https://portfoliomanager.energystar.gov/pm/reports/template/download/%s/XML/false/%s" % (
-            str(template_report_id), template_report_name
+        sanitized_template_report_name = template_report_name.replace('/', '_')
```
Myoldmopar: This is the fourth and final important addition here: The report template name had forward slashes in it (a date to be precise), and that wasn't being sanitized.  So I added that replacement here.  @nllong do we have a single location in SEED for sanitizing data for urls?  Should I use like urllib escape or whatever it is?

seed/views/portfoliomanager.py
```diff
@@ -197,6 +226,7 @@ def get_list_of_report_templates(self):
                 _log.debug("Received the following child JSON return: " + json.dumps(child_object, indent=2))
                 for child_row in child_object:
                     child_row['z_seed_child_row'] = True
+                    child_row['display_name'] = "  -  %s" % child_row['name']
```
Myoldmopar: This is the third important addition here.  This adds a prefix to the child templates and makes them look slick.  This should eliminate, or greatly reduce, any confusion between the template list on ESPM and the list on our dropdown.

Myoldmopar:
@nllong @axelstudios @RDmitchell OK, this is ready for a test on a dev server.  Some enormous improvements in the ESPM import workflow with this PR.  The only thing that I know still needs a little love is the sanitizing of the template name before retrieving a report.  It's done enough that SF should be able to test (based on their template names), but needs a little more robustness before dispatching to everyone.

During our poking around, I was able to get SEED to appear to be importing the entire thing for SF, but it was taking a really long time.  I ultimately had to cancel it.  I tried to work around it, mimicking the process, and it looks like everything was working, but I haven't been able to complete an end-to-end import, so I'm anxious to see if this works for SF.

nllong:
@RDmitchell @Myoldmopar -- this is now of dev1.

RDmitchell:
I can do some testing, but the real test will be for SF to test this on dev1

Myoldmopar:
@nllong this is ready to deploy again for testing.  Temporarily it will only allow import of 10 properties via ESPM, but if this is successful, then the end-to-end ESPM import should be proven out.  Still to do on this PR:
- Verify this works for SF
- Verify the mapping screen looks right during the import process.  I was seeing the columnar data showing up in multiple columns on the UI (5 headings - 1 per file, but then 10 columns of data in the table body)
- Move this process to a celery task that registers a status key
- Open it back up to the full amount of properties
- Create additional unit tests around the import process

Myoldmopar:
@nllong I have written some unit tests that will actually go out to ESPM and test our API views to make sure they operate properly.  The tests will only run if some additional environment variables are available.  I'll send the required variable keys and values separately as they will need to be added to CI for the tests to run (otherwise they are `skipIf`'d)

Myoldmopar:
Also, if @nllong or @axelstudios could deploy this branch to dev1, we could get it tested and be close to wrapping this up once I make the function into a background task.

nllong:
deployed to dev1

Myoldmopar:
So SF now sees Error: undefined on the front end when trying to import.  The JS is trying to show `error.data.message`. and apparently message isn't available.  There were only two spots where there was an erroneous response code with no message.  And only one spot in this particular endpoint.  And it's somehow coming out of a double layered TRY block uncaught.  This is very strange.

Anyway, I added a message coming out, so now there should be Error: *something useful*.

@nllong Please deploy this branch to dev1 at your convenience.

nllong:
redeployed.

Myoldmopar:
@nllong thanks for deploying.  I pushed another commit that might fix the column header issue.  At your leisure, would you be able to go ahead and redeploy at 6639ad8 before we test?

seed/data_importer/views.py
```diff
@@ -316,17 +316,21 @@ def create_from_pm_import(self, request):
...
+        #     u'city': u'City',
+        #     u'state_province': u'State',
+        #     u'postal_code': u'Zip',
+        #     u'county': u'County',
+        #     u'country': u'Country',
+        #     u'property_name': u'Property Name',
+        #     u'property_id': u'Property ID',
+        #     u'year_built': u'Year Built',
+        # }
+        # so now it looks like we *don't* need to override these, but instead we should leave all the headers as-is
```
nllong: this makes sense since after downloading and importing the user is dropped at the mapping screen.

RDmitchell:
thx !

PR #1662, opened 2018-06-11
Return inventory records with only the columns specified

axelstudios:
- Previously we had an option to request only specific columns be returned from the backend to improve performance and reduce data transfer.  This PR restores that functionality.
#### What's this PR do?
- Passes an optional profile id to the inventory filter API to return data for the columns in that list settings profile
- Fetches all columns for only selected rows when opening the merge modal
- Added an API resource to return only columns that can be mapped
- Removed deprecated code that prepended `tax_` and `property_` to field names
- Commented unused `calculated_taxlot_ids` and `primary` fields
- On an org with no saved list settings, check that the data returned includes all columns
- Create a list setting (like with `Show only populated columns` and check that only those specific columns are returned.
- Try merging some records - check that only the specific selected records were loaded with all columns in the API response.

### DIFF
--- a/seed/models/tax_lot_properties.py
+++ b/seed/models/tax_lot_properties.py
@@ -74,7 +74,7 @@ class TaxLotProperty(models.Model):
         for f in chain(opts.concrete_fields, opts.private_fields, opts.many_to_many):
             if not getattr(f, 'editable', False):
                 continue
-            if fields and f.name not in fields:
+            if fields is not None and f.name not in fields:
                 continue
             if exclude and f.name in exclude:
                 continue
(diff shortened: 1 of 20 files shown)


RDmitchell:
@axelstudios -- does this fix something that wasn't working with the implementation of the settings profile in the release version?

axelstudios:
@RDmitchell No, this is a performance improvement

seed/models/tax_lot_properties.py
```diff
@@ -234,15 +251,15 @@ def get_related(cls, object_list, show_columns, columns_from_database):
 
                 join_dict = related_map[getattr(join, lookups['related_view_id'])].copy()
                 join_dict.update({
-                    'primary': 'P' if join.primary else 'S',
-                    'calculated_taxlot_ids': '; '.join(jurisdiction_tax_lot_ids),
+                    # 'primary': 'P' if join.primary else 'S',
```
nllong: need to find all vestiges of this and eradicate.

seed/views/properties.py
```diff
@@ -228,9 +230,15 @@ def _get_filtered_results(self, request, columns):
...
                 })
 
-        property_views_list = PropertyView.objects.select_related('property', 'state', 'cycle') \
-            .filter(property__organization_id=org_id, cycle=cycle) \
-            .order_by('id')  # TODO: test adding .only(*fields['PropertyState'])
+        # Return property views limited to the 'inventory_ids' list.  Otherwise, if selected is empty, return all
+        if 'inventory_ids' in request.data and request.data['inventory_ids']:
+            property_views_list = PropertyView.objects.select_related('property', 'state', 'cycle') \
+                .filter(property_id__in=request.data['inventory_ids'], property__organization_id=org_id, cycle=cycle) \
+                .order_by('id')  # TODO: test adding .only(*fields['PropertyState'])
```
nllong: the .only method should speed up the query request significantly. We can test later.

PR #1674, opened 2018-07-11
Report proper filename for uploaded_filename during PM import

Myoldmopar:
PM import filenames were being named properly, but one of the views reported back a static "PortfolioManagerImport" string instead of the filename.  This fixes it.

#### What's this PR do?
Reports back a proper filename with date timestamp so the user can see when it was imported.

Test a PM import process, autogenerated filename reported on the site should have today's date in it.

#1548 

Mapping Page:
image

DataSet Detail Page:
image

#### Definition of Done:

### DIFF
--- a/seed/data_importer/views.py
+++ b/seed/data_importer/views.py
@@ -402,7 +402,7 @@ class LocalUploaderViewSet(viewsets.ViewSet):
 
         # Create a new import file object in the database
         f = ImportFile.objects.create(import_record=record,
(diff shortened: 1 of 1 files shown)


Myoldmopar:
@RDmitchell this will fix it per the screenshots I showed above.

@nllong we can deploy this to have it tested, but I am hoping to have something to test regarding the PM import errors later today or tomorrow morning so maybe we hold off?  Your call.

nllong:
merging down to deploy

Issue #1667, opened 2018-06-22
DQ Export not working from Inventory List -- server failed error


RDmitchell:
### Expected Behavior
Export the results from the DQ results modal when generated in the Inventory list view
### Actual Behavior 
Get "Failed - server problem" message
![image](https://user-images.githubusercontent.com/6314950/41796273-acea9026-761a-11e8-8da9-65ccf95cc6d6.png)

### Steps to Reproduce
Run DQ results from Inventory list view, click export on the modal with the results
### Instance Information
instance: dev1
SHA: 6560a1a

RDmitchell:
Export DOES work when doing it from the DQ Results modal generated during the mapping process

nllong:
I'm not able to recreate this error. What org ID are you on?

RDmitchell:
instance: seeddemostaging
SHA: SHA: 1a8cd10 (2.3.3 Release)
Org: LBNL Test 200
Cycle: 2018 Compliance Cycle

This still seems to be an issue
![image](https://user-images.githubusercontent.com/6314950/42401706-c5094942-812b-11e8-96c3-7871463a950e.png)

nllong:
does this work on dev1? Wondering if this is a server configuration...

RDmitchell:
I'll check

RDmitchell:
I get the same error on dev1

RDmitchell:
And this is the DQ results from the Inventory screen, not the Data Import world.

RDmitchell:
instance: seeddemostaging
SHA: SHA: 1a8cd10 (2.3.3 Release)
Org: LBNL Test 200
Cycle: 2018 Compliance Cycle

I thought that the export of DQ results was working from the Data import, but it is not -- just confirmed that it also gives a server error.

nllong:
This is fixed in the PR listed above. It was a unicode error due to units having a superscript 2.

https://github.com/SEED-platform/seed/pull/1676/files#diff-34d96703c22d340bee50188a76add6caR185

RDmitchell:
instance: dev1
SHA: ??
Org: LBNL 20

DQ export worked from Inventory List & Data Import

Issue #1608, opened 2018-03-28
Matching Performance -- needs work


RDmitchell:
### Expected Behavior
Matching should work even for large files (12K records)
### Actual Behavior 
Program seems to hang for these large files
### Steps to Reproduce
see issue #1297 
from that issue
See this doc (in the issue folder referenced at the beginning of this issue)
https://drive.google.com/open?id=1NJsdPueCJSbbxvD0TJKOH-8ratQB4n21QIvD9g97Sgw
### Instance Information
Instance: dev1
SHA: 6077134

nllong:
The issue is caused by Cloudflare's hardwired time out of 100 seconds.

RDmitchell:
@nllong -- so what does that mean? Is it only an issue on the dev servers?

nllong:
Yes, the error is only on our dev servers because of the DNS that we are
using. We are still going to fix this because waiting 2 minutes is
unreasonable.

RDmitchell:
great, thx for the explanation!

nllong:
@axelstudios -- I think we can add in the need to reenable show only specific columns to this task, right?

axelstudios:
@nllong Yes, that definitely needs to be reenabled.

RDmitchell:
Seems like I can't test this on dev1 because of the cloudflare issue? I will test on staging when we get the code there? 

Or is there something here to test on dev1?

RDmitchell:
Instance: seeddemostaging (LBNL)
SHA: 49e40e0
User: RD+55

This is no longer an issue -- I can import large building list (12,000+ records) and large ESPM data files (15,000+ records) and the program can import, map, and match them.

![image](https://user-images.githubusercontent.com/6314950/46505770-07617380-c7e7-11e8-9807-cf26363b8941.png)

Issue #1498, opened 2017-11-09
Problems importing large file


RDmitchell:
instance: staging
File:

This is a very large file (177,000 records) but not many fields.

RDmitchell:
Here is the error message
![image](https://user-images.githubusercontent.com/6314950/32585498-904efdea-c4b2-11e7-9624-3f406c2866bd.png)

There is an issue with newlines in one of the fields, but I think that was deleted from the file being imported by PNNL.

nllong:
I can test locally first.

I wonder if this is a @mmclark question since it says upload failed. If the upload failed then it is most likely a configuration issue with S3/local file storage.

nllong:
#1565 #852

RDmitchell:
@nllong -- I can try importing this file again. Is it ok to test this on dev1?

RDmitchell:
@nllong -- I just added it to the Project Tracker under Test and assigned it to myself.

RDmitchell:
instance: dev1
SHA:
Org: LBNL 33
sf13m_bldg_footprint_attribZ_pgz_data.xlsx
This file has the "geom" data. 

This time it got a bit further on the import, but failed at 25%.
![image](https://user-images.githubusercontent.com/6314950/40806049-2507d138-64d5-11e8-95f4-20d41c23ab2a.png)

I will try again with the same number of records but with a reduced set of fields.

RDmitchell:
instance: dev1
SHA: 90e4a7d
Org: LBNL 33
sf13m_bldg_footprint_attribZ_pgz_data_no_polygons-MBLRParcelID.csv
* Doesn’t have “geom” data (last field in the previous file)
* Half the size of the file with the “geom” data

See this doc (in the issue folder) which shows the details of this process (in the section dated 5/31/2018)
https://drive.google.com/open?id=1uAnRJ_cyxP5arwhr1psfFTUWqFbWDvorl1w40aHH6yc

This file got through import, mapping, but finally (after many hours) got this error on matching
![image](https://user-images.githubusercontent.com/6314950/40855057-2e478c94-6588-11e8-89ee-fc2c8d931fbc.png)

RDmitchell:
instance: dev1
SHA: 90e4a7d
Org: LBNL 33
sf13m_bldg_footprint_attribZ_pgz_data_no_polygons-MBLRParcelID.csv

[...]

Issue #873, opened 2016-03-25
"NoneType object is not iterable" in _match_buildings


mmclark:
Seen in the celery log on the seed-timeseries instance:

```
[2016-03-24 20:50:22,786: ERROR/MainProcess] Task seed.tasks._match_buildings[b08a550f-b526-4805-a23f-901207f6bc5a] raised unexpected: TypeError("'NoneType' object is not iterable",)
Traceback (most recent call last):
  File "/home/ubuntu/.virtualenvs/seed/local/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/ubuntu/.virtualenvs/seed/local/lib/python2.7/site-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/ubuntu/seed/seed/decorators.py", line 62, in _wrapped
    response = fn(import_file_pk, *args, **kwargs)
  File "/home/ubuntu/seed/seed/tasks.py", line 1375, in _match_buildings
    results, i, can_rev_idx, unmatched_buildings, user_pk
  File "/home/ubuntu/seed/seed/tasks.py", line 916, in handle_results
    default_pk=building_pk,
TypeError: 'NoneType' object is not iterable
[2016-03-24 20:50:30,163: INFO/MainProcess] Task seed.tasks._match_buildings[4a3dd051-1f5a-4b5a-9089-d01786ae1a93] succeeded in 244.049537786s: {'status': 'success', 'progress': 100, 'progress_key': u':1:SEED:_match_buildings:PROG:107'}
```

RDmitchell:
This might have been me trying to unmatch some records in the matching screen. It seemed to put the program into endless spinner mode, which is why I had @mmclark restart the ts instance.

RDmitchell:
This would have been for the SEED App Contest org, which has 22,000+ records, a few hundred of which have timeseries data associated with them.

nllong:
this should not be an issue anymore. not a clear test.

RDmitchell:
@nllong -- do you think it's worth testing? We don't really deal with time series data anymore so I wouldn't know what data set to test for this ... Or maybe you can suggest a test file (or the type of file that would test this?)

nllong:
I am going to close. There is no easy way to test at the moment.

RDmitchell:
agreed. thx

PR #1676, opened 2018-07-11
239 progress bar

nllong:

The progress bar code was difficult to manage due to the various methods that needed to be called in the correct order. For example:

```
prog_key = get_prog_key('check_data', identifier)

data_quality_results = get_cache_raw(DataQualityCheck.cache_key(identifier))
result = {
	'status': 'success',
	'progress': 100,
	'progress_key': prog_key,
    'message': 'data quality check complete',
    'data': data_quality_result
}
set_cache(prog_key, result['status'], result)
return result
```

now becomes

```
cache_key, dq_id = DataQualityCheck.initialize_cache(import_file_id)
progress_data = ProgressData(func_name='check_data', unique_id=dq_id)
return progress_data.result()
```

There is also the ability to initialize the ProgressData object from an existing key (in the redis db) by calling `progress_data = ProgressData.from_key(progress_key)`

The ability to save the results of any progress is much easer now too by calling `progress_data.finish_with_error('StopIteration Exception', traceback.format_exc())`. There are various helper methods that set all the required status, progress, etc fields.

#### What's this PR do?

* General
	* Upgrade celery to 4.2. Required updating tasks that called sub-tasks. 
	* Remove scipy, ipython, and numpy from local.txt requirements. These are now in local-scipy.txt if needed (e.g. for manual migrations)
	* Added new ProgressData class to manage tracking the progress of background tasks
	* Update test_seed_host_api.py (seed_readingtools.py) to support secondary call to data_quality_checks.
* Matching
	* Move equivalence partitioner code into its own file
* MCM
	* Removed old FakeModel, FakeManager, etc code
* Testing
	* New FakePropertyStateFactory method for returning data in the extra_data field to be used for testing raw_data import and mapping
	* Break up the data_importer tests into integration, views, and unit tests. Integration tests require the loading of CSV/XLSX files
* API
	* Removed endpoint `/api/v2/import_files/{id}/data_quality_results`. Use `/api/v2/data_quality_checks/results/?organization={org_id}&data_quality_id={dq_id}`
	* Add endpoint for starting data quality on import file. `/api/v2/import_files/{id}/start_data_quality_checks/?organization={org_id}`
	* Progress endpoints/results no longer return data such as the number of records matched or the data quality results. You must request those from subsequent result APIs.

#### TODOs

* There are several TODO 239 comments that need to be addressed by @nllong
* @axelstudios - can you check the export data from the inventory page. The modal dialog doesn't appear anymore.

* Import files with and without always eager
* Verify data quality works on both mapping and within inventory
* Verify data quality export works in both contexts
* Verify mapping and matching
* Verify deleting inventory and deleting organization with and without always eager

#239 
#1667
#873 - Is no longer applicable, or at least shouldn't happen again.
Working towards #1498 and #1608

#### Definition of Done:
- [X] Does this add new dependencies? If so, does PIP, npm, bower requirements need to be updated?


axelstudios:
A couple issues so far:

With always-eager turned on, the UI shows 200% after clicking `Save Mappings`.  The network responses all have progress at `100` so it must be a front end issue:



Moving the upload status out of the progress response is good, but there are a few missing details:
##### Before (Upload covered-buildings-sample.csv into empty cycle)
image 
image 
image

##### Before (Re-upload covered-buildings-sample.csv into the same cycle)
image
image
image

##### Now (Upload covered-buildings-sample.csv into empty cycle)
image
image

##### Now (Re-upload covered-buildings-sample.csv into the same cycle)
image
image

nllong:
I'm fixing the data results from matching now. Is there any way you can look into the progress bar @axelstudios?

nllong:
@axelstudios -- I fixed the returning of the matching results. it requires a migration.

axelstudios:
When always eager is turned on, requests like `save_raw_data` return with progress == 100 in the response as opposed to 0.  That progress number is then used as a starting point for when it hits the progress api afterward to get the status, thus 100 + 100.  Since it's really just an issue for always eager I clamped the progress values

seed/data_importer/tasks.py
```diff
@@ -954,223 +817,21 @@ def filter_duplicated_states(unmatched_states):
     :return:
     """
 
+    # TODO #239: Should we save the hash in the database, wouldn't that be faster
```
axelstudios: +1

nllong: this will happen in the performance branch. Thanks for your help with this one. It was a doozy, but I think it is much nicer.

Issue #1698, opened 2018-08-24
Add the ability to not overwrite the data by column on merge


nllong:
* Add the functionality to manage column names
* Add field to allow merged data to decide which data set to use, the old or the new

RDmitchell:
See #1646 (closed because it replicates this issue) which is the description from a user of what they are trying to do that this will solve.

RDmitchell:
is this on dev1 for testing?

nllong:
Yes

RDmitchell:
instance: seeddemostaging
SHA: 095736a
Org: LBNL 20

Set the field called "Custom APN" to not be overwritten
![image](https://user-images.githubusercontent.com/6314950/46321199-d1c64b80-c596-11e8-9032-6395b15be0d6.png)
Then imported a file with that value different than what was in the Master record, and the program kept the history of the import, with the different data, but did not change the master record with that new data.
Protected fields in Column Settings also works well
![image](https://user-images.githubusercontent.com/6314950/46117399-e8932980-c1b5-11e8-85c2-ff476ac822ee.png)

RDmitchell:
I have tested this quite a bit. It looks good to me. I have also asked SF to test, although I'm not sure they will have time. 
I summarized my tests (data and doc) in this folder
https://drive.google.com/drive/folders/1EstVKtA2nezaib4DIIlbDNJUc_TTO_1D?usp=sharing

Am closing this issue.

PR #1699, opened 2018-08-24
Field by field merging preferences

nllong:
Allow the user (from the backend at the moment) to specify if columns should be merged based on the existing or the newest. 

#### What's this PR do?
* Adds a new field to the columns that specify "favor new" or "favor existing". It defaults to "favor new". When data are imported and there are records to be merged, then the favor new for the column will choose the most recent data. Favor existing will keep the existing data in the column.
* Clean up the _merge_extra_data method to not use the extra_data_source
* Update tests

@axelstudios -- let me know if you are planning on doing the front end of this. It would be part of the listing of columns page. 

Initially, just verify that nothing broke. Unit tests test the functionality until we have the frontend.

#1698 

We will need to update this screen.

screen shot 2018-08-24 at 3 04 43 pm

#### Definition of Done:


nllong:
Tested 
* new org
* went to column and column mapping pages - verified column list existed and no column mappings. 
* loaded example data
* went to column page, changed display name -- verified it showed up in inventory.
* column mapping -- verified can delete one or all
* column page -- checked filtering
* added merge protection to site EUI and imported new data -- data did not overwrite

Looks good!

Issue #711, opened 2016-02-02
Associating and disassociating orgs causes account to become disfuctional


RDmitchell:
For March, when we dive into the Parent/ Sub Org world in earnest. 

For Robin LBNL 77 account, I associated it with Robin LBNL 6 account, then disassociated it from 6 by just deleting the 77 user from 6. (on seedtest version as of 2/2/2016, 2:58 pm)

This seems to have caused havoc in 77 -- I am now getting errors when trying to go to any of the Nav Bar pages (Data, buildings, projects). @StephenCzarnecki suspects that the Default organization problem didn't get updated when I removed 77 from 6. We will do more rigorous testing when we get to this issue. 

Here is a screen shot of the errors.
![image](https://cloud.githubusercontent.com/assets/6314950/12767335/7044ed6c-c9bd-11e5-9ab8-31325747ff24.png)

RDmitchell:
@nllong -- I don't think we need to deal with this for the upcoming release. So it can move out of In Progress and back to the To Do column

nllong:
Test 1:
* Create 2 users u01 and u02 with two independent orgs (org01, org02).
* Login as u01 and add u02 to org01 as owner
* Refresh page
* Remove user u01 (yourself) from org01
* Error should pop up on page refresh (this is as expected and okay since the user no longer has any organization)
* Need another ticket to have a nice display message if the user has no organizations.

![image](https://user-images.githubusercontent.com/1907354/45059695-16270000-b05a-11e8-99d9-da8caa6f0683.png)

Test 2:
* Create 2 users u01 and u02 with two independent orgs (org01, org02).
* Login as u02 and add u01 to org02 as member
* Logout u02
* Login as u01 and add u02 to org01 as owner
* Refresh page
* Remove user u01 (yourself) from org01
* Error should pop up on page refresh

RDmitchell:
instance: dev1
SHA:

Test 1:
Followed all the steps and at the end, after a page refresh, I get this message, which I think is as expected
![image](https://user-images.githubusercontent.com/6314950/45923013-1fec9800-be8f-11e8-9f87-0863add92c22.png)

Test 2:
Followed all the steps and at the end, after a page refresh, I get this message, which I think is as expected
![image](https://user-images.githubusercontent.com/6314950/45923056-1f083600-be90-11e8-8dc4-16246dac2be6.png)

PR #1704, opened 2018-09-04
change default org to user first org available

nllong:
If a user removes themselves from an organization, then SEED cannot become unresponsive for that organization. 

#### What's this PR do?
* sets the default org to the next one the user has available. 

Note that if you remove yourself, then you must select an available org from the drop down, then click on the left nav to reset your current view.

See instructions at the of #711 

#711

#### Definition of Done:


axelstudios:
I've updated the error message to say `Your SEED account is not associated with any organizations. Please contact a SEED administrator.`

axelstudios:
@nllong Test out the UI changes I added, and then this is good to merge

nllong:
Looks good! 

image

Issue #1601, opened 2018-03-13
Performance is quite slow moving between different views (Detail to List, etc)


RDmitchell:
### Expected Behavior
Relatively "snappy" response when I click on a record in the list view to display the detail view.
### Actual Behavior 
It takes 13 seconds to move from the detail view to the list view. That doesn't seem like too long but if all you are doing all day in SEED is going from list to detail and back to list, it can be quite annoying and time consuming. I would think it would be about 1-2 seconds (or less?) to move between the two in an ideal world
### Steps to Reproduce
in the organization account listed below, Cycle = 2016 Calendar year, click into a detail view for a property, then click the Property link in the upper left to get back to the list view
I can make an account for this org if needed when someone looks at this issue
### Notes
The proposed approach is to use pagination again which will make the moving back to the inventory page much faster.

### Instance Information
instance: staging (LBNL)
SHA: 36cc59d (2.3.1 release)
Org ID: 277

RDmitchell:
is this ready for testing on dev1 ?

nllong:
yes, this is on dev1

RDmitchell:
instance: dev1
SHA: 6639ad8
Org: ID 49
Property records: 16,102

* Selected only show populated columns to get the number of columns reduced
* Resulted in 260 populated columns
* Took approx 50 seconds for the program to refresh the list of properties (260 fields) after reducing the columns
* Time to get to detail view of one of the records: 4 seconds -- VERY FAST (!)
* Time to get back to List view (260 columns): 52 seconds
* Reduced number of fields to: 16
* Time to get from Detail View to List View: 22 seconds
So reducing the # of fields cut the time in half (or slightly better)

Another org (ID=20)
* 1151 records in 2014 Cycle
* 218 fields in property view
* Time to get to Detail View: 2 seconds
* Time to get from Detail view back to List view: 8 seconds
* Reduce # of fields to: 13
* Time to get from Detail view back to List view: 4 seconds

This seems much improved. Am closing

adrian-lara:
@nllong and @axelstudios I think work was done towards this in two separate efforts. On dev3, Loading 1.6k-1.7k records by 145 columns takes 10-15s. My initial thought is that this is reasonable given that column count, but let me know what you all think.

nllong:
Yeah, we need to find a way to load 177,000 records assuming #1498 happens.

nllong:
> Yeah, we need to find a way to load 177,000 records assuming #1498 happens.

[...]

PR #1705, opened 2018-09-10
Inventory View Fixes

axelstudios:
#### What's this PR do?
- Fixed a bug where extra_data columns weren't returned for related taxlot records
- Fixed the show-populated-columns modal to request all columns to properly filter them
- Restored property/taxlot "list" resources that now return all columns for all records
- Passing -1 as the profile id when requesting inventory returns all columns
- Also fixed csv line endings

- Try out the `Only Show Populated Columns` dialog.  Make sure that it shows the expected amount of columns for both properties and taxlots, and check that extra data columns are actually populated in the inventory list view

#1601


seed/views/properties.py
```diff
@@ -324,6 +329,33 @@ def _move_relationships(self, old_state, new_state):
 
         return new_state
 
+    @api_endpoint_class
+    @ajax_request_class
+    @has_perm_class('requires_viewer')
+    def list(self, request):
```
nllong: it's a bummer that you had to add this back in.

axelstudios: Technically we don't need it, and we're not actually using it, but I think it's helpful for documenting the use of passing `-1` as a profile id to return all columns

PR #1706, opened 2018-09-11
Create default columns managed task

nllong:
During migration, some organizations may not have received the default columns. Without the default columns many of the screens won't work correctly.

#### What's this PR do?
Adds a managed task to allow the user to manually create the default columns. Default columns will only be created in organizations that have no existing columns.

In a test database, remove all the columns of an existing organization. It is easiest to create a new org and remove the columns since the column list settings and column mappings do not exist in a vanilla organization.
```
# Check and create default columns in all organization that do not have columns
./manage.py create_default_columns

# Check and create default columns for a specific organization
./manage.py create_default_columns --org_id 8
```

None
#### Definition of Done:

### DIFF
--- a/seed/lib/mappings/mapping_columns.py
+++ b/seed/lib/mappings/mapping_columns.py
@@ -14,6 +14,7 @@ _log = logging.getLogger(__name__)
 def sort_duplicates(a, b):
     """
     Custom sort for the duplicate hash to decide which raw column will get the mapping suggestion
+    based on the confidence.
     """
     if a['confidence'] > b['confidence']:
         return -1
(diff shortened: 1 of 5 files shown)


seed/lib/mappings/mapping_columns.py
```diff
@@ -163,35 +174,38 @@ def duplicates(self):
...
+        column, then select the next available one on the duplicate column. The one with the highest
+        confidence will 'win' the duplicate battle.
 
         :param dup_map_field: String, name of the field that is a duplicate
         :param columns: list, raw columns that mapped to the same result
         :return: None
 
         """
-        _log.debug("resolving duplicate field for {}".format(dup_map_field))
+        _log.debug("resolving duplicate field for %s" % dup_map_field)
```
nllong: this is the actual fix to the unicode issue. Everything thing else were tests and code cleanup.

Issue #1707, opened 2018-09-11
Site EUI (kBtu/ft²) & Source EUI (kBtu/ft²) fields are not imported with autoimport from ESPM


RDmitchell:
### Expected Behavior
All the fields in the ESPM report template would be imported into SEED via the new Import Portfolio Manager Data, which allows the user to log into their ESPM account and pick the report to download data from
### Actual Behavior 
In the test ESPM account I set up, all the fields except Site EUI (kBtu/ft²) & Source EUI (kBtu/ft²) were imported. It could have to do with the units? 
### Steps to Reproduce
Org: LBNL 20
Cycle: 2017 Calendar Year
ESPM Downloaded report is in this issue folder:
https://drive.google.com/open?id=1CHJBkUgU3W4uYta0chZowN7suvyjVybp

From the LBNL ESPM test account that I set up, I imported data from the "Data Request:SEED City Test Report April 24 - Test 2" under "SEED City Test Report".
![image](https://user-images.githubusercontent.com/6314950/45391050-7c6dce80-b5d6-11e8-8829-703efaca84c4.png)
The 4 records imported properly into SEED

However, when I compare, field by field, the data imported into SEED vs the data in the file downloaded from ESPM, two fields are missing from the SEED import --
* Site EUI (kBtu/ft²)
* Source EUI (kBtu/ft²)
![image](https://user-images.githubusercontent.com/6314950/45391226-07e75f80-b5d7-11e8-9a69-de0f335cfe4b.png)
Here is the mapping for those two fields
![image](https://user-images.githubusercontent.com/6314950/45391254-20577a00-b5d7-11e8-9030-0d4da06d40e4.png)
![image](https://user-images.githubusercontent.com/6314950/45391274-32391d00-b5d7-11e8-8311-a044b7482b41.png)
In the Inventory List, View by Property, if I select "Only Show Populated Columns" the program reports back that it has 32 populated columns
If I look at Property List Settings, the two fields called Site EUI and Source EUI (no units) were not selected as populated columns (even though 2 of the records have data).
![image](https://user-images.githubusercontent.com/6314950/45391462-dd49d680-b5d7-11e8-9403-b15a65ec28d9.png)
If I select those two fields to see if they have data
![image](https://user-images.githubusercontent.com/6314950/45391492-fd799580-b5d7-11e8-84b7-4e0c752994e1.png)
And then go back to the Inventory list for Property tab, those fields do not have data.
[...]

PR #1710, opened 2018-09-27
Fix csv import

Myoldmopar:
Doing ESPM imports, an issue was found that was actually related to the data import cleaning process.  The Site EUI was missing after the mapping page was completed.  (The preview -- first 5 rows -- page showed the values properly)

#### What's this PR do?
1. Sanitizes the property notes field when converting from XML to CSV, cleaning newlines so that the csv looks nice.
2. A quick hack to get the EUI and other similar fields to import properly.

Import from ESPM, make sure EUI is imported all the way to the DB

#1707

#### Definition of Done:

### DIFF
--- a/seed/data_importer/views.py
+++ b/seed/data_importer/views.py
@@ -391,6 +391,10 @@ class LocalUploaderViewSet(viewsets.ViewSet):
                         if pm_variable == u'property_name':
                             this_row.append(this_pm_variable)
                             added = True
+                        elif pm_variable == u'property_notes':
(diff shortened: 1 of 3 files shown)


Myoldmopar:
@nllong confirmed your fix works!  I now see EUI being imported.  I'll let you decide how to fix the flake8 issue there, but I approve the functional change here.

RDmitchell:
@nllong -- let me know when this is on dev1 and I will test. thx

nllong:
on dev1 for testing

PR #1725, opened 2018-10-12
use alpine linux

nllong:
Docker has historically been running on Ubuntu 16.04.

#### What's this PR do?
Move over to using Alpine Linux 3.8 for running SEED in Docker. Images are roughly 30% smaller.

Run docker-compose build and docker-compose up. Monkey test the website.

N/A

#### Definition of Done:


docker/nginx-seed.conf
```diff
@@ -1,6 +1,6 @@
 # the upstream component nginx needs to connect to
 upstream seed_upsteam {
-    server unix:/tmp/uwsgi-seed.sock;
+    server unix:///tmp/uwsgi-seed.sock;
```
axelstudios: @nlong Three slashes?

nllong: yes. similar to file:///

nlong: I agree w/ @nllong

PR #1732, opened 2018-11-09
Update the update datetime field when changing tax lot or property

nllong:
When manually updating a property or taxlot would not update the updated date time.

#### What's this PR do?
* When manually changing a field via the UI, then make sure the update datetime changes.

* Upload data
* Select a random property and view the detailed page
* Edit any of the values in the detailed page
* Verify that the date time changed to most recent timestep
* Do it again on the same data and make sure it still works (the second time is actually a slightly different path through the code)
* Repeat the whole procedure for a tax lot

N/A

N/A

#### Definition of Done:


nllong:
@axelstudios can you take a look at this when you get a chance? Thanks!

Issue #1734, opened 2018-11-10
segfault when loading a lot of records to view in Inventory List


RDmitchell:
### Expected Behavior
Able to view the records for a cycle in Inventory List view
### Actual Behavior 
Get segfault when trying to do this
### Steps to Reproduce
See this issue folder with doc for details
https://drive.google.com/open?id=1Mf04IgrL6eeLWbNAXjB7DcyH__erBaqc
Direct link to the doc
https://docs.google.com/document/d/1rIkmSxiwHPwEAWNx2BIqRDBXkjh1rTalu_3GA06aTOU/edit?usp=sharing

### Instance Information

Instance: LBNL production server
SHA: 2.4.0 release (getting 500 error trying to get the SHA @mmclark looking into this)

nllong:
@RDmitchell closing for release notes, test and reopen if needed.

PR #1746, opened 2018-12-04
Updated uWSGI for the aws config to match the test config

axelstudios:
#### What's this PR do?
Updates the uWSGI dependency
Retest the ticket when deployed on https://seedv2.lbl.gov with Python 3
#1734


nllong:
This file may eventually disappear if we move to docker only deployment.

RDmitchell:
@nllong -- is this what may help with the segfault I was seeing on staging with NYC?

PR #1747, opened 2018-12-05
GIS - 1st release

adrian-lara:
**Please Note:**
 - See updates to `setup_osx.rst` for installation notes.
 - Dependencies were added so please run
    - `./bin/install_javascript_dependencies.sh`
    - `pip install -r requirements/local.txt`

This PR contains the first release of GIS features.

#### What's this PR do?
Main components added:
 - PostGIS field attached to properties/tax lots
 - Map rendered with points from PostGIS fields
 - (more features to be mentioned)
 - Travis CI -related changes made to `tox.ini`, `.travis.yml`, etc. for this feature
 - Also, as requested, the removal of coverage results for FE tests occurs here
     - The FE tests do still run.

Import properties and tax lots via files and use the map page.

#### Definition of Done:
   Note that the removal of FE tests from the coverage report increased the total coverage percentage (as the `seed/static` folder was no longer being included).


seed/models/tax_lot_properties.py
```diff
@@ -326,6 +330,8 @@ def get_related(cls, object_list, show_columns, columns_from_database):
             obj_dict[lookups['obj_state_id']] = obj.state.id
             obj_dict[lookups['obj_view_id']] = obj.id
 
+            obj_dict[lookups['long_lat']] = long_lat_wkt(obj.state)
```
adrian-lara: This adds the field to the data that gets returned to the FE.

seed/static/seed/js/seed.js
```diff
@@ -1064,6 +1065,33 @@ SEED_app.config(['stateHelperProvider', '$urlRouterProvider', '$locationProvider
           }]
         }
       })
+      .state({
```
adrian-lara: This was essentially copied from the 'inventory_list' state. I left a lot of the `resolve` components, as I expected them to be used later for filtering.

seed/static/seed/partials/inventory_map.html
```diff
@@ -0,0 +1,135 @@
+<div class="page_header_container" ng-cloak>
```
adrian-lara: I completely copied this from the inventory_list.html partial. I commented out components that I believe will not be necessary (going forward for filtering, etc). I'll go in later and remove them completely once filtering is added or if we end up pushing this to the `develop` branch.

seed/utils/geocode.py
```diff
@@ -0,0 +1,118 @@
...
+# encoding: utf-8
+
+import requests
+
+from django.conf import settings
+
+from django.contrib.gis.geos import GEOSGeometry
+
+
+def long_lat_wkt(state):
```
adrian-lara: This is a helper method I found myself using in two different spots, so I placed it here. Is this the appropriate place?

seed/utils/geocode.py
```diff
@@ -0,0 +1,118 @@
...
+    if state.long_lat:
+        return GEOSGeometry(state.long_lat,srid=4326).wkt
+
+def geocode_addresses(buildings):
```
adrian-lara: I abstracted away a lot of the steps into helper methods thinking it would help with readability. I'm not sure if it helped, but feel free to provide feedback on readability (or anything, of course) either way.

seed/utils/geocode.py
```diff
@@ -0,0 +1,118 @@
...
+
+def _id_geocodings(id_addresses, address_geocodings):
+    return {
+        id: address_geocodings.get(address)
+        for id, address
+        in id_addresses.items()
+        if address_geocodings.get(address) is not None
+    }
+
+def _batch_addresses(addresses, n=100):
```
adrian-lara: This isn't necessarily an `addresses` specific batching method, but this was the only place I used it, so I'm not sure if this is best kept here or elsewhere.

seed/tests/test_utils_geocode.py
```diff
@@ -0,0 +1,207 @@
...
+    FakePropertyStateFactory,
+    FakeTaxLotStateFactory,
+)
+
+from seed.utils.geocode import geocode_addresses
+from seed.utils.geocode import long_lat_wkt
+from seed.utils.organizations import create_organization
+
+
+def batch_request_uri_length_matcher(r1, r2):
```
adrian-lara: In the case of multiple locations being batched into one API/HTTP request, it was possible for the locations to be listed in a different order in between test runs. Because of that, VCR would consider this different order as a different request, and throw errors.

Rather than changing the method to preserve order, this method was created to only compare the length of the URI when considering if the API/HTTP request is the same as a previous run's.

seed/utils/geocode.py
adrian-lara: After some thought, this is probably better off as a Serializer. Let me know if you agree or disagree. (I think `rest_framework` might have some features for GIS fields.)

docs/source/setup_osx.rst
```diff
@@ -117,6 +118,32 @@ default use password `seedpass` when prompted
     psql -c 'ALTER USER seeduser CREATEDB;'
     psql -c 'ALTER USER seeduser CREATEROLE;'
 
+
+PostGIS 2.5
```
adrian-lara: I have more confidence in the single Homebrew command as I installed PostGIS that way. I'm less confident in the MacPorts install as I haven't used MacPorts prior. 

If no one here reaches out with corrections/feedback from their setup, I'll likely have someone else run through this portion of the setup for feedback.

danielbressan: looks great to me, we can see if nrel team has any thoughts.

seed/static/seed/js/controllers/inventory_map_controller.js
```diff
@@ -0,0 +1,78 @@
...
+
+        var vectors = new ol.layer.Vector({
+          source: vectorSources(),
+          style: vector_style
+        });
+
+        // Consider making center be the centroid of shown points
+        // and make zoom dynamic (just large enough to see all points)
+        center_zoom = {
+          center: ol.proj.fromLonLat([-104.986292, 39.765566]),
```
danielbressan: @adrian-lara Is this a hard code of the center?

seed/utils/geocode.py
```diff
@@ -0,0 +1,116 @@
...
+import requests
+
+from django.conf import settings
+
+from django.contrib.gis.geos import GEOSGeometry
+
+
+def long_lat_wkt(state):
+    if state.long_lat:
+        return GEOSGeometry(state.long_lat,srid=4326).wkt
```
danielbressan: @adrian-lara what does srid=4326 mean?

seed/utils/geocode.py
danielbressan: I'm not sure what this is doing, what does GEOSGeometry().wkt do?

seed/utils/geocode.py
danielbressan: Seems fine.

seed/utils/geocode.py
```diff
@@ -0,0 +1,116 @@
...
+def _address_geocodings(id_addresses):
+    addresses = list(set(id_addresses.values()))
+
+    batched_addresses = _batch_addresses(addresses)
+    results = []
+
+    for batch in batched_addresses:
+        locations = "location=" + "&location=".join(batch)
+        response = requests.get(
+            'https://www.mapquestapi.com/geocoding/v1/batch?' +
```
danielbressan: @adrian-lara maybe make this an env or at least global variable...? Do we use it anywhere else? I'm just afraid mapquest will change their API and we'll have to change this code just for that. Thoughts? Happy to discuss...

seed/static/seed/js/controllers/inventory_map_controller.js
adrian-lara: For now, yes. PostGIS has the ability to perform different calculations, and I believe that finding a centroid is possible. I'm less sure about the zoom level, but I left the comment above this `center_zoom` variable (which I'm now realizing I forgot to locally scope, so I'll be changing that too).

For now, this is hard coded to the center of the US, but I'll update the location and zoom to be more dynamic.

danielbressan: @adrian-lara cool, maybe make the comment a `TODO: ` Just so we know we need to do something with that later :)

seed/utils/geocode.py
danielbressan: seems fine to me!

config/settings/local_untracked.py.dist
```diff
@@ -42,7 +42,7 @@ import os
 # postgres DB config
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.postgresql_psycopg2',
+        'ENGINE': 'django.contrib.gis.db.backends.postgis',
         'NAME': 'seed',
         'USER': 'your-username',
         'PASSWORD': 'your-password',
```
danielbressan: @adrian-lara we need the MAPQUEST api key setting right? probably worth adding that to the local untracked and instructions (I thought it was but now I don't see it).

seed/utils/geocode.py
adrian-lara: From my understanding, `GEOSGeometry()` initializes a geometry object with, in this case, a long_lat input held in the DB as a Well-Known Binary (WKB) that looks like this `0101000000000000000000F03F000000000000F03F`. The `.wkt` part returns the Well-Known Text (WKT) version that looks like 'Point(104.54 39.78)`.

All this info coming from here: https://docs.djangoproject.com/en/2.1/ref/contrib/gis/geos/#django.contrib.gis.geos.GEOSGeometry

seed/utils/geocode.py
adrian-lara: The Spatial Reference Identifier references which coordinate system to use. What I found from my research was that 4326 is pretty common around the world. It's used for GPS and is also the default for PostGIS and GeoJSON.

danielbressan: @adrian-lara ok that's fine then. Worth an inline comment for that then (maybe not even a bad idea to include a url link to the data you used to find it or something). Just so if someone if reading the code a year from now and they're like what's this hard coded number, maybe that's my problem! lol

seed/utils/geocode.py
adrian-lara: @danielbressan I left a comment in the code. Let me know if you think it should be rephrased or generally changed.

.travis.yml
```diff
@@ -40,12 +42,14 @@ before_script:
...
   - mv config/settings/test_local_untracked.py config/settings/local_untracked.py
   - psql -c "CREATE DATABASE seeddb;" -U postgres
+  - psql -d seeddb -c "CREATE EXTENSION postgis;" -U postgres
 env:
   global:
     - DOCKER_COMPOSE_VERSION=1.16.0
     - DJANGO_SETTINGS_MODULE=config.settings.travis
     - DISPLAY=:99.0
     - COVERALLS_REPO_TOKEN=y8UqJm8Bri5ZP8hr3YZM3guBaUKpsfoCv
+    - secure: HtOAAxPiA2NnSxXTBsmmXlFURH6WxnXXeaJLkjKc8cPWMsPX2QSouggtFYCPJxikrvN6bOYvDii67v2++E0qpuPO28LtywaLWu7AdDYxJ+rp4ysVjdi2Pp1bh723SZj2Xi3OMOLPz/kekT41DLKtoEzmBUEFEBIuJ27oU/phNGU=
```
adrian-lara: This is my encrypted API key. Since VCRpy is being used and cassettes were saved to GitHub (after filtering out the key in the URL), Travis builds won't actually trigger API requests.

Notably, when/if the cassettes are replaced in this repo, the encrypted key should get replaced with whatever key was used to build the cassettes.

seed/static/seed/js/controllers/inventory_map_controller.js
```diff
@@ -0,0 +1,78 @@
...
+
+      // Render Map
+      var renderMap = function () {
+        var raster = new ol.layer.Tile({
+          source: new ol.source.OSM()
+        });
+
+        var vector_style = new ol.style.Style({
+          image: new ol.style.Icon({
+            src: urls.static_url + "seed/images/favicon.ico",
```
adrian-lara: This is slated to change to be a pin vs the NREL SEED icon which was just being used as a placeholder.

Issue #1733, opened 2018-11-09
Newly created column will not show up in API until data exists


nllong:
### Expected Behavior
A newly created column (via the API) should show up in the API even if the data are null.

### Actual Behavior 
The column does not appear in the API. It will show up in the detailed page and inventory list if selected in the list profile setting (and will be empty as expected).

### Steps to Reproduce
Create new column using API 

Post the following to `https://localhost:8000/api/v2/columns/`:

```
{
  "shared_field_type": "None",
  "data_type": "string",
  "is_extra_data": true,
  "table_name": "PropertyState",
  "display_name": "Created from API Display Name",
  "merge_protection": "Favor New",
  "column_name": "Created from API"
}
```

Go to the property api for a single property and verify that the column does not appear `https://localhost:8000/api/v2/properties/722615/?organization_id=<org_id>`

Go to the detailed page, add the column to the view, then add a value to the column. Verify that it now appears in the API request.

### Instance Information

Server Instance: All serers

adrian-lara:
@nllong - Is this only an issue in the case that the new column is `extra_data`?

adrian-lara:
Also, in the case where the new column is an `extra_data` field, the most straightforward path I'm seeing towards forcing it's existence on a record is to update the `extra_data` value (more specifically, update the JSON/dict contained in `extra_data`) for that record. 

This would have to be repeated for every record (tax lot or property) within a user's organization. Though this doesn't seem the most efficient, I'm happy to go down that path. But before proceeding, I'm curious if you had any other ideas for potential solutions.

nllong:
@adrian-lara yes, you get the problem. It would have to be created for every record. The other option is to do it on the display side, which may be faster. That is, when the user requests a field, we go and grab all of the potential columns and create a null dict, then populate it with actual values if they exist.

nllong:
Tested fix branch and it now returned this:
![image](https://user-images.githubusercontent.com/1907354/50562988-65163680-0cd6-11e9-93f3-fa2b8a8398d5.png)

Looks good.

PR #1750, opened 2018-12-18
Return never-populated 'extra_data' columns for properties and tax lots

adrian-lara:
Issue #1733 

#### What's this PR do?
Returns previously unpopulated `extra_data` columns when a property/ tax lot show API endpoint is called. This was done by updating PropertyState- and TaxLotState- Serializers.

  - Run server
  - Create a new column via API:
     Post the following to https://localhost:8000/api/v2/columns/:
```
{
  "shared_field_type": "None",
  "data_type": "string",
  "is_extra_data": true,
  "table_name": "PropertyState",
  "display_name": "Created from API Display Name",
  "merge_protection": "Favor New",
  "column_name": "Created from API"
}
```
  - Get the property data via the API endpoint:  
    `127.0.0.1:8000/api/v2/properties/<property_id>/?organization_id=<organization_id>`
  - Verify that the column does appear 

Issue #1733 

#### Definition of Done:


seed/serializers/properties.py
```diff
@@ -171,8 +172,25 @@ class Meta:
         }
 
     def to_representation(self, data):
-        """Overwritten to handle time conversion"""
+        """Overwritten to handle time conversion and extra_data null fields"""
         result = super(PropertyStateSerializer, self).to_representation(data)
+
+        if data.extra_data:
+            organization = data.organization
+            extra_data_columns = Column.objects.filter(
```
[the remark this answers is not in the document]
adrian-lara: I made progress in exploring this option a bit more, but I'm actually not sure which endpoints use this serializer. 

As a quick check, I added a `print` statement to the PropertyStateSerializer's `to_representation` method and hit the "127.0.0.1:8000/api/v2/properties/?organization_id=`<id>`" endpoint. I believe the `to_representation` method wasn't run as evidenced by not seeing output in the terminal where my server was running. Also, the output for that endpoint did not show pre-populated `extra_data` fields for returned properties.

I'm still learning the ins and outs of Django, so it's very possible that I'm missing something. I plan on investigating more, but could you easily point me to an endpoint that you know uses the PropertyStateSerializer so that I can check that the argument flag approach works as expected?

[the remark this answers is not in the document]
nllong: I think adding an additional argument that passes in the `extra_data_columns` is probably the best. That way, the PropertyViewListSerializer can execute the query just once, instead of it being executed multiple times in the PropertyStateSerializer.

Yeah, I think you are seeing that there are more than one path through the Serializer world. The /api/v2/poperties goes through the PropertyViewSet (below), which does not explicitly call the PropertyStateSerializer (but should).

https://github.com/SEED-platform/seed/blob/f965e464f9c15d8e0e555c35f874304c9a8572ac/seed/views/properties.py#L203-L207

seed/serializers/properties.py
```diff
@@ -276,12 +293,20 @@ def to_representation(self, data):
             iterable = data
             view_ids = [view.id for view in iterable]
             results = []
+
+            # Grab extra_data columns to be shown in the result
+            organization = data[0].state.organization_id
```
adrian-lara: What is the expected value/instance type of `data` in this context?

nllong: data is the django db object instance. `result` would be a dict.

seed/serializers/properties.py
```diff
@@ -276,12 +293,22 @@ def to_representation(self, data):
             iterable = data
             view_ids = [view.id for view in iterable]
             results = []
+
+            # If data is provided, grab extra_data columns to be shown in the result
+            if iterable:
```
nllong: Ah, yes, this works too. Just grab the columns on the first iteration... well done.

PR #1781, opened 2019-01-07
GIS link to show page

adrian-lara:


seed/static/seed/js/controllers/inventory_map_controller.js
```diff
@@ -125,6 +125,27 @@ angular.module('BE.seed.controller.inventory_map', [])
       });
       $scope.map.addOverlay(popup_overlay);
 
+      var detailPageIcon = function(point_info) {
+        var link_html = ''
+        var icon_html = '<i class="ui-grid-icon-info-circled"></i>'
+
+        if ($scope.inventory_type == 'properties') {
+          link_html = '<a href="#/properties/' +
```
adrian-lara: We tried the following two for the value of `link_html`
```
link_html = '<a ui-sref=\"inventory_detail({inventory_type: \'properties\', view_id: ' +
  point_info.property_view_id +
  '})\">' +
  icon_html +
  '</a>'
```
```
link_html = `<a ui-sref="inventory_detail({inventory_type: \'properties\', view_id: ${point_info.property_view_id}})">` +
  icon_html +
  '</a>'
```

adrian-lara: Notably, after the page loads, I click a pin to see a popup, inspect the icon element, and copy the HTML into the HTML template (map page partial). That hardcoded icon produces an href, so it makes me think this dynamically generated HTML just isn't being processed by AngularJS or ui-router since it's generated after the page has finished loading. 

Let me know if you see anything clearly wrong.

PR #1787, opened 2019-01-09
Skip PM tests while PM is down

Myoldmopar:
Due to reasons, PM is currently unavailable.  The SEED-PM connection is tested with several unit tests.  Once PM went down, these tests starting failing, as they should.  But this issue with PM should not really cause SEED's test dashboard to be red, so we'll skip these for now.

#### What's this PR do?
Adds some logic to try to determine whether ESPM is actually available.  Even during this "down" time, the site is still running, so we can't just check the return code from the request.  

Just run the entire test suite of SEED, it should now all pass, with a few skipped.

I don't believe there are any yet, I was just informed the tests were failing, and why.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,9 @@ src/
 *node_modules*
 bower_components/
 
+# JetBrainz tends to suggest a venv directory in the project root for virtual environments
+venv/
+
 package-lock.json
 dump.rdb
 log/*.log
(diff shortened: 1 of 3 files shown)


Myoldmopar:
@nllong I updated the logic so it properly skips.  I also used the Django Response's built-in `.json()` method instead of parsing manually.  I expect it to pass now and be ready to go in.

Issue #1741, opened 2018-11-29
BuildingSync batch import from UI


RDmitchell:
### Expected Behavior
Add ability to import multiple BuildingSync files (batch import) from the SEED UI

Discussed in meeting with Harry Bergmann and @nllong 11/29/2018
### Actual Behavior 
n/a
### Steps to Reproduce
n/a
### Instance Information
n/a

nllong:
Here is a test zip file with 5 bsyncs.

https://drive.google.com/file/d/1V8ZVmgu3BT1fgSRAqwRyhHKeHoqovKC-/view?usp=sharing

nllong:
@RDmitchell I'm closing this issue to get it into the release notes. Feel free to still test and make sure it works as expected.

PR #1805, opened 2019-02-06
1741 buildingsync batch import

kflemin:
#### What's this PR do?
* Enable batch import of BuildingSync xml files.  XML files must be in compressed in .zip format.
* Allow both single xml and zip formats on the Datasets -> Add Data Files -> Upload BuildingSync Data button.
* Batch import works for BuildingSync xml files only.
* Modified data upload modal to show errors/warnings output across all files in the batch import.  This lists the files that had errors and must be resubmitted.  Valid files are imported even when there are invalid files within the batch.
* Modified data upload modal to always show warnings for single file or batch import even when import was successful. (the warnings were suppressed)
(Modified data upload modal results text specific to whether the upload was buildingsync, batch, and successful to provide more feedback to the user.
* Added a test for zip files and fixed the existing building_file api tests to pass with the new success response.
* And extended the width of the Action dropdown button on the inventory detail view so that "Update with BuildingSync" doesn't get split over 2 lines

Submit a zip file containing multiple buildingsync xmls via the Datasets-> Add Data Files -> Upload BuildingSync Data button.  You can also test with zip files containing a mix of xmls that are valid, invalid, and have warnings.

#1741


nllong:
Here is a test zip file with 5 bsyncs.

PR #1825, opened 2019-02-19
Fix Read the Docs and swagger

nllong:

* Read the docs was not building since we migrated to python 3.x
* Swagger broke during the pipeline upgrade

#### What's this PR do?

* fixes the read the docs. Note that past versions of SEED will not be rebuilt in RTD.
* find the static assets using the AppDirectoriesFinder instead of specifying in the STATICFILES_DIRS.

Best to test by building in docker from scratch and ensuring that all the static assets are found when running SEED (and swagger).

N/A

#### Definition of Done:
- [x] Does this add new dependencies? If so, do pip or npm requirements need to be updated?

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,12 +15,17 @@
 import os
 import sys
 import json
+from shutil import copyfile
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 
-# sets the default settings module for read the docs
+# Copy over the local_untracked if it does not exist yet.
+if not os.path.exists('../../config/settings/local_untracked.py'):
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -15,12 +15,17 @@
...
 import json
+from shutil import copyfile
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 
-# sets the default settings module for read the docs
+# Copy over the local_untracked if it does not exist yet.
+if not os.path.exists('../../config/settings/local_untracked.py'):
```
nllong: I also made a change on the RTD website to build with python3. RTD was also crashing because it couldn't file the local_untracked.py. I am not sure how this was working before.

PR #1826, opened 2019-02-22
AWS/S3 Cleanup

axelstudios:
#### What's this PR do?
- Removes all non-SES AWS and S3 references (like unused S3 upload)
- Updates translations
- Code review, standard tests, monkey tests to be sure that there are no regressions
- Check that nothing of value was removed


nllong:
Looks like you disabled ZIP files on upload which are needed. You can upload multiple BuildingSync files in a zip file.

image

axelstudios:
@nllong Hmm, I'm pretty sure the zip upload still works the same.  Zips are only accepted via the BuildingSync upload button - each upload button has specific allowed file types:

PR #1827, opened 2019-02-22
Fix PM

nllong:

 File "./seed/data_importer/views.py", line 442, in create_from_pm_import
    pm_csv_writer.writerow(row)
TypeError: a bytes-like object is required, not 'str'

#### What's this PR do?

Update csv write to python3 syntax.

Import PM data using username and password.

N/A

#### Definition of Done:
- [x] Does this add new dependencies? If so, do pip or npm requirements need to be updated?


nllong:
Do not update the base branch until we test on staging and production.

PR #1846, opened 2019-03-25
Footprint import

adrian-lara:
`..._footprint` fields are exposed to user but users were not able to upload data to insert data into these fields.

#### What's this PR do?
The intent of this PR is to implement the following flow for footprint imports:
1. Starting on the initial mapping page, where the user types in their fields. “Tax Lot Footprint” or “Property Footprint” should be a suggested, green options (that autocompletes like “Address Line 1”, etc).

2. To consider the case of a user using the “wrong”, non-polygon-formatted field or if that field contains non-polygon-formatted values for certain rows, give them a warning on the next page. Specifically, Data Quality Results will be used to include “Error Message”s for the rows with non-polygon-formatted values.

3. If the user proceeds with these erroneous rows along with any rows having correct polygon-formatted values, we’ll process and save any footprints that are correctly formatted. For the erroneous fields, The `…_footprint` field will be empty, and we'll keep a record of the incorrectly-formatted contents as extra_data with the file field name + ' (Invalid Footprint)'.

Rerun migrations or add footprint fields to an org's columns (as shown in DATABASE_COLUMNS).
Add properties and tax lots with valid and invalid footprints. Check for DQ results and that footprints were, in fact, saved if valid.


seed/models/columns.py
```diff
@@ -139,6 +137,7 @@ class Column(models.Model):
         'DateTimeField': 'datetime',
         'BooleanField': 'boolean',
         'JSONField': 'string',
+        'PolygonField': 'geometry',
```
adrian-lara: Note that I introduced a new `data_type`. I believe this didn't negatively impact areas of the codebase outside of the work that I was doing, but I think this is important to note

axelstudios:
Looks good.  My wishlist includes tweaking the popover view (ideally just showing the actual drawn polygon) to hide gigantic amounts of coordinates:

image

RDmitchell:
Yes it would be very cool to see the drawn footprint (!)

nllong:
Maybe not tomorrow but it will output to the GeoJSON file soon.

PR #1853, opened 2019-04-01
Meters

adrian-lara:
Implementation of Meter uploads using a time-series enhanced database (Postgres + TimescaleDB extension).

Notable details/caveats:
- Documentation for how Thermal Conversions are brought into app has been shared with NREL via 'NREL-Devetry' shared directory.
- As a reminder, it's been decided that this Meters implementation needs to be supplemented with a follow up script and code update to allow PropertyState models to be associated to each other by Property models.
- Successfully imported meter files, importfiles, show up in importrecord detail without associated cycles as their data is not bounded by cycles.
- **An issue exists** when/if duplicate times are provided in GreenButton import across different batches. This comes from batch upserting meters in groups of 1000, where duplicate meter entries could exist in different batches. One of the duplicate readings will be chosen according to which batch finishes last (randomly). If duplicates exist with the same batch, the app correctly identifies this as an error and does not upload these entries since it isn't known which entry should take precedence. (This is an edge/corner case that I decided to punt on for now due to time constraints.)
- GreenButton types are assumed to be either "Electricity" or "Natural Gas". Corresponding units are parsed to the best of our ability. Only units known in Portfolio Manager imports (and their metric conversions, m, d, k, M, G etc.) are accepted from GreenButton as these are the only ones for which we have thermal conversions.

Assumptions:
- A GreenButton meter import contains data for one meter of a single property.
- There cannot be 2 different meters of the same type for a single property.

#### What's this PR do?
Key features include:
- the ability to import Portfolio Manager monthly meter data
- the ability to import GreenButton (sub-monthly) meter data
- the ability to view exact meter data (w/ start and end times) as well as monthly and annually aggregated meter data

Import PM monthly meters for multiple properties and GB meters for single properties. View the uploaded meters using the new "Energy" tab in the property detail pages.


seed/data_importer/views.py
```diff
@@ -539,25 +539,6 @@ def retrieve(self, request, pk=None):
         if not import_file.uploaded_filename:
             f['uploaded_filename'] = import_file.filename
         f['dataset'] = obj_to_dict(import_file.import_record)
-        # add the importfiles for the matching select
```
adrian-lara: Note that these are `importfiles` of a dataset (ImportRecord) of an ImportFile. For example, given an ImportFile `test.xlsx`, these are the ImportFiles belonging to the same ImportRecord that the ImportFile `test.xlsx` belongs to.

I haven’t been able to trace any FE or BE code that uses those specific `importfiles`. Upon removing them from the request, the bug is fixed, all tests still pass, and monkey testing importing/matching files all seem to be OK.

Note that this `importfiles` logic seems to have originated in the initial commit of the app https://github.com/SEED-platform/seed/blame/cd6edee5778d7d117e5469e4cc8a875bf7093259/seed/views/main.py#L1566

I suspect it might be remnants of an old architecture.

seed/migrations/0102_auto_20190415_1115.py
```diff
@@ -0,0 +1,119 @@
+# -*- coding: utf-8 -*-
```
adrian-lara: @nllong Thanks for combining the two migrations! I ran into errors trying to let Django create this combined migration (letting Django read the models), so I assumed the two migrations I ended up creating needed to be kept separate. Travis test passed so it turns out the migrations can be combined. Thanks for catching this! 😄

nllong:
yeah, you will need to rerun the migrations on your database. Easiest to
just recreate a clean db.

nllong:
Can you move the Energy tab to be just to the right of the "Property Detail"?
image

nllong:
Fix the spacing between the last two items:

image

seed/api/v2/urls.py
```diff
@@ -79,12 +79,12 @@
     # data uploader related things
     url(r'get_upload_details/$', get_upload_details, name='get_upload_details'),
     url(r'^schema/$', get_api_schema, name='schema'),
-    url(r'meters/(?P<pk>\w+)/timeseries/$',
-        MeterViewSet.as_view({'get': 'timeseries'}),
-        name='meters-get-timeseries'),
-    url(r'meters/(?P<pk>\w+)/timeseries/$',
-        MeterViewSet.as_view({'post': 'add_timeseries'}),
-        name='meters-add-timeseries'),
+    # url(r'meters/(?P<pk>\w+)/timeseries/$',
```
nllong: should we just remove these? Seems like these are old.

seed/data_importer/tests/test_greenbutton_import.py
```diff
@@ -0,0 +1,259 @@
...
+        self.assertEqual(meter_1.source, Meter.GREENBUTTON)
+        self.assertEqual(meter_1.source_id, 'User/6150855/UsagePoint/409483/MeterReading/1/IntervalBlock/1')
+        self.assertEqual(meter_1.is_virtual, False)
+        self.assertEqual(meter_1.meter_readings.all().count(), 2)
+
+        meter_reading_10, meter_reading_11 = list(meter_1.meter_readings.order_by('start_time').all())
+
+        self.assertEqual(meter_reading_10.start_time, make_aware(datetime(2011, 3, 5, 21, 0, 0), timezone=self.tz_obj))
+        self.assertEqual(meter_reading_10.end_time, make_aware(datetime(2011, 3, 5, 21, 15, 0), timezone=self.tz_obj))
+        self.assertEqual(meter_reading_10.reading, 1790 * 3.412 / 1000 * ureg('kBtu'))
```
nllong: nice

seed/data_importer/tests/test_greenbutton_import.py
```diff
@@ -0,0 +1,259 @@
...
+        refreshed_meter = refreshed_property_1.meters.get(type=Meter.ELECTRICITY)
+
+        meter_reading_10, meter_reading_11, meter_reading_12 = list(refreshed_meter.meter_readings.order_by('start_time').all())
+        self.assertEqual(meter_reading_10.reading, 1790 * 3.412 / 1000 * ureg('kBtu'))
+        self.assertEqual(meter_reading_11.reading, 1791 * 3.412 / 1000 * ureg('kBtu'))
+
+        # Sanity check to be sure, nothing was changed with existing meter reading
+        self.assertEqual(meter_reading_12, existing_meter_reading)
+
+    def test_existing_meter_reading_has_reading_source_unit_and_conversion_factor_updated_if_import_file_references_previous_entry(self):
```
nllong: this_is_the_longest_function_name_ive_ever_seen

RDmitchell:
And how about putting all the ESPM choices together? Which would mean the ESPM meter option would go right below the Upload Portfolio Manager Data button

seed/lib/superperms/orgs/models.py
```diff
@@ -119,12 +136,17 @@ class Meta:
     created = models.DateTimeField(auto_now_add=True, null=True)
     modified = models.DateTimeField(auto_now=True, null=True)
 
+    # Default preferred all meter units to kBtu
+    display_meter_units = JSONField(default=_default_display_meter_units)
```
nllong: 👍

seed/migrations/0102_auto_20190415_1115.py
```diff
@@ -0,0 +1,119 @@
...
+        migrations.AlterIndexTogether(
+            name='timeseries',
+            index_together=set([]),
+        ),
+        migrations.RemoveField(
+            model_name='timeseries',
+            name='meter',
+        ),
+        migrations.RemoveField(
+            model_name='meter',
```
nllong: thanks for cleaning these up.

nllong:
This interface is a bit strange... but i think we leave it for now. It is quite cool that users can specify the way to want to view meter data by units.

image

nllong:
> Fix the spacing between the last two items:
> 
> image

Per @RDmitchell  -- And how about putting all the ESPM choices together? Which would mean the ESPM meter option would go right below the Upload Portfolio Manager Data button

adrian-lara:
> hi @adrian-lara can you verify the timestamp imports for GreenButton? It looks like the data are in epoch time, but when importing and converting, it looks like that we are at least 1 hour off.
> 
> Raw file (example-GreenButton-data.xml)
> 
> ```
> <IntervalReading>
>           <timePeriod>
>             <duration>900</duration>
>             <start>1299387600</start>
>           </timePeriod>
>           <value>1790</value>
>         </IntervalReading>
>         <IntervalReading>
>           <timePeriod>
>             <duration>900</duration>
>             <start>1299388500</start>
>           </timePeriod>
>           <value>1791</value>
>         </IntervalReading>
> ```
> 
> Epoch Calc
> image
> 
> Seed
> image

@nllong, I specified GB epoch times to be read and saved using the default timezone of the app `America/Los_Angeles` specified here: https://github.com/SEED-platform/seed/blob/41a307c78d31929d421e728fa814b4385f2e4d4a/config/settings/common.py#L34

I used the same epoch converter site but changed the timezone (https://www.epochconverter.com/timezones?q=1299387600&tz=America%2FLos_Angeles) and got the time shown in your SEED screenshot.

nllong:
Oh. Good. I'll double check that makes sense!

RDmitchell:
Do we have a good source for greenbutton data? When we were implementing this before with CMU I generated greenbutton files from my house, which I can do again if that would be useful.

adrian-lara:
The samples I used came from here: https://www.energy.gov/downloads/green-button-sample-data-pge 

> Do we have a good source for greenbutton data? When we were implementing this before with CMU I generated greenbutton files from my house, which I can do again if that would be useful.

nllong:
> Fix the spacing between the last two items:
> 
> image

Looks Good. Thanks for the fixes.

nllong:
I think this is good to go. Any last comments before I merge?

nllong:
> Can you move the Energy tab to be just to the right of the "Property Detail"?
> image
bingo. 

image

PR #1854, opened 2019-04-08
ULID, UBID, and Footprints

nllong:
SEED can now handle more advanced geometry-based operations now that the database is PostGIS-based.

#### What's this PR do?
* Add ULID as a matching field
* Parse building footprints in WKT format 
* Add centroid, lat, and long to taxlot
* Add popup of building footprint

* import san jose data
* map Property Coordinates to Property Footprint
* verify that the property footprint popup works
* verify that the ulid, ubid are parsed correctly by verifying the existence of lat/long in the data
* test exporting the GeoJSON and importing into geojson.io.

#1855 #1856


RDmitchell:
@nllong / @axelstudios -- is this where the program displays the footprint in the WKT field? I can import a file with that WKT info, but don't see the footprint shape.

Is the the correct format for a WKT field that SEED interprets? It is what was in the file I imported
image

PR #1876, opened 2019-05-07
Fix to include extra_data fields when loading the pairing page

### LINKED ISSUE #1875: Extra data not showing up in the pairing page
If the user wants to view an extra data field in the pairing page, then the data aren't actually showing up.

axelstudios:
#### What's this PR do?
extra_data field data wasn't included on page load of the pairing page.  This PR changes it to include all fields

fixes #1875


nllong:
Thanks for getting this in there so quick!

PR #1877, opened 2019-05-09
Labels to views

adrian-lara:
**WARNING** This PR involves making irreversible changes to the DB. See comments below.

Labels are currently attached to canonical Properties and TaxLots (the base Property and TaxLot models). Since canonical Properties and TaxLots would potentially be shifted around during the upcoming year-over-year changes, it was decided to move Label associations to PropertyViews and TaxLotViews respectively. 

#1893 

#### What's this PR do?
This PR moves Label associations from canonical Properties and TaxLots to PropertyViews and TaxLotViews respectively.  

Specifically, there's code in the migration that goes through all the existing -Views, and for each, the code captures all the corresponding Property's (or TaxLot's) Label associations and applies these associations to the -View itself. Then, the migration runs to delete the join tables `seed_property_labels` and `seed_taxlot_labels` as well as all the records within them.
**WARNING: This is irreversible, and a backup of your database should be made.**

Upon changing Label relationships from Property/TaxLot to the corresponding -Views, the following existing **features** were transitioned to use this new relationship.
- Property & TaxLot Inventory List page View, Create, Add, & Remove Labels
- Property & TaxLot Inventory Detail page View, Create, Add, & Remove Labels
- Label can be associated to PropertyView or TaxLotView due to DQ Rule during post-import DQ check
- Property & TaxLot merges/unmerges combine and distribute Labels
- Property & TaxLot Labels are exported in CSV & GeoJSON exports
- Organization checks between Labels and target -View

Additionally, the following were checked to be sure nothing needed to be changed:
- Property & TaxLot Inventory List and Map pages filtering by Label
- Property/TaxLot Pairing
- Matching (unused "match" and "unmatch" methods in data_importer/tasks.py)
- Labels are not associated on import (according to DQ Rules). This is by design and is only mentioned here to say that this remained the same.

- Using an existing database with some data, add and/or make note of Properties/TaxLots and their associated Labels.
- Run migrations.
    - Again... **WARNING:** This is irreversible, and a backup of your database should be made.
- Be sure that from the user's perspective, Labels are still associated to their Properties/TaxLots.
- Monkey test any features involving Labels. See the list of features in the above section.


RDmitchell:
@nllong -- I am a bit puzzled by the statement in the PR that says:

* Labels are not associated on import (according to DQ Rules)

But labels ARE associated on import when apply the DQ rules, if those rules has labels associated with them. 

Presumably this still works and I am just mis-interpreting the description in the PR?

docs/source/data_quality.rst
```diff
@@ -1,5 +1,9 @@
 Data Quality
 ============
 
-Data quality checks are run after the data are paired or on-demand by selecting rows in the inventory
-page and clicking the action button.
+Data quality checks are run after the data are paired, during import of Properties/TaxLots, or on-demand by selecting rows in the inventory
```
adrian-lara: @RDmitchell in reference to your question. https://github.com/SEED-platform/seed/pull/1877#issuecomment-491019562
This elaborates what I found to be true in the app. Let me know if I misrepresented anything or got anything wrong, and I'd be happy to update this document and PR as needed.

RDmitchell:
@adrian-lara -- the code comment does seem to be correct. I guess I was just mis-interpreting what you had written in the PR -- so basically nothing in that part of the functionality of the program changes. Thx

adrian-lara:
> @adrian-lara -- the code comment does seem to be correct. I guess I was just mis-interpreting what you had written in the PR -- so basically nothing in that part of the functionality of the program changes. Thx

@RDmitchell You're right, the wording in my PR comment is unclear/confusing. Updated it 😄

axelstudios:
Running the migration against the production database (320,401 applied property labels, 300,377 applied taxlot labels) took about 2.5 hours to run.

The commit I just added reduces the migration time to less than 2.5 minutes (but it will need ~2GB of memory).

PR #1886, opened 2019-05-22
Pairing hotfix label hooks

adrian-lara:
The goal of this PR is to include checks that ensure a Label organization and Property/TaxLot organization matches before they are associated together.

1. A pre_save check has been added for when the code starts from the Property (or TaxLot) model and associates labels that way `property.labels.add(label)`.
2. Custom logic has been added where associations were made with a `bulk_create` which bypasses any pre_save type signals. This happens via the Label API endpoint.
3. Custom logic has been added where an association is created using the through model (Property_label or TaxLot_label). This happens during DQ checks and Labeled Rules are broken.

Follow up Issue: #1887

### DIFF
--- a/seed/models/data_quality.py
+++ b/seed/models/data_quality.py
@@ -927,27 +927,29 @@ class DataQualityCheck(models.Model):
             label_org_id = rule.status_label.super_organization_id
 
             if rule.table_name == 'PropertyState':
-                property_org_id = Property.objects.get(pk=linked_id).organization_id
-                if property_org_id == label_org_id:
+                property_parent_org_id = Property.objects.get(pk=linked_id).organization.get_parent().id
+                if property_parent_org_id == label_org_id:
                     label_class.objects.get_or_create(property_id=linked_id,
                                                       statuslabel_id=rule.status_label_id)
                 else:
                     raise IntegrityError(
-                        'Label with org_id={} cannot be applied to a record with org_id={}.'.format(
+                        'Label with super_organization_id={} cannot be applied to a property with parent '
+                        'organization_id={}.'.format(
                             label_org_id,
-                            property_org_id
+                            property_parent_org_id
                         )
                     )
             else:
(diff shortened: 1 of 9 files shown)


axelstudios:
Looks good - I just added a commit to make sure this works for sub-organizations as well (prior to this commit you would be a 500 integrity error if you tried to add a label to a property in a sub org).

Since labels have a `super_organization_id` and properties/taxlots just have an `organization_id` this calls `org.get_parent()` to make sure the id equality is consistent

adrian-lara:
> Looks good - I just added a commit to make sure this works for sub-organizations as well (prior to this commit you would be a 500 integrity error if you tried to add a label to a property in a sub org).
> 
> Since labels have a `super_organization_id` and properties/taxlots just have an `organization_id` this calls `org.get_parent()` to make sure the id equality is consistent

Ah, good catch! Makes sense and changes look good to me.

seed/models/data_quality.py
```diff
@@ -922,12 +924,34 @@ def update_status_label(self, label_class, rule, linked_id):
         """
 
         if rule.status_label_id is not None and linked_id is not None:
+            label_org_id = rule.status_label.super_organization_id
+
             if rule.table_name == 'PropertyState':
-                label_class.objects.get_or_create(property_id=linked_id,
-                                                  statuslabel_id=rule.status_label_id)
+                property_parent_org_id = Property.objects.get(pk=linked_id).organization.get_parent().id
+                if property_parent_org_id == label_org_id:
```
nllong: 👍

seed/utils/generic.py
```diff
@@ -105,3 +106,16 @@ def json_serializer(obj):
     if isinstance(obj, datetime):
         serial = obj.isoformat()
         return serial
+
+
+def compare_orgs_between_label_and_target(sender, pk_set, instance, model, action, **kwargs):
```
nllong: nice

PR #1891, opened 2019-06-02
Update the backup script to delete files older than 30 days

mmclark:

The backup script previously deleted database and media file backups older than 45 days.  This is eating up more disk space on the servers than we'd like.

#### What's this PR do?

This alters the backup script to delete files older than 30 days.

Run the script and verify that files older than 30 days in ~/seed-backups are indeed removed.

N/A


nllong:
Why don’t we make this an argument with a default of 30?

mmclark:
Good idea.  I'll update the PR

PR #1892, opened 2019-06-04
Meters merge update

adrian-lara:
Meters and subsequent MeterReadings could potentially be lost when merging and unmerging Properties manually. 

#1894 

#### What's this PR do?
The following has been implemented:
- Users are warned when any properties being merged have associated Meters.
- Merge Meters and Readings if any or all properties being merged have them. In the case of merging overlapping Readings (by start_time and end_time), priority is given to Readings belonging to the 'priority' Property.
- During unmerging, Meters are copied to both of the new (canonical) Properties.

Via Inventory List and Detail Property pages:
- Add overlapping and non-overlapping meters to properties.
- Check that ONLY when merging properties with meters, a warning pops up. This should not happen if none of the selected have Meters or taxlots are being merged.
- Merge while checking that meter readings persist and/or are combined.
- Unmerge while checking that both of the new properties now have identical readings.
- When merging overlapping readings, be sure that the priority order between properties is inherited amongst the meter readings.  Properties selected most recently get more priority, so their readings should be used when overlaps occur.


RDmitchell:
@adrian-lara -- will this fix #1895 ?

adrian-lara:
> @adrian-lara -- will this fix #1895 ?

No - this addresses the potential to have Meters and Readings lost/deleted on merge/unmerge.

seed/models/meters.py
```diff
@@ -79,6 +81,46 @@ class Meter(models.Model):
 
     type = models.IntegerField(choices=ENERGY_TYPES, default=None, null=True)
 
+    def copy_readings(self, source_meter, overlaps_possible=True):
+        """
+        Copies MeterReadings of another Meter. By default, overlapping readings
+        are considered possible so a SQL bulk upsert is used. But if overlapping
```
nllong: 👍 on SQL bulk

seed/static/seed/js/services/inventory_service.js
```diff
@@ -118,6 +118,14 @@ angular.module('BE.seed.service.inventory', []).factory('inventory_service', [
      *
      */
 
+     inventory_service.properties_meter_check = function (inventory_ids) {
+       return $http.post('/api/v2/properties/meter_check/', {
```
nllong: seems like this method would be better named meters_exist...

adrian-lara:
> The PR does what it states it should do--look great!
> 
> Few other, small items:
> 
> It appears that the `example-GreenButton-data-1002-1-dup` file has values that are really large (for three intervals). I think it is behaving as I would expect, but I am wondering the reason for this file/test?
> 
> ![image]
> 
> When loading the meter page with more than ~20 rows, the other rows aren't immediately populated until the user scrolls. Is this just an artifact on my machine?
> 
> [image]

Regarding the `example-GreenButton-data-1002-1-dup` file. During import, SEED should not have been able to upload all the records, because within this file, there are overlapping readings that end up in the same Celery batch (of 1000 readings). Since, it isn't clear which reading should take precedence, none in that batch are uploaded.

This is tested here: https://github.com/SEED-platform/seed/blob/develop/seed/data_importer/tests/test_greenbutton_import.py#L229

Regarding the meter page with +20 rows, I think that may have been added by @axelstudios to speed things up from the UX perspective.

seed/static/seed/js/services/inventory_service.js
adrian-lara: Agreed. I'll jump in the PR and change that.

axelstudios:
Yeah the visible rows `minRowsToShow` is an expected behavior from the ui-grid (defaults to 10), but it can be increased without any performance side effects here because there are so few columns

Issue #1879, opened 2019-05-10
profile_id as query param and part of v2.1 API


nllong:
profile_id is presently required to be a data param in the post query for version 2.0 of the API. It is proposed to move this to be both a query param and a data parameter. 

Also, the functionality needs to exist in the 2.1 version of the API as well.

PR #1900, opened 2019-06-18
1879 profile id v21

nllong:
Note that this is based off of master (called 2.5.2-patch0).

#### Any background context you want to provide
The listing of properties between version 2 and 2.1 of the API differed in the sense that there was not a way to remove fields in the 2.1 results using the profile_id setup.

#### What's this PR do?
* Adds the profile_id to the properties list for version 2.1 API
* Adds new helper method for extracting the show_columns from a profile_id

* Import data
* Create a profile on the properties list
* Go to the URL for the v2.1 API call and include profile_id in query param  
  `{{SEEDURL}}/api/v2.1/properties/?organization_id={{organization_id}}&profile_id={{profile_id}}
* Make sure that the state data are only the fields in the profile_id.

#1879 

n/a

### DIFF
--- a/seed/api/v2_1/views.py
+++ b/seed/api/v2_1/views.py
@@ -28,12 +28,13 @@ from seed.models import (
 from seed.serializers.properties import (
     PropertyViewAsStateSerializer,
 )
+from seed.utils.api import OrgMixin
 from seed.utils.viewsets import (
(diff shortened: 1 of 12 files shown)


nllong:
deploy to staging and alert DOEE when finished.

Issue #1901, opened 2019-06-18
403 Forbidden error -- trying to remove column mappings


RDmitchell:
### Expected Behavior
log in as admin/superuser, click the "Remove Column Mappings" button, program deletes column mappings from the organization
### Actual Behavior 
Got 403 Forbidden error
### Steps to Reproduce
See this doc
https://docs.google.com/document/d/1nbgmo91-I_ijgZA1xskpVMuG39I2KR50OsaHbnYidN8/edit?usp=sharing
### Instance Information
instance: LBNL production
SHA: 56e5ea6b

adrian-lara:
I was not able to reproduce this on a local deployment of production.

Every time I tried to remove column mappings for an org as a superuser, I was able to do so successfully.
![Screen Shot 2019-10-24 at 9 31 48 AM](https://user-images.githubusercontent.com/30608004/67503081-3fcb2e00-f644-11e9-97c5-e15a7addbe2c.png)

RDmitchell:
I am reopening this and moving it to Test so that I can verify that I can't reproduce it.

adrian-lara:
@RDmitchell Sorry, I wasn't 100% sure on the process. In cases like this, should I let you know I couldn't reproduce then let you verify and close on your side?

RDmitchell:
@adrian-lara / @nllong -- This doesn't seem to be a problem on dev1, or on LBNL production, so I will close this issue.

RDmitchell:
I think the way it was handled was fine -- I will test anything that is in the Test column that I think I can test, even if it is just to see if I can reproduce something that you couldn't. So in this case it was in the Test column, I saw you couldn't reproduce it, so I tested it on both dev1 and LBL production, verified that I also couldn't reproduce it anymore, so then I just closed it. That seems like a good process to me !

PR #1902, opened 2019-06-19
Allow superuser permissions

nllong:
The newer viewsets do not allow for superusers to call commands on the columns and columnmappings.

#### What's this PR do?
Add the superuser logic to the SEEDOrgPermissions.has_perm.

Verify that users don't have excess permissions.
Verify that a user can delete column mappings as an admin to an organization that they don't belong to.

#1901


nllong:
Merge into develop when finished.

Issue #1903, opened 2019-06-19
500 Internal Server Error trying to rename columns in Column Settings if the column already exists 


RDmitchell:
### Expected Behavior
Be able to rename fields in the Column Settings screen
(Maybe it is not possible to do what I am trying to do, but if not, the program should have a message for the user, rather than an internal server error).
### Actual Behavior 
In Column Settings, there are several fields with slightly different names, but all represent the same data. The attempt is to try to rename one field to the name of a field that already exists. 
![image](https://user-images.githubusercontent.com/6314950/59881146-a32a1400-9363-11e9-9808-6688aa803d1f.png)
Clicking the rename button, I set the new name to be the same as the 2nd field, which has most of the words in lower case, as opposed to upper case. 
![image](https://user-images.githubusercontent.com/6314950/59881291-0caa2280-9364-11e9-8146-42ee5e8ae233.png)
Clicking the Confirm button, I get an Internal server error
![image](https://user-images.githubusercontent.com/6314950/59881461-79bdb800-9364-11e9-98b8-f597d5af844a.png)
### Steps to Reproduce
Import the same file twice, once as a Portfolio Manager file, accepting the default field names, which makes extra data fields have capitalized letters for the field names
![image](https://user-images.githubusercontent.com/6314950/59881686-0b2d2a00-9365-11e9-89e7-0964c5ff2c8c.png)
Import the file again, as a Portfolio Manager file, this time clicking the "Map SEED fields to imported file fields" 
![image](https://user-images.githubusercontent.com/6314950/59881714-197b4600-9365-11e9-9842-9eefabe003f0.png)
 (and then changing at least the first field, Property ID, to PM Property ID, so the program will accept the mapping. Now the SEED field has the same capitalization as the ESPM field.
![image](https://user-images.githubusercontent.com/6314950/59881830-66f7b300-9365-11e9-9a14-03d1fe62051a.png)
And now in the Column Settings screen, there are two fields with slightly different names (due to Capitalization differences) which represent the same data
![image](https://user-images.githubusercontent.com/6314950/59881146-a32a1400-9363-11e9-9808-6688aa803d1f.png)

### Instance Information
instance: seeddemostaging
SHA: 98a1fb00

RDmitchell:
There is a sample file in the gdrive folder for this issue which can be used for testing
Here is the link to the folder
https://drive.google.com/open?id=14KS1Zsg_fdNVNA5m2a8KFfFdGTZl3rmv
The filename is 
[...]

PR #1905, opened 2019-06-21
Fixes column renaming issues

axelstudios:
#### What's this PR do?
Fixes two issues:
- New columns were not being created if even a single organization had a column with that name.  If the current org didn't have the column but another org did it could lose the data by adding it to extra_data without an actual column to access the data.  If no orgs had that column it would be created for the wrong org.
- Fixes a KeyError when trying to access extra_data keys that may not exist
- Have multiple orgs, and using the most-recently created org test renaming columns and verify that the data is correct afterwards.
- Rename an existing extra_data column to a different existing extra_data column, and then do the exact same rename a second time.  This should result in all `null` values for the new column
#1903 
#### Relevant screenshots?
Extra_data diff:
image

### DIFF
--- a/seed/models/columns.py
+++ b/seed/models/columns.py
@@ -643,7 +643,8 @@ class Column(models.Model):
         try:
             with transaction.atomic():
                 # check if the new_column already exists
-                new_column = Column.objects.filter(table_name=self.table_name, column_name=new_column_name)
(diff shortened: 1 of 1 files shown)


nllong:
This appears to be the same as #1906.

PR #1906, opened 2019-06-21
Column renaming hotfix

axelstudios:
Production hotfix:
See #1905

### DIFF
--- a/seed/models/columns.py
+++ b/seed/models/columns.py
@@ -643,7 +643,8 @@ class Column(models.Model):
         try:
             with transaction.atomic():
                 # check if the new_column already exists
-                new_column = Column.objects.filter(table_name=self.table_name, column_name=new_column_name)
+                new_column = Column.objects.filter(table_name=self.table_name, column_name=new_column_name,
+                                                   organization=self.organization)
                 if len(new_column) > 0:
                     if not force:
                         return [False, 'New column already exists, specify overwrite data if desired']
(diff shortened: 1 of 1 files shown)


seed/models/columns.py
```diff
@@ -643,7 +643,8 @@ def _serialize_for_extra_data(column_value):
         try:
             with transaction.atomic():
                 # check if the new_column already exists
-                new_column = Column.objects.filter(table_name=self.table_name, column_name=new_column_name)
+                new_column = Column.objects.filter(table_name=self.table_name, column_name=new_column_name,
+                                                   organization=self.organization)
```
nllong: nice, so making sure that the new_column is part of the right organization!

seed/models/columns.py
```diff
@@ -680,14 +681,14 @@ def _serialize_for_extra_data(column_value):
                 # go through the data and move it to the new field. I'm not sure yet on how long this is
                 # going to take to run, so we may have to move this to a background task
                 orig_data = STR_TO_CLASS[self.table_name].objects.filter(
-                    organization=new_column.organization,
+                    organization=self.organization,
```
nllong: this doesn't need to be changed, but go ahead and leave it.

seed/models/columns.py
```diff
@@ -680,14 +681,14 @@ def _serialize_for_extra_data(column_value):
...
                     data_state=DATA_STATE_MATCHING
                 )
                 if new_column.is_extra_data:
                     if self.is_extra_data:
                         for datum in orig_data:
-                            datum.extra_data[new_column.column_name] = datum.extra_data[self.column_name]
-                            del datum.extra_data[self.column_name]
+                            datum.extra_data[new_column.column_name] = datum.extra_data.get(self.column_name, None)
```
nllong: much better... emtpy data would be a problem.

seed/models/columns.py
axelstudios: Yeah, but I figured for consistency I'd go ahead and just reference `self` instead

PR #1910, opened 2019-07-08
minor issues with PropertyStateWritableSerializer and GreenAssessment…

fablet:
…PropertySerializer

Since the "state" field in the PropertyViewAsStateSerializer is a nested writable serializer field which allows the creation of a PropertyState record simultaneously with the creation of a PropertyView, the type validation based on PropertyView model fields was causing the nested creation to break with "int() argument must be a string, a bytes-like object or a number, not 'dict'".
Likewise, since the serializer automatically renders data sent to "state" as PropertyStateWritableSerializer data, the  json.loads call was also raising a TypeError: "the JSON object must be str, bytes or bytearray, not 'dict'"


nllong:
Thanks for fixing this. Sorry that our tests didn't catch this issue :/. I'll merge once the tests pass.

PR #1911, opened 2019-07-08
Matching update

adrian-lara:
**Note/Reminder**:This PR includes changes to pre-existing migrations. Deployments that had migrations after `0104_auto_20190509_1854` will need to have any proceeding migrations rolled back, then the migrations within this PR will need to be applied. Presumably, this only made it as far `dev1` - further considerations would likely be needed if other deployments are affected.

In order to more easily implement year-over-year functionality, it was decided to revamp matching. 

The purpose of these changes were to simplify the matching logic while giving users the power to decide what constitutes a match. Along the way, a pre-emptive, though still related, update was made to merge roll-up matches on different occasions.

Note: Technically, any of the merge roll-ups are reversible.

#### What's this PR do?
The following changes have been made:
- User can specify matching criteria columns with `is_matching_criteria` field on Column model.
- Extra data columns can't be matching criteria.
- There must be at least one matching criteria.
- The mapping page during file import lists these matching criteria columns in the texts and warnings.
- heavy refactor of record **import** matching logic to be more DB dependent, (hopefully) improve transparency, and reusability
- Using code mentioned in bullet above, match-merge logic created for a single, **existing** record sets (-States, -Views, Canonical Record).
- Match-merge round run post record-edit - roll ups can occur here.
- Match-merge round run post records-merge - roll ups can occur here.
- Match-merge round run on incoming records - roll ups can occur here. (This existed before but only the 1st of many matches was merged.)
- During any merge roll ups, priority is given to most recently updated records (by corresponding AuditLog dates), then, if applicable, the record in focus (on edit or merge).
- Merging should still combine/preserve labels, notes, pairings, and if applicable, meters.
- (Pre-emptive update) Merging and unmerging records creates new canonical IDs, while only preserving the old ones if they're used in other -Views.

With a dataset of properties and tax lots having labels, notes, pairings, and meters, take the following actions in a variety of orders. While doing this, make note of any areas that could use more/less warning or feedback for users.
- update the matching criteria via the Column Settings page
- import matching properties and tax lots ("matching" according to your specified criteria)
- manually merge properties and tax lots
    - merges should result in roll ups if there are records that match the resulting record
- via the inventory detail page, edit a record to match or that already matches other record(s) and ensure roll ups occur.
- Unmerging any records should be the same (even though canonical records are usually changed).


seed/tests/test_match_existing.py
```diff
@@ -0,0 +1,796 @@
...
+
+        self.assertEqual(TaxLot.objects.count(), 3)
+        self.assertEqual(TaxLotState.objects.count(), 3)
+        self.assertEqual(TaxLotView.objects.count(), 3)
+
+        state_ids = list(TaxLotView.objects.all().values_list('state_id', flat=True))
+        self.assertCountEqual([tls_1.id, tls_2.id, tls_3.id], state_ids)
+
+
+class TestMatchingExistingViewFullOrgMatching(DataMappingBaseTestCase):
```
adrian-lara: Note, this method and tests will be heavily changed in year-over-year work. It might not be the most useful to review now.

seed/utils/match.py
```diff
@@ -0,0 +1,155 @@
...
+
+    if count > 1:
+        merged_state = merge_states_with_views(state_ids, org_id, 'System Match', StateClass)
+        view_id = ViewClass.objects.get(state_id=merged_state.id).id
+        return count, view_id
+    elif count == 1:
+        return 0, None
+
+
+def whole_org_match_merge(org_id):
```
adrian-lara: Note, this method and tests will be heavily changed in year-over-year work. It might not be the most useful to review now.

seed/utils/merge.py
```diff
@@ -0,0 +1,261 @@
+# !/usr/bin/env python
```
adrian-lara: Much of the logic here existed previously. These changes were made to isolate it for reuse. Testing was set up before this refactor to help ensure functionality remained the same.

seed/utils/match.py
```diff
@@ -0,0 +1,155 @@
+# !/usr/bin/env python
```
adrian-lara: Much of the methods were moved from the matching file within the `data_importer` module. I took the approach of separating matching logic specific to incoming data and general matching logic that includes logic for existing data.

seed/data_importer/tasks.py
```diff
@@ -1138,350 +1137,6 @@ def _get_field_from_obj(field_obj, field):
     return m.hexdigest()
 
 
-def filter_duplicated_states(unmatched_states):
```
adrian-lara: These methods were moved to the `/data_importer/match.py` file and heavily refactored.

seed/data_importer/match.py
```diff
@@ -0,0 +1,429 @@
...
+    )
+    matched_count = StateClass.objects.filter(pk__in=merged_state_ids).update(
+        data_state=DATA_STATE_MATCHING,
+        merge_state=MERGE_STATE_MERGED
+    )
+
+    return list(set(processed_views)), duplicate_count, new_count + matched_count
+
+
+def save_state_match(state1, state2, priorities):
```
adrian-lara: This was moved, but not actually changed.

seed/views/properties.py
```diff
@@ -623,18 +527,26 @@ def unmerge(self, request, pk=None):
...
+        new_property_2 = Property.objects.get(pk=new_property.id)
+        new_property_2.id = None
+        new_property_2.save()
+
         Property.objects.get(pk=new_property.id).copy_meters(old_view.property_id)
+        Property.objects.get(pk=new_property_2.id).copy_meters(old_view.property_id)
+
+        # If canonical Property is NOT associated to a different -View, delete it
+        if not PropertyView.objects.filter(property_id=old_view.property_id).exclude(id=old_view.id).exists():
+            Property.objects.get(pk=old_view.property_id).delete()
```
adrian-lara: Note: This and the same logic for Tax Lots are done to "clean up" canonical records. This technically isn't relevant until linking is introduced as canonical records only belong to one -View for now.

axelstudios:
@adrian-lara It looks like you can currently uncheck all matching fields and then save, we definitely want to prevent that.

Also, would it be possible to dynamically update the help text and error conditions on the mapping screen to reflect the correct matching fields?

image

adrian-lara:
@axelstudios just to document our conversation/issue and fix.

Currently in this PR, automatic merging of existing records makes use of the `Favor New/Merge Protection` column setting. Since this setting was only really meant for incoming records or manually merging records (where the user can manually change the merge order), ultimately, we realized this made the outcome of the automatic merge for existing records inconsistent and difficult to predict.

Because of that, I’ll work on including updates for merging existing records. Specifically, we'll be ignoring merge protection settings and prioritizing merges by most recently updated according to audit logs.

adrian-lara:
Seeing as there's one approval already, it's worth noting here that I'm still working to get the `Favor New/Merge Protection` related update into this PR. So as of now, the DO NOT MERGE label is definitely appropriate.

adrian-lara:
> ...Currently in this PR, automatic merging of existing records makes use of the `Favor New/Merge Protection` column setting. Since this setting was only really meant for incoming records or manually merging records (where the user can manually change the merge order), ultimately, we realized this made the outcome of the automatic merge for existing records inconsistent and difficult to predict.
> 
> Because of that, I’ll work on including updates for merging existing records. Specifically, we'll be ignoring merge protection settings and prioritizing merges by most recently updated according to audit logs.

I was able to implement logic to override the matching protection settings for existing records.

During the implementation and testing of this, I noticed that an AuditLog is created for an existing record in the following situations:
1. When a record is _**first**_ edited. Specifically, nothing was done to a record since import to create an AuditLog.
2. Whenever a record is merged with another.

To elaborate the first case above, an AuditLog is not created on edit after the first time it's edited. This doesn't quite allow the (match merge) roll-ups to prioritize records by most recently updated (edited or manual merged) by using the corresponding AuditLogs. 

My first thought is to change the edit endpoints to create an AuditLog on every edit, not just on the _first_ edit. @nllong or @axelstudios do you think this is the correct next step? Is it necessary? Are there any consequences to making this change?

RDmitchell:
@nllong / @adrian-lara -- does this affect the "meter" release, or is this for the release after that?

adrian-lara:
> @nllong / @adrian-lara -- does this affect the "meter" release, or is this for the release after that?

@RDmitchell I understand this to be for the next release (not this "meter" release).

seed/migrations/0108_auto_20190730_1212.py
```diff
@@ -0,0 +1,59 @@
...
+from __future__ import unicode_literals
+
+from django.db import connection, migrations, models
+import django.utils.timezone
+
+
+def forwards(apps, schema_editor):
+    property_sql = (
+        "UPDATE seed_propertystate " +
+        "SET created = seed_propertyauditlog.created, updated = seed_propertyauditlog.created " +
```
adrian-lara: Note: In order to populate these fields with a reasonable base value, I make an attempt to update these new `created` and `updated` -State fields with the value from the corresponding -AuditLog. 

If a -State doesn't have an -AuditLog, the "current' time of when the migration is run is used for both values. I'm only aware of this (a -State without an -AuditLog) happening is when the preliminary, extra_data-only records are created during import.

seed/tests/test_columns.py
```diff
@@ -1088,7 +1088,7 @@ def test_retrieve_db_field_name_from_db_tables(self):
         """These values are the fields that can be used for hashing a property to check if it is the same record."""
         expected = ['address_line_1', 'address_line_2', 'analysis_end_time', 'analysis_start_time',
                     'analysis_state_message', 'block_number', 'building_certification',
-                    'building_count', 'campus', 'city', 'conditioned_floor_area', 'created',
+                    'building_count', 'campus', 'city', 'conditioned_floor_area',
```
adrian-lara: I'm not sure why `created` and `updated` were expected to be used (if they existed) for hashing a property.

seed/models/columns.py
```diff
@@ -53,8 +53,10 @@ class Column(models.Model):
     # Do not return these columns to the front end -- when using the tax_lot_properties
     # get_related method.
     EXCLUDED_COLUMN_RETURN_FIELDS = [
+        'created',
```
adrian-lara: The inclusion of `created` and `updated` within 
- EXCLUDED_COLUMN_RETURN_FIELDS
- COLUMN_EXCLUDE_FIELDS
- EXCLUDED_MAPPING_FIELDS

...all seemed appropriate. Can anybody else confirm this?

seed/tests/test_columns.py
axelstudios: Yeah, I don't think we would want those considered when hashing.  That said, since they were previously being returned by `retrieve_db_field_name_for_hash_comparison()` that may mean we need a migration to recompute all hashes.  @nlong thoughts?

adrian-lara: Interestingly, I would expect that hashes would need to be recomputed because of this line: https://github.com/SEED-platform/seed/blob/develop/seed/data_importer/tasks.py#L1117

But after importing the same file twice under the following circumstances:
1. including `created` and `updated` in COLUMN_EXCLUDE_FIELDS
2. excluding those fields from COLUMN_EXCLUDE_FIELDS

... the hash values were the same and the records were ignored on the second import? I'm planning to investigate more, but this wasn't what I expected.

adrian-lara:
In order to test the latest changes to the hashing process, take a DB updated with the latest `develop` migration - upload different files. Then, roll back `seed` migrations to `104...` switch to this branch, and run these latest migrations (`105`... to `109`). Finally, upload the same fiels same files.

seed/data_importer/tests/test_match_incoming.py
```diff
@@ -0,0 +1,1019 @@
+# !/usr/bin/env python
```
nllong: great set of tests here, thank you!

docs/source/matching.rst
```diff
@@ -0,0 +1,69 @@
...
+In-Cycle Merging
+****************
+For records within the same Cycle, there really shouldn't be more than one
+representation of the same property (or tax lot). As much as possible, the program
+is set up to prevent this from happening by automatically **merging** matched
+records together whenever it might occur.
+
+Specifically, a search for matches occurs for a given record in the following cases:
+
+1. The record has been manually edited.
```
nllong: this is a new feature, right?

docs/source/matching.rst
```diff
@@ -0,0 +1,69 @@
...
+via merge as described in the above section the next time the system finds them during a match search.
+
+Match Searching in Depth
+************************
+Though they accomplish the same goal, the process for merging is very different between the last case, importing,
+and the first 2 cases, manual edit or manual merge.
+
+In the case of manual merging or editing, this process accounts for the fact that these are records that already exist.
+Specifically, they may have associations such as labels, notes, pairings, and for properties, meters.
+So during a subsequent match search leading to a merge of two or more records, all of these "old" associations are be
```
nllong: `are be` -> `are to be`

RDmitchell:
I can't tell from this description if this is the old or the new functionality but merging back together records that the user has unmerged by hand may be a problem, if that is what is being implemented (which I can't really tell if that's the case).
image

docs/source/matching.rst
adrian-lara: Yes

docs/source/matching.rst
adrian-lara: Changed it to just `are`

adrian-lara:
> I can't tell from this description if this is the old or the new functionality but merging back together records that the user has unmerged by hand may be a problem, if that is what is being implemented (which I can't really tell if that's the case).
> image

This is new functionality and is done purposely to prevent one of the key situations that would break linking (cross-cycle associations).

Ideally, if a user wants to unmerge records that were automatically merged, they absolutely can. Though if they don't take further action that includes editing any of the matching fields so that the two records no longer match, they will be rolled up again once the app discovers they still match.  As mentioned above, this is necessary to allow for linking, and since users now have the ability to choose matching fields, we've enabled them to dictate what exactly causes records to be automatically matched.

seed/data_importer/tests/test_match_incoming.py
adrian-lara: 👍 I absolutely would not have attempted a refactor of the existing logic to match on import without these.

nllong:
Can you make this column 'sortable'? Some users have hundreds of columns.

image

seed/models/columns.py
nllong: yeas this sounds right  based on our discussions. The property.updated and propety.created are not included in the hash, nor should any created/updated.

seed/migrations/0109_auto_20190730_1212.py
```diff
@@ -0,0 +1,87 @@
...
+from django.db import connection, migrations, models
+from seed.data_importer.tasks import hash_state_object
+
+import django.utils.timezone
+
+
+def forwards(apps, schema_editor):
+    property_sql = (
+        "UPDATE seed_propertystate " +
+        "SET created = seed_propertyauditlog.created, updated = seed_propertyauditlog.created " +
```
nllong: tahnks for adding these!

adrian-lara:
> this looks good. tested locally and things appear to work as stated. Still more testing needed with more datasets, but I suggest we merge down.

👍 Sounds great! Just a reminder - any deployments (i.e. dev1) will likely need to have their migrations rolled back (likely to `0104_auto_20190509_1854 `) and then updated with the migrations coming from this PR.

nllong:
Thanks for the reminder. On dev1 we will just redeploy the db.

Il giorno mer 14 ago 2019 alle 08:14 Adrian Lara <notifications@github.com>
ha scritto:

Issue #1912, opened 2019-07-10
Error importing Green Button meter data


RDmitchell:
### Expected Behavior
Import green button data
### Actual Behavior 
Error occurred at import. Also triggered a Sentry error
### Steps to Reproduce
See files and doc in this folder
https://drive.google.com/drive/folders/1Ar7hVHkmIPdjUPQl0dERgAbB6WYRZ2oe?usp=sharing
Server Instance: dev1
Server SHA: 93c4bdca

RDmitchell:
Do we need a specification doc (that we can put up on our website) that shows the format of green button data that SEED will accept? It could be that this green button file, which is from 2013-2014, is old and the format changed ??

RDmitchell:
@adrian-lara -- I was able to upload a much newer PGE green button file, so maybe the older files have a formatting issue. I have put the file that I was able to import into this issues directory and added that info to the issue doc.

RDmitchell:
@adrian-lara -- could it be that SEED doesn't import gas green button data? The first file that got an error was gas, the 2nd file that imported was electric data.

RDmitchell:
@adrian-lara -- it seems like SEED can only import electric green button data. The two electric data files I have both import, the one gas file I have does not. So that's probably the issue. 

@nllong -- seems like SEED should be able to import gas green button data? Not sure what the original design / intent was.

adrian-lara:
> @adrian-lara -- it seems like SEED can only import electric green button data. The two electric data files I have both import, the one gas file I have does not. So that's probably the issue.
> 
> @nllong -- seems like SEED should be able to import gas green button data? Not sure what the original design / intent was.

SEED is set up to accept gas green button data, but it looks like it's tripping up on receiving this particular combination of units - "milli" therm.  I'm looking into a general fix now.

adrian-lara:
@RDmitchell In the meantime, if you'd like to see imported gas meters, you can upload this version of the file you shared with me: https://drive.google.com/open?id=1lDfiPbVIqtVAMZ8UsqT_ply18pV7_ONW

**I manually changed the contents.**

Specifically, I changed the units specified in line 49
from: `<powerOfTenMultiplier>-3</powerOfTenMultiplier>`
to: `<powerOfTenMultiplier>0</powerOfTenMultiplier>`

RDmitchell:
thx!

RDmitchell:
[...]

PR #1914, opened 2019-07-10
GreenButton meter import unit reading fixes and improvements

adrian-lara:
SEED's GreenButton meter importer couldn't read files with specific units and power-of-ten-multiplier combinations. 

Specifically, this was an issue for units that the app only understands without prefixes. For example, the app understands the unit "therms" (as opposed to something like "kBtu", which has a prefix included in the name).  In that example, the application's GB meter import logic was not prepared to handle one of these un-prefixed units with a non-zero power-of-ten-multiplier. 

Notably, as of this writing, the only unit that falls in this case **and** can be found by GreenButton is actually "therms".

#### What's this PR do?
Logic is added to the GreenButton meter file parser to address the scenario described above.

Import GreenButton gas meters of "therms" along with 0 and non-0 power-of-ten-multipliers.
Import GreenButton meters of other types, units, and multipliers.

#1912

### DIFF
--- a/seed/lib/mcm/reader.py
+++ b/seed/lib/mcm/reader.py
@@ -134,57 +134,94 @@ class GreenButtonParser(object):
 
     def _parse_type_and_unit(self, raw_data):
         """
-        Uses the kind and uom/powerOfTenMultiplier to parse type and
-        raw unit, respectively.
+        Parses raw XML to read the kind and uom/powerOfTenMultiplier. Using
+        those, an attempt is made to validate the type and unit as a combination
+        that the application accepts.
 
-        For the given type, it first scans the valid units for that type to see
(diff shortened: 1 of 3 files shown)


adrian-lara:
> Do you want me to merge and deploy this to dev? @adrian-lara @RDmitchell

@nllong Yes, except that I'm not sure how much time/effort it takes to get the changes to dev1 and there might be some more changes needed for the meters feature. I'll reach out separately.

adrian-lara:
@nllong following up, this is good to merge to develop, but updating dev1 with it might not be worth.

nllong:
@adrian-lara -- this has been deployed.

RDmitchell:
Tested and closed #1912 -- thanks !

PR #1925, opened 2019-07-18
Sticky table headers

axelstudios:
#### What's this PR do?
A number of miscellaneous fixes I've had in my working directory:
- Adds sticky table headers to most standard tables, including:
    - dataset list, dataset detail, superuser orgs, organizations page (both tables), cycles, labels, sub organizations, org members
- Includes js lint cleanup
- Fixes the lodash vulnerability, along with all node module audit issues

- Go to the pages listed above and scroll down
- Check that the GitHub vulnerable dependency warning is gone after merging

#1927

Sticky headers


RDmitchell:
sticky table headers !!! Yipppeee !!! 

We should try to do this to all screens that don't have them. One in particular is the Admin screen for superusers. I will make an issue for that (and maybe try to list all the other places where sticky table headers would be nice).

Thanks !!

nllong:
@axelstudios - can you create an "issue" for this and add the label "feature" so this ends up in the change log?

Looks good though. I'll test it later today!

axelstudios:
Issue created 👍

Issue #1919, opened 2019-07-12
Meter Feature Improvements


RDmitchell:
### Expected Behavior
n/a
### Actual Behavior 
n/a
### Steps to Reproduce
Here is a link to a doc with suggestions for improving the meter feature, with suggested priority. When we decide to deal with these we can make separate issues for them as needed. 
https://docs.google.com/document/d/1RalxFMoQOxdwCkRwNC8Y5lU4zrmP3yLIIRyRhNhwJAo/edit?usp=sharing
### Instance Information
instance: dev1
SHA: 04c106fa

RDmitchell:
@RDmitchell -- update the doc with the changes @adrian-lara will make in the next day or so, then make issues for the ones that are not complete, then close this issue.

RDmitchell:
instance: dev1
SHA: 
org: LBNL 200
Cycle: 2019 Compliance

1 - ESPM Meter Data Import - Add PM Property ID - Priority = High - **DONE** !
Add the PM Property ID field to the confirmation dialog box for each meter when as ESPM meter data is being imported. 
![image](https://user-images.githubusercontent.com/6314950/61570866-3d08dd80-aa44-11e9-9ac8-72aa318ae3e7.png)

2 - ESPM import dialog box - Make columns wider or expandable  - Priority = High - **DONE** !
* Make this dialog box larger so that the fields names can be displayed, or allow the right hand side of the column to be dragged to the right to widen the column to see the column names. 
* Add PM Property ID as well as the existing Meter ID
They are wider and the columns are resizable -- EXCELLENT !!
![image](https://user-images.githubusercontent.com/6314950/61570915-ab4da000-aa44-11e9-9636-0ca6b37e2acd.png)

RDmitchell:
@adrian-lara -- Really excellent improvements, thank you so much !!

I will extract the lower priority items from the referenced doc, and make a new issue (or issues) for them for future. But these improvements are great for the upcoming Meter Data release.

Issue #1918, opened 2019-07-12
ESPM Meter Types that SEED doesn't recognize (what are they?)


RDmitchell:
### Expected Behavior
When importing an ESPM spreadsheet with Meter data in it, SEED knows about certain Meter Types
### Actual Behavior 
I got an error importing my sample ESPM meter data spreadsheet
### Steps to Reproduce
* I downloaded an ESPM spreadsheet with Meter data in it from my test ESPM account, and there were records in it with Meter Types that don't seem to be standard "energy" meter types, and which I don't know what they represent. 
Here is a link to the file (in the folder for this issue).
https://drive.google.com/drive/folders/1y9sdSvwvkKlCABc5fFgCzy_G5PEcDCzH?usp=sharing
* Here is a screenshot of the Meter Types that were in the spreadsheet, with the odd types highlighted in yellow.
![image](https://user-images.githubusercontent.com/6314950/61158501-f6e4d480-a4ad-11e9-9e49-e584bc7a64b0.png)
* SEED couldn't recognize them (which is understandable), and gave an error when I tried to import the file.
* I deleted all records with those meter types and then could import the file (see #1917 )
Not sure how we should deal with this issue -- for now I will tell users to delete any records with those Meter Types. In the future we may want to ignore them on import (and tell users we are ignoring them?), or understand what they mean and whether we should be importing them (?)
### Instance Information
Instance: dev1 (NREL)
SHA: 04c106fa

adrian-lara:
@RDmitchell I think it'd be ideal if what you suggested was in place - that unknown meter types are ignored and users are shown these ignored meters. I'll let @nllong weigh in on that and it's general priority before adding it on my list of fixes/changes to implement right away.

RDmitchell:
Instance: dev1 (NREL)
SHA: e120119f
Org: LBNL 200
Cycle: 2019 Compliance
ESPMCityBldgs-all tabs.xlsx

This file is a "Portfolio" download spreadsheet from my test ESPM account with 10 buildings and lots of miscellaneous meter data, including records with the odd Meter Types
![image](https://user-images.githubusercontent.com/6314950/61571622-90c9f580-aa49-11e9-88b3-a50411b44cb0.png)

The program imported the file without issue, ignoring the odd meter types, only importing the meters that we have decided SEED should import (at least for now).
![image](https://user-images.githubusercontent.com/6314950/61571664-ccfd5600-aa49-11e9-9ccc-409dbbca2860.png)

Issue #1917, opened 2019-07-12
Importing ESPM Meter data for multiple buildings


RDmitchell:
### Expected Behavior
Import a spreadsheet from ESPM with only a Meter Entries tab but with multiple buildings (Portfolio Manager ID)
### Actual Behavior 
* It seemed like it was going to import properly, showing the list of meters 
* But then I get an error that it can't complete the import. 
* Then I tried importing it again and it worked (!!). 
So not sure you will be able to reproduce this but wanted to log it.
### Steps to Reproduce
See this folder for file to import (ESPMCityBldgs-MeterEntriesOnly.xlsx)
https://drive.google.com/drive/u/0/folders/1DaLNcy9MIiImf8JuamfXg6E5E5WKXVw2
* Import the file into a Data Set
* Should see a confirmation dialog box showing the meters
* Click the Confirm button
![image](https://user-images.githubusercontent.com/6314950/61157209-53de8b80-a4aa-11e9-9643-53ced8b4fe22.png)
* Got an error (that I didn't record) and the file wouldn't import
* I believe it generated a Sentry error (1109124477) (but this issue does not address that), and it is unclear what actually triggered that Sentry error)
* But then I tried to import the file again and it worked (!!).. **_sigh_** ...
* Here is the final dialog box confirming the import
![image](https://user-images.githubusercontent.com/6314950/61157489-2219f480-a4ab-11e9-94cc-d3781dbfaef1.png)
* And the meter data is matched to the properties in the Inventory Detail Meter view
![image](https://user-images.githubusercontent.com/6314950/61157779-f77c6b80-a4ab-11e9-9639-d2dff8da2b87.png)

### Instance Information
Instance: dev1 (NREL)
SHA: 04c106fa
Org: LBNL 200
Cycle: 2019 Compliance Cycle
Dataset: 2019 Compliance

adrian-lara:
@RDmitchell just a heads up, I'll be creating a separate issue for the Sentry error you provided. Likely, no action is needed from you unless you want to remove the text about the SentryError here as to avoid anyone thinking it was resolved once this is issue closed.

RDmitchell:
@adrian-lara -- thanks -- I deleted the image, but left the Sentry number in. 

And feel free to make a new issue. Sometimes I put too many things in one issue and they need to be separated out !!

RDmitchell:
Instance: dev1 (NREL)
SHA: e120119f
Org: LBNL 200
Cycle: 2019 Compliance

[...]

PR #1926, opened 2019-07-18
Meter feature improvements

adrian-lara:
NREL-side testing was done and feedback was given on potential improvements.

#### What's this PR do?
The following feature improvements are contained in this PR. The reasoning behind each change is also given:
- [x] Meter import modal is larger and columns are resizeable (#1919 item 2)
- [x] Meter detail page and notes detail pages show record address (#1919 item 4)
- [x] PM Property IDs shown during PM meter import process (#1919 item 1 and #1917)
- [x] PM Meter import supports workbook with multiple sheets (#1919 item 6)
- [x] Inventory detail meter page - sort list of meters to filter (this change came from my own testing)
- [x] Unknown Meter types and/or units are ignored - valid types and units are still displayed (#1918)

Go through the list above and ensure each has been updated and is working. 
Also, it might be useful to double check that other, non-meter related still work. I made some minor changes to the logic for reading Excel files.

For actual code review, it might be easier to analyze files by commit.

#1917, #1918, #1919 


RDmitchell:
@adrian-lara -- lots of great improvements !! Thank you !!

seed/static/seed/js/controllers/data_upload_modal_controller.js
```diff
@@ -244,6 +251,9 @@ angular.module('BE.seed.controller.data_upload_modal', [])
         $scope.uploader.in_progress = false;
         $scope.uploader.progress = 0;
 
+        var modal_element = angular.element(document.getElementsByClassName('modal-dialog'));
```
adrian-lara: Given this `modal-dialog` element is created by the parent controller, I took this approach to showing a bigger modal only for Meter uploads. Let me know if there was a better way to go about it.

nllong: @axelstudios - any comments on this?

nllong:
Can we remove these files from seed/tests/data?

```
example-pm-monthly-meter-usage-2-cost-meters.xlsx
example-pm-monthly-meter-usage-with-unknown-types-and-units.xlsx
example-pm-monthly-meter-usage.xlsx
```

The ones that we should use for testing live in seed/data_importer/tests, correct?

adrian-lara:
> Can we remove these files from seed/tests/data?
> 
> ```
> example-pm-monthly-meter-usage-2-cost-meters.xlsx
> example-pm-monthly-meter-usage-with-unknown-types-and-units.xlsx
> example-pm-monthly-meter-usage.xlsx
> ```
> 
> The ones that we should use for testing live in seed/data_importer/tests, correct?

One set is used to test the Meter confirmation endpoint (which is a meter endpoint), and the other is used to test the actual importing of meter (data_importer endpoint).

seed/static/seed/js/controllers/data_upload_modal_controller.js
axelstudios: Good enough for now - `$uibModal` has very limited options (and no access to the host element), and because modals are direct children of the `<body>` element it's a pain to have dynamic styles for them.  If we feel like addressing this in the future my suggestion would be to move the `ng-app="BE.seed"` directive to the `<html>` element and create a service to manage classes that apply to `<body>`.  That way you could create a style like `body.modal-lg .modal-dialog` that's set with ngClass on the body from a service.

Issue #1950, opened 2019-08-27
ULID-only mapping - Geocoding confidence text is wrong


RDmitchell:
### Expected Behavior
Import Tax Lot data matching only on ULID, expected Geocoding confidence to report "Manually geocoded (N/A)"
### Actual Behavior 
However, it reported "Missing address components (N/A)". I did the same type of import for Property (mapping to UBID) and the program DID populate the Geocoding Confidence field with "Manually geocoded (N/A)"
![image](https://user-images.githubusercontent.com/6314950/63732356-ba382700-c828-11e9-80dd-a2bb171cb213.png)
### Steps to Reproduce
See doc and files in this issue folder
https://drive.google.com/open?id=1Fm9CEp4SqX_MEKLXuKlhUC_Mbt7m_yEw

### Instance Information
Instance: seeddemostaging
SHA: e120119f (2.6.0)

RDmitchell:
Instance: Dev1
SHA: 6c178358

This is fixed (see #1945 
![image](https://user-images.githubusercontent.com/6314950/65648317-aa8e3880-dfb6-11e9-8872-9987b66e2592.png)

Issue #1944, opened 2019-08-14
ULID doesn't report that it could generate Lat/Long, but then does generate them


RDmitchell:
### Expected Behavior
Import a file that has ULID (for Tax Lot) but no address info. I expected the program to be able to generate lat/long values based on that ULID
### Actual Behavior 
After the import, program reported that "geocoding results for tax lots" were "unsuccessful due to missing address fields"
### Steps to Reproduce
* Import file that has (I believe) real ULIDs representing tax parcels
* Map to Tax Lot: ULID (although I also had to map to Jurisdiction Tax Lot ID due to #1943 )
![image](https://user-images.githubusercontent.com/6314950/62985932-42acd580-bdee-11e9-9449-27a2b0b26d19.png)
* Save mapping, program reports geocoding unsuccessful
![image](https://user-images.githubusercontent.com/6314950/62985913-26a93400-bdee-11e9-8134-6335179a1af9.png)

### Instance Information
Instance: seeddemostaging (LBNL)
SHA: e120119 (2.6.0 release)

RDmitchell:
Test file is here (issue folder)
https://drive.google.com/open?id=1RiAue3-Jqpw5bmCKZCzAX28mrVlKsZ7f&authuser=rdmitchell@lbl.gov&usp=drive_fs

RDmitchell:
It appears that even though the program reported in results that there geocoding was not successful, the program did actually populate the lat/long fields in the Tax Lot data.
![image](https://user-images.githubusercontent.com/6314950/62986428-6bce6580-bdf0-11e9-81cd-766d3249bf80.png)

**So seems like the Results dialog box needs to be more explicit about what the program actually generated, ie, the geocoding was successful from "address data" or "UBID" or "ULID" or whatever**
*******
Also, the "note" for "geocoding confidence" was different for the same situation in Property (ie, only UBID, no address info). 

For Tax Lot / ULID, it says:
"Missing address components (N/A)
![image](https://user-images.githubusercontent.com/6314950/62986387-3164c880-bdf0-11e9-8d7d-b2239434b825.png)

For Property / UBID, it says:
"Manually geocoded (N/A)"
![image](https://user-images.githubusercontent.com/6314950/62986329-ed71c380-bdef-11e9-8e32-604b2d3ca161.png)

Seems like that "note" should be the same for both cases.

RDmitchell:
The same case for UBID, where there is no address info but just UBID, the results dialog reports that geocoding was successful
[...]

Issue #1942, opened 2019-08-12
On import, duplicates are not flagged when ULID/UBIDs are involved


adrian-lara:
### Expected Behavior
After importing `covered-buildings-sample-with-UBID.csv` multiple times using the same mappings, records from imports after the first should be flagged as duplicates and ignored.

File located here: https://drive.google.com/drive/folders/1tNQeehIVqcWNsVo13jYjP_qS2xv5CszS

### Actual Behavior 
Treating these records as TaxLots with raw UBID column considered ULID, only 55 of the 512 are flagged as duplicates every time this file is uploaded.
<img width="539" alt="Screen Shot 2019-08-11 at 7 35 19 PM" src="https://user-images.githubusercontent.com/30608004/62842868-a6aa8f00-bc72-11e9-8154-6569c6d9553d.png">

Treating these records as Properties mapping UBID to UBID (with everything else as a non-matching field), only 10 of the 512 are flagged as duplicates every time this file is uploaded.

Separately, if this file does not have any columns mapped to ULID or UBID, duplicates are caught or flagged as expected.

### Steps to Reproduce
See above.

### Instance Information
N/A

RDmitchell:
Instance: dev1
SHA: 6c17835

**ULID Case:**
Using the same data set as for #1944 and imported the ULID file (Covered_Buildings_UBID-ONLY.xlsx) 2nd time (all fields mapped to Tax Lot, only two fields, ULID and Parcel ID mapped to Jurisdiction Tax Lot ID), the program reported that all the imported records were duplicates:
![image](https://user-images.githubusercontent.com/6314950/65648547-8717bd80-dfb7-11e9-842e-d80dc4130919.png)
Note: It would be nice to show the filename in this results screen to verify what was just done, but that's another issue (!)

**UBID Case:**
I imported another file from #1944 (Covered_Footprints_ULID-ONLY-50KUp.xlsx) that had the fields FacilityID, UBID and Area. I mapped all fields to Property Table -- Facility ID to Custom ID 1, UBID to UBID and Area to Area.
![image](https://user-images.githubusercontent.com/6314950/65648648-fab9ca80-dfb7-11e9-9d6c-9e98601cb3e3.png)

Seems to be stuck on 33% complete (there are only 25 records) even after several minutes.
![image](https://user-images.githubusercontent.com/6314950/65648847-df02f400-dfb8-11e9-8385-2d0ba82009e5.png)
[...]

PR #1954, opened 2019-09-04
Taxlot geocoding update

adrian-lara:
There were multiple issues with Tax Lot geocoding - specifically involving ULID.  These essentially came from the previous (and now invalid) assumption that Tax Lots would not be eligible for manual geocoding via provided long/lat or ULID.

#### What's this PR do?
Addresses the issues listed below.
- Geocoding logic update to enable manual geocoding for tax lots
- Fix import dedup fail with pre_save to sync taxlot lat and long
- Import feedback provided for manual geocoding of taxlots via ULID (update relevant view, controller, and partial)

1. Import TaxLots with ULID, checking to see that each record with a valid ULID is accounted for in the import results feedback.
2. Import that same file again and be sure that everything was flagged as a duplicate.
3. Check that any/all of the entries were geocoded manually - Geocoding confidence should be set to "Manually geocoded (N/A)".

- #1950 
- #1944 
- #1942 


seed/models/tax_lots.py
```diff
@@ -488,4 +488,26 @@ class Meta:
...
+        pass  # Occurs on object creation
+    else:
+        # Sync Latitude, Longitude, and long_lat fields if applicable
+        latitude_change = original_obj.latitude != instance.latitude
+        longitude_change = original_obj.longitude != instance.longitude
+        long_lat_change = original_obj.long_lat != instance.long_lat
+        lat_and_long_both_populated = instance.latitude is not None and instance.longitude is not None
+
+        # The 'not long_lat_change' condition removes the case when long_lat is changed by an external API
+        if (latitude_change or longitude_change) and lat_and_long_both_populated and not long_lat_change:
```
nllong: hmm. i get this logic, but I wonder if there is an easier way. Approving for now.

Issue #1946, opened 2019-08-16
Meter data visible on some cycles but not all 


RDmitchell:
### Expected Behavior
Import a set of meter data for a record, then view it in the cycle that the data spans, ie, in this case the data is from 2018, and the user tried to view it through the Property Detail / Meter view coming from Cycle == 2018
### Actual Behavior 
The meter data does not display in the cycle == 2018, but in two other cycles, 2014 and 2016. There is no way to tell what cycle the meter data will be associated with. My understanding is that meter data is not associated with a cycle, but that seems like a problem given that users in general use cycles to organize their data, will have many years of data, and don't know which cycle the meter data will be visible from. 
### Steps to Reproduce
See the doc in this issue folder
https://drive.google.com/open?id=1SfEBa4iLFkRzgLYdzMY695s25ANNrvrL
### Instance Information
Instance: seeddemostaging
SHA: e120119 (2.6.0 release)

RDmitchell:
I think this is a problem that I didn't discover in testing because even though I had many cycles defined, and imported into all of them, none of them had the same data, ie, the same building list with the same ESPM ID numbers. In this case, there are several years of cycles, each with the same ESPM IDs in them. So maybe that is part of the why we are seeing this behavior. We will have to figure out how to deal with this, because as it is currently implemented, it's not really going to work for most users.

RDmitchell:
I am also checking with the user to make sure that they didn't inadvertently set the cycle to 2016 when they imported the meter data -- I don't know if that is relevant or not.

RDmitchell:
More info from the user.
* The cycle was set to 2010 when they uploaded the data.
* It is visible from cycles 2014 and 2016, but not the most recent cycle, 2018, which is the year range that the meter data is from. 

More details in the doc referenced above.

RDmitchell:
In a test org I made 4 cycles, 2015-2018, imported ESPM data with 2018 dates, then meter data (cycle set to 2015 on upload), and the data appeared in the Property Detail via the 2018 cycle. So with that simple set up, I couldn't replicate the behavior seen in the description above. I added this test info to the referenced doc.

adrian-lara:
moved this to "In Progress" since this won't be ready to deploy until the linking PR is merged

RDmitchell:
@adrian-lara -- I can't seem to import ESPM Meter data. 

See the beginning of the doc for this issue for today's testing results

[...]

Issue #1591, opened 2018-03-08
Detail Screen: Highlight the changes in the columns


nllong:
This code was already written by Daniel McQuillen and Paul Munday, but not sure it ever got into develop. See the discussion in the Slack channel called version-2-0-ui dated Sept 15, 2016.

See page 3 of doc: BlueSky Design Doc - Property - Tax Lot Detail View
https://drive.google.com/open?id=1KXsJvaWTXmRiER2B1ekQMF0bFq09huo3lNiVqHbBocc

From that convo, from Paul:
note however that feature relies on the presense of changed_fields (which comes from the description filed in the Audit log, which AKAIK, is only added for fields that are updated via the API, directly

Comment from Daniel in private slack channel, 10/20/2016
There used to be a property_taxlot_service.js in this folder but I don’t see it anymore

https://github.com/SEED-platform/seed/tree/develop/seed/static/seed/js/services (edited)

That used to have some comments about how we looked for changed_fields in the data coming from the server

this is an earlier file that had some of that code:

https://github.com/SEED-platform/seed/blob/dmcq-add-cycle-ui/seed/static/seed/js/services/property_taxlot_service.js (edited)

The branch was dmcq-add-cycle-ui

In that same branch I imagine my controller and html view show how that data was used to highlight cells.

https://github.com/SEED-platform/seed/blob/dmcq-add-cycle-ui/seed/static/seed/js/controllers/property_detail_controller.js

https://github.com/SEED-platform/seed/blob/3c4aa8e2fd029052374b3b6a657a19ddcae757a7/seed/static/seed/partials/inventory_detail.html#L78 (edited)

Here is a link to the image provided by Daniel showing the feature
https://files.slack.com/files-pri/T0C6U7VPC-F2C69SLQK/pasted_image_at_2016_09_15_01_42_pm.png

This would be nice to get into the program soon, but is not required for V 2.0 release.

nllong:
Some of this is happening in the #1957 PR

RDmitchell:
@adrian-lara / @nllong 

Instance: dev1
SHA: 6a6d4ba0
Cycle: 2018 Compliance

I went to the detail view of a somewhat randomly selected property (PM 5233255)

There is only one file that has been imported, so just one column in addition to the Master column. However, for all the fields that have no data in either the imported file or the master, the program has highlighted them yellow, which doesn't make sense.

[...]

Issue #1232, opened 2017-02-01
Need easy way to view data year over year


RDmitchell:
From the SEED partner call today (2/1/2017), feedback indicated that the ease of comparing year over year was much easier in V 1.5, where in the detail view you could look at one record and see the different columns representing different years, than it is in V 2.0, where the years are broken out by cycle so that in the detail view you only see one cycle.

We need to add a feature to SEED to easily allow users to compare data year over year:
* on a record by record basis
* over multiple records

This needs design. Even a simple implementation might be nice for V 2.1

RDmitchell:
This could be an option in the Detail view to show only the master records for the cycles selected, rather than the history for the current cycle.

RDmitchell:
instance: dev1
SHA: 32b8c7

I tried testing on some older orgs that had multiple years of data and I didn't quite get the results I was expecting but that could have to do with the data, how it was originally mapped, etc. 

Here is a doc where I will put my various attempts in different orgs to see what the issues might be with older data. 

https://docs.google.com/document/d/1vXovzxfTJy_wxwVIoxveHk9XlCQudVLs4vYCex4MC4c/edit?usp=sharing

RDmitchell:
I think this generally looks good as a first start -- I have several comments for refinement at the top of the linked doc above, but I think the best thing to do is to get user feedback on this before trying to refine it anymore.

adrian-lara:
Reading through your write up above, you provided get feedback on the list Cross-Cycle page! I just wanted to make sure you saw the detail view version for this issue. It'd be good to get your feedback on that as well 
![Screen Shot 2020-01-28 at 8 20 03 PM](https://user-images.githubusercontent.com/30608004/73325610-6a60f000-420c-11ea-9b06-f75aee7fde27.png)

PR #1957, opened 2019-09-11
Linking

adrian-lara:
WARNING: This PR includes a DB migration that automatically runs a match-merge-link round for all records across all cycles across all organizations. (Done to establish a baseline of consistent data)

In order to support year-over-year analysis within SEED, it's necessary to establish `links` between records across cycles. This requires certain record consistency conditions be met, and logic has been included to allow SEED to continually meet these conditions.

Specifically, if two records are linked across cycles, they must match. Implications of this include unlinking records if they no longer match.

For more information on matching, merging, and linking, see the `docs/source/matching.rst` file.

#### What's this PR do?
Quick side note - inventory detail page highlights columns with values changes amongst master and historical was added into this PR.

This PR implements linking between records across Cycles using the canonical Property and TaxLot models (which are associated via the corresponding -Views).

Backend:
- Establishes a baseline of consistent data via DB migration (see warning above).
- Incorporates a linking round whenever there was previously a match-merge round, effectively replacing the match-merge round with the notion of a match-merge-linking round.  This happens in the following cases:
  - post record edit
  - post records merge
  - during the import of incoming records (only affects matches amongst existing records)
- Removes links whenever records are unmerged.
- Removes links whenever records no longer match. This happens in the following cases:
  - post unmerge of records (neither are linked and a match-merge-round isn't run on them right away)
  - post edit of a record
  - user changes matching criteria for an organization. This can be an intensive calculation that can create a state of the database that would be difficult to reverse. Given that, a preview has been created that will show users what their database would look like before confirming the changes.

Frontend UI/UX:
- Links for an individual record can be viewed via Cross-Cycles detail page/tab.
- Links can be viewed in aggregate via a new Cross-Cycles list page/tab.
- As mentioned in the above section, when a user changes matching criteria, they are able to view a preview of the changed data before confirming. The speed of the preview generation and actual calculations depend heavily on the number of record merges needed.
- Notifications are provided when merges are executed and links are created during match-merge-linking rounds.

Also, reprioritization of automatically merged records is included in this PR. The -State  `updated` fields are now used to prioritize these automatic merges. See https://github.com/SEED-platform/seed/pull/1957/commits/f99827df32d308abe8426ebd6461398003f45c94

- Generally, testing should involve going through all the match-merge-linking round scenarios mentioned above and verify links are established (or broken) via the two Cross-Cycle pages. 

#1946 
#1232 
#1591


seed/data_importer/match.py
```diff
@@ -364,6 +376,30 @@ def states_to_views(unmatched_state_ids, org, cycle, StateClass):
     return list(set(processed_views)), duplicate_count, new_count + matched_count
 
 
+def link_views(merged_views, ViewClass):
```
adrian-lara: This is the most notable change in this file. The other changes are mostly from a refactor of matching criteria getters and variable/function renaming.

seed/static/seed/js/controllers/confirm_column_settings_modal_controller.js
```diff
@@ -0,0 +1,326 @@
...
+        warnings_acknowledged: true,
+      }
+
+      // Check if PM Property ID is actually being removed from matching criteria
+      if (_.find($scope.change_summary_data, {column_name: "pm_property_id", is_matching_criteria: false})) {
+        $scope.checks.matching_criteria_excludes_pm_property_id = true;
+        $scope.checks.warnings_acknowledged = false;
+      }
+
+      // Preview
```
adrian-lara: There are some duplicate lines of code between this and the cycles list controller but there are definitely some necessary and nuanced differences. It's likely possible to DRY this up a bit by abstracting the two but I didn't place a high priority on it as it's maybe not the most cost efficient talk. I'm happy to revisit this later though

RDmitchell:
Older Issues that should be checked: 
#6

config/settings/common.py
```diff
@@ -207,7 +207,7 @@
          CeleryDatetimeSerializer.seed_loads,
          content_type='application/json', content_encoding='utf-8')
 CELERY_WORKER_MAX_TASKS_PER_CHILD = 1
-CELERY_ACCEPT_CONTENT = ['seed_json']
+CELERY_ACCEPT_CONTENT = ['seed_json', 'pickle']
```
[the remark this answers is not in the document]
nllong: 👍

RDmitchell:
@adrian-lara / @nllong -- the statement in the description about the db migration seems a bit "scary" to me, depending on how users have mapped their data in the past. I guess we will have to test this on real user data, but I am wondering if doing a forced migration is the best strategy. 

I might want to have a session with you all to have you go over the implications of this.

PR #1958, opened 2019-09-19
Migration Performance

axelstudios:
Migration 0109 can take 4ish hours to recompute all property/taxlot hashes
#### What's this PR do?
This wraps the expensive operations in atomic transactions to save only 2 transactions instead of millions, and cuts the time by at least half
Run the migrations against a database that hasn't already been migrated to this point

**Note that this will require local developers to start over with their database since we changed the order of the migrations.**


adrian-lara:
Code changes look good - I'll be running these migrations on a local copy of the production database tonight when I can leave my machine running for ~4 hours

PR #1959, opened 2019-09-19
Data Quality Label Removal

axelstudios:
Running data quality checks against existing data could appear to run indefinitely.
#### What's this PR do?
Fixes a bug when running data quality checks against an existing record with a label that doesn't violate a rule that has that same label assigned.  It would attempt to remove the label and fail due to the recent label refactor
@nllong It would be good to create a unit test and remove oep-city from the default deployment
Apply a rule to a record, then create a data quality rule with that same label that the record doesn't violate.  Run data quality checks and it should successfully remove the label.


RDmitchell:
@axelstudios -- thanks for looking into this !!

I guess at some point @mmclark will deploy this fix to production and staging.

Issue #1571, opened 2018-02-09
Update verbiage in dialog box(s) to reflect merging and pairing rather than matching


RDmitchell:
SHA: 68bdd74 (2.3.0 release)

The program currently talks about "matching" in at least the matching results screen. We might want to update the verbiage to reflect our new vocabulary of merging and pairing.

There are probably more instances in the program where this should be updated, but here is an example.

![image](https://user-images.githubusercontent.com/6314950/36046365-80af690a-0d8d-11e8-9195-a77b4ccd6d02.png)

adrian-lara:
@RDmitchell @nllong Looks like this was opened over a year ago. 

Regarding the example provided, I would say that this is message is still relevant - though I have come across a case where "SEED could not locate any existing matches" comes up even though matches were located. I'll address that.

But more generally, I'll scan the whole app to fix inconsistencies between dialog boxes/descriptions/text and how the app currently functions. Let me know if I should address this GH Issue any differently from that.

RDmitchell:
@adrian-lara -- sounds good ! Thanks

RDmitchell:
@nllong -- I am moving this to Test so I can test it once it is on dev1 or wherever.

RDmitchell:
Seems like there were a lot of changes, looking at the PR. I don't think I can thoroughly test them all, so as I do other testing I will try to pay attention to notifications / messages, and if I see any that I think need work, I'll make another issue for them for the Dec work. Thanks Adrian !!

PR #1971, opened 2019-09-30
Update verbiage relating to recent matching changes

adrian-lara:
#### What's this PR do?
- On file import, remove stale statement about matches not being found
- On column settings page, update text detailing the possible ways to change columns.
- For both of the above changes, translate text.

- Review text on Org Column Settings page
- Successfully import a file and see that the **"SEED could not locate any existing matches"** text has been removed.

#1571


seed/static/seed/partials/data_upload_modal.html
```diff
@@ -233,7 +233,7 @@ <h4 class="modal-title" ng-switch-when="16" translate>PM_METER_IMPORT_RESULTS</h
             </div>
             <div class="alert alert-{$ step_10_style $}">
                 <span ng-show="step_10_error_message">{$ step_10_error_message | translate $}</span>
-                <span ng-hide="step_10_error_message" translate="COULD_NOT_LOCATE_MATCHES"></span>
+                <span ng-hide="step_10_error_message" translate="IMPORT_ANOTHER_QUESTION"></span>
```
adrian-lara: I'm not sure what the previous logic was, but I completely removed the "SEED could not locate any existing matches" text. 

Given that the summary in the lines above this include information on whether there were matches, this seemed like a reasonable approach.

adrian-lara:
@nllong This inadvertently adds a bunch of upcoming lokalise translations (from my linking branch). Is that OK?

RDmitchell:
@adrian-lara -- this seems like a good improvement. Thanks !

Issue #1963, opened 2019-09-19
Allow user to specify which fields are to be used for geocoding


nllong:
Currently SEED uses predefined database fields to geocode (e.g., address_line_1, city, state, postal_code), however, not all users are storing their address data in these fields.

This enhancement would allow the user to set up the address fields in the column settings page to be used for geocoding (properties and taxlots).

adrian-lara:
@nllong Should extra_data columns be eligible to be used for geocoding? 

Also, I'm thinking users should be able to specify the order in which their chosen fields are concatenated in order to build the "full address" that'll be geocoded. I'm not sure how this'll be implemented just yet, but I think this is necessary. Let me know your initial reaction, if any.

nllong:
Yes, extra_data should be eligible. Concatenation sounds right.

Nick

adrian-lara:
@nllong or @RDmitchell
I'm wondering how users that don't use SEEDs provided address fields (address_line_1, city, state, etc) store their address data. Is it ever the case that they are using just one field with a "full address"? Or are they still storing address information in multiple fields?

I'm asking because of the following: 
For geocoding, there are cases where MapQuest API gives SEED valid and specific-enough geocoding locations that are potentially "incorrect". For example, something like `100 Market Street` can exist in many cities. MapQuest API can provide one of several valid and specific-enough geolocations that SEED would accept.

In the initial setup of geocoding in SEED, I avoided this by requiring that at least 3 address components were provided before attempting to geocode. With the default geocoding columns being:
- address_line_1
- address_line_2
- city
- state
- postal_code
Requiring 3 of those ensured that at least one of city, state, or postal_code would be used in a valid geocoding result. (Having the 3 be exactly city, state, and postal_code wouldn't return a geolocation with a high enough specificity rating for SEED to accept.)

I wouldn't expect to handle every possible case, so if all remains the same, at the very least, I'll be adding text in the column settings page mentioning this requirement of 3 fields. But if there are common cases that should be addressed (like one would that would prevent geocoding entirely by storing everything in one field), I'm wondering if we need to reconsider how SEED is set up to prevent accepting valid but potentially "incorrect" geolocations.

adrian-lara:
Capturing conversation: There are current cases where users store complete, geocode-able addresses in one field. Requiring 3 or more fields be used for geocoding can't be enforced. A different approach must be taken to address the case of valid but ambiguous cases.

[...]

PR #1986, opened 2019-10-16
Geocoding fields

adrian-lara:
This PR includes migrations.

There are cases where users don't store address information in the provided DB address columns such as `address_line_1`, `city`, `postal_code`, etc.

In those cases, geocoding wouldn't be possible as geocoding was "hardcoded" to use the provided DB addresses.

#### What's this PR do?
Geocoding columns are now customizable. Users can toggle columns to be a part of geocoding as well as choose the order in which the corresponding values are concatenated to construct a full address.

Since there's now no way to pre-filter records that will provide ambiguous geocoding results, say if an address like "101 Market Street, California" provides multiple valid geocoding results, any time multiple geocoding results are provided for an address, the record is not geocoded and `Geocoding Confidence` is populated with "Low - check address (Ambiguous)"

Along with this, text/warnings/advice for geocoding has been added or updated in both the "Geocode Selected" action modal on the inventory list pages and the column mapping page during the import file flow.

With a valid MapQuest API key, update geocoding columns via the column settings page, then import files while using these columns to hold address fields. **Note**: For an organization, when no columns are used for geocoding an inventory type, no geocoding is attempted at all for that org's inventory type. When the org has enough geocoding columns specified for an inventory type, each record must still have at least 1 column with non-empty string or number values.

**Follow these 'with and without API key' steps to view updated flows along with text and advice.**

Without a MapQuest API key attached to an org, 
1. Visit the column mapping page within the import file flow and see that no text is provided regarding geocoding
2. Visit the inventory list pages and see that geocoding isn't possible when you select records and try to "Geocode Selected".

With a MapQuest API key, 
1. Visit the column settings page for an org; view the geocoding advice at the top of the page; and alter the geocoding columns (use DB columns and extra_data columns).
2. Visit the column mapping page within the import file flow, and see the columns you've selected for geocoding as well as text/advice about geocoding.
3. Revisit the column settings page, and remove geocoding columns for any inventory type so that that inventory type has no geocoding columns.
4. Revisit the column mapping page and see the columns are no longer listed for the inventory type that has no geocoding columns. (If you deactivated them from both properties and tax lots, then all the geocoding-related text should be removed.)
5. Visit the inventory list page for the inventory type with no geocoding columns and try to "Geocode Selected", a warning should appear that geocoding isn't possible since there aren't enough columns.

#1963 

**Note**:  In addition to the above changes, I used this PR to...
1. refresh the VCR cassettes for the geocoding tests that actually hit the MapQuest API. See that several .yaml files were updated.
2. create lokalised translations for geocoding-related text that was previously untranslated.
3. addressed relevant "Geocode selected" (from inventory list pages) UI/UX improvements


seed/static/seed/js/controllers/column_settings_controller.js
```diff
@@ -109,18 +160,36 @@ angular.module('BE.seed.controller.column_settings', [])
...
       };
 
-      // Matching Criteria sorting
-      $scope.column_sort = 'default';
-      $scope.matching_criteria_sort = function() {
-        if ($scope.column_sort === 'default') {
-          $scope.columns = _.sortBy($scope.columns, 'is_matching_criteria');
-          $scope.column_sort = 'is_matching_criteria';
-        } else if ($scope.column_sort === 'is_matching_criteria') {
+      // Table Sorting
```
adrian-lara: These changes involved a refactor of the previous matching criteria sorting. The columns are now presorted by ID and this state is considered "default".

seed/migrations/0113_column_geocoding_order.py
```diff
@@ -0,0 +1,43 @@
...
+            'address_line_1',
+            'address_line_2',
+            'city',
+            'state',
+            'postal_code',
+        ]
+
+        # Start at 1 since 0 is considered deactivated.
+        for i, field in enumerate(default_geocoding_fields, start=1):
+            Column.objects.filter(
```
adrian-lara: These filters are an attempt to only target the original, DB address fields associated to each org's inventory types. I messed something like this up last time for default matching fields, so this deserves another pair of eyes.

seed/utils/geocode.py
```diff
@@ -7,6 +7,7 @@
```
adrian-lara: Just noting that changes to this page are the relatively more impactful than other changes in this PR, and they deserve a closer review.

seed/views/organizations.py
```diff
@@ -840,3 +840,42 @@ def matching_criteria_columns(self, request, pk=None):
         )
 
         return JsonResponse(matching_criteria_column_names)
+
+    @api_endpoint_class
+    @ajax_request_class
+    @has_perm_class('requires_member')
+    @detail_route(methods=['GET'])
+    def geocoding_columns(self, request, pk=None):
```
adrian-lara: Used to ensure geocoding columns do exist for an org's inventory type as well as to list the geocoding columns in the column mappings page within the import file flow.

seed/views/geocode.py
```diff
@@ -82,6 +82,10 @@ def confidence_summary(self, request):
                     id__in=tax_lot_ids,
                     geocoding_confidence__startswith='Low'
                 )),
+                'manual': len(TaxLotState.objects.filter(
```
adrian-lara: Part of the fix for manually geocoded taxlots not being noted when "Geocoding Selected" from the inventory list page.

seed/models/columns.py
```diff
@@ -583,6 +583,9 @@ class Column(models.Model):
     import_file = models.ForeignKey('data_importer.ImportFile', blank=True, null=True)
     units_pint = models.CharField(max_length=64, blank=True, null=True)
 
+    # 0 is deactivated. Order used to construct full address.
+    geocoding_order = models.IntegerField(default=0, blank=False)
```
nllong: I'm excited to see that we are starting to more thoroughly use columns! There was a time not too long ago where columns were an abstract idea :)

seed/migrations/0113_column_geocoding_order.py
nllong: this looks right assuming you are adding 1,2,3,4,5 to address_line_1,address_line_2,city,state, postal_code for both taxlot and property fields. You have `is_extra_data` false which ensures that these are the database fields (and not extra data). Looks good!

seed/static/seed/js/controllers/mapping_controller.js
```diff
@@ -312,6 +316,28 @@ angular.module('BE.seed.controller.mapping', [])
...
+
+          $scope.taxlot_geocoding_columns_array = _.map(geocoding_columns.TaxLotState, function (column_name) {
+            return _.find($scope.mappable_taxlot_columns, {column_name: column_name}).display_name;
+          });
+          $scope.taxlot_geocoding_columns = $scope.taxlot_geocoding_columns_array.join(', ');
+        });
+      };
+
+      var org_id = user_service.get_organization().id;
+      geocode_service.check_org_has_api_key(org_id).then(function(result) {
```
nllong: 👍

seed/tests/data/vcr_cassettes/geocode_base_case.yaml
```diff
@@ -2,69 +2,113 @@ interactions:
 - request:
     body: null
     headers:
-      Accept: ['*/*']
-      Accept-Encoding: ['gzip, deflate']
-      Connection: [keep-alive]
-      User-Agent: [python-requests/2.20.0]
+      Accept:
```
nllong: Any reason why this file changes so much?

seed/tests/test_views.py
```diff
@@ -1469,6 +1469,7 @@ def test_get_property_columns(self):
             'column_name': 'pm_property_id',
             'display_name': 'PM Property ID',
             'data_type': 'string',
+            'geocoding_order': 0,
```
nllong: thanks for fixing all the test cases for this change!

seed/utils/geocode.py
```diff
@@ -128,15 +148,18 @@ def _full_address(building):
...
+            address_value = getattr(building, col['column_name'])
+
+        # Only accept non-empty strings or numbers
+        if (isinstance(address_value, (str, Number))) and (address_value != ""):
+            address_components.append(str(address_value))
 
-    if address_components.count("") < 3:
+    if len(address_components) > 2:
         full_address = ", ".join(address_components)
         return re.sub(r'[;/?:@=&"<>#%{}|["^~`\]\\]', '', full_address)
```
nllong: there a test for this one? I'm sure I'll run into it as i'm going through it...

seed/tests/data/vcr_cassettes/geocode_base_case.yaml
adrian-lara: I'm not sure why the format changed. 

I did expect a few of these to change because the date/time, etc of the request is captured, and I did change some addresses behind the requests (effectively completely changing the request &response)

nllong: okay. doesn't seem like this will impact anything. so move along...

seed/utils/geocode.py
adrian-lara: https://github.com/SEED-platform/seed/pull/1986/files#diff-10e49244e46b722e692f9065896cc2faR340

Note: Your comment was coming in on a different commit than the latest commit, but I think you were still referring to the `len(address_components > 0:` version.

Screen Shot 2019-11-04 at 1 00 24 PM
(non-latest comment on the left)

Issue #1979, opened 2019-10-07
Detail View Edit of Custom ID 1 sometimes produces an error


RDmitchell:
### Expected Behavior
Should be able to edit a Custom ID 1 field, even if it has no data
### Actual Behavior 
Get an error, at least on one record, when trying to save a new value in the Custom ID 1 field.

However, on other records with blank Custom ID 1 fields, the program saves the new data without issue. 
### Steps to Reproduce
See this doc
https://docs.google.com/document/d/1JQdvDJHInDS_BVaemqoJ_BA-OaXesdRgPSaNF1DaOQg/edit?usp=sharing
### Instance Information
Instance: LBNL production and staging
SHA: 86255f35

axelstudios:
After quite a bit of testing I was able to narrow down the cause of this error to the minimal-reproducible example:

This error will always occur if you try to edit a new record (never been merged) that came from an import file that has been deleted from the dataset (or the dataset has been deleted).  That obviously shouldn't be the case, so I'll have a fix shortly

nllong:
so an edge case of an edge case

RDmitchell:
@axelstudios -- so maybe for the San Jose record, they should delete the problem record and reimport it? And there may be others that also fall into this category in their database.

axelstudios:
You could delete the records and reimport, but it would be easier to just wait until this fix is merged and deployed

RDmitchell:
@axelstudios -- ok, so it sounds like this fix when deployed will fix existing records. Thx !

nllong:
@RDmitchell -- i already notified you in this PR, but please test on staging. https://github.com/SEED-platform/seed/pull/1987

RDmitchell:
@nllong -- I tested the one record that was an issue, and it now saves edits to the record. I emailed the San Jose person working on this, to see if there were any other records with a similar issue. I will wait a bit to hear from her, but it looks like this is fixed, and I will close it soon. Then the patch can be deployed to production. Thx !

RDmitchell:
I am going to close this. Will make another issue if San Jose reports another issue.

PR #1987, opened 2019-10-17
#1979 - Record edit failure

axelstudios:
Attempting to edit a new record that has never been merged that came from an import file that has since been deleted from the dataset will cause the edit to fail.

#### What's this PR do?
Instead of saving the `import_file_id` from the original PropertyState or TaxlotState on the first edit this change will remove it.  It's not used anyway, and all other actions, like merging, similarly set this field to None when that happens.

1. In an empty organization import a file (empty org just to ensure the records don't merge with anything)
2. Delete the import file from the dataset (or delete the dataset)
3. Attempt to edit any of the imported records - it should now succeed

#1979

### DIFF
--- a/seed/views/properties.py
+++ b/seed/views/properties.py
@@ -1035,6 +1035,9 @@ class PropertyViewSet(GenericViewSet, ProfileIdMixin):
                 if log.name == 'Import Creation':
                     # Add new state by removing the existing ID.
                     property_state_data.pop('id')
+                    # Remove the import_file_id for the first edit of a new record
+                    # If the import file has been deleted and this value remains the serializer won't be valid
+                    property_state_data.pop('import_file')
                     new_property_state_serializer = PropertyStateSerializer(
                         data=property_state_data
                     )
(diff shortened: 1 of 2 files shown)


axelstudios:
@nlong We should chat about why this is happening since the import_file record technically does still exist, but just has the `deleted` boolean set to true

nllong:
don't we want this based off of 2.6.0-patch00 branch?

axelstudios:
> don't we want this based off of 2.6.0-patch00 branch?

Fixed 👍

RDmitchell:
@axelstudios -- so is this something that can be retroactively fixed for the problem record that San Jose has on the LBNL production server? Or is this fix just for going forward with new records?

axelstudios:
This change fixes the behavior for all records, including existing

nllong:
@RDmitchell -- I deployed this to your staging server and is ready for you to test. If it is good, let me know (or Miles), and we can deploy to prod.

RDmitchell:
@nllong -- ok, I will try to test today. Thanks !

RDmitchell:
Tested and closed issue.

Issue #1989, opened 2019-10-17
Unmerging records loses pairing


axelstudios:
### Expected Behavior
Unmerging records should have any paired properties/taxlots to be paired with both resulting records

### Actual Behavior 
The two newly-unmerged records have no pairing

### Steps to Reproduce
1. Import `example-data-taxlots.xlsx`
2. Merge two properties together
3. Go to the detail view of the new property and click `Unmerge Last`
4. Check if the two unmerged records are no longer paired with taxlots

RDmitchell:
@adrian-lara / @axelstudios -- should this be in the Dec 19 project? Seems like it should get fixed before the next release ??

nllong:
yes, this is being address now.

RDmitchell:
Instance: dev1
Org: LBL 404

* Merged 2 property records with different Tax Lot ID pairings together.
* Merged property was linked to both the tax lots of the original 2 properties
* Unmerged the properties (Unmerge Last in Detail view)
* Each unmerged property was still linked to the two tax lots from the merge
* I was able to un pair the tax lots as needed from the unmerged properties to get the records back in their original state. 

This is how I would expect the program to behave.

PR #1995, opened 2019-10-20
#1989 - Unmerge loses pairing

adrian-lara:
Pairings should persist to both records produced from an unmerge. This was not happening, though the logic was in place - the execution wasn't quite right.  Specifically, deleting the original "target" record (the one being unmerged) occurred before the pairing relationships could be captured. 

#### What's this PR do?
Fixes the error above by capturing pairings sooner in the unmerge process.

Unmerge records having pairs and see that each resulting record has the pairs of the "target" record. 

Oddly enough, there is a passing test for this in the test suite, but that test seems to have been giving false positives. 
https://github.com/SEED-platform/seed/blob/200cca45bd5538b988ff9836b8133b993789a383/seed/tests/test_taxlot_views.py#L116
I suspect that this is caused by the automatic use of transactions in the test suite.

#1989

### DIFF
--- a/seed/views/properties.py
+++ b/seed/views/properties.py
@@ -503,6 +503,10 @@ class PropertyViewSet(GenericViewSet, ProfileIdMixin):
                 'message': 'property view with id {} does not exist'.format(pk)
             }
 
+        # Duplicate pairing
+        paired_view_ids = list(TaxLotProperty.objects.filter(property_view_id=old_view.id)
+                               .order_by('taxlot_view_id').values_list('taxlot_view_id', flat=True))
+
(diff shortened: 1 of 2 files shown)


nllong:
should we base this PR off of patch-00 so that it gets on the production system?

RDmitchell:
Not knowing all the implications of @nllong 's comment, I would say Yes, we want this fixed before we move this to production, so including it in a patch for production seems to make sense to me.

axelstudios:
This bug hasn't made it to production yet (just `develop`) so we're probably fine with the current base branch

Issue #1992, opened 2019-10-17
Data Mapping for imported files does not reflect the actual mapping for that data file


RDmitchell:
### Expected Behavior
In Data Sets, for files that have already been imported, mapped, merged, paired, linked, etc, clicking the Data Mapping button should show the mapping for the file that was imported
### Actual Behavior 
However, clicking that button shows the latest mapping for a given field, ie, based on how the latest file was mapped. We need to have an accurate way to see the actual mapping for the original file.
### Steps to Reproduce
* Import a file where all fields are mapped to Tax Lot table
1 - example-data-taxlots.xlsx
Mapping -- all fields mapped to Tax Lot
![image](https://user-images.githubusercontent.com/6314950/67049046-2ed16880-f0ea-11e9-81bc-38eb2a35eea7.png)
* Import another file where all fields are mapped to Property table, where there is one field, in this example "City", that is in both files. 
1 - example-data-ESPM-Alerts-UBID.xlsx
Mapping -- all fields mapped to Property
![image](https://user-images.githubusercontent.com/6314950/67049272-b0c19180-f0ea-11e9-867a-1dba6f2444f6.png)
* Click on the Data Mapping button for the original file where City was mapped to Tax Lot
![image](https://user-images.githubusercontent.com/6314950/67049384-f1b9a600-f0ea-11e9-8751-4037580453a3.png)
* You will see that the City field is shown as being mapped to Property, because that was the last mapping for the field called City. But it was really mapped originally to Tax Lot
![image](https://user-images.githubusercontent.com/6314950/67048791-b4084d80-f0e9-11e9-86bb-de92de50f6b5.png)
 
### Instance Information
Instance: dev1
SHA: 46d759f6

RDmitchell:
Sample files are in the folder for this issue
https://drive.google.com/drive/folders/1oCN5E22NbMC_V4HsWN175lu_n61PEU9a?usp=sharing

adrian-lara:
The "bulk" of the planned solution is to add a new JSON field on the ImportFile model that will be populated with the given mapping at the time of import.

@nllong let me know what you think about backfilling this field. I'm thinking these should be empty as backfilling is impossible and would be misleading if it weren't the actual mappings used at the time.

RDmitchell:
I think backfilling this field (from previously imported files) would be tricky so I agree that we probably shouldn't do it.

adrian-lara:
Turns out historical data was being captured on each import file. 

[...]

PR #2002, opened 2019-10-28
ImportFile's post-matching mapping page shows historical mapping

adrian-lara:
See issue linked below. The approach initially planned in the issue of creating a new ImportFile field to capture mapping settings during initial import was not needed as there was already a field capturing that info.

#### What's this PR do?
This PR uses ImportFile objects' `cached_mapped_columns` field to render the column mappings by user. 

**Very Notable Caveat**
The cached mappings makes use of raw, DB-level column names, so the current Column Display Name Settings are used to present cached mappings in a way they're accustomed to. For example, `PM Property ID` is shown instead of `pm_property_id`. The caveat here is that users can change Column Display Name Settings after the fact.

If this were to happen, technically, the value in SEED Header might not be _exactly_ what the user originally entered during import. **But** they will always get a reference to the DB-level column that the data file header was mapped to. 

The alternative is to provide the original text input as it was when the user imported the file.  This could be misleading or, worse, represent a completely different DB-level column that the data file header was mapped to on import.  

For example, if users change the display name of the DB column `custom_id_1` to `Some Unique ID` but they've previously imported files to `custom_id_1` when the display name was the default `Custom ID 1`. Users would need historical context to know that the data file header on these previously imported files are mapped to the column they currently understand as `Some Unique ID`.  

In this example, the current set up would show `Some Unique ID` as the display name instead of `Custom ID 1` that the user originally used on import.

Visit the mapping page of different ImportFiles and see the previously used mappings are used. This includes unit choices where applicable.

Also, import more files with the same columns as existed in previously imported files and map these data file columns to different SEED columns OR change column mappings for the Org. Then, verify that previously used mapping persist.

#1992


seed/static/seed/js/controllers/mapping_controller.js
```diff
@@ -567,9 +567,31 @@ angular.module('BE.seed.controller.mapping', [])
...
+          // If available, use display_name, else use raw field name.
+          var mappable_column = _.find(
+            $scope.mappable_property_columns.concat($scope.mappable_taxlot_columns),
+            {column_name: cached_col.to_field, table_name: cached_col.to_table_name}
+          )
+          if (mappable_column) {
+            col.suggestion = mappable_column.display_name;
+          } else {
+            col.suggestion = cached_col.to_field;
+          }
```
adrian-lara: The only time the `else` case would be used is if a Column was deleted. As I understand it, this should never never happen as users don't have the ability to delete Columns, but the fail safe was put in place just in case. Let me know if it actually is possible for users to delete Columns for an org.

nllong: Yes, at the moment, users can't delete columns. So I think this is okay.

RDmitchell:
The real test is to map the same field in two different imports, one to Property and then the next import map the same field to Tax Lot. Then if you look at this history of the Property import (the first one), it used to show the table for the latest import, in this case Tax Lot. Hopefully now it would show the original mapping, for the original file import, of Property.

I will test this case when this fix is on dev1. 

Thx

Issue #1845, opened 2019-03-20
DQ Checking hanging in mapping due (possibly) to units issue


RDmitchell:
### Expected Behavior
In the Mapping Review screen, I expected the DQ Results to finish and return the results in a modal.
### Actual Behavior 
The spinner kept spinning and DQ results never materialized. Also seemed to generate some Sentry errors, possibly about a units issue. 
### Steps to Reproduce
See this google doc 
https://docs.google.com/document/d/1myTiAKcLWzemNtgclQRfKRcc-8ssSL_dVO_SiqLo9AM/edit?usp=sharing
### Instance Information
Server Instance: dev1
Server SHA: 0469a1dc

RDmitchell:
@nllong / @axelstudios -- turns out the field Site EUI did not have the units set in the DQ rules. After I fixed that, the DQ results worked fine in mapping.

We should add some error checking during the DQ checking, to let the user know if there is a problem like this, ie, units not set (and whatever other cases we can come up with that would cause the DQ checking to hang).

RDmitchell:
see also #1817

axelstudios:
I wasn't able to reproduce this with the units, but I strongly suspect it's related to #1873

RDmitchell:
Instance: dev1
SHA: 6dad9f771
Org: LBNL Test 200
Cycle: 2018 Compliance Cycle
1 - example-data-ESPM-Alerts-UBID.xlsx

Program is able to do the DQ checking even if units are not set for the fields in the DQ rules.

Issue #1817, opened 2019-02-12
DQ: Program "hangs" when DQ rules can't be applied


RDmitchell:
### Expected Behavior
Ability to run the DQ rules from the Inventory List
### Actual Behavior 
When trying to run the DQ rules from the Inventory List, the program "hangs", ie, the spinner spins forever and never finishes, no errors in the Developer tools (Chrome), etc. When I "uncheck" rules that are probably invalid for this data set, the program can run the DQ checks. Seems like somewhere in the process, the program should let the user know that their DQ rules are valid and need to be corrected. 
### Steps to Reproduce
see the doc in this issue folder
https://drive.google.com/drive/folders/1OrPV9JncoqNhXeySgR9kYpFxZLtRIpVv?usp=sharing
### Instance Information
Instance: staging (LBNL)
SHA: 214f0f5d

RDmitchell:
see also #1845

axelstudios:
After some testing, this has the same error reported in celery as #1873

RDmitchell:
Instance: dev1
SHA: 6dad9f771

The program now seems to be able to apply the DQ checks, ie, it doesn't hang anymore. 

However, one of the results doesn't make sense to me but I will put that on a new issue and close this one.

Issue #1375, opened 2017-06-05
DQ Check - Save makes labels disappear on Admin Screen


RDmitchell:
instance: staging
SHA: bc91031
org: LBNL Test 200

Defined 3 new rules with labels
![image](https://cloud.githubusercontent.com/assets/6314950/26805380/88ede1f6-4a00-11e7-87c7-3d6ab73c36f1.png)

Clicked Save changes, first two labels disappeared.

Left screen and came back and the labels were there. 

Now can't reproduce on the same org or rules.

Should be tested on completely new rules.

RDmitchell:
Just verified that if you add a label to a rule, by defining the label "on the fly" in DQ Admin, when you save the label is no longer displayed. But it is actually defined, because when you refresh the screen it is there.

RDmitchell:
Instance: dev1
SHA: 6dad9f771
Org: LBNL Test 200

Fixed !

Issue #1511, opened 2017-11-22
DQ results: Program doesn't tell user that a rule can't be applied due to data definition


RDmitchell:
instance: seeddemostaging
SHA:
Org: LBNL 22
Cycle: Default 2016 Calendar Year
covered-buildings-sample.csv

I set up this DQ check -- I want the program to report any records that have the Tax Lot Gross Floor Area = blank; but I haven't entered any minimum and maximum values
![image](https://user-images.githubusercontent.com/6314950/33105612-da7f7202-cee2-11e7-9e86-4deb2f2fc186.png)

When I run this DQ check on the View by Tax Lot table (all the fields are mapped to tax lot, I was hoping it would show the blank records.
![image](https://user-images.githubusercontent.com/6314950/33105651-1c85b31e-cee3-11e7-8711-5a5ae15384b4.png)

But when I select all the records in that tab, and then select the Data Quality Check Action pulldown
![image](https://user-images.githubusercontent.com/6314950/33105678-4344995c-cee3-11e7-9ec9-1b75ecc7d1f2.png)

The program doesn't report any errors
![image](https://user-images.githubusercontent.com/6314950/33105704-5e19a1be-cee3-11e7-9de2-7e5ec53603f1.png)

RDmitchell:
If I add criteria for the min and max, still no results
![image](https://user-images.githubusercontent.com/6314950/33105810-fb3723a4-cee3-11e7-9660-81dfd3a7ffa4.png)

**And it's because the field, Gross Floor Area, isn't a number, ie, I can't do the query I want to do as a filter in the Inventory List (which I didn't first check before setting up the DQ results).**

Seems like we need to do the following:
* User shouldn't enter the field type in DQ setup -- the program knows what it is and it should automatically set it
* If the type of DQ check that the user has set up can't be done on the field type (ie, a numeric query on a string field) the program should tell the user that (maybe when they save the DQ setup ??)
* But in the case where a minimum and maximum aren't set, and the user just wants the program to check to see if there are any blank "null" data in the field, the program should be able to do that even on a string ... ?

RDmitchell:
![image](https://user-images.githubusercontent.com/6314950/33106278-3c58f8d8-cee6-11e7-85c5-32a08af3ed0f.png)

RDmitchell:
Instance: dev1
Org: LBNL 22
Fixed !
[...]

PR #2004, opened 2019-11-05
Handle type casting in data quality checks

nllong:
The typecasting within the data quality checks were not handling the case where a string was defined as a number and has a min/max. This caused the data quality checks to fail ungracefully preventing future checks to occur

#### What's this PR do?
* Update the minimum_valid and maximum_valid checks to correctly handle an invalid typecasting call.
* Add tests

Test 1:
* Modify the data quality rule of a text non-extra-data field (e.g. Address Line 1) so that the rule data type is set to Number, Date, Year, Area, or EUI and it has a minimum and/or maximum value set
* Upload the file and proceed to the Mapping Review screen where data quality results are loaded

Test 2:
* Set an EUI min and max with the unit being either Number or EUI.
* Verify that the DQ's run and report issues appropriately.

Test 3:
* Set the units of the site EUI field to feet squared. This makes no sense, but SEED needs to handle this nicely
* Run a DQ on a record and verify that the DQ results in a warning about mismatched dimensions.

#1817 
#1845 
#1873 
#1375 (alex)


nllong:
This may have been resolved with this PR as well: https://github.com/SEED-platform/seed/issues/1511

axelstudios:
@nllong The tests all look good, and I've added a couple of commits with further improvements (see commit messages for details)

That said I'm seeing some really weird behavior when I go back and forth from the Mapping screen to the Mapping Review screen (to review data quality results).  When I change what field I'm mapping to frequently the field disappears entirely, even when switching it between properties/taxlots

nllong:
I think the weird behavior of the data clearing out is as expected. The results of cleansing is stored in redis and I know that we’ve had issues with data not clearing out and getting redundant results. The other reason could be the cache/redis is timing out. 

I’ll test your last changes and merge if all good. Unless you think the issue is a showstopper.

axelstudios:
Test my changes and then we can probably merge this.  The other problem isn't strictly data quality, so it can be a new issue.  I definitely think we should be able to review mapping results, go back and change something, and have it continue to work

nllong:
Yeah it should continue to work, but it would require rerunning the mapping results.

axelstudios:
Hmm, I didn't know it wasn't doing that.  It's definitely re-running data quality checks each time (I'm using always-eager fwiw)

Issue #1480, opened 2017-10-12
Five digit zip codes with leading 0


ClearlyEnergy:
Five digit zip codes with a leading 0, for example 05055 in Vermont are often mangled by excel into four digits numbers. If they are loaded into SEED as four digit numbers, they are not corrected back to five, which leads to erroneous zip codes. This is really an excel problem, but it should be an easy check/fix when loading data into SEED.

RDmitchell:
@nlong -- Has this been fixed?

nllong:
No this still needs to get resolved. Can you add this to the June project?

Il giorno lun 17 giu 2019 alle 17:00 RDmitchell <notifications@github.com>
ha scritto:

RDmitchell:
Added to To Do column of June 2019 project

RDmitchell:
@nllong / @lainsworth8801 -- do we have an example file for testing? I can't figure out if I should have a 5 digit zip with a leading zero or just a 4 digit zip in a test file.

lainsworth8801:
There’s a example-data-properties-postal.xslx under data_importer/tests/data. Should have all valid combos for within the US

RDmitchell:
Instance: dev1
SHA: 
Org: LBNL 200
Cycle: Test zip code

Spreadsheet postal code field data looks like this (ugh !)
![image](https://user-images.githubusercontent.com/6314950/71302198-281ad800-235d-11ea-9968-d8406a79ea6c.png)

SEED interpreted the data like this -- very nice !!
![image](https://user-images.githubusercontent.com/6314950/71302192-1a655280-235d-11ea-87af-11b21846f328.png)

PR #2012, opened 2019-11-12
Zip

lainsworth8801:
Issue #1480 
Added leading zeros for postal codes under 5 digits;
Added leading zeros for postal codes with 4 digits extension;
Added test excel sheet with postal codes added (fake column name for user mapping purpose).

original uploaded:
Screen Shot 2019-11-12 at 10 29 31 AM

| Map Data   | Property Saved  |
|------------|---|
| Screen Shot 2019-11-12 at 10 30 01 AM  | Screen Shot 2019-11-12 at 10 36 41 AM |


nllong:
@lainsworth8801 -- thanks for the PR. Can you add in the issue ID into the PR description?

seed/data_importer/tests/integration/test_data_import.py
```diff
@@ -389,3 +389,92 @@ def import_exported_data(self, filename):
...
+        if ts is None:
+            raise TypeError("Invalid Taxlot Address!")
+
+        # regular or w/ ext case:
+        for row in range(self.parser.sheet.nrows):
+            if self.parser.sheet.cell(row, col_addr).value == '93029 Wellington Blvd':
+                if '-' in ts.postal_code:
+                    self.assertEqual(len(ts.postal_code.split('-')[0]), 5)
+                    self.assertEqual(len(ts.postal_code.split('-')[1]), 4)
+                    self.assertEqual(ts.postal_code.split('-')[0].lstrip('0'),
```
nllong: making these explicit checks would be nice, (i.e., something like `self.assertEqual(ts.postal_code.split('-')[0].lstrip('0'), '05005')`)

seed/lib/mcm/mapper.py
```diff
@@ -99,6 +99,15 @@ def apply_column_value(raw_column_name, column_value, model, mapping, is_extra_d
     if raw_column_name in mapping:
         table_name, mapped_column_name, display_name, is_extra_data = mapping.get(raw_column_name)
 
+        # special postal case:
+        if mapped_column_name == 'postal_code':
```
nllong: looks good.

seed/data_importer/tests/integration/test_data_import.py
lainsworth8801: I can change that. Was thinking about doing a general format check instead of spot checking.

nllong:
That would be nice. Thank you. It allows the developer to see what it is expecting instead of having to .split and .zfill.

Da: Lin Ainsworth <notifications@github.com>
Risposta: SEED-platform/seed <reply@reply.github.com>
Data: lunedì novembre 18 2019 09:34
A: SEED-platform/seed <seed@noreply.github.com>
Cc: Nicholas Long <Nicholas.Long@nrel.gov>, Review requested <review_requested@noreply.github.com>
Oggetto: Re: [SEED-platform/seed] Zip (#2012)


@lainsworth8801 commented on this pull request.

________________________________

In seed/data_importer/tests/integration/test_data_import.py<https://github.com/SEED-platform/seed/pull/2012#discussion_r347483104>:

> +                col_addr = col

+            elif self.parser.sheet.cell_value(0, col) == 'postal code':

+                col_postal = col

+

+        # get mapped taxlot postal_code

+        ts = TaxLotState.objects.filter(address_line_1='93029 Wellington Blvd').first()

+        if ts is None:

+            raise TypeError("Invalid Taxlot Address!")

+

+        # regular or w/ ext case:

+        for row in range(self.parser.sheet.nrows):

+            if self.parser.sheet.cell(row, col_addr).value == '93029 Wellington Blvd':

+                if '-' in ts.postal_code:

+                    self.assertEqual(len(ts.postal_code.split('-')[0]), 5)

+                    self.assertEqual(len(ts.postal_code.split('-')[1]), 4)

+                    self.assertEqual(ts.postal_code.split('-')[0].lstrip('0'),

I can change that. Was thinking about doing a general format check instead of spot checking.

seed/lib/mcm/mapper.py
```diff
@@ -99,8 +99,8 @@ def apply_column_value(raw_column_name, column_value, model, mapping, is_extra_d
     if raw_column_name in mapping:
         table_name, mapped_column_name, display_name, is_extra_data = mapping.get(raw_column_name)
 
-        # special postal case:
-        if mapped_column_name == 'postal_code':
+        # special postal case
+        if mapped_column_name in ['postal_code', 'owner_postal_code']:
             if column_value:
                 if '-' in str(column_value):
                     postal = str(column_value).split('-')[0].zfill(5)
```
lainsworth8801: Thanks for catching the owner_postal_code.

seed/data_importer/tests/integration/test_data_import.py
lainsworth8801: For whether '0' zip should stay '0' or be packed with 0s, my reason of packing it with 0s is to show users that this is supposed to be a zip code but it's 0. Although I'm not sure if this is a better idea than leaving it as it is which is just a single digit 0.

nllong: Yeah, I think we want this to be packed, that is 0 -> 00000. Right now it doesn't appear to do that.

seed/lib/mcm/mapper.py
lainsworth8801: @nllong I recalled wrong...the reason it's not packing 0 zip with 0s is because i checked column_value here before I convert. Will change that.

adrian-lara:
> Also, do you think we should go back and update all the zip codes in the database with a migration (and rehash)? I think this is mostly a question for @adrian-lara, @axelstudios and/or @ClearlyEnergy.

I think a migration to update zip codes on pre-existing records and recalculating `hash_object` would probably be needed.

seed/lib/mcm/mapper.py
```diff
@@ -101,12 +101,11 @@ def apply_column_value(raw_column_name, column_value, model, mapping, is_extra_d
 
         # special postal case:
         if mapped_column_name == 'postal_code':
-            if column_value:
-                if '-' in str(column_value):
-                    postal = str(column_value).split('-')[0].zfill(5)
-                    ext = str(column_value).split('-')[1].zfill(4)
-                    column_value = postal + '-' + ext
-                column_value = str(column_value).zfill(5)
+            if '-' in str(column_value):
```
lainsworth8801: Screen Shot 2019-11-18 at 11 04 44 AM

axelstudios:
Yeah, I like the idea of a migration to fix existing data and then rehash

RDmitchell:
@axelstudios -- I would like testing the migration on existing data to be something that is in the Test column of the Dec 19 Github project. Not sure how to do that, but if a PR comes through for that migration, I will add it to the Test column.

nllong:
@lainsworth8801 -- can you add in a migration for this. See this as an example: https://github.com/SEED-platform/seed/blob/develop/seed/migrations/0111_rehash.py

nllong:
> @lainsworth8801 -- can you add in a migration for this? See this as an example: https://github.com/SEED-platform/seed/blob/develop/seed/migrations/0111_rehash.py
> Thanks!

Hey @lainsworth8801 -- can you implement this rehashing migration before we merge this PR?

lainsworth8801:
@nllong yes sorry was gonna work on this next

lainsworth8801:
I only ran it on my local db which is a very small size db. i'll add a run on the production.

nllong:
> I only ran it on my local db which is a very small size db. i'll add a run on the production.

Thanks. Mainly concerned because you are updating the postal codes [here](https://github.com/SEED-platform/seed/pull/2012/files#diff-c331009bf9a69bc088380eea1bea3241R19-R24) and want to make sure that it works. What if postal code is a number/string/double/etc.

adrian-lara:
> I think a migration to update zip codes on pre-existing records and recalculating `hash_object` would probably be needed.

Back tracking on this previous comment after conversations with Lin and a closer look at the proposed migration.  I don't think an additional rehashing is needed after updating the `postal_code` values since the `.save()` is being run for each -State object. SEED already triggers an update of hashes given this command.

@nllong @axelstudios What do you guys think?

nllong:
I think you are right @adrian-lara. Since we are calling .save(), then we only need to find the records that need to be updated. I think we should run with SQL (or Django)... something like:

```
from django.db.models import Q
PropertyState.objects.filter(Q(postal_code__iregex=r'\b\d{4,5}-\d{1,4}\b/') | Q(postal_code__iregex='\b\d{4,5}\b'))
```

I did not test the above command... but this would save iterating over all the records.

seed/migrations/0114_rehash_postal_code.py
```diff
@@ -0,0 +1,96 @@
...
+        field.owner_postal_code = zero_fill(field.owner_postal_code)
+        field.save()
+
+    for field in ts.objects.all():
+        field.postal_code = zero_fill(field.postal_code)
+        field.save()
+
+    property_sql = (
+        "UPDATE seed_propertystate " +
+        "SET created = seed_propertyauditlog.created, updated = seed_propertyauditlog.created " +
```
adrian-lara: I think we'd only need to restore the `created` field and not the `updated` field here

adrian-lara:
> I think you are right @adrian-lara. Since we are calling .save(), then we only need to find the records that need to be updated. I think we should run with SQL (or Django)... something like:
> 
> ```
> from django.db.models import Q
> PropertyState.objects.filter(Q(postal_code__iregex=r'\b\d{4,5}-\d{1,4}\b/') | Q(postal_code__iregex='\b\d{4,5}\b'))
> ```
> 
> I did not test the above command... but this would save iterating over all the records.

Yeah something like that would help decrease the number of records getting updated.

seed/migrations/0114_rehash_postal_code.py
nllong: I actually think we don't need to SET the `created` nor `updated` field in either of these... just the `postal_code`, right?

adrian-lara: I updated my comment (meant to switch `created` and `updated`). I'd think we'd need to "revert" the `updated` field since changing postal_codes one at a time using `save()` will change the `updated` time for each record.

adrian-lara: I guess, I'm just assuming we wouldn't actually want the `updated` times to change, but I could definitely understand wanting those values to change.

nllong: Ah, that makes sense. I think it is okay if the updated times change as it really is a change.

nllong:
@lainsworth8801 -- did you have a chance to update this rehash method to filter down to only the fields that are going to be updated? Let me know if you can get to it otherwise I'll work on it.

lainsworth8801:
@nllong was trying to fix the Xcode version issue for timescales on Friday to work on this but didn’t get far. I’ll be back on Monday it could wait through the weekend. My apologies.

nllong:
@lainsworth8801 I just updated the migration to fix the dependency error. I think this PR is really close can you update the migration to only find records that have 4 zip code characters and update only those records? As Adrian points out you only need to run .save() to have the hash get updated.

Note that you probably will need to recreate your local database since I merged in some other db migrations.

lainsworth8801:
@nllong definitely. Thanks nick.

Issue #1759, opened 2018-12-21
Save mappings for specific input file data sources (ESPM, Tax Lot, Building List)


RDmitchell:
### Expected Behavior
Implement saved mappings for defined input file data sources (PM, Tax Lot, Building List, etc) that the user would have control over through the UI. In other words, the user could define a mapping, save it with a name, and associate it with a file type (such as ESPM), if it was a file type that SEED knew about (from the import screen). 

Alternatively, the user could select from the list of saved mappings when importing a specific file. They could have a saved mapping for their Tax Lot files and their Building List files, and when importing a specific file, they could choose which mapping to use. 
### Actual Behavior 
n/a
### Steps to Reproduce
n/a
### Instance Information
n/a

nllong:
Make sure to address the PM changing field names. That is, we need to version the PM field names as part of this work.

RDmitchell:
look at #201 to make sure this addresses the concerns in that issue.

adrian-lara:
@RDmitchell and @nllong 

I'm hoping to identify an acceptable import file flow should be in terms of column mapping. Would the following be an acceptable user workflow in the column mapping page? 
> On the column mapping page of the import flow, there's a dropdown for the user to select a user-defined column mapping setting. Changing this option will require a confirmation click as current mapping "best guesses" by SEED or any of the user changes will be overwritten.
> 
> Once the column mapping setting is chosen, SEED populates the values according the mapping wherever there are exact matches between the setting and the import file column names. For anything that's not an exact match, SEED goes through it's "best guess" algorithm.
> 
> From here the process is similar to before where the user can make changes before continuing.
> They could also change the column mapping settings they chose to start.

adrian-lara:
Per conversations, I will be coming up with more details and mocking out the above-mentioned flow. I'll provide more details on the proposed changes soon in hopes of getting feedback from both of you.

RDmitchell:
@adrian-lara -- this issue represents the mapping presets implementation, right?

adrian-lara:
Yes - that's right. 

I think you may have seen and used this feature already, but just in case, we've replaced the contents of the organization column mapping setting page with a page that allows for the management of the presets. So that settings page and the actual import mapping page should be tested here.

RDmitchell:
@adrian-lara -- where is the new "page that allows for the management of the presets" -- I have seen in on the mapping page, but is it also somewhere else?

adrian-lara:
[...]

Issue #94, opened 2015-01-22
Document how mapping works in the SEED-Platform / seed wiki


RDmitchell:
We need to understand how mapping works in order to fix some of the issues related to that functionality. Right now, mapping a Portfolio Manager file is relatively arduous for the user, because there are usually a lot of fields to map and the program doesn't map all of them correctly. Also, because theoretically we know all the PM field name possibilities, we should be able to set the program up to get the mapping 100% right without the user doing anything. 

In order to understand how mapping works so that we can develop solutions to make it less painful to map files, in particular Portfolio Manager files, we need to answer (at least) these questions:

1 - In a "virgin" account, where no data has been loaded or mapped, when the first Portfolio Manager file is mapped, how is the program assigning the SEED fields ? Is there a "hard coded" list of field mappings (or are they stored in the database?), or is the program using a matching algorithm to determine what data fields map to the SEED fields?

2 - When the first file has been mapped, if the user has made any changes to the original, suggested mappings, the program remembers that mapping for future data uploads. Where is that new mapping stored, and how is that the same or different from how the initial mapping suggestions are stored.

3 - For a file format such as Portfolio Manager, we should be able to determine the exact mappings to the SEED fields, because we can determine all the PM fields that the program will encounter. If we had such a "hard coded" mapping between SEED and PM files, how hard would it be to implement, give how the program currently works? Would it involve a lot of changes, or would it be relatively straightforward.

4 - Based on work with BE last summer, if we want to "clear out" the field mapping list and start over with just the original suggestions, ie, all the new, remembered field mappings would be deleted, it seemed like that data base had to be deleted and recreated. Understand why this is, so that we can figure out another solution, as this one currently does not work. 

I will make a separate google doc, stored in the issue directory for this issue, which shows with example files, some of the issues surrounding mapping. I will update this issue when that doc is ready.

RDmitchell:
Here is the link for the google doc that shows examples of these mapping issues
https://docs.google.com/a/lbl.gov/document/d/1PsC6NIhgY7sFXcVfV2pfdyu9Cztnvj4qGQa7-ZTzx3c/edit?usp=sharing

I am still editing it, but I believe there is enough information there to at least start on question # 1. I have put the sample files referenced in it in the issue gdoc folder

RDmitchell:
There is the issue that some of the PM files we have gotten from cities have underscores and some do not. There are examples (from 2 cities) in the gdrive directory for each of these.

[...]

PR #2022, opened 2019-12-02
1759 mapping presets

adrian-lara:
This PR is ready for review, but note that once new user advice/text has been reviewed, **these texts need to be translated via lokalise**. The DO NOT MERGE label should probably be applied but I was concerned that this wouldn't be passed over for review because of it

**This PR contains migrations.**

#### Considerations/Notes:
- As a side-effect, this PR removes the old "incorrect" mapping history issue 1992 (not linked as it's not solved here). It is solved by this PR, #2002, which should both be deployed together.
- SEED-generated mapping suggestions and the corresponding flows (i.e. 'Upload Portfolio Manage data' button) are not being used anymore. The logic to generate and use these remain in the app for now in case parts of it do want to be used. This includes the ColumnMapping model which I recommend be deprecated once users have fully transitioned to using presets.

Users wanted a greater level of consistency, customization, and reliability when mapping import file columns to SEED columns. Allowing them to maintain and apply their own presets is an attempt to provide this.

#### What's this PR do?
Remove old Column Mapping concept where a single set of Column Mappings is kept and continually updated for an organization by their most recent imports. This concept was replaced by the concept of Column Mapping **Presets**, where users of an org can define, maintain, and apply their own set of presets.

Each organization is given 2 default presets:
1. A snapshot of the current (old) Column Mappings (if applicable).
2. Default PM column mappings as defined in SEED (using https://github.com/SEED-platform/seed/blob/develop/seed/lib/mappings/data/pm-mapping.json)

Also, "Remove Column Mapping" button on Admin page is removed.

CRUD mapping presets in the old and now repurposed Column Mapping page.
CRU (not delete) mapping presets in the File Import process.
Go through any flows that users might encounter involving the above two.

#1759 
#94 - removed no longer applicable documentation


adrian-lara:
> Is it possible for the program to use the previous 'find suggested column mappings' when we paste in the headers into the box? Or did that functionality get completely removed?
> 
> image
> 
> That is use this method:
> 
> https://github.com/SEED-platform/seed/blob/32bf352bf090669ab76f0f8df76df44ba6ef3560/seed/lib/mcm/mapper.py#L26
> 
> .

**Did you envision the suggestions would pre-fill the SEED headers once "Copy into Data File headers" is clicked? or did you have something else in mind?**

The code still exists, but there's a few things to consider that could make it work slightly different. Finding suggested column mappings can be summarized by these lines of proposed deleted documentation: https://github.com/SEED-platform/seed/pull/2022/files#diff-df98e7e0427152ab152ef4cf76362fccL31-L37

**TL;DR - possible that only step 2c will be run.**

As of now, previous mappings are still in the app, but I was going to propose their removal once users have fully transitioned to the presets. If we wanted completely remove them from the app, that first step, 2a, would be skipped.

Also, step 2b, was only "triggered" when the PM specific upload was happening during the old file import process. I'm not sure if we'd want this triggered all the time? Or skipped completely? Or given it's own button on this page?

I'm happy to implement the suggestions into this page, but wanted to get your thoughts/reactions to the above.

nllong:
What I’m thinking is to provide a button on the mapping profile page where the user can click and get the best matches based on all the column mappings that exist across all of their profiles. This includes best mapping to the known database fields. The reason why I want something like this is to allow the user to quickly get up and running when importing a new file... they still need to check the mappings, but at least they’ll have a good start. Does this make sense?

adrian-lara:
Yeah that makes sense. Just to be sure, you're saying you essentially want to get suggestions based on only the previous step 2c? `...matches between raw and existing database columns using Jaro-Winkler distance...[for priority]`

nllong:
> Yeah that makes sense. Just to be sure, you're saying you essentially want to get suggestions based on only the previous step 2c? `...matches between raw and existing database columns using Jaro-Winkler distance...[for priority]`

yep, exactly. And make it a button that the user clicks to return "best known matches".

adrian-lara:
@nllong I definitely plan on adding the button to the Column Mapping Settings page (org settings), and am thinking it can't hurt to have on the actual import mapping page as well. 

Just wanted to confirm, that what you had in mind? Or would it somehow be "a little too much" customizability for the actual import mapping page maybe since that page was more for tweaks?

adrian-lara:
> @nllong I definitely plan on adding the button to the Column Mapping Settings page (org settings), and am thinking it can't hurt to have on the actual import mapping page as well.
> 
> Just wanted to confirm, that what you had in mind? Or would it somehow be "a little too much" customizability for the actual import mapping page maybe since that page was more for tweaks?

@nllong following up on this and sharing the current iteration of the button. Let me know your thoughts on both. Thanks! 😄 
Screen Shot 2019-12-18 at 12 19 18 PM

nllong:
> > @nllong I definitely plan on adding the button to the Column Mapping Settings page (org settings), and am thinking it can't hurt to have on the actual import mapping page as well.
> > Just wanted to confirm, that what you had in mind? Or would it somehow be "a little too much" customizability for the actual import mapping page maybe since that page was more for tweaks?
> 
> @nllong following up on this and sharing the current iteration of the button. Let me know your thoughts on both. Thanks! 😄
> Screen Shot 2019-12-18 at 12 19 18 PM

I think the screen shot. This is exactly what I had in mind.

seed/migrations/0116_columnmappingpreset.py
```diff
@@ -71,7 +71,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ('orgs', '0011_auto_20190714_2159'),
-        ('seed', '0114_auto_20191211_0958'),
+        ('seed', '0115_rehash_postal_code'),
```
nllong: nice catch!

nllong:
i think the translations are ready too. verbiage seems reasonable.

nllong:
@adrian-lara can you update the django.mo files again, sorry 😢

Issue #1977, opened 2019-10-07
Review and update post-import file summary


adrian-lara:
### Expected Behavior
After successful file import, a summary of the number of files imported should be provided.

### Actual Behavior 
At the very least, a summary of matching files is missing.

Other import outcomes (non-matching, successful & unsuccessful geocoding, etc) situations should be reviewed as well.

### Steps to Reproduce
Import 2 slightly different versions of a file that would cause match-merges:
<img width="872" alt="Screen Shot 2019-10-06 at 7 58 38 PM" src="https://user-images.githubusercontent.com/30608004/66280608-c0ccac00-e874-11e9-9b33-b82d7aaab9c9.png">

See the import summary and see there is no indication of matching occurring:
<img width="739" alt="Screen Shot 2019-10-06 at 7 55 58 PM" src="https://user-images.githubusercontent.com/30608004/66280656-ed80c380-e874-11e9-98cc-e6f08ce70faf.png">

### Instance Information
local deployment of `develop`

adrian-lara:
@nllong and @axelstudios After more research, I think changes should be made directly to the information provided to the frontend and/or, at the very least, the backend processes that provide that information. 

See the following sheet for more details. **If you're both short on time, the most important piece I'm hoping for feedback on is the question in red.** Of course, any feedback on my proposed changes or notes/comments is appreciated.
https://docs.google.com/spreadsheets/d/1x9UEVGu38KWfrQvvISKw77pVMXlWxMGj_Xoam3iT_BA/edit#gid=0

RDmitchell:
Results summary seems much more informative, so these were good improvements. 
![image](https://user-images.githubusercontent.com/6314950/71128468-1b9e5000-21a2-11ea-808d-df5bf2225203.png)

RDmitchell:
Another example
![image](https://user-images.githubusercontent.com/6314950/71129854-48079b80-21a5-11ea-9f67-9217c1e5314c.png)

Issue #656, opened 2016-01-12
Add filename to modal reporting matches when there are no matches


RDmitchell:
At the end of the matching step, the program reports the number of non-duplicate and duplicate matches. This is the modal that appears when you import the first file (no other files exist to match to, so it is just reporting on matches within the file being loaded)

We should add the file name that the program was working on to the modal

![image](https://cloud.githubusercontent.com/assets/6314950/12277524/a1c8d512-b930-11e5-924d-263213b65f92.png)

RDmitchell:
This would still be a good thing to add -- most of the other screens show the filename.

RDmitchell:
Instance: staging (LBNL)
SHA: 36cc59d (2.3.1 release)

This is still an issue -- the filename is reported on all the other mapping screens, just not the final results screen for the matching.

RDmitchell:
The results modal now shows the filename in the "title"

![image](https://user-images.githubusercontent.com/6314950/71125939-c01d9380-219c-11ea-802e-becc23aacec3.png)

PR #2026, opened 2019-12-05
Post import summary cleanup

adrian-lara:
Import results didn't seem to be coming through during the post import summary, especially for matching record counts. Upon more research, I discovered that, while building out the recent matching changes, I didn't fully understand and update the results of the import. 

Also, the file name wasn't showing up when there weren't any matches.

#### What's this PR do?
The results needed to be updated to accommodate for all the possibilities on import. The new possibility was that merges could happen between existing records on import of new records.

Along with that, the variable names were made more descriptive and consistent upon initializing and transferring between BE and FE code.

Lastly, the filename is added to modal reporting matches when there are no matches.

After an import without matches, check that file name appears.

Import records with different combinations of the following:
- with and without matches...
- with and without duplicates..
- within the file being imported
- against existing records
- between existing records (records that were auto merged but were then manually unmerged and unchanged)
Verify the counts during these imports.

Here are the list of possibilities with notes next to some:
- New properties  _(includes records that were in-import-file merges as well)_
- New tax lots  _(includes records that were in-import-file merges as well)_
- Merged properties within import file _(can be included in 'New properties' or even 'Merged properties against existing records')_
- Merged tax lots within import file _(can be included in 'New tax lots' or even 'Merged tax lots against existing records')_
- Merged properties against existing records
- Merged tax lots against existing records
- Merged properties between existing records
- Merged tax lots between existing records
- Ignored property duplicates within the import file
- Ignored tax lot duplicates within the import file
- Ignored duplicates of existing properties
- Ignored duplicates of existing tax lots

#1977 
#656


seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -173,7 +173,7 @@ angular.module('BE.seed.controller.inventory_list', [])
             inventory_type: function () {
               return $stateParams.inventory_type;
             },
-            single_record: function () {
+            provided_inventory: function () {
```
adrian-lara: @nllong had a typo in there - thanks for catching this!
https://github.com/SEED-platform/seed/pull/2035/files#diff-0aab274639148dc7bdec29d433d9150dR176

nllong: Cool. I’ll test again

nllong:
worked now. thanks!

PR #2033, opened 2019-12-06
Data quality check for valid data type

lainsworth8801:
1. Changed 'Severity' tab to 'Severity Level';
2. Added 'valid data' for data quality check;
3. Added valid data labeling for inventory filtering;
3. Made labeling for valid data type mandatory (return bad_request if no labels applied);
4. Under data quality, added error message if no labels applied for valid data.

Screen Shot 2019-12-04 at 3 45 43 PM

Screen Shot 2019-12-04 at 3 46 23 PM

Screen Shot 2019-12-04 at 3 46 48 PM

Screen Shot 2019-12-04 at 3 47 04 PM

Screen Shot 2019-12-06 at 9 36 38 AM

Screen Shot 2019-12-06 at 9 37 17 AM


seed/models/data_quality.py
```diff
@@ -746,6 +754,29 @@ def _check(self, rules, row):
                         self.add_result_dimension_error(row.id, rule, display_name, value)
                         continue
 
+                    # Check for mandatory label for valid data:
+                    try:
+                        if rule.minimum_valid(value) and rule.maximum_valid(value):
+                            if rule.severity == 1:
```
nllong: likewise, this should be `if rule.severity == Rule.SEVERITY_VALID:`

seed/models/data_quality.py
```diff
@@ -714,9 +720,10 @@ def _check(self, rules, row):
                     # check the min and max values
                     try:
                         if not rule.minimum_valid(value):
-                            s_min, s_max, s_value = rule.format_strings(value)
-                            self.add_result_min_error(row.id, rule, display_name, s_value, s_min)
-                            label_applied = self.update_status_label(label, rule, linked_id)
+                            if rule.severity == 0 or rule.severity == 2:
```
nllong: I think we want this to read.
```
if rule.severity == Rule.SEVERITY_ERROR or rule.severity == Rule.SEVERITY_WARNING:
```

I'll make the change.

seed/models/data_quality.py
```diff
@@ -927,6 +958,18 @@ def add_result_missing_req(self, row_id, rule, display_name, value):
             'severity': rule.get_severity_display(),
         })
 
+    def add_result_missing_label(self, row_id, rule, display_name, value):
+        if rule.severity == 1:
```
nllong: same here.

seed/models/data_quality.py
```diff
@@ -109,9 +113,11 @@ class Rule(models.Model):
     ]
 
     SEVERITY_ERROR = 0
-    SEVERITY_WARNING = 1
+    SEVERITY_VALID = 1
```
nllong: I think this is going to break the consistency of the database. Presently there are 1's in the database, which means that any rule that was a warning, is not a valid data check.

image

Need to change validity check to be 2. 

I'll make this change. This will also require a migration. Need to call `./manage.py makemigrations`. If you run this command you will notice that it is not going to renumber any of the Rules in the database.

seed/views/data_quality.py
```diff
@@ -402,6 +402,12 @@ def save_data_quality_rules(self, request, pk=None):
         dq = DataQualityCheck.retrieve(organization.id)
         dq.remove_all_rules()
         for rule in updated_rules:
+            if rule['severity'] == Rule.SEVERITY_VALID and rule['status_label_id'] is None:
+                return JsonResponse({
+                    'status': 'error',
+                    'message': 'Label must be assigned when using Valid Data Severity.'
```
nllong: @adrian-lara -- does this need to be translated?

nllong:
@lainsworth8801 -- okay, this is ready for you to test locally again. Make sure to run the migrations `./manage.py migrate`. Thanks!

lainsworth8801:
Sounds good thanks Nick!

seed/views/data_quality.py
adrian-lara: I actually haven't checked/accounted for translations of text coming from responses like this.

But yes, if the message is parsed and displayed to the user within the UI (not just in the dev tools), a translation would require a new lokalise entry.

lainsworth8801:
@nllong all changes are working functionality wise. For the error message, not sure if I'm understanding your concern right but on the UI side it is interpreting either the root load_error_message or "there was an error loading the page" if error_message is empty so I assigned load_error_message to response's data error message. Not familiar with Lokalise so maybe some info on that plz?

seed/views/data_quality.py
nllong: @adrian-lara can you add a translation for this one? Sorry to assign it to you. We can circle back with @lainsworth8801 on how it works post-release. Thanks!

adrian-lara: No worries! Just added it 👍

Issue #2015, opened 2019-11-14
Importing certain column names can break all future imports


axelstudios:
### Expected Behavior
You should be able to upload any file with any column names and expect the import to succeed
### Actual Behavior
Importing a file with a column name that matches a field on a table that does not support extra_data fields (`Property` and `TaxLot` rather than `PropertyState` and `TaxlotState`) can cause the current and all future imports to fail with 500 errors.
### Steps to Reproduce
1. Create an import file with a column named `Campus`.  (The common scenario for this is to export data from seed and then attempt to reimport that file)
2. On the initial Mapping screen it will recognize that column and erroneously suggest mapping to the `Property` table, and the Inventory Type dropdown will have an invalid value
![2019-11-14_115940](https://user-images.githubusercontent.com/411466/68889101-7c38f980-06d9-11ea-8967-01a84fd085b0.png)
3. If you leave the Inventory Type dropdown set to the invalid value, then regardless of the column name the Review Mappings step will immediately fail with a 500 and it will create an invalid extra_data column associated with the `Property` table:
![image](https://user-images.githubusercontent.com/411466/68889895-177e9e80-06db-11ea-85ab-b6a26b0a1e02.png)
4. Attempting to import any file after this, regardless of columns or mapping, will result in a 500 error when you click `Save Mappings`.

It's very likely that this can be reproduced with other column names from the `Property` and `Taxlot` tables, but so far this is untested

axelstudios:
I hit this again - uploaded a file to a new org with the column name `Random` that got the mapping suggestion of `Created` on the `Property` table....

nllong:
Okay. I can work on this one

RDmitchell:
I don't see this issue anymore, so will close this but have come across another issue #2054

PR #2034, opened 2019-12-06
Do not map to Property or TaxLot model fields (only PropertyState and TaxLotState)

nllong:

There are a few fields that map to Property and TaxLot models (not PropertyState and TaxLotState) including campus, created, and updated. These fields were being returned in the mapping suggestions, but the name of the table was invalid (and unmappable).

#### What's this PR do?
* Force the mapping suggestions to map to only PropertyState and TaxLotState fields by forcing the table name in the result.
* This fixes the issue with the invalid mapping column, but it is still **not** advised to use the same name.

See #2015 

Note that if a user still wants to map the fields, then they will get two fields named the same. With one being on the property/taxlot and the other being on the propertystate/taxlotstate. 
image


RDmitchell:
@nllong -- I don't quite understand how a user, through the GUI, would map to Property or TaxLot models rather than PropertyState or TaxLotState. I assumed that this was all handled "behind the scenes".

Is this an issue when using the API calls because you can map to specific models?

nllong:
It was possible... see #2015. It was definitely an edge case.

RDmitchell:
So this would only happen in the UI if the "table" pulldown list was blank, ie, not set to either table?

The ESPM files almost always contain a Campus field and I have never see the field not set to Property in the mapping screen. But clearly it can happen, I just have never seen it.

nllong:
That’s correct. Yeah it can happen, but it is rare.

Da: RDmitchell <notifications@github.com>
Risposta: SEED-platform/seed <reply@reply.github.com>
Data: venerdì dicembre 6 2019 13:09
A: SEED-platform/seed <seed@noreply.github.com>
Cc: Nicholas Long <Nicholas.Long@nrel.gov>, Mention <mention@noreply.github.com>
Oggetto: Re: [SEED-platform/seed] Do not map to Property or TaxLot model fields (only PropertyState and TaxLotState) (#2034)


So this would only happen in the UI if the "table" pulldown list was blank, ie, not set to either table?

The ESPM files almost always contain a Campus field and I have never see the field not set to Property in the mapping screen. But clearly it can happen, I just have never seen it.

nllong:
@adrian-lara or @axelstudios  -- can you test this one and merge if it looks good?

nllong:
@adrian-lara can you test this and merge if good?

Issue #1991, opened 2019-10-17
Column Settings - Sort by MATCH CRITERIA needs some tweaks


RDmitchell:
### Expected Behavior
Want to sort by the MATCH CRITERIA column, so that the order is:
* checked box(es)
* unchecked boxes
* Ineligible
### Actual Behavior 
Right now the sort doesn't put all the "Ineligible" fields together. They are interspersed with the unchecked boxes
![image](https://user-images.githubusercontent.com/6314950/67048152-8c64b580-f0e8-11e9-82ba-455f2d39e370.png)

### Steps to Reproduce
Go to the Column Settings link in Organizations, and click on the MATCH CRITERIA column until the checked boxes are at the top.
### Instance Information
Instance: dev1
SHA: 46d759f6

adrian-lara:
@lainsworth8801 - heads up, I made some changes to the logic driving this sorting. In the `geocoding_fields` PR (not tagging here to avoid "referencing" the issue in the PR). I'd wait for that PR to get merged to develop before working on this just to make your life easier.

RDmitchell:
This seems to be working. Thanks !

PR #2038, opened 2019-12-10
Sort is_matching_criteria by order of checked, unchecked, ineligible

lainsworth8801:
Glyphicon-list icon (next to 'MATCH CRITERIA' label) does the sorting but was in a slight different order. It is modified to sort columns by matching criteria in an order of checked, unchecked, and ineligible.

Screen Shot 2019-12-09 at 10 11 59 PM


nllong:
Addresses  #1991

Issue #709, opened 2016-02-02
Allow user to disable columns in import files and handle redundant column names in import files


RDmitchell:
The program can't deal automatically with a file that has two fields with the exact same name. We should require the user to fix the file so that there are not duplicated field names, and then reimport it. 
So in the case where there are two field names in a data file that are identical, the program should detect that and put up a message to the effect of:

The program has detected the following fields that have identical names <xx,xx>. Please make the field names unique in your data file and reimport. 

Once we get to this issue, we can refine the message and possibly the behavior.

RDmitchell:
@bravacheLBL will make a sample file and a more complete description of the program's current behavior to add context to this issue.

bravacheLBL:
The sample file is here:
https://drive.google.com/a/lbl.gov/file/d/0B-Qu-Oc76NpJdlYtRVV2eDM5RkU/view?usp=sharing

In that file I duplicated the address column and left the same field name. I also deleted some of the data in the duplicate to be able to differentiate the two.

It seems that the way this is handled by SEED is to detect that there are two fields with the same name, but to only attach the last column to that name (twice).

![image](https://cloud.githubusercontent.com/assets/14808873/12766497/0760fd94-c9b9-11e5-87cb-7733a5d08d49.png)

If you try to name the two fields differently, only one of them will contain the data while the other will be created but empty.

![image](https://cloud.githubusercontent.com/assets/14808873/12766696/c66ffdb6-c9b9-11e5-952c-b9427010a499.png)

(the field "Address Line 1" is completely empty)

The easiest solution would be to block that type of file and print an error for the user to fix it, so that he can differentiate the two fields manually, knowing what they represent.

RDmitchell:
see issue #781

RDmitchell:
instance: staging (LBNL)
SHA: 36cc59d (2.3.1 release)

I'm not sure the title of this issue really captures the problem, but decided to leave it. 

The program still exhibits this behavior, and I am not even sure what to do as a user. So I think the "Duplicate" indication is left over from other code. 

So we need to figure out how we want the program to handle this and then bring it into the current world. Seems like the program should let you map the fields to different SEED names, but maybe that's not possible given the way the mapping table works. 

So maybe the user just has to abandon the mapping, fix the file outside the program, and then reimport it? 

![image](https://user-images.githubusercontent.com/6314950/37236894-7f0b2c52-23c1-11e8-8720-07bdeac896f7.png)

nllong:
[...]

PR #2042, opened 2019-12-11
709 dup import file headers

adrian-lara:
SEED hangs when attempting to map columns for a file with duplicate file headers

#### What's this PR do?
Prevent mapping attempts on a file that has duplicate file headers.

Import mapping page disables the "Map Your Data" button and provides an updated warning if there are duplicate file headers.

Also, in case mapping for such a file is attempted via API - Python is set to throw an error.

#709


locale/en_US/LC_MESSAGES/django.po
```diff
@@ -1668,8 +1668,8 @@ msgstr "Please go to the following page and setup your account:"
 msgid "Please join the SEED-Platform Dev forum where you can connect with other developers."
 msgstr "Please join the SEED-Platform Dev forum where you can connect with other developers."
 
-msgid "Please review SEED headers.  Duplicate values are not allowed."
-msgstr "Please review SEED headers.  Duplicate values are not allowed."
+msgid "Please review SEED headers or Data File headers. Duplicate values are not allowed for either."
```
nllong: Can we reword to 'Please review SEED column names or the headers of the file being imported. Duplicate values are not allowed in either case.'

adrian-lara:
Updated the alert text.

Also, I am seeing the warning persist even when the SEED column value is changed
Screen Shot 2019-12-12 at 9 59 01 AM

nllong:
> Updated the alert text.
> 
> Also, I am seeing the warning persist even when the SEED column value is changed
>

Let me test again. It definitely wasn't updating.

nllong:
Looks good. I'll merge down once CI passes.

Issue #1998, opened 2019-10-23
First edit including lat long change doesn't trigger manual geocoding logic


adrian-lara:
### Expected Behavior
When a user edits an existing record and changes the `latitude` and `longitude` values, the `Geocoding Confidence` and `long_lat` should be updated. 

### Actual Behavior 
In most cases, this does happen via a pre_save hook. But when a user edits a record for the first time, a completely new -State object is used in order for the originally imported data to persist. This new -State object does not get affected by the pre_save hook.

The affected record would not be pinned on the map page until the user triggers a "Geocode Selected" event on the record. 

Also, if the intent of an edit is to remove geocoding, the user would not be able to remove the `Geocoding confidence` value with the initial edit. This is relevant in the case where a user wants to trigger geocoding by MapQuest for a previously manually geocoded record. This re-geocoding is only allowed on records that don't have a `Geocoding confidence` value of "Manually geocoded (N/A)".

### Steps to Reproduce
Import a record without geocoding it - don't provide a MapQuest API key or don't import a file with address data. On the first edit of this record, change the lat and long. See that the geocoding confidence isn't populated.

### Instance Information
all instances

### Possible Solution
On first-edit, create a copy of the -State and run that copy through the existing not-first-edit code.

adrian-lara:
Even without an API key, you should still see the "Manually Geocoded..." within geocoding_confidence once you make that initial edit of populating latitude and longitude.

RDmitchell:
Instance: dev1
SHA:
Org: LBNL 201
Cycle: Geocoding test

@adrian-lara -- is this the behavior you are expecting? I couldn't quite tell from the testing desciption. 

* Imported sample ESPM file that has only Address Line 1 and City (neither are real)
* No Mapquest API specified
* File imported, Lat, Long, Geocoding confidence are all blank
![image](https://user-images.githubusercontent.com/6314950/71218121-3beb0f00-2275-11ea-8891-bd966a36fc92.png)
* Click on the "i" icon to go into the detail view of PM Property ID # 6798215 (first record in screenshot above) -- Lat/Long/Geocoding confidence fields are blank before editing
![image](https://user-images.githubusercontent.com/6314950/71218384-29250a00-2276-11ea-8f5c-bbbdb8d849db.png)
* Click the Edit link, add values for Lat and Long
![image](https://user-images.githubusercontent.com/6314950/71218433-52de3100-2276-11ea-82a6-d9e6c52a523d.png)
[...]

PR #2049, opened 2019-12-16
1998 lat long initial edit

adrian-lara:
See issue linked below.

#### What's this PR do?
This PR adjusts the initial edit process for records (both properties and tax lots).

Just like before, for an initial edit, a copy of the record is created and used to preserve the initial import history. Now, this copy is created as the very first step, without the "initial edit" changes **first**, as opposed to creating the copy with the changes included. The new copy is then run through the normal (non-initial) edit flow.

This allows the presave hook to trigger.

For "new" (just imported) records that were geocoded on import, edit them while removing latitude and longitude. See that the `geocoding_confidence` and `long_lat` are no longer populated.

For "new" records that were not geocoded on import, edit them while changing a latitude and longitude. See that the `geocoding_confidence` and `long_lat` are populated.

#1998

### DIFF
--- a/seed/tests/test_property_views.py
+++ b/seed/tests/test_property_views.py
@@ -135,6 +135,33 @@ class PropertyViewTests(DataMappingBaseTestCase):
         self.assertGreater(datetime.strptime(data['property']['updated'], "%Y-%m-%dT%H:%M:%S.%fZ"),
                            datetime.strptime(db_updated_time, "%Y-%m-%dT%H:%M:%S.%fZ"))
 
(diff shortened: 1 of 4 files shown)


adrian-lara:
Yep! Seeing the "Manually geocoded (N/A)" upon adding lat and long values on first edit is the expected functionality. 

The issue being fixed was that "Manually geocoded (N/A)" wasn't being populated nor was the record creating a GIS Geometry `POINT (...` entry (and subsequently not showing up on the map).

Issue #994, opened 2016-06-24
It is possible to have orphaned user accounts without an associated organization


RDmitchell:
This sentry error was caused by the condition that the user account did not have an associated organization. This happened when I added a user to an org, then later deleted that user from that org, but forgot that I had done that. So when I logged in as the user that had been removed from the org, the program had no associated org and generated this error.
https://app.getsentry.com/lawrence-berkeley-national-lab/seed-django-production/issues/128196867/

Seems like we shouldn't allow orphaned user accounts. This might be a bit complicated to enforce, but if a user account is removed from an org, either by that org or by a superuser, seems like seed should delete that user account from wherever it is stored, as long as it isn't associated with any other organization. 

This is somewhat related to #13, which sort of the opposite case, ie, delete users if they are associated with an organization being deleted.

RDmitchell:
related to #1024

RDmitchell:
instance: dev1
SHA:
Org: lbnl

* I logged in as superuser, went to admin page
* Found one of my test orgs that only has one user
![image](https://user-images.githubusercontent.com/6314950/73319679-f6b1e980-41f1-11ea-89de-0b2d9bb64f66.png)
* Clicked on the Remove User button
* This dialog box appears, which seems to allow me to delete the only user from this org
![image](https://user-images.githubusercontent.com/6314950/73319724-15b07b80-41f2-11ea-9dbf-bbd4213bbd17.png)
* Then received this error
![image](https://user-images.githubusercontent.com/6314950/73319815-5ad4ad80-41f2-11ea-941e-292851ea5c92.png)

Is this the expected behavior?

Should there be organizations without users? 

Not sure what the fix was for this.

nllong:
Yes, an organization has to have at least one user.

RDmitchell:
@nllong --so is the current behavior what is desired? 

I guess the current state is that the program eventually gets an error if you try to remove the only user from an organization, although it's not that elegant. 

You can close this if you think the current state is good enough.

nllong:
I think we should close as the typical user won't get this message (mainly only admins).

RDmitchell:
agreed !

PR #2050, opened 2019-12-17
Deactivate orphaned users when removed from last org

axelstudios:
#### What's this PR do?
If a user only belongs to one org and is removed from that org (either from the admin page or the org member page) the user will be prompted to confirm that that will delete the user (rather than orphaning them).

Also:
- Fixes add/remove user form validation and console error
- Fixes a variety of small bugs on the admin page
- Fixes a bug with updating the default org for a user that was removed

1. Add a user to at least two orgs
2. Delete the user from those orgs and confirm that the last one prompts you to delete the user
#994


docs/source/developer_resources.rst
```diff
@@ -305,7 +305,7 @@ To make a release do the following:
     python docs/scripts/change_log.py –k GITHUB_API_TOKEN –s 2018-02-26 –e 2018-05-30
 
 4. Paste the results (remove unneeded Accepted Pull Requests) into the CHANGELOG.md. Make sure to cleanup the formatting.
-5. Make sure that any new UI needing localization has been tagged for translation, and that any new translation keys exist in the lokalise.co project. (see :doc:`translation documentation <translation>`).
+5. Make sure that any new UI needing localization has been tagged for translation, and that any new translation keys exist in the lokalise.com project. (see :doc:`translation documentation <translation>`).
```
nllong: i saw that they changed their URL... thanks for fixing this.

seed/views/organizations.py
```diff
@@ -460,14 +464,15 @@ def remove_user(self, request, pk=None):
...
-            # assign them to a new organization if they want to use the account again.
-            first_org_user = OrganizationUser.objects.filter(user=user).order_by('id').first()
-            if first_org_user:
-                request.user.default_organization = first_org_user.organization
-                request.user.save()
+        user_orgs = OrganizationUser.objects.filter(user=user)
+
+        if user_orgs.count() == 0:
+            # Remove orphaned user
+            user.delete()
```
nllong: are we sure we want to delete the user? Or should we just deactivate their account? If we deactivate the account, then we would need to make sure that the is_active flag is respected throughout the code. My main concern is that there are models that reference user ids (e.g., notes)

https://github.com/SEED-platform/seed/blob/4587ae1e8a1eec589677086ce1c887898521d13b/seed/landing/models.py#L53-L56

nllong:
@axelstudios is this ready for review/testing?

axelstudios:
@nllong Updated to deactivate users instead of deleting (with migration for existing org-less users)

Issue #1713, opened 2018-09-28
Return progress status immediately when uploading large files


nllong:
### Expected Behavior
The user should be able to upload large files.

### Actual Behavior 
The web server takes *a long* time to return the progress when uploading a large file. This is causing a timeout error on the CDN. 

The issue is due to the following code block:

```
progress_data = ProgressData(func_name='save_raw_data', unique_id=file_pk)
    # save_raw_data_run.s(file_pk, progress_data.key)
    try:
        # Go get the tasks that need to be created, then call them in the chord here.
        tasks = _save_raw_data_create_tasks(file_pk, progress_data.key)
        if tasks:
            chord(tasks, interval=15)(finish_raw_save.s(file_pk, progress_data.key))
        else:
            finish_raw_save.s(file_pk, progress_data.key)
    except StopIteration:
        progress_data.finish_with_error('StopIteration Exception', traceback.format_exc())
    except Error as e:
        progress_data.finish_with_error('File Content Error: ' + e.message, traceback.format_exc())
    except KeyError as e:
        progress_data.finish_with_error('Invalid Column Name: "' + e.message + '"',
                                        traceback.format_exc())
    except Exception as e:
        progress_data.finish_with_error('Unhandled Error: ' + str(e.message),
                                        traceback.format_exc())
    _log.debug(progress_data.result())
    return progress_data.result()
```

The `return progress_data.result()` happens after all the data have been loaded into the celery tasks. The system should return the progress_data (and key) before all the data are loaded.

### Steps to Reproduce
* Upload large file on dev1
* In inspector notice that the `save_raw_data` returns a 524 after 100seconds.

### Instance Information

Server Instance: `dev1`
Server SHA: `any`

RDmitchell:
@nllong -- I was just about to test the 177,000 record file from PNNL with the big geometry / polygon field in it. 

Should I hold off, given this issue that needs to be addressed?

nllong:
Yeah, let's focus on NYC first, then we can look at this.

RDmitchell:
sounds good.

RDmitchell:
@nllong -- what file should I use for testing? 

Also, what progress bar is at issue here? Is it the very first "Upload your data" progress bar? 
![image](https://user-images.githubusercontent.com/6314950/73320133-37f6c900-41f3-11ea-8743-332162490472.png)

Or something else? 

Imported file with 13,000+ records.
* First progress bar appeared immediately
[...]

PR #2055, opened 2019-12-19
Background tasks return response immediately

nllong:
When uploading a file the data gets chunked and assigned to background tasks to parse. The chunking happens upon upload and is not done in the background. When this is a large file (many rows/columns) then the chunking can take some time.

#### What's this PR do?
* Make the save_data methods to be background tasks that call other background tasks (chords)
* Update some dependency versions including Django (not required for this PR)

* Testing uploading CSV, ESPM, Meter, GeoJSON, and GreenButton data
* Test with both always eager and normal background tasks

#1713


adrian-lara:
Regarding manual testing... By ESPM import, are you referring to the flow where users enter their ESPM creds? If so, I haven't gone through that before (as I don't have access to an account).

Issue #1913, opened 2019-07-10
Add Notes info to export


RDmitchell:
### Expected Behavior
Count of Notes + Notes content for each record in an exported file
### Actual Behavior 
That data is not contained in the exported file

It would be nice to have the following in the exported file
* Notes Count -- this would be the number of Notes associated with a record, similar to the way it is displayed in the inventory list view
![image](https://user-images.githubusercontent.com/6314950/61004964-14823480-a31c-11e9-803a-7210adb225b8.png)
* Notes Content -- this would be one field that would contain all the info in all the notes associated with a record, preferably separated by a semicolon or some other delimiter (?); this might be tricky because there will be cases with lots of notes that might have a lot of characters in them. But hopefully with a CSV download the program could dump all that info into one field (don't really know). 
  * If this isn't possible, at least having the Notes Count field would still be useful.
### Steps to Reproduce
n/a
### Instance Information
Server Instance: dev1
Server SHA: 93c4bdc

RDmitchell:
This is a request from the City of San Jose who is using SEED for their Benchmarking ordinance.

RDmitchell:
From @axelstudios -- would we want to include a timestamp for each note? 

That would seem useful, if possible.

RDmitchell:
From @axelstudios -- how the notes contents would look in Excel; this seems great.
![image](https://user-images.githubusercontent.com/6314950/61006314-e2be9d00-a31e-11e9-8f1e-747e15258369.png)

nllong:
Oops. You can delete it. Thanks!

Il giorno lun 26 ago 2019 alle 14:00 RDmitchell <notifications@github.com>
ha scritto:

RDmitchell:
instance: dev1
SHA: d32b8c7

This appears to be working for both Property and Tax Lot

Export from Property table includes Notes (and Labels)
![image](https://user-images.githubusercontent.com/6314950/73226355-9c972280-4124-11ea-851c-2b74d127d410.png)

Export from Tax Lot table includes Notes (and Labels -- although see #2092)
![image](https://user-images.githubusercontent.com/6314950/73226532-4aa2cc80-4125-11ea-9a6a-0e6441eb76a2.png)

PR #2056, opened 2019-12-19
Notes are added to record exports

adrian-lara:
See attached issue
#### What's this PR do?
Appends Notes to records when exporting records.

**Issue/Caveat:** Since the file exports are generated in the backend, the "created" time for each note is always provided in the app's default timezone, 'America/Los_Angeles'. I've spent some time trying to prevent this, but was not successful. Since the users requesting this are from San Jose (within that timezone), I'll just be creating an issue for this.

Extra details regarding this: When a note is created via the browser, the DB entry for this note seems to use the machine's current local timezone. When that note is read from the DB into Django (and subsequently, into the export file), it's automatically converted to 'America/Los_Angeles' regardless of my own actual timezone.

**Also**, this hides the tax lot excel export button for now since conversations are ongoing surrounding tax lots and BuildingSync.

Add notes to records, export them, and see that notes are added

#1913 - add notes to exports
#2057 - hide tax lot excel export button for now

### DIFF
--- a/seed/tests/test_tax_lot_property.py
+++ b/seed/tests/test_tax_lot_property.py
@@ -7,7 +7,6 @@
 import json
 
 from django.core.urlresolvers import reverse_lazy
-from django.test import TestCase
 
 from seed.landing.models import SEEDUser as User
 from seed.models import (
@@ -15,6 +14,7 @@ from seed.models import (
     PropertyView,
     TaxLotProperty,
     Column,
+    Note,
 )
 from seed.test_helpers.fake import (
     FakePropertyFactory,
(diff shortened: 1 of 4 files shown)


seed/tests/test_tax_lot_property.py
```diff
@@ -110,6 +111,43 @@ def test_csv_export(self):
...
+            data=json.dumps({'columns': columns, 'export_type': 'csv'}),
+            content_type='application/json'
+        )
+
+        # parse the content as array
+        data = response.content.decode('utf-8').split('\r\n')
+        notes_string = (
+            multi_line_note.created.astimezone().strftime("%Y-%m-%d %I:%M:%S %p") + "\n" +
+            multi_line_note.text +
+            "\n----------\n" +
```
nllong: On my machine, this causes the file to import into excel across mulitple lines

image

Did this work for you when opening in Excel?

adrian-lara: I uploaded an export from my machine called "2propswithnotes.xlsx" into our shared drive:
https://drive.google.com/drive/folders/1d2xdSEVcwpO1TFaItXA_1r-9WDmd0v2W

Admittedly, I don't have Excel on my machine. But I do see the notes contained in one cell when opening in both Apple Numbers (🤢 I know) and Google Sheets.
Screen Shot 2019-12-20 at 10 09 02 AM
Screen Shot 2019-12-20 at 10 11 37 AM

Issue #2018, opened 2019-11-26
Geocoding Confidence in Master record are blank for merged records


RDmitchell:
### Expected Behavior
I would expect that the program would show the Geocoding Confidence for the latest record in the Inventory Detail view when 2 or more records are merged
### Actual Behavior 
If a record is not merged with another record, the Geocoding Confidence field is displayed in the Master in the Detail View (and thus in the List View). This seems to be true in the Detail view of both the Tax Lot and Property tables
**Property Tab:**
If there is just one record in the Detail View, the program displays the Geocoding Confidence value from that record in the Master record
![image](https://user-images.githubusercontent.com/6314950/69589823-264e3680-0fa2-11ea-873e-9e941adac52b.png)

However, if 2 records are merged, whether they have the same Geocoding Confidence field value or not, the Master record for that field is blank (this screen shot is actually from the production server, but the behavior is the same on dev1).
![image](https://user-images.githubusercontent.com/6314950/69590552-bdb48900-0fa4-11ea-8e3d-8eed963b8f65.png)

And here is even a more anomalous case -- the Geocoding Confidence is displayed in the Property List view (and so I am assuming it would be in the Master record)
![image](https://user-images.githubusercontent.com/6314950/69590775-6bc03300-0fa5-11ea-9947-fd8f6fd9d472.png)
But when you look at the Master record, the Geoconfidence field is blank (!)
![image](https://user-images.githubusercontent.com/6314950/69590884-d1142400-0fa5-11ea-8b4b-469ab67983cc.png)

**Tax Lot Tab:**
Record that has merged records -- Geocoding Confidence field is blank in Master record
![image](https://user-images.githubusercontent.com/6314950/69590312-e0926d80-0fa3-11ea-8634-10b657b29e7b.png)
Record does not have merged records -- Geocoding confidence field in the Master record has the same value as the 1 file that has been imported
![image](https://user-images.githubusercontent.com/6314950/69590378-1afc0a80-0fa4-11ea-8c2a-bf561abc0f4a.png)

### Steps to Reproduce
Import 2 files where some records merge and some don't.
### Instance Information
Instance: dev1
SHA: 6dad9f771

Instance: seedv2 (production)
SHA: df21680b
Org: LBNL Test 200

Files:
1 - example-data-ESPM.xlsx
1 - example-data-ESPM-UBID-NoAlerts.xlsx
1 - example-data-taxlots.xlsx

They can be found in the folder for this issue
[...]

Issue #2054, opened 2019-12-18
Matching: Program doesn't recognize matching fields (sometimes)


RDmitchell:
### Expected Behavior
Importing a file, if there are the appropriate fields set to matching I would expect the program to recognize them.
### Actual Behavior
In this particular instance, the program doesn't recognize them. 
### Steps to Reproduce
See this folder with doc and campus test.csv file
https://drive.google.com/drive/folders/1bfzvJqZxarsSCMEVog7Gy9AuSUC_D4Bq?usp=sharing
Direct link to doc
https://docs.google.com/document/d/1XAz_UWf-uca7HlnZCGWoetlSIIesL6Rjdmcs-grcdaU/edit?usp=sharing
### Instance Information
Instance: dev1
SHA: bbab51f0

RDmitchell:
Came across this when testing #2015

adrian-lara:
Switching this from Alex to myself because I've already started investigating before we realized there was already an issue for it.

adrian-lara:
@RDmitchell I wasn't able to replicate this on dev1 nor within a local deployment of the deb1 DB.

I made numerous attempts using a single preset, switching between presets, and using the "Copy Data File Headers..." button. Poking around at the logic that's driving the warning to come up, I didn't find anything that could have thrown this bug. Also, there doesn't seem to be any issues with the specific organization you were using.

Unless you can provide more info that might be useful in helping replicate, my recommendation would be to close this if you can't replicate it again, and we'll have to reopen it with (hopefully) some details that'll help identify what's causing this bug.

@nllong - I think you hit this bug too, so tagging you hear in case you can recall any details that'll shed light on what caused this to happen.

RDmitchell:
@adrian-lara / @nllong -- I encounter this issue every now and again, but have no idea what I have done to cause it. 

Can we leave it open, but maybe in To Do (?) with a moderate priority ??

adrian-lara:
No strong opinion from me. In either case, I'd just reiterate that we should add more details when/if this issue comes up again.

If we do get more details, I'm happy to look into this again.

RDmitchell:
@nllong -- what do you think? Close this issue and wait for it to happen again and make a new issue? 

Or keep it open in the To Do list. 

Not sure how we are going to get more details on why this happens. But it does happen

adrian-lara:
[...]

PR #2068, opened 2020-01-06
Merging geocoded records

adrian-lara:
Merging geocoded records did not trigger the merging of geocoding results properly.

The solution to this involves using the complete geocoding results entirely from one record. For example, we're preventing the use of the Latitude of one record with the Longitude of the other.

See the linked issue for more details.

#### What's this PR do?
Allows for geocoding results between records to be merged when the two records are merged.

From the user perspective, adding merge protection to one of these geocoding result columns adds merge protection to the other geocoding result columns.

Merge geocoded records between changes of the merge protection for the geocoding result columns - Longitude, Latitude, and Geocoding Confidence.

#2018


seed/lib/merging/tests/test_merging.py
```diff
@@ -151,6 +156,167 @@ def test_merge_state_favor_existing(self):
         self.assertEqual(result.address_line_2, 'new')
         self.assertEqual(result.extra_data['field_1'], 'orig_value')
 
+    def test_merge_geocoding_results_no_merge_protection(self):
```
nllong: woah! TDD.. love it! 😜

seed/lib/merging/merging.py
```diff
@@ -84,6 +88,50 @@ def get_state_attrs(state_list):
...
+        existing_results_empty = existing_results_empty and can_attrs[geocoding_col][state1] is None
+        new_results_empty = new_results_empty and can_attrs[geocoding_col][state2] is None
+
+        geocoding_favor_new = geocoding_favor_new and priorities.get(geocoding_col, 'Favor New') == 'Favor New'
+
+        # Since these are handled here, remove them from canonical attributes
+        del can_attrs[geocoding_col]
+
+    # Multiple elif's here is necessary since empty checks should be first, followed by merge protection settings
+    if new_results_empty:
```
nllong: seems like possible performance issues here... we'll cross that bridge when we get there.

adrian-lara:
> Maybe this isn't related, but I can't finish testing due to the following.
> 
> I've created a geocoding profile but the site isn't allowing me to map asserting that I need a Custom ID 1... which I have.
> 
> image
> 
> Maybe this is fixed elsewhere?

I wasn't able to reproduce this. I'm thinking there's a lower chance that this is related, but would you be able to share the file you were using during import? 

Also, it looks like you're mapping both "Custom ID 1" and "Address Line 1", so it's even more surprising that you're getting that error.

RDmitchell:
@adrian-lara / @nllong  -- I have seen this problem. See issue #2054

nllong:
Going to accept this one as the issue associated with it has been addressed. The other issue #2054 has been captured and in the queue to be fixed.

Issue #2017, opened 2019-11-19
Column Settings: Add ability to sort by Display Name or Column Name


RDmitchell:
### Expected Behavior
Sort by most of the columns in Column Settings
### Actual Behavior 
Can't sort by Display Name or Column Name
![image](https://user-images.githubusercontent.com/6314950/69107052-090aec80-0a25-11ea-836f-c7042d9665f9.png)

### Instance Information
Instance: dev1
SHA: 6dad9f771

RDmitchell:
instance; dev1
SHA: 6a6d4ba0
Org: LBNL Test 200

Seems to work fine. The icon for the up/down is a bit close to the label name, might be nice to have a bit more spacing but that's a tiny detail. 

Closing

PR #2088, opened 2020-01-24
2017 column settings

aviveiros11:
First PR, Woo!  This is addressing the column settings table sorting on the organizations page
#### What's this PR do?
Allows the user to now sort the Display name and Column name columns alphabetically. 
Click the header of either column to sort them and click again to return to the default state.  
Solves issue [#2017](https://github.com/SEED-platform/seed/issues/2017)


seed/static/seed/partials/column_settings.html
```diff
@@ -56,8 +56,12 @@ <h3 translate>MODIFYING_COLUMN_SETTINGS</h3>
                 <table id="column-table" class="table has_no_btm_margin">
                     <thead>
                         <tr>
-                            <th translate>Display Name</th>
-                            <th translate>Column Name</th>
+                            <th class="text-left" ng-click="toggle_display_name_order_sort()">
+                                <span>Display Name<i class="glyphicon glyphicon-sort"></i></span>
```
adrian-lara: I'm thinking we should make the icons all the same. I kind of like the `sort` icon you're using vs the `list` icon that existed before, but I don't have a strong opinion either way.

@axelstudios any thoughts?
Screen Shot 2020-01-24 at 12 38 29 PM

nllong: I agree with @adrian-lara -- that is the sort icon is a bit nicer!

axelstudios: Same - I like _sort_

nllong:
Changes look good. Merging once tests pass.

Issue #2023, opened 2019-12-02
Data Import: when importing a new file to a dataset, default to the last cycle imported rather than the first cycle in the pulldown list


RDmitchell:
### Expected Behavior
I imported a file into what I thought was one cycle, but it turned out that I wasn't paying attention and let the program default to the first cycle rather than selecting the correct cycle myself. 
### Actual Behavior 
When importing a file into a data set, the program always defaults the cycle to the first cycle in the pulldown list. It would be nice if the program could remember the last cycle that was used in the dataset -- I suspect that most people organize their datasets by cycle and it's easy to make a mistake by not selecting the correct cycle each time you import a file
### Steps to Reproduce
Import a file into a dataset
In this case, the file in the dataset that was previously imported used Cycle = "2018 Compliance Cycle". 
![image](https://user-images.githubusercontent.com/6314950/69990250-2dbd9480-14fa-11ea-9888-2f18eb9a6113.png)
I then import a new file, and without paying close enough attention, I let the program default the cycle to the first one on the list, which has a very similar name. 
![image](https://user-images.githubusercontent.com/6314950/69990461-97d63980-14fa-11ea-8ce5-574650a4f689.png)

I don't think we need to add this to the Dec 2019 project, but wanted to capture the thought, and maybe in January when we check in with users, we can make a list of issues like this and see if they think they would be good usability improvements. So I am setting it to Priority = 3
### Instance Information
Instance: dev1
SHA: 6dad9f771

aviveiros11:
Hey, @RDmitchell I have added the persistent choice functionality so it will locally cache the choice in storage and remember what the user has chosen!  Did you any other more pressing usability concerns?

RDmitchell:
Instance: dev1
SHA: c97042d2

This is working nicely. Thanks

PR #2097, opened 2020-01-30
Persistant drop down cycle choice

aviveiros11:
#### What's this PR do?
Stores the choice of an item from the cycles dropdown menu in local storage.  
`Data` > `Add Data to Files` > `Select` > `Dismiss` Repeat to find your selection is cached
Addressing #2023

### DIFF
--- a/seed/static/seed/js/controllers/data_upload_modal_controller.js
+++ b/seed/static/seed/js/controllers/data_upload_modal_controller.js
@@ -64,12 +64,9 @@ angular.module('BE.seed.controller.data_upload_modal', [])
       organization
     ) {
       $scope.cycles = cycles.cycles;
-      var cached_cycle = inventory_service.get_last_selected_cycle();
-      if ($scope.cycles.length && cached_cycle == null){
-        $scope.selectedCycle = cycles.cycles[0];
-      } else {
-        $scope.selectedCycle = cached_cycle;
-      }
+      var cached_cycle = inventory_service.get_last_cycle();
+      $scope.selectedCycle = _.find(cycles.cycles, {id: cached_cycle}) || _.first(cycles.cycles);
+
       $scope.step_10_style = 'info';
       $scope.step_10_title = 'load more data';
       $scope.step = {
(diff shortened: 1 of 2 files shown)


seed/static/seed/js/controllers/data_upload_modal_controller.js
```diff
@@ -64,7 +64,7 @@ angular.module('BE.seed.controller.data_upload_modal', [])
       organization
     ) {
       $scope.cycles = cycles.cycles;
-      if ($scope.cycles.length) $scope.selectedCycle = $scope.cycles[0];
+      if ($scope.cycles.length) $scope.selectedCycle = inventory_service.get_last_selected_cycle();
```
adrian-lara: You might want to add a fallback to the previous `$scope.cycles[0]`. Since there might not be anything in the cache, it'll get the import modal in a weird state in that case.

Screen Shot 2020-01-30 at 6 09 09 PM

There's probably a few examples of a similar pattern in the codebase somewhere.

adrian-lara: Note, ☝️ happens consistently when you switch orgs. I'm sure it happens when you've just got nothing cached as well. 

There's a localstorage pattern in SEED that caches historical "choices" by, say, org, if you can find it, that might be even better to implement. But I'm not convinced it's necessary since most users likely aren't jumping back and forth between organizations like developers might be.

aviveiros11:
@adrian-lara thank you for catching that!

seed/static/seed/js/controllers/data_upload_modal_controller.js
```diff
@@ -64,7 +64,12 @@ angular.module('BE.seed.controller.data_upload_modal', [])
       organization
     ) {
       $scope.cycles = cycles.cycles;
-      if ($scope.cycles.length) $scope.selectedCycle = $scope.cycles[0];
+      var cached_cycle = inventory_service.get_last_selected_cycle();
+      if ($scope.cycles.length && cached_cycle == null){
```
adrian-lara: When switching between Organizations, this logic still allows for a Cycle from a different organization to be selected.  It'll look like this:
Screen Shot 2020-01-31 at 11 27 43 AM
...which is similar to before except all the buttons are still clickable.

adrian-lara:
Oh also, remember to update this branch with the latest develop

seed/static/seed/js/controllers/data_upload_modal_controller.js
aviveiros11: I will follow the pattern linked above!

Issue #2080, opened 2020-01-21
Historical mapping page bug from mapping preset changes


adrian-lara:
### Actual Behavior
The historical import mapping page doesn't load.

### Steps to Reproduce
Import a file, and revisit the mapping page that should show the historical mappings used.

### Instance Information
`develop`

aviveiros11:
Hey @adrian-lara could I get a little more context for this?  By historical mappings do you mean current column mapping detail?  Thanks!

adrian-lara:
Yep! If you import a file, then revisit the mapping page via the Mapping
button in the Data Set page, the page should fail to load.

aviveiros11:
So after importing a file, if you don't choose to `continue to data mapping` and you `dismiss` the window and then click `data mapping` the page should fail to load?

RDmitchell:
Instance: Dev1
SHA: getting a 500 error, so don't know, probably c97042d

I tested this on several mappings that were quite old, up to 3 years old, and when I clicked on the Data Mapping button in a data set "list view", I was always able to see the mapping screen, and it looked like the mappings made sense (I didn't necessarily know what the original mappings were, but looking at the changes made between the original field names in the Data File Header column and the changed names in the SEED Header column, they all made sense).

adrian-lara:
This might be the same issue, so I just reopened. See related [Issue folder](https://drive.google.com/drive/u/0/folders/1zTicid8x_Jqo5E5ftXIQYtxVkAcGAES4)

aviveiros11:
This error occurred during an API anomaly that has since been patched

adrian-lara:
Makes sense. Haven't seen this pop up for a while 👍🏽

PR #2101, opened 2020-02-10
Found an undefined variable bug

aviveiros11:
Originally unable to return to mappings page from defaults
#### What's this PR do?
solves the undefined variable issue preventing the page from loading
Add data/import/map/pair and then return to the data mappings page
#2080

### DIFF
--- a/seed/static/seed/js/controllers/mapping_controller.js
+++ b/seed/static/seed/js/controllers/mapping_controller.js
@@ -182,7 +182,6 @@ angular.module('BE.seed.controller.mapping', [])
       };
 
       $scope.import_file = import_file_payload.import_file;
-      console.log($scope.import_file);
       $scope.import_file.matching_finished = false;
       $scope.suggested_mappings = suggested_mappings_payload.suggested_column_mappings;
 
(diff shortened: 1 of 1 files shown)


seed/static/seed/js/controllers/mapping_controller.js
```diff
@@ -720,23 +720,26 @@ angular.module('BE.seed.controller.mapping', [])
       var display_cached_column_mappings = function () {
         var cached_mappings = JSON.parse($scope.import_file.cached_mapped_columns);
         _.forEach($scope.mappings, function (col) {
-          var cached_col = _.find(cached_mappings, {from_field: col.name, to_table_name: col.suggestion_table_name})
-          col.suggestion_column_name = cached_col.to_field;
-          col.from_units = cached_col.from_units;
-
+          var cached_col = _.find(cached_mappings, {from_field: col.name})
+          if (cached_col == null) {
```
adrian-lara: I think the only possible values of `cached_col` is an object for the column or `undefined`, so I think this always enters the `else` logic.

I think it's likely that the only fixed needed was to remove the "to_table_name" key and value from the `_.find`, but I haven't look at this more closely, so let me know if I misunderstood anything.

aviveiros11: Hey, @adrian-lara thank you for the review!  The main reason I did the misc formatting push was to try and clear the License Compliance however Alex told the that it sometimes just doesn't pass.  I will remove the console log and also check out the null/undefined logic now that it should contain values

adrian-lara:
Blocks the testing of #1992

Issue #1933, opened 2019-07-23
BRICR BuildingSync exports after Update with BuildingSync


macumber:
[results.xml.txt](https://github.com/SEED-platform/seed/files/3423994/results.xml.txt)

### Steps to Reproduce

Upload attached XML file to create a new record in SEED.  From property page, select Update with BuildingSync action.  Re-upload the same XML file.

### Expected Behavior

Export of BuildingSync Excel would remain constant.

### Actual Behavior 

Timeseries meter data is no longer exported on Meter Readings tab.  Baseline scenario is duplicated on Scenarios tab.

### Instance Information

Server Instance: `bricr-dev`

macumber:
@nllong FYI

macumber:
This affects simulations run in BRICR since we first upload the BuildingSync, then reupload after running

https://bricr-dev.seed-platform.org/app/#/properties/37

macumber:
Additionally, when a property is deleted the associated scenarios are not deleted.

RDmitchell:
@nllong -- also see my comments in #1888 -- I think I am seeing similar issues to @macumber

PR #2109, opened 2020-02-13
Fix/update bsync

macintoshpie:
#### What's this PR do?
Fixes issue #1933 
Follow the reproduction steps in #1933 -- the meters tab should not be empty
#1933

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,19 +42,20 @@ install:
   - npm --version
   - nvm install stable
 before_script:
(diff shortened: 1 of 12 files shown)


macintoshpie:
That is a separate issue (the error saying it couldn't parse a time from `None`) due to the changing of TimeStamp -> Timestamp for Bsync schema v2, which is being addressed by other branches/issues. Currently SEED only supports schema v1

Issue #2102, opened 2020-02-11
Add "Freeze Master Column" feature in Detail View


RDmitchell:
### Expected Behavior
It would be nice to be able to freeze the first column in the detail view in the case where the history of the imported files scrolls off to the right. 
### Actual Behavior
The first column doesn't freeze to you lose track of the field names and the Master record values when you have to scroll to the right to see the history of the data imports.
### Steps to Reproduce
See this doc
https://docs.google.com/document/d/10gtRcg-keXeFEhJ5ryBuWGFfh8rKta-qOXjsj1wSPZQ/edit?usp=sharing
### Instance Information
Instance: dev1
SHA: 6a6d4ba0

aviveiros11:
Hey, @RDmitchell I am having trouble navigating to that page to reproduce the issue and map out the desired behavior.  Could you tell me how to navigate to the detail view? Thanks!

RDmitchell:
@aviveiros11 -- click on the "i" icon to the left of a record in the Inventory list view, either Tax Lot or Property
![image](https://user-images.githubusercontent.com/6314950/74483993-8bf8e300-4e6c-11ea-9719-63a2c4474584.png)
The idea would be to "pin" or "freeze" the first two columns so that when the user scrolls to the right, they stay visible. I suppose it would be nice to give the user complete control over which columns to "freeze" but that seems complicated, and it seems like we could start by just freezing the first two.
![image](https://user-images.githubusercontent.com/6314950/74484133-e2feb800-4e6c-11ea-9ac0-f881724557fb.png)

aviveiros11:
Good Morning, @RDmitchell I have a `pin` feature that I would like to add to the master column.  The User clicks the pin Icon and freezes the column when scrolling.  Let me know about the color I just threw something in to identify if the column is pinned or not.  

![Screen Shot 2020-02-17 at 10 07 15 AM](https://user-images.githubusercontent.com/31779424/74673921-771e9700-516d-11ea-9c30-f21c164bf0dd.png)
![Screen Shot 2020-02-17 at 10 07 24 AM](https://user-images.githubusercontent.com/31779424/74673918-76860080-516d-11ea-94ef-d8fc6388e49e.png)

aviveiros11:
Also @RDmitchell reading this again - would you like one toggle to freeze both of the columns? An option for each? Or just the ability to freeze the master column.

RDmitchell:
@aviveiros11 -- The pin feature looks good. 

[...]

PR #2112, opened 2020-02-18
2102 freeze column

aviveiros11:
modifying table capability in detail view 
#### What's this PR do?
Allows the user to pin  the first two columns in detail view
toggle the pin icon on the column header 
#2102


aviveiros11:
### NOTE: looking for advice from @axelstudios on the UX/UI for the user

Issue #1985, opened 2019-10-09
Reconcile SEED with BSync file for Audit Template Tool compatibility


lainsworth8801:
Known issues from seed generated BSync files:
1. Custom ID 1, 2;
2. Need a 'Portfolio Manager Building ID';
3. Some of the elements have been renamed or removed by AT;
4. Address is added on Building level besides site level;
5. SEED's os modeled  measures time series data won't be logged by AT. Only metered (baseline) time series data is;
6. All metered electricity and natural gas time series data is measured at 'Peak Voltage' in kWh and therms, not 'Total Energy' in ktBu;
7. All metered time series data interval frequency type is 'Other', not 'Monthly;
8. All metered time series data will need an unique id to link to its available energy resource to be able to populate on AT; 
9. StartTimeStamp -> StartTimestamp;
10. EndTimeStamp -> EndTimestamp;
11. Missing fields e.g. available energy resources, dates, potential energy for savings even if it's 0, 'LinkedPremisesOrSystem' and UDFs like 'Audit Template Report Type/San Francisco Report' to Report...all indicated by AT's '!'.

Existing AT tool bugs:
1. SF assessor parcel nuber is renamed to 'City Custom Building ID' and it only logs the first 8 characters (the rest gets truncated). However PDF shows the entire string;
2. Meter Years Start Dates is hard double-quoted in BSync file. If removed, 'Utility Data and Benchmarking' goes 400;
3.  Facility Description - Lighting: LED Application Type is marked as a must-have (with a asterisk) for all lighting types besides LED for city reporting;
4. On Safari, if contact info is missing, PDF files won't get generated. This issue goes away by clicking one of: the 'save' button for 'Add notes about this building' in 'Building information', or the 'save' buttons for 'Additional Comments' in either 'Audit Details' and 'Audit Team and Building Staff'. Sometimes these 'save' buttons redirect to pdf url upon clicking instead of being marked 'saved' (supposed case);
5. Same issue is shown on Chrome but differently: if contact info is not all the way filled out, pdf won't generated unless click on the 'save' buttons for 'Additional Comments' in either 'Audit Detail' or 'Audit Team and Building Staff' ('save' in 'Building Information' doesn't work in this case). Pdfs will then be pushed to generation, but not re-directed;
6. 4 and 5 go away if contact info is complete.

lainsworth8801:
Issues are reported to AT help desk.

lainsworth8801:
[...]

Issue #1935, opened 2019-08-02
BuildingSync V2.0 Updates


nllong:
Add new scale of application per https://github.com/BuildingSync/schema/pull/122
Timestamp: https://github.com/BuildingSync/schema/pull/133

nllong:
I think the test for this would be simply the ability to update this file:

You'll need to unzip first.
[buildingsync_ex01_measures.xml.zip](https://github.com/SEED-platform/seed/files/4646591/buildingsync_ex01_measures.xml.zip)

RDmitchell:
Instance: dev1
SHA: 
Org: LBNL 200

Imported the BuildingSync V2 file linked in this issue.

Here is the default mapping -- I am wondering where these defaults are set? It seems like the fields that are mapped to "field style names" might want to be changed to something else, ie, "floors_above_grade" should probably be "Floors Above Grade". And they are not editable.
![image](https://user-images.githubusercontent.com/6314950/82977904-12f34680-9f98-11ea-8dde-16d4906edaff.png)

macintoshpie:
Thanks for testing things!

> Here is the default mapping -- I am wondering where these defaults are set? It seems like the fields that are mapped to "field style names" might want to be changed to something else, ie, "floors_above_grade" should probably be "Floors Above Grade". And they are not editable.

These defaults are set on the backend, ie the user cannot change this in the version of SEED you are looking at. However, we just added a new feature where users can change these header values using column mapping presets. You can then format/change them however you like. See here: https://github.com/SEED-platform/seed/pull/2213. Chatting with Adrian, it sounds like this will be available on dev1 soon.

> Looking at the XML file, seems like the field mapped to "premise_identifier" is a

Correct, it gets premise_identifier from Assessor parcel #. I'm happy to change this for the non-BRICR parser, @nllong do we want to use something else here?

> I didn't find anywhere in Column Settings or Column Mappings where these default BuildingSync mappings are defined. Maybe I am missing something?

You're not missing anything. In this first implementation we were not allowing users to change anything from the mappings. This was solely to allow v2.0 files and to begin integrating the BuildingSync import flow with the other file types.

> Also, SEED didn't pick up the UBID field that is in the BuildingSync file

@nllong is this something we want to start doing for BuildingSync files? If so let's create an issue

> It also didn't pick up the latitude and longitude values from the XML file

[...]

PR #2128, opened 2020-02-27
Feat/bsync parser

macintoshpie:
The BuildingSync parser and general import flow needs to be refactored to handle multiple schema versions.
This is one part of of a couple steps to improve the import process of these files
#### What's this PR do?
- Replaces the BuildingSync parser and exporter with more generalized solutions, allowing us to handle different schema versions
- allows uploading and exporting of v2.0 buildingsync files
- adds new application scales to property measures from bsync schema v2 (common areas, tenant areas, entire building)

NOTE: this will BREAK the exporting of files already on BRICR. After merging this PR, you must run the transformations on the file in the repo [here](https://github.com/BuildingSync/transformations).

##### Setup
Pull this branch. A python package and migration was added so apply migrations and reinstall requirements.
##### Test
- upload ex_1_no_schemaLocation.xml (this branch at seed/building_sync/tests/data). This file is missing the expected BuildingSync schema URL so it should fail with an error about no schema location being found
- uploade ex_1_no_address.xml. The file is missing address info, so should make it past the mapping page, but fails once you try to map it.
- upload buildingsync_v2_0_bricr_workflow.xml
    - check property info looks correct
    - check scenarios are there, check meter data
    - export the BuildingSync file, there should be no significant differences between the original and the exported file
- upload buildingsync_ATT_export.xml; tests that we can upload files that were exported from Audit Template Tool.
    - verify property info looks correct
    - export BuildingSync file, there should be no significant differences between the original and the exported file
- upload a file with non-required fields (ex1.xml, it has lat and long), then _updating_ that property with another file which doesn't have those fields (ex1_no_lat_long.xml). Export the property and it should have the lat and long in the xml. Then run the exported file thorough a schema validator with schema 2.0 (this schema can be found at /seed/building_sync/schemas, and if on OSX you can run `xmllint <exported_file> --schema <schema file>`). This is testing that we can pull data from the property state and that we can build valid xml.
- upload a NON buildingsync property, then export it as a BuildingSync xml. Verify that any property-state data the source file included is also a part of the exported xml.
#1935 
#1985


macintoshpie:
currently I just have 3 buttons for uploading different buildingsync files. Looking for UI/design feedback. @nllong @axelstudios 
Screen Shot 2020-02-27 at 2 12 27 PM

RDmitchell:
@macintoshpie / @adrian-lara / @axelstudios 

The most recent change to this modal was to reduce the number of buttons. 

It seems like having 3 buttons for building sync import files is not great. And I don't think a lot of people will be importing building sync files. And most people won't know the difference between the different options anyway. 

Is there some way that there could be a 2nd level to the building sync file import to specify the file version, so just keep one building sync import button, and in the next step pick the version (with some help as to what each version represents?)

seed/models/building_file.py
```diff
@@ -45,6 +46,16 @@ class BuildingFile(models.Model):
         (HPXML, 'HPXML')
     )
 
+    BUILDINGSYNC_V1 = 1
+    BUILDINGSYNC_V2 = 2
+    BUILDINGSYNC_V2_ALT = 3
```
nllong: what is the meaning of V2_Alternative? Why is this different than V2?

seed/building_sync/mappings.py
```diff
@@ -0,0 +1,976 @@
...
+                        'type': 'value',
+                        'value': '@IDref'
+                    }
+                }
+            }
+        }
+    }
+}
+
+BASE_MAPPING_V2_ALT = {
```
nllong: can we just name this V2 and deprecate to V2_ALT?

seed/models/property_measures.py
```diff
@@ -57,6 +60,9 @@ class PropertyMeasure(models.Model):
         (SCALE_MULTIPLE_PREMISES, 'Multiple premises'),
         (SCALE_ENTIRE_FACILITY, 'Entire facility'),
         (SCALE_ENTIRE_SITE, 'Entire site'),
+        (SCALE_ENTIRE_BUILDING, 'Entire building'),
+        (SCALE_COMMON_AREAS, 'Common areas'),
+        (SCALE_TENANT_AREAS, 'Tenant areas'),
```
nllong: I think you need to run `./manage.py makemigrations` to create the updates needed for these changes.

seed/building_sync/mappings.py
macintoshpie: The problem is the different locations for some address info, this is why there are two mappings. e.g. some files put address_line_1 at `./Address/StreetAddressDetail/Simplified/StreetAddress`, while others have it at `./Buildings/Building/Address/StreetAddressDetail/Simplified/StreetAddress`. I guess this is part of a broader question of how we want to handle these slight but predictable structural differences. Something worth discussing on Thursday

seed/models/building_file.py
macintoshpie: See my comment on deprecating V2.

seed/building_sync/mappings.py
nllong: okay, let's discuss Thursday.

macintoshpie: After discussing, we found out that the current versions in BuidingSync were actually v2.0-pr1. The mapping names have been updated accordingly.

seed/building_sync/building_sync.py
```diff
@@ -5,230 +5,139 @@
...
+        self.version = self._parse_version(default=self.BUILDINGSYNC_V2_PR1)
+
+        # if the namespace map is missing the auc prefix, fix the tree to include it
+        if self.element_tree.getroot().nsmap.get('auc') is None:
+            self.fix_namespaces()
+
         return True
 
-    def export(self, property_state, process_struct=ADDRESS_STRUCT):
+    def fix_namespaces(self):
```
macintoshpie: Any better ideas what to do here? The issue is that a user might inculde the right namespace, but not have the prefix as auc, e.g. `xmlns:other_ns="<correct buildingsync uri>"`. I was only able to dig up [this reference](https://bugs.launchpad.net/lxml/+bug/555602) for changing namespace prefixes, and the only thing I got to work was just recreating the tree.

This solution just copies over the entire tree as best as it knows how. The problem with this is I think comments will get dropped. If that's a problem I can spend more time investigating

seed/building_sync/building_sync.py
```diff
@@ -5,230 +5,139 @@
...
             raise Exception("File not found: {}".format(filename))
 
+        # save element tree
+        with open(filename) as f:
+            self.element_tree = etree.parse(f)
+
+        # TODO: once xml translator has been implemented and used to convert
+        # files from 2.0-pr1 to 2.0, REMOVE the default argument so it fails
+        # if no proper version is found
+        self.version = self._parse_version(default=self.BUILDINGSYNC_V2_PR1)
```
macintoshpie: For now I'm allowing us to import files (either preexisting SEED bsync files _or_ newly uploaded files)  that don't have a schemaLocation that matches our expectations, and just uses v2.0-pr1 if it can't find one.

Just wanted to confirm that that's what we want to do for now until we are able to translate the files to 2.0.

macintoshpie: Decided that preexisting SEED files will be assumed to be 2.0-pr1 until the transformation is implemented. Also, any _newly_ imported files will fail to be imported if they don't have a parsable version.

seed/migrations/0110_auto_20200323_1509.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
adrian-lara, line 1: @macintoshpie think you could shift this migration to the end and delete `seed/migrations/0111_merge_20200324_0837.py`?

seed/lib/xml_mapping/mapper.py
```diff
@@ -8,115 +8,15 @@
 from __future__ import absolute_import
```
adrian-lara, line 8: Since this has gotten so slim and imports heavily from building_sync.mapping, what's your thoughts on bringing this into the building_sync/mappings.py file. 

Per our conversation - we can probably do this later, but leaving this here as a note/documentation.

seed/building_sync/mappings.py
```diff
@@ -0,0 +1,1005 @@
...
+        last_element[:] = sorted(last_element, key=getkey)
+    else:
+        # element already exists, update its value
+        update_element(last_element, target, value)
+
+
+# Base mapping for BuildingSync schema version 2.0-pr1
+# NOTE: this is actually a mix of 2.0-pr1 and 1.0 as it uses 'TimeStamp' rather than 'Timestamp'
+# and it allows the 'Reports' element to exist or not exist.
+BASE_MAPPING_V2_PR1 = {
```
adrian-lara, line 392: What do you think about splitting these mappings out into their own files for organization purposes?

seed/building_sync/building_sync.py
```diff
@@ -237,534 +154,185 @@ def export(self, property_state, process_struct=ADDRESS_STRUCT):
...
+            seed_scenario['annual_site_energy_use_intensity'] = scenario['annual_site_energy_use_intensity']
+            seed_scenario['annual_source_energy'] = scenario['annual_source_energy']
+            seed_scenario['annual_source_energy_use_intensity'] = scenario['annual_source_energy_use_intensity']
+            seed_scenario['annual_electricity_energy'] = scenario['annual_electricity_energy']
+            seed_scenario['annual_peak_demand'] = scenario['annual_peak_demand']
+            seed_scenario['annual_natural_gas_energy'] = scenario['annual_natural_gas_energy']
+            seed_scenario['measures'] = [id['id'] for id in scenario['measure_ids']]
+
+            seed_scenario['meters'] = list(meters.values())
+            scenarios.append(seed_scenario)
```
adrian-lara, line 244:
```suggestion
            scenario['meters'] = list(meters.values())
            scenario['measures'] = [id['id'] for id in scenario['measure_ids']]
            
            # remove key values that are no longer needed
            del scenario['time_series']
            del scenario['resource_uses']
            del scenario['measure_ids']
            del scenario['is_virtual']            
                        
            scenarios.append(scenario)
```

I'm thinking that you could trim the existing scenario dict and add key values as needed rather than construct a new scenario dict. Potentially, it could look like this but definitely let me know if I'm missing something both in general or in the key/values needed in the proposed change.

seed/building_sync/building_sync.py
```diff
@@ -237,534 +154,185 @@ def export(self, property_state, process_struct=ADDRESS_STRUCT):
...
+            # process the scenario meters (aka resource uses)
+            meters = {}
+            for resource_use in scenario['resource_uses']:
+                meter = {}
+                meter['source'] = Meter.BUILDINGSYNC
+                meter['source_id'] = resource_use['source_id']
+                meter['type'] = resource_use['type']
+                meter['units'] = resource_use['units']
+                meter['is_virtual'] = scenario['is_virtual']
+                meter['readings'] = []
```
adrian-lara, line 208: You could also set these key values during the construction of the meter dict. I'd imagine it'd be faster to set them all at once if possible, but that's just my guess.
```
                meter = {
                    'source': Meter.BUILDINGSYNC,
                    ...
                }
```

seed/building_sync/building_sync.py
```diff
@@ -237,534 +154,185 @@ def export(self, property_state, process_struct=ADDRESS_STRUCT):
...
+                meters[meter['source_id']] = meter
+
+            # process the scenario meter readings
+            for series_data in scenario['time_series']:
+                reading = {}
+                reading['start_time'] = series_data['start_time']
+                reading['end_time'] = series_data['end_time']
+                reading['reading'] = series_data['reading']
+                reading['source_id'] = series_data['source_id']
+                reading['source_unit'] = meters[reading['source_id']].get('units')
```
adrian-lara, line 219: similar thought as the meter dict logic

seed/building_sync/building_sync.py
adrian-lara: Might be better answered by @nllong or someone else that understands BuildingSync files better.

seed/building_sync/building_sync.py
macintoshpie: I think your approach is definitely correct and simpler. My thought process was that parsing is already kind of magic and loosey goosey (we're using dicts for standardized structure!) I wanted to be as explicit as possible here about what we expect our incoming data to look like and what our outgoing data should contain, so I might prefer keeping it very explicit

seed/building_sync/building_sync.py
```diff
@@ -237,534 +154,185 @@ def export(self, property_state, process_struct=ADDRESS_STRUCT):
...
+            seed_scenario['annual_natural_gas_energy'] = scenario['annual_natural_gas_energy']
+            seed_scenario['measures'] = [id['id'] for id in scenario['measure_ids']]
+
+            seed_scenario['meters'] = list(meters.values())
+            scenarios.append(seed_scenario)
+
+        res = {'measures': measures, 'scenarios': scenarios}
+        # property fields are at the root of the object
+        for k, v in result['property'].items():
+            res[k] = v
```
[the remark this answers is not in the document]
macintoshpie: I think that's a good method. Now, looking at my comment above about explicitness, I wonder why I did it this way...

seed/building_sync/building_sync.py
```diff
@@ -6,229 +6,146 @@
...
+        a custom mapping.
 
         :param property_state: object, PropertyState to merge into BuildingSync
-        :param process_struct: dict, mapping from PropertyState to BuildingSync
+        :param custom_mapping: dict, user-defined mapping (used with higher priority over the default mapping)
         :return: string, as XML
         """
-
-        # if property state is not defined, then just return the BuildingSync unparsed
         if not property_state:
```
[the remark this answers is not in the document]
macintoshpie: Yes, this condition was already there. I'm actually not sure when export would ever get called without a property state, and I don't see that case in the code anywhere. @nllong do you recollect why this case exists?

seed/building_sync/building_sync.py
```diff
@@ -237,534 +154,185 @@ def export(self, property_state, process_struct=ADDRESS_STRUCT):
...
-        it will only set the first path for now. The future could allow a variable to pass in the
-        index (or other constraint) before setting the value.
-
-        :param path: string, path which to navigate to set the value
-        :param data: dict, dictionary to process
-        :param value: value to set, could be any type at the moment.
-        :return: boolean, true if successful
-        """
+            if value is None:
+                continue
```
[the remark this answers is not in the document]
macintoshpie: Yes, I think that's correct. My understanding of the way the original exporter worked is that we only ever update or create values in the xml, we never remove elements. @nllong can you comment if this is the desired behavior?

seed/building_sync/building_sync.py
```diff
@@ -6,229 +6,146 @@
...
+        # - else, ignore it
+        base_path = merged_mappings['property']['xpath']
+        field_mappings = merged_mappings['property']['properties']
+        for field, mapping in field_mappings.items():
+            if mapping['xpath'].startswith('./'):
+                mapping_path = mapping['xpath'][2:]
+            else:
+                mapping_path = mapping['xpath']
+            absolute_xpath = os.path.join(base_path, mapping_path)
+
```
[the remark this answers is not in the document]
macintoshpie: I think that's a good call 👍

seed/building_sync/building_sync.py
```diff
@@ -6,229 +6,146 @@
...
-                    ]
-                )
-
-        for field, v in process_struct['return'].items():
+            return etree.tostring(self.element_tree, pretty_print=True).decode()
+
+        if not self.element_tree:
+            self.init_tree(version=BuildingSync.BUILDINGSYNC_V2_0)
+
+        merged_mappings = merge_mappings(self.VERSION_MAPPINGS_DICT[self.version], custom_mapping)
```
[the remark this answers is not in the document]
macintoshpie: That's a good question, I can think of two cases. For the export tests I don't pass a custom mapping because it's simpler. Also, if a user has never imported a buildingsync file, but wants to export one of their properties as buildingsync, they will not have a mapping saved (this happens only when importing a bsync file), so their custom mapping would be `None` as well.

seed/building_sync/mappings.py
```diff
@@ -0,0 +1,694 @@
...
+    if '[' in tag:
+        tag, xpath_condition = tag.split('[')
+        child_tag, child_value = xpath_condition.replace(']', '').split('=')
+        child_tag = child_tag.replace('auc:', '')
+        child_value = child_value.replace('"', '')
+
+    return tag, child_tag, child_value
+
+
+def _build_path(element, xpath_list):
```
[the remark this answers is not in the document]
macintoshpie: I think I'd prefer to keep it as just a string since that's how we pass and store xpaths elsewhere, might be a little more consistent.

seed/building_sync/mappings.py
```diff
@@ -0,0 +1,694 @@
...
+        raise Exception(f'Unable to find path in schema: "{element_path}"')
+
+    ordered_children = [child.name for child in schema_element.iterchildren()]
+
+    # construct a function for sorting an element's children by returning the index of the
+    # child according to the ordering of the schema
+    def _getkey(element):
+        if isinstance(element, etree._Comment):
+            # put comments at the end
+            return 100
```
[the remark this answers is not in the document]
macintoshpie: Haha that's a good point, let's make it reeeal big

seed/lib/xml_mapping/mapper.py
macintoshpie: I think I'll hold off on this, but do agree that it would be nice to consolidate and reorganize a few of these files

seed/building_sync/mappings.py
macintoshpie: Agreed, though holding off for now

seed/models/building_file.py
```diff
@@ -100,12 +100,16 @@ def process(self, organization_id, cycle, property_view=None):
...
-        if self.file_type == self.BUILDINGSYNC:
-            parser_args.append(BuildingSync.BRICR_STRUCT)
-        data, messages = parser.process(*parser_args, **parser_kwargs)
+        try:
+            parser.import_file(self.file.path)
+        except Exception as e:
+            return False, None, None, {'errors': [str(e)], 'warnings': []}
+
+        kwargs = {}
+        if self.file_type == 'BuildingSync':
```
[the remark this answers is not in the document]
macintoshpie: good catch!

seed/building_sync/building_sync.py
nllong: From my understanding the namespace monikers can be mapped to whatever we choose (e.g., auc, n1, n2, etc). I'm not familiar with the impacts of this though. I would like to keep the comments if possible.

seed/building_sync/building_sync.py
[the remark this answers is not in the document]
nllong: no recollection.

seed/building_sync/building_sync.py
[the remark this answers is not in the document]
nllong: Yes, previously we did not delete elemetns in the XML. Only create or update. 

I am not sure we want to deal with deletions at the moment.

seed/building_sync/mappings.py
```diff
@@ -0,0 +1,733 @@
+import copy
+from datetime import datetime
+
+import pytz
+from lxml import etree
+
+
+BUILDINGSYNC_URI = 'http://buildingsync.net/schemas/bedes-auc/2019'
```
nllong, line 8: I'm guessing that this URI is going to be updated to 2020 when we release 2.1/3.0 this year (June). Any issues with that?

seed/building_sync/mappings.py
```diff
@@ -0,0 +1,733 @@
...
+# -- GENERAL functions
+#
+def table_mapping_to_buildingsync_mapping(table_mapping):
+    """Converts a mapping returned from ColumnMapping.get_column_mappings_by_table_name
+    into the structure expected by the BuildingSync functions
+
+    :param table_mapping: dict, a table mapping
+    :return: dict, a buildingsync style mapping
+    """
+    # expected structure of table_mapping:
```
nllong, line 54: move this into the docstring.

seed/building_sync/mappings.py
```diff
@@ -0,0 +1,733 @@
...
+
+
+# Base mapping for BuildingSync schema version 2.0
+BASE_MAPPING_V2_0 = {
+    'property': {
+        'xpath': '/auc:BuildingSync/auc:Facilities/auc:Facility/auc:Sites/auc:Site',
+        'type': 'object',
+        'properties': {
+            'address_line_1': {
+                'xpath': './auc:Buildings/auc:Building/auc:Address/auc:StreetAddressDetail/auc:Simplified/auc:StreetAddress',
```
nllong, line 435: sweet xpaths!

seed/building_sync/mappings.py
```diff
@@ -0,0 +1,733 @@
...
+            },
+            'time_series': {
+                'xpath': './auc:TimeSeriesData/auc:TimeSeries',
+                'type': 'list',
+                'items': {
+                    'start_time': {
+                        'xpath': './auc:StartTimestamp',
+                        'type': 'value',
+                        'value': 'text',
+                        'formatter': to_datetime
```
nllong, line 711: i like these formatters

seed/building_sync/tests/test_buildingsync.py
```diff
@@ -0,0 +1,192 @@
+# !/usr/bin/env python
+# encoding: utf-8
+"""
+:copyright (c) 2014 - 2019, The Regents of the University of California, through Lawrence Berkeley National Laboratory (subject to receipt of any required approvals from the U.S. Department of Energy) and contributors. All rights reserved.  # NOQA
```
nllong, line 4: fix copyright to 2020

seed/data_importer/tests/integration/test_data_import.py
```diff
@@ -154,8 +154,8 @@ def setUp(self):
         self.maxDiff = None
 
         # setup the ImportFile for using the example zip file
-        filename = 'example-bsync-single.zip'
-        filepath = osp.join(osp.dirname(__file__), '..', 'data', filename)
+        filename = 'ex_1_and_buildingsync_ex01_measures.zip'
+        filepath = osp.join(BASE_DIR, 'seed', 'building_sync', 'tests', 'data', filename)
```
nllong, line 158: If `BASE_DIR` is the best approach to manage dirs, then we should update this everywhere. But not now.

seed/building_sync/mappings.py
macintoshpie: That will probably a bit of refactoring, but I don;'t think it should be too bad. I will hold off on making those changes for now

nllong:
👍

Issue #2127, opened 2020-02-26
Add functionality to see entire SEED HEADER field name in Mapping


RDmitchell:
### Expected Behavior
When reviewing the Mapping fields, it would be nice to be able to see the entire field name
### Actual Behavior
Right now it is not possible to pull the SEED HEADER right hand column border to the right to expand the width to see the entire name for long field names.
![image](https://user-images.githubusercontent.com/6314950/75383301-f65c4b00-5890-11ea-9652-f0110ab81247.png)

It would be nice if either 
* the program automatically sized the SEED HEADER column width to fit the longest field name (although that may not be a good solution because you don't know what the user might type in by hand).
* allow the user to pull the right hand side of the SEED HEADER column to the right to expand the width so that they can make it the right size to see the complete text of all the fields. 

This is for the case when the user is actively mapping their file.

The other case for this is when viewing the field mapping after the file has been imported -- this is another case where it would be nice to be able to see the full field name. Here is an example where that is a problem -- but fixing this doesn't have to be included in this issue -- I can make a separate issue for that functionality if that is desired.
![image](https://user-images.githubusercontent.com/6314950/75384449-fd845880-5892-11ea-8ff2-a7116d301c8c.png)

### Steps to Reproduce
Import an ESPM file with all the standard 250 fields many of which have really long names that include units at the end. 
### Instance Information
instance: dev1
SHA: c97042d

RDmitchell:
instance: dev1
SHA: 9f8ec3a7

Works great -- I imported an ESPM file with some very long fields names and I could stretch the field column to see them in their entirety.

Thanks !

![image](https://user-images.githubusercontent.com/6314950/77794953-fd11e580-7029-11ea-99b3-179a7fea7821.png)

PR #2129, opened 2020-02-28
Added resize func with HEAVY help from Alex

aviveiros11:
Addressing column resizing based on window width
#### What's this PR do?
Allows the user to now resize column regardless of window width
Click and drag the column header to the desired length
Directly address #2127


nllong:
@axelstudios did you need to check something on this before merging?

Issue #2131, opened 2020-03-02
Bricr-dev testing for potential merge into develop


adrian-lara:
### Steps to Reproduce
Using the files contained here: https://drive.google.com/drive/folders/1w9d1mX6MxMaZi0fsTP3HX_WItP2G6WCk

- Upload "Monthly Metrics_Sample Data Request Response - 4 Properties" the "Upload a Spreadsheet" button.On the mapping page, use the "Copy Data File Headers..." button but change the first column mapping to "PM Property ID". 
- Upload the "example-pm-monthly-meter-usage.xlsx" using the "Upload Portfolio Manager Meter Usage" button.
- Upload the "buildingsync_v2_0_bricr_workflow.xml" file using "Upload BuildingSync Data" button on the datasets page
- Visit one of the inventory detail pages of the "4 properties" and upload the same BuildingSync file using "Actions" -> "Update with BuildingSync".

Also, after merging PM and BS records both having meters (via inventory list page, check/choose records, "Actions" -> "Merge selected", there may also be issues with how Meters and Meter Readings are being merged.

### Branch Information
`chore/merge-bricr-dev` - `bricr-dev` that's had `develop` merged into it

macintoshpie:
I was able to replicate the bug. In the process I've discovered there are two issues with some of our BuildingSync code:
1. If you only upload the `buildingsync_v2_0_bricr_workflow.xml` file and go to the meters page, you'll see two meters in the table, but 6 meters listed in the filter tab. The issue is due to how we "export" meter usage here: https://github.com/SEED-platform/seed/blob/b80d9a5011fb7f547d8466873d0dd05ea9c0c8f6/seed/utils/meters.py#L105
The `field_names` are not unique for the different meters of the file, as it seems we assume ResourceUse ID to be a unique identifier within the scope of a property.

2. After uploading the `Monthly Metrics_Sample Data Request Response - 4 Properties` and `example-pm-monthly-meter-usage.xlsx`, then uploading the BuildingSync file as a new property. On the properties page, then select one of the properties with meter reading with the buildingsync property. It will fail (currently) b/c buildingsync meter readings don't have the field `conversion_factor` set. If this is changed to default to 1 when importing the file, it you should be able to merge the properties.
[...]

PR #2132, opened 2020-03-03
Fix/bsync meterreading import

macintoshpie:
#### What's this PR do?
- makes MeterReading's `conversion_factor` a required field. Any existing meter readings which have a NULL value and have kbtu for units will have conversion factor set to 1.0. If the migration finds a record that has a NULL value and it's units are not kbtu, it raises an exception.
- Fixes example file (buildingsync_v2_0_bricr-workflow.xml) by making ResourceUse IDs unique
- Fixes merging of properties (Actions -> Merge Selected) with BuildingSync files
  - Merging previously assumed all meter readings had converion_factor field set, but imported building files did not. Fixed by setting it to 1.0 for all meter readings (assumed kBtu)
  - fixes bug where meter-scenario information was dropped when merging properties
https://drive.google.com/drive/folders/1w9d1mX6MxMaZi0fsTP3HX_WItP2G6WCk
1. Upload properties along with meter data (Monthly Metrics_Sample Data Request Response - 4 Properties  &  example-pm-monthly-meter-usage).
2. Upload a BuildingSync file (buildingsync_v2_0_bricr-workflow.xml)
2. From the Properties listing, select a property and the building sync property and click Actions -> Merge Selected. It should be successful, and you should verify that on the meters page:
  - meter data from both files is present.
  - scenario information for the building sync meters is retained (click the filter by, and select 'scenario')
#2131


seed/models/properties.py
```diff
@@ -609,7 +611,7 @@ def coparent(cls, state_id):
         return coparents, len(coparents)
 
     @classmethod
-    def merge_relationships(cls, merged_state, state1, state2):
+    def merge_relationships(cls, merged_state, state1, state2, property_id=None):
```
adrian-lara:
```suggestion
    def merge_relationships(cls, merged_state, state1, state2):
```

seed/models/scenarios.py
```diff
@@ -91,11 +91,17 @@ def copy_initial_meters(self, source_scenario_id):
         from seed.models.meters import Meter
 
         source_scenario = Scenario.objects.get(pk=source_scenario_id)
+        try:
+            property_ = PropertyView.objects.get(state=self.property_state).property
```
adrian-lara: Potentially no change here - just curious when a -State we care about doesn't have a corresponding -View? Is this a mid-merge type of situation?

seed/models/scenarios.py
```diff
@@ -91,11 +91,17 @@ def copy_initial_meters(self, source_scenario_id):
...
+        except PropertyView.DoesNotExist:
+            property_ = None
 
         for source_meter in source_scenario.meter_set.all():
             # create new meter and copy over the readings from the source_meter
             meter = Meter.objects.get(pk=source_meter.id)
             meter.pk = None
             meter.scenario_id = self.id
+            if property_ is not None:
+                meter.property = property_
```
adrian-lara: I think having a meter without a property association makes sense in the bigger picture, but as it currently stands, I've never seen a meter without a property associated to it. Maybe this is something we chat about in person.

One thought: If we don't have a related property, would the newly created meter be visible on the frontend?

seed/tests/test_property_views.py
```diff
@@ -517,6 +518,79 @@ def test_properties_merge_without_losing_meters_when_some_meters_from_same_sourc
         # Overlapping reading that wasn't prioritized should not exist
         self.assertFalse(MeterReading.objects.filter(reading=property_2_reading).exists())
 
+    def test_properties_merge_combining_bsync_and_pm_sources(self):
```
adrian-lara: Testing! Thanks 🙏

seed/views/properties.py
```diff
@@ -480,8 +480,32 @@ def merge(self, request):
...
+            # Find unique notes for the views for later
+            notes = list(Note.objects.values(
+                'name', 'note_type', 'text', 'log_data', 'created', 'updated', 'organization_id', 'user_id'
+            ).filter(property_view_id__in=view_ids).distinct())
+
+            cycle_id = views.first().cycle_id
+
+            # create new property state for merging into
             merged_state = state.objects.create(organization_id=organization_id)
+            merged_state.save()
```
adrian-lara: Do we need the .save() here?

seed/models/scenarios.py
adrian-lara: Per conversation, we'll removing this conditional

seed/models/scenarios.py
adrian-lara: Per conversation, we'll replace the exception with a informative comment. But we do want to keep it here in case there are cases we didn't consider.

seed/models/meters.py
```diff
@@ -104,11 +104,12 @@ def copy_readings(self, source_meter, overlaps_possible=True):
         bulk_create is used.
         """
         if overlaps_possible:
-            reading_strings = [
-                f"({self.id}, '{reading.start_time}', '{reading.end_time}', {reading.reading}, '{reading.source_unit}', {reading.conversion_factor})"
-                for reading
-                in source_meter.meter_readings.all()
-            ]
+            reading_strings = []
```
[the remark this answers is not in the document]
adrian-lara: per conversation, we'll revert this back and have an error thrown in the really low chance that  conversion_factor = null

adrian-lara:
@nllong - Tagging you here as I don't have merge auth for `bricr-dev`

Issue #2114, opened 2020-02-19
Deleting a rule after saving previous changes does not reactive save buttons


lainsworth8801:
### Expected Behavior
Any changes including deleting a rule/rules should activate save buttons for user to decide whether to save changes or not.

### Actual Behavior
"Save settings" button stays grayed out after user deletes a rule if user has previously made and saved changes.

### Steps to Reproduce
1. Delete a rule, hit "save";
2. Delete another rule, "save" button stay gray out with a check mark.

### Instance Information
![Screen Shot 2020-02-19 at 9 00 24 AM](https://user-images.githubusercontent.com/49968203/74850228-f12e5780-52f6-11ea-8f52-16edb6dc6e4b.png)

![Screen Shot 2020-02-19 at 9 02 07 AM](https://user-images.githubusercontent.com/49968203/74850237-f390b180-52f6-11ea-8444-17008d735ab2.png)

Server Instance: `dev1, dev2, production, staging, etc`
Server SHA: `abcd123`

RDmitchell:
instance: dev1
SHA: 9f8ec3a
Org: LBNL 200

When I delete a rule, the Save Changes button remains grayed out. Seems like each time I made a change I should have the option to change it. 

This does not appear to be fixed so I am reopening this and putting it into In Progress

Here are my steps:
* Go to Tax Lot Tab
* Click "Restore Default Rules", which puts up the 2 default rules, the Save Changes button is grayed out
![image](https://user-images.githubusercontent.com/6314950/77796086-e53b6100-702b-11ea-879a-825e2a320161.png)
* I delete the first rule, Address Line 1, and would like to save that change but the "Save Changes" button is still grayed out so I can save the change
![image](https://user-images.githubusercontent.com/6314950/77796294-3ea39000-702c-11ea-9b5d-decc1d678e8e.png)
* I then delete the last rule, and the Save Changes button is still grayed out. I thought when I tested this before, after I deleted the 2nd rule, the Save Changes button was active, but I am not seeing this now.

RDmitchell:
@lainsworth8801 -- also, on the Tax Lot tab, if I click the Restore the Defaults button, the Save Changes field is still grayed out. So I can't save that change either

RDmitchell:
And if I uncheck the boxes, so that the rules are there but not active, the Save Changes button is still grayed out. 

If I go to the Properties tab and then back to the Tax Lot tab, my changes were not saved, and the rules remain checked.

[...]

Issue #2020, opened 2019-11-27
DQ: Defining multiple rules for the same field is limited


RDmitchell:
### Expected Behavior
I would like to be able to set multiple DQ rules for the same field
### Actual Behavior 
Because of the way that the DQ rule definitions are currently constructed, I don't believe there is a way to do what I want to do. 
### Steps to Reproduce
See this doc
https://docs.google.com/document/d/1uOgXDJQtYf4XcTg57AYPE_TKQtpw8agg29pZgNRW628/edit?usp=sharing
### Instance Information
Instance: dev1
SHA: 6dad9f771

lainsworth8801:
@RDmitchell talked with team and decided that having 2 different criteria, required/not null and not required/null, for the same rule does not seem to make sense. I added an extra column for filtering null field value if users' desire is to separate low, high and null field data (field data is 'required' or 'not null' but field includes null data instead). This will have all case scenarios reported separately and labels should be applied accordingly. There's a PR created for this along with #1614 and #1615 and is await for review.

RDmitchell:
@lainsworth8801 -- If I have a rule that is set up as shown below, does the label get added when those criteria are NOT met, ie, if the field does not contain the text "OK" then the program says that it is an error and applies the label to the record, correct? (I assume that is the case but wanted to check).

![image](https://user-images.githubusercontent.com/6314950/85802287-81433a00-b6f9-11ea-8205-e353e6b2066d.png)

RDmitchell:
@lainsworth8801 / @adrian-lara -- if you have never imported a file into SEED, how does it know what fields are available for defining the DQ rules? Right now, it seems that if you are importing a file with fields in it that SEED doesn't know about, there is not a way to define DQ rules for those fields until you have imported the file once ? 

I guess my main question is where does the list of fields come from in the DQ rule definition screen?

lainsworth8801:
> @lainsworth8801 -- If I have a rule that is set up as shown below, does the label get added when those criteria are NOT met, ie, if the field does not contain the text "OK" then the program says that it is an error and applies the label to the record, correct? (I assume that is the case but wanted to check).
> 
> ![image](https://user-images.githubusercontent.com/6314950/85802287-81433a00-b6f9-11ea-8205-e353e6b2066d.png)

Correct as this is an error severity level. If it was set "Valid", label would be applied if it contains the text "OK".

lainsworth8801:
[...]

Issue #1615, opened 2018-04-17
DQ Required vs Not Null


RDmitchell:
The issue is what is the difference between Required and Not Null. For a specific data file, if the DQ checking is done on import, the Required field makes sense, ie, check to make sure that the fields marked "Required" are in the imported file. 

For the case where the DQ checking is done on the inventory list, the Required field may not actually make much sense, at least on the level of checking each record. Presumably the field is either there or not for all the records in the Inventory List (?).

### Instance Information
instance: LBNL production
SHA: 36cc59d

nllong:
Required and Not Null are quite similar. From [this chunk of code](https://github.com/SEED-platform/seed/blob/develop/seed/models/data_quality.py#L655-L668) it looks like if the field is just missing and it is required, then it add the "missing field". If the field does exist and it is either None or empty string, then it will either be "value is missing and none/empty" or "value is null". 

I don't know the best way to resolve this without a couple specific cases that we want to test and the result that we want for each of them. Maybe we just get rid of not null?

RDmitchell:
I think users want to know if a field that they want to have data in doesn't have data in it, so in some ways, it might be better to get rid of the "Required" and just leave the "Not Null".

But before we do anything let me talk to the users who were having difficulty with this and see if I can get some specific test cases from them.

nllong:
@RDmitchell -- can you verify that this is still up for discussion?

RDmitchell:
Am getting feedback from users, will let you know

Ammon-R:
@RDmitchell @nllong - Figured I'd chime in here rather than an email. I agree with Robin, removing the "required" column and leaving "not null" would probably be the best. From a benchmarking perspective, we are uploading the same data set over an over again and we know what fields are in the benchmarking data set that we are uploading. The "not null" field is more informative since it will tell us if there is an inherent issue with the data. (Example: benchmarks that don't have energy use data will have a null for the EUI but that EUI field will always be in the upload.) I haven't really used SEED for anything other than benchmarking so there might be a use case for the "required" field that I'm not aware of.

BenchmarkDC:
Hi All,

Andrew from DC here. Chiming in to agree with Ammon for the exact same reason: we wouldn't upload a field for one record but exclude it for another, it's always the same spreadsheet. It might be useful to keep around in the case that a user has to upload a spreadsheet that is generated by the user (i.e. data verification forms created by a reporter)? 

[...]

PR #2133, opened 2020-03-05
#1614, #1615, #2020, #2114 - DQ refactor

lainsworth8801:
Added condition check column for null field values filtering.
Currently when multiple rules set up for the same field, properties missing values for this field will be reported as warning or error multiple times.

#### What's this PR do?
1. Removed report redundancy on null values;
2. Applied labels in data quality report:
Screen Shot 2020-05-07 at 12 14 18 PM
3. Removed 'required' and 'not null' columns;
4. Added rule condition check for value existence (required/null), range (min/max), and text (contain/not contain)
5. Removed required/not null checkbox columns and add condition checks for both;
6. Alert users of empty input by outlining text box in red if rule must include/exclude a string;
7. Migration added to update current db. **'Required' and 'not null' fields should not be removed in migration until this PR is approved.**
Screen Shot 2020-03-20 at 9 45 17 AM
Screen Shot 2020-04-30 at 2 30 19 PM

Select a condition for check:
1. Required/Not Null checks for null field values and report once if captured. This option disables min/max value and units columns;
3. If condition checks is selected for range, min and/or max sections are enabled for user to fill in with min/max values;
4. If condition check is selected for 'include' or 'not include', rule checks for included/excluded string.

#1614, #1615, #2020, #2114   

Set up a blank range rule for not null value check for Site EUI:
Screen Shot 2020-03-04 at 6 42 00 PM

Null site eui is reported once for the properties that are missing it along with other properties that have eui but are invalid:
Screen Shot 2020-03-04 at 6 45 28 PM


nllong:
Hey @lainsworth8801, is this PR ready to go or are we going to wait to refactor the DQ rules into individual rules. I'm hoping the former... let me know.

lainsworth8801:
Hey nick this PR is for what I showed you and Nick on Wednesday so sorry
this is still waiting for refactor. As of now I am shooting for the end of
March.

seed/migrations/0123_rule_condition.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('seed', '0122_remove_rule_condition'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='rule',
+            name='condition',
```
nllong: It looks like there are three migrations to add, remove, then add the condition again. Is that required? Seems like 0119 through 0123 can be squashed?

nllong:
> Hey nick this PR is for what I showed you and Nick on Wednesday so sorry this is still waiting for refactor. As of now I am shooting for the end of March.

Oh, okay. I just added the DO NOT MERGE flag. My comment about the migrations is still valid though in my review. Thanks!

seed/migrations/0125_dq_refactor.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 2.2.10 on 2020-04-13 23:03
+
+from django.db import migrations, models
+
+
+def forwards(apps, schema_editor):
```
adrian-lara, line 6: I think we may need to handle cases where a Rule had `min` and/or `max` as well as True for `required` or `not_null`. Similarly, we probably need to do something similar for cases where a Rule had `text_match` as well as True for `required` or `not_null`.

What do you envision this would look like in the new DQ world? Maybe something like: Each of those individual cases gets two new Rules created, one Rule with `condition` of "required" or "not null" and the other Rule with `condition` of "range" or "include"/"exclude".

seed/models/data_quality.py
```diff
@@ -129,31 +135,36 @@ class Rule(models.Model):
             'not_null': True,
             'rule_type': RULE_TYPE_DEFAULT,
             'severity': SEVERITY_ERROR,
+            'condition': RULE_INCLUDE,
```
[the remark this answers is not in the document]
lainsworth8801: 'not_null' in this context came from the original 'not null' check column which is different from the newly defined 'not null' and 'required'. Rule.not_null is to check if the values exists, which disables the min/max column (here for address_line_1 would be 'text must contain'). Since by default, all DEFAULT_RULES check for valid values (valid address line 1 that contains a user defined string). That said, default rules should all check the validity of a field (include, exclude, range), not existence (required, not null). Line 'not_null': True should be removed (leads back to my remove filed question this morning).

seed/models/data_quality.py
```diff
@@ -306,6 +335,8 @@ class Rule(models.Model):
     table_name = models.CharField(max_length=200, default='PropertyState', blank=True)
     field = models.CharField(max_length=200)
     enabled = models.BooleanField(default=True)
+    condition = models.CharField(max_length=200, default='', blank=True)
+    # placeholder = models.CharField(max_length=200, default='(field must contain this text)')
```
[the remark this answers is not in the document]
lainsworth8801: Min/max fields were disabled when condition check select 'required' or 'not null'. Must be some code i accidentally removed. I will double check.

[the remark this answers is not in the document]
adrian-lara: Sorry - I may be missing something. I'm not sure what you mean here though - I was just referring to the commented out `placeholder` attribute.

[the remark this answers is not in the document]
lainsworth8801: My bad I was replying to your one comment before placeholder where min/max values aren't disabled for required condition for address line 1.

> :

seed/models/data_quality.py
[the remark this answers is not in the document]
adrian-lara: Are you saying the Rule.not_null attribute was only being used to disable the min/max column on the frontend? 

My impression is that the effective purpose of this current implementation (of `condition='not null'`) and the previous implementation (`Rule.not_null = True`) should remain the same - to identify when the value for this field is null in the DB for any particular record. Am I misunderstanding the purpose either before or after the changes we have in this PR?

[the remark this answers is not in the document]
lainsworth8801: Rule.not_null and Rule.required are used to: 1. disable range values on the front end and ;2. filter out null values in the model as error fields so that detailed error message only report 'null' error once instead of multiple times in case it falls in other rules for the same field. Previous 'not_null' is equivalent to now 'range', 'include' and 'exclude'.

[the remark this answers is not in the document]
adrian-lara: I'm a little fuzzy on this still. Are you saying that, before, rule.not_null and rule.required were just fallbacks for whenever other criteria like min and max checks or "contain/does not contain" checks were not met? Sorry, I'm not quite sure what you meant by item 2. 

I don't know how the previous set up of rule.not_null being True or False is equivalent to now `condition` being set to "range", "include", or "exclude".

seed/migrations/0125_dq_refactor.py
lainsworth8801: My understanding of the plan is on when min and max are both null, not_null or required condition applies depending on 'required' and 'not null' check box columns. If both are checked, it's 'required' ('required' must also be 'not_null', 'not_null' doesn't have to be 'required'). If a rule has a min and/or max, it'd be checking for assumably existing values which would be checking for range (for numbers, year, eui, etc.), and 'include'/'exclude' a sub string for text ('exclude' as a new condition does not exist in the database yet). That said, 'required' or 'not null' is regardless if min and/or max has/have value(s).

lainsworth8801: Sorry for the confusion. We should revisit this soon as we can (Monday the 20th?).

adrian-lara: Yep - let's chat Monday. That explanation makes sense to me for the cases you presented. 

For the conversation we have on Monday. The cases that I was thinking about was when users had both min and max criteria as well as required (or not_null) criteria for one Rule. I was thinking that in this new DQ setup, we can't just ignore the min and max criteria whenever we have required or not_null criteria for a Rule being converted from the old setup.

seed/migrations/0125_dq_refactor.py
```diff
@@ -0,0 +1,62 @@
...
+    #                |                       |--> text_match not null ---> text_match == '': not_null/required/'' #
+    #                |                                                |--> text_match != '': include              #
+    #                |--> min / max not null ---> range                                                           #
+    # =========================================================================================================== #
+    Rule.objects.filter(data_type=1, text_match=None, not_null=True).update(condition='not_null')
+    Rule.objects.filter(data_type=1, text_match=None, required=True, not_null=False).update(condition='required')
+    Rule.objects.filter(data_type=1, text_match=None, required=False, not_null=False).update(condition='')
+    Rule.objects.filter(data_type=1, text_match='').filter(not_null=True).update(condition='not_null')
+    Rule.objects.filter(data_type=1, text_match='').filter(required=True, not_null=False).update(condition='required')
+    Rule.objects.filter(data_type=1, text_match='').filter(required=False, not_null=False).update(condition='')
```
adrian-lara, line 26: I didn't see records in prod that fulfilled this case, but similar question to the text_match=None version of this query..., is there a point to a Rule meeting this criteria?

seed/migrations/0125_dq_refactor.py
```diff
@@ -0,0 +1,62 @@
...
+    Rule.objects.filter(data_type=1).exclude(text_match=None).exclude(text_match='').update(condition='include')
```
adrian-lara, line 27: I notice we never set the condition to "exclude" in any of these updates. Did SEED not previous have the notion of checking that text **did not** exist for a text field?

seed/migrations/0125_dq_refactor.py
```diff
@@ -0,0 +1,62 @@
...
+    Rule.objects.exclude(data_type=1).exclude(data_type=None).filter(min=None, max=None, required=False, not_null=False).update(condition='')
+    Rule.objects.filter(data_type=0).exclude(min=None, max=None).update(condition='range')
+    Rule.objects.filter(data_type=2).exclude(min=None, max=None).update(condition='range')
+    Rule.objects.filter(data_type=3).exclude(min=None, max=None).update(condition='range')
+    Rule.objects.filter(data_type=4).exclude(min=None, max=None).update(condition='range')
+    Rule.objects.filter(data_type=5).exclude(min=None, max=None).update(condition='range')
+
+    Rule.objects.filter(data_type=None, min=None, max=None, text_match=None, not_null=True).update(condition='not_null')
+    Rule.objects.filter(data_type=None, min=None, max=None, text_match=None, required=True, not_null=False).update(condition='required')
+    Rule.objects.filter(data_type=None, min=None, max=None, text_match=None, required=False, not_null=False).update(condition='')
```
adrian-lara, line 40: For this line and L43, see comments about L23 and L26.

seed/migrations/0125_dq_refactor.py
lainsworth8801: Not as i know of.

seed/migrations/0125_dq_refactor.py
```diff
@@ -0,0 +1,62 @@
...
+    #                              |--> min / max not null: range                                                 #
+    #                                                                                                             #
+    # data_type null -------> min & max null ---> text_match null: not_null/required/''                           #
+    #                |                       |--> text_match not null ---> text_match == '': not_null/required/'' #
+    #                |                                                |--> text_match != '': include              #
+    #                |--> min / max not null ---> range                                                           #
+    # =========================================================================================================== #
+    Rule.objects.filter(data_type=1, text_match=None, not_null=True).update(condition='not_null')
+    Rule.objects.filter(data_type=1, text_match=None, required=True, not_null=False).update(condition='required')
+    Rule.objects.filter(data_type=1, text_match=None, required=False, not_null=False).update(condition='')
```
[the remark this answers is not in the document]
lainsworth8801: Currently this condition does not check anything. And yes my intent for leaving the condition blank is: 1. to alert and prompt users that no condition has been selected for this rule and there should be one; 2. to prompt users to hopefully clean up rule group since There are rules in the db where there's only data type and nothing else.

seed/migrations/0125_dq_refactor.py
lainsworth8801: There are 2 records shown in 20200323 dump. Text_match == '' and text_match is null are deemed as 'not_null' or 'required' if corresponding condition is checked for text type rules. If neither condition is checked, it's left blank.

seed/migrations/0125_dq_refactor.py
```diff
@@ -0,0 +1,62 @@
...
+    Rule.objects.filter(data_type=1).exclude(text_match=None).exclude(text_match='').update(condition='include')
+
+    Rule.objects.exclude(data_type=1).exclude(data_type=None).filter(min=None, max=None, not_null=True).update(condition='not_null')
+    Rule.objects.exclude(data_type=1).exclude(data_type=None).filter(min=None, max=None, required=True, not_null=False).update(condition='required')
+    Rule.objects.exclude(data_type=1).exclude(data_type=None).filter(min=None, max=None, required=False, not_null=False).update(condition='')
+    Rule.objects.filter(data_type=0).exclude(min=None, max=None).update(condition='range')
+    Rule.objects.filter(data_type=2).exclude(min=None, max=None).update(condition='range')
+    Rule.objects.filter(data_type=3).exclude(min=None, max=None).update(condition='range')
+    Rule.objects.filter(data_type=4).exclude(min=None, max=None).update(condition='range')
+    Rule.objects.filter(data_type=5).exclude(min=None, max=None).update(condition='range')
```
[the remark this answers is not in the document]
lainsworth8801: Thank you!

seed/models/data_quality.py
```diff
@@ -702,76 +735,64 @@ def _check(self, rules, row):
                 if (rule.table_name, rule.field) not in self.column_lookup:
```
adrian-lara, line 735: After doing some manual testing of this logic, here's what I'm seeing when comparing the old setup and this new setup:

**I think the following are working as expected:**

- In old and new setup, when "Address Line 2" only has `contain` Rules, mapping that column but values are null;  this does not trigger Rules. before_after - contains THEN mapped - GOOD

- In old setup, when "Conditioned Floor Area" has `not null` checked for the two `min` and `max` Rules, not mapping triggers both Rules. In new setup, when "Conditioned Floor Area" has three separate `not null, `min`, and `max` Rules, not mapping that column triggers only the `not null` Rule. before_after - minmax not_null THEN not mapped - GOOD

- In old and new setup, when "Address Line 2" only has `contain` Rules, not mapping that column does not trigger Rules. before_after - contains THEN not mapped - GOOD

**I'm wondering if the following two cases are not what we'd expect:**

- In old setup, when "Conditioned Floor" area has only `min` and `max` Rules, mapping that column but values are null, this **does not trigger** Rules. In new setup, when "Conditioned Floor" area has only min and max Rules, mapping that column but values are null - this **does trigger** Rules. before_after - minmax THEN mapped to empty - BAD?

- In old setup, when "Conditioned Floor Area" has only `min` and `max` Rules, not mapping that column **does not trigger** Rules. In new set up, when "Conditioned Floor Area" has only `min` and `max` Rules, not mapping that column **does trigger** Rules. before_after - minmax THEN not mapped - BAD?

adrian-lara, line 735: After doing some manual testing of this logic, here's what I'm seeing when comparing the old setup and this new setup:

**I think the following are working as expected:**

- In old and new setup, when "Address Line 2" only has `contain` Rules, mapping that column but values are null;  this does not trigger Rules. before_after - contains THEN mapped - GOOD

- In old setup, when "Conditioned Floor Area" has `not null` checked for the two `min` and `max` Rules, not mapping triggers both Rules. In new setup, when "Conditioned Floor Area" has three separate `not null, `min`, and `max` Rules, not mapping that column triggers only the `not null` Rule. before_after - minmax not_null THEN not mapped - GOOD

- In old and new setup, when "Address Line 2" only has `contain` Rules, not mapping that column does not trigger Rules. before_after - contains THEN not mapped - GOOD

**I'm wondering if the following two cases are not what we'd expect:**

- In old setup, when "Conditioned Floor" area has only `min` and `max` Rules, mapping that column but values are null, this **does not trigger** Rules. In new setup, when "Conditioned Floor" area has only min and max Rules, mapping that column but values are null - this **does trigger** Rules. before_after - minmax THEN mapped to empty - BAD?

- In old setup, when "Conditioned Floor Area" has only `min` and `max` Rules, not mapping that column **does not trigger** Rules. In new set up, when "Conditioned Floor Area" has only `min` and `max` Rules, not mapping that column **does trigger** Rules. before_after - minmax THEN not mapped - BAD?

lainsworth8801: @adrian-lara you are correct. I overlooked some of the older front and back end code when there was no dedicated conditions assigned to individual rule. Null values should only be reported if 'not null' condition is applied. Null values should not be reported under rules dedicated for range check.

seed/migrations/0125_dq_refactor.py
```diff
@@ -0,0 +1,98 @@
...
+    def create_condition(rule, condition):
+        Rule.objects.create(name=rule['name'], description=rule['description'], table_name=rule['table_name'], field=rule['field'],
+                            enabled=rule['enabled'], data_type=rule['data_type'], rule_type=rule['rule_type'], min=rule['min'],
+                            max=rule['max'], severity=rule['severity'], units=rule['units'],
+                            data_quality_check_id=rule['data_quality_check_id'], status_label_id=rule['status_label_id'],
+                            text_match=rule['text_match'], condition=condition)
+
+    Rule.objects.filter(data_type=1, text_match=None, not_null=True).update(condition='not_null')
+    Rule.objects.filter(data_type=1, text_match=None, required=True, not_null=False).update(condition='required')
+    Rule.objects.filter(data_type=1, text_match=None, required=False, not_null=False).update(condition='')
```
adrian-lara, line 39: For cases like this where we're setting `condition=''`, would it make sense to go disable these for users to further make the point that these aren't doing anything and requires action?

seed/migrations/0125_dq_refactor.py
```diff
@@ -0,0 +1,98 @@
...
+    # data_type null:
+    #     min and max null -> text_match null: not_null -> not_null;
+    #                                         !not_null -> required: required;
+    #                                                   -> !required: '';
+    #                      -> text_match not null: include -> not_null: (+)not_null;
+    #                                                      -> !not_null -> required: (+)required;
+    #     min or max not null: range -> not_null: (+)not_null;
+    #                                -> !not_null -> required: (+)required;
+
+    def create_condition(rule, condition):
```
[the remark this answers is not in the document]
lainsworth8801: I agree with you here thanks @adrian-lara !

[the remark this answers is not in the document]
lainsworth8801: @adrian-lara Here's some before/after cross-check on one of "District of Columbia Department of Energy and Environment":
Screen Shot 2020-05-05 at 9 44 45 PM

Screen Shot 2020-05-05 at 9 45 13 PM

Get by everything field of a rule EXCPET variables such as severity level and label to avoid duplicates. Create (and update) by not_null/required/min/max/text status.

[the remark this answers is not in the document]
lainsworth8801: Here are snapshots of before and after migration:
Screen Shot 2020-05-05 at 9 45 13 PM
Screen Shot 2020-05-05 at 9 44 45 PM

seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -231,16 +276,46 @@ angular.module('BE.seed.controller.data_quality_admin', [])
...
+        $scope.defaults_restored = false;
+        $scope.rules_reset = false;
+        if (!rule.condition) rule.condition = null;
+        if (rule.condition === 'include' || rule.condition === 'exclude' && rule.data_type !== 'string') rule.data_type = 'string';
+        if (_.isMatch(rule, {condition: 'range', data_type: 'string'})) rule.data_type = null;
+        if (_.isMatch(rule, {condition: 'not_null', data_type: 'string'})) rule.text_match = null;
+
+        var condition = rule.condition;
+        if (condition === 'required') {
+          rule.required = false;
```
adrian-lara, line 300: Should we stop referring to and relying on the `required` and `not_null` attributes on the frontend? It looks like we're still referring to it in some capacity throught out the this controller.

This likely part of a larger discussion, but I assumed those particular Rule attributes would be removed from the database, model, etc. once we're fully converted over to using the `condition` values. I can imagine that removal happening down the road, and I could see an argument for making this refactor (of not using `required` and `not_null` attributes anymore) during that time.

seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -231,16 +276,46 @@ angular.module('BE.seed.controller.data_quality_admin', [])
...
+        } else if (condition === 'not_null') {
+          rule.required = false;
+          rule.not_null = rule.required;
+          $scope.change_not_null(rule);
+        }
+        if (rule.condition !== 'range') {
+          rule.units = '';
+          rule.min = null;
+          rule.max = null;
+        }
```
adrian-lara, line 311: Should we do something similar for text_match and include/exclude?

seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -223,6 +258,16 @@ angular.module('BE.seed.controller.data_quality_admin', [])
                   r.label = match.id;
                 }
               }
+              if(!(r.min === '' || r.min === null) && !(r.max === '' || r.max === null)) {
+                if (r.max < r.min) {
+                  var min = r.min;
+                  r.min = r.max;
+                  r.max = min;
+                }
```
adrian-lara, line 266: Nice! I like this feature but would generally suggest that small, unrelated cleanups be done in a separate PR when the current PR is as large as this one - unless, of course, somebody specifically requested it be done in this one 😄

seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -193,18 +215,30 @@ angular.module('BE.seed.controller.data_quality_admin', [])
...
               };
+              if (rule.condition === 'not_null') {
+                r.min = null;
+                r.max = null;
+              }
+              r.condition = rule.condition;
+
+              if (rule.condition === 'units') {
+                r.units = '';
+              }
```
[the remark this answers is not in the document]
lainsworth8801: This should be removed. As a matter of fact there's a deep cleaning needing to happen.

seed/static/seed/js/controllers/data_quality_admin_controller.js
lainsworth8801: I believe we should remove not_null/required columns from database and model. And removing these 2 attributes is the intention. I agree that once the feature is stable we will no longer need these 2 columns for condition reference.

seed/static/seed/js/controllers/data_quality_admin_controller.js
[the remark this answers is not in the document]
lainsworth8801: This seems like some old leftover code to me. It's been removed locally

seed/migrations/0125_dq_refactor.py
```diff
@@ -0,0 +1,98 @@
...
+    #                                                      -> !not_null -> required: (+)required;
+    #     min or max not null: range -> not_null: (+)not_null;
+    #                                -> !not_null -> required: (+)required;
+
+    def create_rule(rule, condition, min, max, text_match):
+        Rule.objects.get_or_create(name=rule['name'], description=rule['description'], table_name=rule['table_name'],
+                                   field=rule['field'], enabled=rule['enabled'], data_type=rule['data_type'],
+                                   rule_type=rule['rule_type'], min=min, max=max, units=rule['units'],
+                                   data_quality_check_id=rule['data_quality_check_id'],
+                                   text_match=text_match, condition=condition)
```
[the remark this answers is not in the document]
adrian-lara: What's the reasoning for not disabling the fields like for not_null?

Also - what do you think about the the DB migration method being changed as suggested to avoid essentially duplicate Rules?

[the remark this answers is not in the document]
lainsworth8801: So my understanding for 'required' for now is this could be for potential column mapping which to me works as a place holder and is allowed to include nullable values. There are records found where 'required' is true but 'not_null' is false and some of these rule_fields are 'notes'. If we disabled required, it'll be no difference from not_null. The only difference is it will not be reported if it receives a null value.

seed/static/seed/js/controllers/data_quality_admin_controller.js
lainsworth8801: This is done about 11 lines up:
Screen Shot 2020-05-06 at 2 21 29 PM

seed/migrations/0125_dq_refactor.py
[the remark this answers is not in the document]
lainsworth8801: > Also - what do you think about the the DB migration method being changed as suggested to avoid essentially duplicate Rules?
It makes sense to me. I do want to confirm with you that these are the essential elements for properly creating additional rules without being duplicated. Here i used all elements except for severity level and label id as these two definitely create duplicates.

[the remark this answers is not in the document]
lainsworth8801: > Also - what do you think about the the DB migration method being changed as suggested to avoid essentially duplicate Rules?

It makes sense to me. I do want to confirm with you that these are the essential elements for properly creating additional rules without being duplicated. Here i used all elements except for severity level and label id as these two definitely create duplicates.

seed/static/seed/js/controllers/data_quality_admin_controller.js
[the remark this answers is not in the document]
lainsworth8801: Nothing. This shall be removed.

seed/migrations/0125_dq_refactor.py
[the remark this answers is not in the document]
adrian-lara: Per our call, 
- min, max, and text_match will be disabled on the frontend when Rule has condition "Required".
- min, max, and text_match will be set to None for newly created Rules with condition 'required' or 'not_null'.
- bring back "required" check on None values as it was in develop. Any potential changes to the underlying required logic will be made in a separate PR.

seed/static/seed/js/controllers/data_quality_admin_controller.js
adrian-lara: Per our conversation, we're removing that stale checkbox-related test.

seed/models/data_quality.py
```diff
@@ -702,76 +735,63 @@ def _check(self, rules, row):
...
                         self.add_result_missing_req(row.id, rule, display_name, value)
                         label_applied = self.update_status_label(label, rule, linked_id, row.id)
                 elif value is None or value == '':
-                    # Empty fields
-                    if rule.required:
-                        self.add_result_missing_and_none(row.id, rule, display_name, value)
-                        label_applied = self.update_status_label(label, rule, linked_id, row.id)
-                    elif rule.not_null:
-                        self.add_result_is_null(row.id, rule, display_name, value)
+                    if rule.severity == Rule.SEVERITY_ERROR or rule.severity == Rule.SEVERITY_WARNING:
```
adrian-lara, line 742: What's the motivation behind this initial severity level check?

Couldn't there be a case where users would consider it "Valid Data" and just want a label automatically applied if a field is null or missing?

seed/models/data_quality.py
```diff
@@ -702,76 +735,63 @@ def _check(self, rules, row):
...
-                        self.add_result_comparison_error(row.id, rule, display_name, s_value, s_max)
-                        continue
-                    except DataQualityTypeCastError:
-                        s_min, s_max, s_value = rule.format_strings(value)
-                        self.add_result_type_error(row.id, rule, display_name, s_value)
-                        continue
-                    except UnitMismatchError:
-                        self.add_result_dimension_error(row.id, rule, display_name, value)
-                        continue
+                    if rule.condition == Rule.RULE_RANGE:
```
adrian-lara, line 754: Any reason why this got turned into a completely separate if-conditional? I'd think the previous else would just turn into an `elif`. Am I missing something?

seed/models/data_quality.py
```diff
@@ -1007,6 +1052,8 @@ def add_invalid_geometry_entry_provided(self, row_id, rule, display_name, value)
             'message': display_name + ' should be in WKT format',
             'detailed_message': detailed_message,
             'severity': rule.get_severity_display(),
+            'condition': rule.condition,
+            'label': rule.status_label,
```
adrian-lara, line 1056: As discussed separately, we'll be moving these potential fixes into another PR/Issue. There's likely other things to check, since I recall DQ labels being updated previously. It'd be worth deeper review/inspection of required changes for that bug fix.

seed/models/data_quality.py
```diff
@@ -1042,12 +1089,13 @@ def update_status_label(self, label_class, rule, linked_id, row_id):
                     raise IntegrityError(
                         'Label with super_organization_id={} cannot be applied to a record with parent '
                         'organization_id={}.'.format(
+                            label_org_id,
```
adrian-lara, line 1092: Why are contents of this error message being changed?
```suggestion
```

seed/models/data_quality.py
```diff
@@ -1042,12 +1089,13 @@ def update_status_label(self, label_class, rule, linked_id, row_id):
...
                             label_org_id,
                             taxlot_parent_org_id
                         )
                     )
-
-            self.results[row_id]['data_quality_results'][-1]['label'] = rule.status_label.name
+            if rule.severity != Rule.SEVERITY_VALID:
+                self.results[row_id]['data_quality_results'][-1]['label'] = rule.status_label.name
```
adrian-lara, line 1098: I'm thinking we should remove this change. Don't users want to be able to auto apply labels without flagging things as errors?

seed/models/data_quality.py
```diff
@@ -129,31 +135,37 @@ class Rule(models.Model):
...
+            'text_match': '',
         }, {
             'table_name': 'PropertyState',
             'field': 'pm_property_id',
-            'data_type': TYPE_STRING,
+            'data_type': TYPE_NUMBER,
             'not_null': True,
             'rule_type': RULE_TYPE_DEFAULT,
             'severity': SEVERITY_ERROR,
+            'condition': RULE_RANGE,
```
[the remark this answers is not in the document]
lainsworth8801: Good find thanks! Didn't see it being defined as CharField in properties.py.

seed/models/data_quality.py
```diff
@@ -129,31 +135,37 @@ class Rule(models.Model):
...
         }, {
             'table_name': 'PropertyState',
             'field': 'custom_id_1',
             'not_null': True,
             'rule_type': RULE_TYPE_DEFAULT,
             'severity': SEVERITY_ERROR,
+            'condition': RULE_RANGE,
```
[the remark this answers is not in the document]
lainsworth8801: You are correct. These changes were made earlier on. All rule without explicit range values with not_null being true should be under 'not null' condition.

seed/models/data_quality.py
lainsworth8801: We had a conversation earlier on about data quality check should only report errors and warning. It does not report valid data. However, a label is mandatory for a rule set for filtering valid data on the UI:
Screen Shot 2020-05-07 at 1 44 25 PM

seed/models/data_quality.py
lainsworth8801: No recall on this one. Must have been 'command + s' turning into 'command + d'.

seed/models/data_quality.py
lainsworth8801: I believe i wasn't aware minimum_valid() and maximum_valid() check both number and string type values. I will remove that.

seed/tests/test_data_quality_checks.py
```diff
@@ -163,8 +163,8 @@ def test_check_property_state_example_data_with_labels(self):
         dq.check_data(ps.__class__.__name__, [ps])
 
         dq_results = dq.results[ps.id]['data_quality_results']
-        labels = [r['label'] for r in dq_results]
-
+        # labels = [r['label'] for r in dq_results]
+        labels = [r.get('label') for r in dq_results if r.get('label') is not None]
```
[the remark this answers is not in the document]
lainsworth8801: If i recall it once errored out for not finding label but I will double check

seed/models/data_quality.py
lainsworth8801: You mean users see a null value as a valid value?

seed/static/seed/js/controllers/mapping_controller.js
```diff
@@ -198,14 +198,16 @@ angular.module('BE.seed.controller.mapping', [])
...
+        if (display_name) display_name = display_name.display_name;
         return {
           column_name: column_name,
           display_name: display_name
         };
       });
       matching_criteria_columns_payload.TaxLotState = _.map(matching_criteria_columns_payload.TaxLotState, function (column_name) {
-        var display_name = _.find($scope.mappable_taxlot_columns, {column_name: column_name}).display_name;
+        var display_name = _.find($scope.mappable_taxlot_columns, {column_name: column_name});
+        if (display_name) display_name = display_name.display_name;
```
adrian-lara, line 210: I can't figure out the possible connection between these change and DQ changes. Think you could provide background here?

seed/static/seed/js/controllers/menu_controller.js
```diff
@@ -53,11 +53,9 @@ angular.module('BE.seed.controller.menu', [])
...
       $scope.$on('app_error', function (event, data) {
         // Keep the first error
-        if (!$rootScope.route_load_error) {
-          $rootScope.route_load_error = true;
-          $scope.menu.error_message = data.message;
-          $rootScope.load_error_message = data.data.message;
-        }
+        if (!$rootScope.route_load_error) $rootScope.route_load_error = true;
+        $scope.menu.error_message = data.message;
+        $rootScope.load_error_message = data.data.message;
```
adrian-lara, line 58: Looks like inadvertent formatting - could you change it back? Probably worth you reviewing the FE code for things like this as I might've missed them.

seed/static/seed/js/controllers/data_quality_modal_controller.js
```diff
@@ -27,10 +27,33 @@ angular.module('BE.seed.controller.data_quality_modal', [])
...
+                  results.data_quality_results = _.reject(results.data_quality_results, group[remove[i]]);
+                }
+              }
+              check_null = false;
+            }
+          });
+        }
+      });
+      $scope.dataQualityResults = originalDataQualityResults;
+      console.log($scope.dataQualityResults);
```
adrian-lara, line 56: console.log

seed/static/seed/js/controllers/data_quality_modal_controller.js
```diff
@@ -27,10 +27,33 @@ angular.module('BE.seed.controller.data_quality_modal', [])
...
+                for (var i = remove.length - 1; i >= 0; i--) {
+                  results.data_quality_results = _.reject(results.data_quality_results, group[remove[i]]);
+                }
+              }
+              check_null = false;
+            }
+          });
+        }
+      });
+      $scope.dataQualityResults = originalDataQualityResults;
```
adrian-lara, line 55: Looks like this prevents multiple not_null/required Rules from being shown. Is that right??

It looks like it's not happening on the frontend, so I think I'm misunderstanding something...

seed/data_importer/tests/integration/test_footprints_import.py
```diff
@@ -177,8 +177,11 @@ def test_demo_v2(self):
         self.assertEqual(tdq.results[tax_lot_3.id]['data_quality_results'][0]['detailed_message'], "'' is not a valid geometry")
 
         pdq = DataQualityCheck.retrieve(self.org.id)
-        pdq.check_data('PropertyState', [property_1, property_2, property_3])
-        self.assertEqual(pdq.results.get(property_1.id, None), None)
+        pdq.check_data('PropertyState', [property_1 , property_2, property_3])
+        # For property_1, no error message of null 'Conditioned Floor Area' reported if no 'not_null' condition is set
+        for rule in pdq.rules.filter(table_name='PropertyState', field='conditioned_floor_area'):
+            self.assertNotEqual(rule.condition, 'not_null')
```
adrian-lara, line 183: I'm thinking this test is unrelated to the footprint import. Is that right? If so, I'm fine to keep it here but worth noting it is very confusing to see it here.

seed/models/data_quality.py
```diff
@@ -702,22 +735,21 @@ def _check(self, rules, row):
...
                         self.add_result_is_null(row.id, rule, display_name, value)
                         label_applied = self.update_status_label(label, rule, linked_id, row.id)
-                elif not rule.valid_text(value):
-                    self.add_result_string_error(row.id, rule, display_name, value)
-                    label_applied = self.update_status_label(label, rule, linked_id, row.id)
+                elif rule.condition == Rule.RULE_INCLUDE or rule.condition == Rule.RULE_EXCLUDE:
+                    if not rule.valid_text(value):
+                        self.add_result_string_error(row.id, rule, display_name, value)
+                        label_applied = self.update_status_label(label, rule, linked_id, row.id)
                 else:
```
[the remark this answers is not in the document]
lainsworth8801: if we want nothing done when condition=='' then yes.

seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -231,16 +271,36 @@ angular.module('BE.seed.controller.data_quality_admin', [])
...
         }).finally(function () {
+          $scope.rules_updated = true;
           spinner_utility.hide();
         });
       };
+      $scope.change_condition = function (rule) {
+        $scope.rules_updated = false;
+        $scope.defaults_restored = false;
+        $scope.rules_reset = false;
+        if (!rule.condition) rule.condition = null;
```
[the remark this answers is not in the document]
lainsworth8801: Line 288 is since we don't have a 'null' condition as an option.
Ln 285 - 287 is to hide the checked mark if condition is being changed:
Screen Shot 2020-05-07 at 4 41 42 PM

seed/static/seed/js/controllers/mapping_controller.js
lainsworth8801: No this is not. This was one of the issues that austin was working on (also seen by Robin during the same time period) that was preventing me from importing a file. This was the work-aound we briefly talked about. It does not directly affect dq check. I will remove it.
image

seed/static/seed/js/controllers/menu_controller.js
lainsworth8801: This was intentional. Or error message doesn't get updated with respect to different errors. As a matter of fact this is a mini bug that needs to be improved in case multiple error on page found.

seed/static/seed/js/controllers/data_quality_modal_controller.js
lainsworth8801: Screen Shot 2020-05-07 at 12 14 18 PM

seed/data_importer/tests/integration/test_footprints_import.py
lainsworth8801: I will remove it and move it to data_quality_tests

seed/static/seed/js/controllers/data_quality_modal_controller.js
lainsworth8801: In this table, there are 1 null address line 1, 4 null energy start scores, 3 null site eui, 1 null gross floor area:
Screen Shot 2020-05-07 at 5 07 13 PM

My rules are set up as:
1. check 'required' and 'not_null' address line 1;
2. check 'not null' energy star score;
3. check 'range' on enery star score;
4. check 'range' on gross floor area;
5. check 'not null' site eui;
6. check 'range' on site eui:
Screen Shot 2020-05-07 at 5 09 18 PM

After running data quality check, the report should show:
1. Address line 1 is required and none for property with custom id 1 of 1;
2. Null Energy star score for property with custom id 1 of 1 (address-less property), 11 ninth, 93031 wellington, and 2660 welstone;
3. Null site eui for 11 ninth, 93031 wellington, and 2660 welstone;
4. and out of range values (site eui of 2700 welstone and energy star of 2655 welstone );
Screen Shot 2020-05-07 at 5 12 58 PM

What is/should not being show:
1. Null energy star scores being reported 3 times for each of all properties that include null energy star scores (3 energy star score rules: 2 for ranges, 1 for not_null);
2. Null site eui being reported 3 times for all properties including null site eui values (3 site eui rules, 2 for range, 1 for not_null).

seed/static/seed/js/controllers/menu_controller.js
lainsworth8801: Screen Shot 2020-05-07 at 7 02 52 PM

seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -249,11 +309,11 @@ angular.module('BE.seed.controller.data_quality_admin', [])
         var newDataType = _.find(columns, {column_name: rule.field}).data_type;
 
         if (_.isNil(newDataType)) newDataType = null;
-
         // clear columns that are type specific.
         if (newDataType !== original) {
           rule.text_match = null;
           rule.units = '';
+          rule.condition = '';
```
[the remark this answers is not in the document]
lainsworth8801: Should be 'range' for a new rule.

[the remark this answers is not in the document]
lainsworth8801: Condition is defaulted to range if data type is number, year, date, and eui; 'include' is text; and '' if no data type.

seed/data_importer/tests/integration/test_footprints_import.py
```diff
@@ -152,21 +152,16 @@ def test_demo_v2(self):
 
         self.assertEqual(len(ps), 3)
```
adrian-lara, line 154: Why'd these changes happen here?

seed/static/seed/js/controllers/menu_controller.js
```diff
@@ -53,19 +53,16 @@ angular.module('BE.seed.controller.menu', [])
...
       $scope.$on('app_error', function (event, data) {
         // Keep the first error
-        if (!$rootScope.route_load_error) {
-          $rootScope.route_load_error = true;
-          $scope.menu.error_message = data.message;
-          $rootScope.load_error_message = data.data.message;
-        }
+        $scope.menu.error_message = data.message;
+        $rootScope.load_error_message = data.data.message.join(' ');
+        if (!$rootScope.route_load_error) $rootScope.route_load_error = true;
```
adrian-lara, line 58: Maybe we can have a quick chat about this one tomorrow. I'm still not sure what exactly this does.

seed/data_importer/tests/integration/test_footprints_import.py
lainsworth8801: When property_1. has a null conditioned. floor area, because the rule was set to check for range, not a non-null value, no error detailed message is generated for property 1. This test is no longer valid. Part of the reason why I edited the test prior to removing it.

seed/static/seed/js/controllers/menu_controller.js
lainsworth8801: Definitely. The point of doing this is to prevent confusion for users if there are multiple errors generated. As of now, we have 2 potential errors: valid rule without a label and must include/must not include an empty string.
Screen Shot 2020-05-08 at 6 52 52 AM

Before, it shows only one error message if multiple errors found on page because on view/models/data_quality.py it returns which error message gets trigger first.

lainsworth8801: @adrian-lara I think i see the purpose of 'Keep the first error' for this comment. Originally i was confused by not seeing all error messages when multiple errors were found on the page and that was the motive of modifying this piece of code. I will change it back to keeping the first error.

seed/data_importer/tests/integration/test_footprints_import.py
lainsworth8801: Reverted test as I misunderstood the original test.

seed/static/seed/js/controllers/menu_controller.js
adrian-lara: Per our discussion, this will likely be updated in an upcoming PR.

RDmitchell:
@nllong -- so I guess the outcome of all this is that you have to have multiple rules for a given field to check for whether or not it is required, blank, etc. 

This seems like a lot of setup for the users, but I guess it makes it very clear what is being checked. 

I will do some testing, based on rules that have been set up by existing users.

lainsworth8801:
@RDmitchell you are correct. The ideal ui should be looking as if each field has expandable and collapsable child rows for all condition. This would be future work and is on my radar.

RDmitchell:
great thx

Issue #2134, opened 2020-03-05
Add new column setting to allow blank/'Not Available' values to overwrite other values


adrian-lara:
### Background
During merges, records are combined by comparing values one column/attribute at a time. For each attribute, if each record has a value, an organization-level column setting is read to decide which value to take into the eventual merged record. If one of the records is "empty" (has a value of None in Python or null in postgres), the non-empty value is used.

In most cases, this is desired, but recently, there was a request to allow for empty values to be treated like non-empty values. 

Specifically, there were records with previously populated values for an energy usage field. The energy usage values were actually never truly available. In this case, the populated values for each record should actually be "emptied". 

#### Expected Behavior
At the organization level, be able to specify whether to recognize empty values as normal, non-empty values. This new setting will be an additional column setting.

**Notable caveat**: On import, if a known (by SEED) column is not provided/mapped to, SEED will think the incoming records have "empty" values for that column. This means, if recognize_blank is set to True for a particular column, and that column is not mapped to during import, any resulting merged records will have an empty value for that column/attribute.

### Instance Information
Production and develop

#### Note about production vs develop implementation
On production, the main use case for this is merging **during import**, particular for a user that mostly accesses SEED via API. 

When implementing into develop, a few notable differences and additional considerations should be taken:
- Develop has the column setting change preview. The new Recognize Empty setting needs to be included in that.
- Develop has a modified help text section on the organization column setting page. Recognize Empty -related help text should be added to that.
- On the inventory list pages, the manual merge preview does not consider this new setting. Since the main user of this feature is concerned with import merges (over manual merges) and they likely won't be using the UI, this "fix" is being pushed off to develop.
- Considerations should be taken for how the merging geocoding results interacts with this new setting.
- The import mapping page should have text mentioning the "Recognize Empty" setting if there are any columns that are set to True. Some version of the "Notable caveat" should be included? Maybe via warning?

RDmitchell:
instance; Dev1
SHA: 9f8ec3a

Works well. Plus Andrew in DC has been already using this. But I test it through the UI and it works nicely.

PR #2135, opened 2020-03-06
2134 blank col setting 261

adrian-lara:
--------This PR contains a migration-------

See tagged issue.

#### What's this PR do?
See commits:
- Column model has new recognize_empty bool field defaulted to False
- Merging records considers whether fields have recognize_empty=True
- Column model's recognize_empty added to serializer (now available to API)
- Org column settings page has Recognize Empty setting

Also:
- update Travis Postgres port

Test on production DB*

Toggle the Recognize Empty column setting. Import records that trigger merges and with Not Available or empty values in a particular field. When the setting is checked, see that empty values (truly blank, missing, or "Not Available") are treated like any other non-empty value during a merge.

**Notable caveat**: On import, if a known (by SEED) column is not provided/mapped to, SEED will think the incoming records have "empty" values for that column. This means, if recognize_blank is set to True for a particular column, and that column is not mapped to during import, any resulting merged records will have an empty value for that column/attribute.

Also, I suggest reading the **Note about production vs develop implementation** in the tagged issue.

#2134 

Screen Shot 2020-03-06 at 4 51 19 PM


seed/lib/merging/merging.py
```diff
@@ -124,11 +125,22 @@ def merge_state(merged_state, state1, state2, priorities, ignore_merge_protectio
...
+        table_name=state2.__class__.__name__,
+        recognize_empty=True,
+        is_extra_data=False
+    ).values_list('column_name', flat=True)
+
     default = state2
     for attr in can_attrs:
-        # Do we have any differences between these fields? - Check if not None instead of if value.
-        attr_values = [value for value in list(can_attrs[attr].values()) if value is not None]
-        attr_values = [v for v in attr_values if v is not None]
```
adrian-lara: I understood these two lines to be doing the same thing. Let me know if I'm misunderstanding something.

seed/lib/merging/merging.py
```diff
@@ -124,11 +125,22 @@ def merge_state(merged_state, state1, state2, priorities, ignore_merge_protectio
     # Calculate the difference between the two states and save into a dictionary
     can_attrs = get_state_attrs([state1, state2])
 
+    recognize_empty_columns = state2.organization.column_set.filter(
```
adrian-lara: Reaching for organization columns off the second PropertyState didn't feel great/clean, but I thought it was better than providing another parameter and updating the code base in multiple places.

Also note the filter criteria - table_name, recognize_empty, and is_extra_data. Given we're starting off with an Organization's columns, I thought this was specific enough, but someone else should consider and confirm.

adrian-lara:
@nllong ready for your review. Also, I wanted to explicitly point out the caveat mentioned above. Let me know if you'd like to discuss.

seed/lib/merging/merging.py
nllong: yeah, those look redundant... clean up later though?

seed/lib/merging/merging.py
nllong: i don't see this as an issue since you are really looking at the organization and not the state2.

seed/lib/merging/merging.py
adrian-lara: I ended up replacing those with a single list comprehension (that now incorporates the "recognize_empty" setting), so I think we're good here.

.travis.yml
```diff
@@ -41,19 +41,19 @@ install:
   - npm --version
   - nvm install stable
 before_script:
-  - psql -c "DROP DATABASE IF EXISTS seeddb;" -U postgres
-  - psql -c "DROP DATABASE IF EXISTS test_seeddb;" -U postgres
-  - mv config/settings/test_local_untracked.py config/settings/local_untracked.py
+  - psql -p 5433 -c "DROP DATABASE IF EXISTS seeddb;" -U postgres
```
nllong: thanks for cherry picking this!

seed/lib/merging/merging.py
```diff
@@ -96,9 +96,10 @@ def _merge_extra_data(ed1, ed2, priorities, ignore_merge_protection=False):
     all_keys = set(list(ed1.keys()) + list(ed2.keys()))
     extra_data = {}
     for key in all_keys:
+        recognize_empty = key in recognize_empty_columns
         val1 = ed1.get(key, None)
         val2 = ed2.get(key, None)
-        if val1 and val2:
+        if (val1 and val2) or recognize_empty:
```
nllong: looks right to me.

seed/models/columns.py
```diff
@@ -588,6 +588,8 @@ class Column(models.Model):
     merge_protection = models.IntegerField(choices=COLUMN_MERGE_PROTECTION,
                                            default=COLUMN_MERGE_FAVOR_NEW)
 
+    recognize_empty = models.BooleanField(default=False)
```
nllong: correct default

PR #2136, opened 2020-03-11
2134 blank col setting dev

adrian-lara:
------ This PR contains a migration -----

See tagged issue

#### What's this PR do?
See commits

Note, when comparing this PR to #2135 which adds this feature to prod as a patch, this PR contains a few additional features:
- The column settings change preview includes this new setting
- This new setting is a sortable field in the column settings page
- The column setting page help text is includes info on this new setting
- The interaction between this new setting and merging geocoding results fields is that they are purposely handled separately and don't interact at all. Merging geocoding results happens before recognize_empty and doesn't consider recognize_empty at all. This is done to avoid pouring effort into what I'm considering to be a low impact edge case.

Toggle the Recognize Empty column setting. Import records that trigger merges and with Not Available or empty values in a particular field. When the setting is checked, see that empty values (truly blank, missing, or "Not Available") are treated like any other non-empty value during a merge.

**Notable caveat**: On import, if a known (by SEED) column is not provided/mapped to, SEED will think the incoming records have "empty" values for that column. This means, if recognize_blank is set to True for a particular column, and that column is not mapped to during import, any resulting merged records will have an empty value for that column/attribute.

#2134 

Screen Shot 2020-03-11 at 2 15 14 PM


nllong:
The migration fails if we already ran the 2.6.1-patch3 prior to this migration. Is there an easy way to deal with this? I am concerned because because the production database will have 2.6.1-patch3 applied and we will then need to run this migration.

```
![Uploading image.png…]()
```

adrian-lara:
> The migration fails if we already ran the 2.6.1-patch3 prior to this migration. Is there an easy way to deal with this? I am concerned because the production database will have 2.6.1-patch3 applied and we will then need to run this migration.

Ah - good catch. I'll have to see if Django has anything built in to handle this - looking into it now.

adrian-lara:
@nllong I'm thinking of using `--fake` flag for that particular migration. Let me know what you think.
Specifically, when prod finally catches up to 0119_column_recognize_empty, assuming all the prior migrations are migrated, we'd run 

`./manage.py migrate seed 0119_column_recognize_empty --fake`

Here are the steps I took to arrive at this suggestion:
1. Using a prod DB with 0113_column_recognize_empty migrated, I checked out the develop branch, and viewed migrations just to verify assumptions:
Screen Shot 2020-03-11 at 4 31 44 PM

2. For the sake of time, I shifted 0119_column_recognize_empty to 0115_... and deleted the previous 0115_... migrations and later:
Screen Shot 2020-03-11 at 4 36 58 PM

3. As expected, trying to run `migrate seed` led me to `django.db.utils.ProgrammingError: column "recognize_empty" of relation "seed_column" already exists`.

4. From there, after showing migrations, I ran the following command: 
Screen Shot 2020-03-11 at 4 40 15 PM
From everything I'm seeing and trying, the migrations are in sync with the actual DB. This includes verifying the field still exists and running a mix of rollbacks and re-migrates.

nllong:
I'm fine with fake as long as we don't have to call it explicitly. Is there a way that this can just happen? Or is there a way to update the SQL to check if column exists?

adrian-lara:
I _was_ envisioning this flag would be provided explicitly, as it would only need to happen "once" during the actual release. And it would throw an explicit error if someone tried to run that migration with that flag. (Quotes on "once" because, of course, any prod DB copies trying to get upgraded would need that flag too - i.e. our local copies or develop.)

That probably isn't great, so I'll continue looking. From what I've read so far, it definitely seems possible to be able to have it done automatically, but I haven't found a quick or obvious route to take.

nllong:
@adrian-lara Okay, can you document this in the release_instructions.md file in case another user hits this?

adrian-lara:
>@adrian-lara Okay, can you document this in the release_instructions.md file in case another user hits this?

@nllong Re: my latest commit, is this what you had in mind?

SEED didn't seem to have a release_instructions.md file, so I created one and placed it at the root. Let me know if I should move it or, of course, if I should rephrase or correct the language within.

nllong:
The release notes that I was referring to were really these: https://github.com/SEED-platform/seed/blob/develop/docs/source/migrations.rst

I've updated them.

Issue #2137, opened 2020-03-12
BuildingSync to handle v1.1 and v2.0


macintoshpie:
### Expected Behavior
SEED should support both 1.1 and 2.0 BuildingSync schema files.

#### Basic user flow
- user starts a file upload, and clicks a button to upload a BuildingSync file, and chooses a zip or a single xml file.
- user is directed to a 'mapping' page, which displays a table for the default mapping for property state information of the determined schema version (see note below in Additional changes). ie on the left is the seed db column name, and on the right is the xpath to the data. It is NOT required to display the data that results from that mapping (ie 'first five rows')
- user clicks 'map data' button, which kicks off the mapping on our backend

#### Additional changes
- BuildingSync file uploads should now use the typical backend flow for importing and mapping files.
- BSync files uploaded will need to specify, via the file's `xsi` attribute on the BuildingSync element, which schema version it adheres to. e.g. `https://raw.githubusercontent.com/BuildingSync/schema/v<VERSION>/BuildingSync.xsd`. The app should error if the `xsi` does not follow that format.
- export will need to account for the imported version, and map accordingly

### Actual Behavior
Only 1.1 is supported on SEED

### Instance Information
Server Instance: `bricr-dev`, to be cherrypicked into `develop`

RDmitchell:
@macintoshpie / @nllong 

Is this to be tested in the normal version of SEED, or only in bricr? If just bricr then I am not the person to do the testing.

nllong:
I think you can skip testing on this. We have BRICR team doing testing.

Issue #1928, opened 2019-07-18
Intermittent deadlock on Meter and MeterReading imports


adrian-lara:
During PM Meter/MeterReading imports, an error occurs somewhat infrequently. Any time the error occurs, re-importing the file has always lead to a successful import.  I haven't heard or seen this happen for GreenButton Meter/MeterReading imports but it could be possible as the logic is similar though not identical. 

Unfortunately, the underlying reason behind the error is not fully understood as the error has been difficult reproduce at will. It is understood that Postgres deadlocks are causing the error. Specifically, an Update Exclusive Lock on the MeterReading model and a Row Exclusive Lock on the Meter model seem to be behind these deadlocks. 

The PM Meter import process can be found here: https://github.com/SEED-platform/seed/blob/develop/seed/data_importer/tasks.py#L806
There's one transaction, with 2 main queries:
1. Get or create a Meter record
2. Reference that Meter record to bulk create the MeterReadings
It's necessary to consider that this method is sent to Celery and run as a background task. For any given import, each of these background tasks should be importing/working with separate Meter records (and, subsequently, separate MeterReading records).

Here are the error messages from two instances of this occuring
![Sentry - Deadlock error](https://user-images.githubusercontent.com/30608004/61549951-935b2980-aa0e-11e9-9fda-52cc962c2dde.png)

```
[2019-07-17 14:40:19,431: ERROR/ForkPoolWorker-2] Chord callback for '77736616-0854-454b-bd74-653397beab51' raised: ChordError("Dependency 8532fab9-edcb-4b70-8400-3154823e646d raised OperationalError('deadlock detected\\nDETAIL:  Process 97539 waits for ShareUpdateExclusiveLock on relation 16945371 of database 16940925; blocked by process 97538.\\nProcess 97538 waits for ShareRowExclusiveLock on relation 16942891 of database 16940925; blocked by process 97539.\\nHINT:  See server log for query details.\\n',)",)
Traceback (most recent call last):
  File "/Users/adrianlara/devetry/projects/seed/venv367-20190509-develop/lib/python3.6/site-packages/celery/backends/redis.py", line 399, in on_chord_part_return
    callback.delay([unpack(tup, decode) for tup in resl])
  File "/Users/adrianlara/devetry/projects/seed/venv367-20190509-develop/lib/python3.6/site-packages/celery/backends/redis.py", line 399, in <listcomp>
    callback.delay([unpack(tup, decode) for tup in resl])
[...]

PR #2144, opened 2020-03-16
refactor(bsync): use tasks for importing

macintoshpie:
Previously BSync files were imported at their own endpoint and parsed in-process (ie blocking).
#### What's this PR do?
Moves the BSync backend import flow to use the flow of other files (ie background tasks). Additionally it adds a mapping page for buildingsync files, though it doesn't currently do anything other than show the mapping from column name to an xpath.
Go through this flow with these different files from the directory `seed/building_sync/tests/data`:
- a supported BuildingSync xml file (v2.0-pr1): `ex_1.xml`
- a zip of supported BSync files: `valid_xml_ex1_ex2.zip`
- a broken BSync file (e.g. missing address info): `ex_1_no_street_address.xml`
1. Navigate to the upload data modal, and click 'Upload BuildingSync Data'.
2. Select the file
3. Click 'continue to mapping', and you should be directed to a mapping page
4. It should show the set of fields from BuildingSync.BRICR_STRUCT and their XPath. Nothing should be editable on this page! Users are not allowed to change mappings currently
5. After clicking "May Your Data", you should be redirected to the mapped properties page. If there was an invalid BSync file included, you should see a modal describing why it failed. If all files failed to be mapped (e.g. a zip with all invalid files, or a single invalid xml file), then you should _not_ see the "save mapping" button, nor should you see any table (it should just say "no data").
6. Go to your properties and the BuildingSync properties should be there.
#2137


seed/lib/xml_mapping/reader.py
```diff
@@ -0,0 +1,39 @@
...
+        _, file_extension = os.path.splitext(filename)
+
+        # grab the data from the zip or xml file
+        self.data = []
+        if file_extension == '.zip':
+            with zipfile.ZipFile(file_, 'r', zipfile.ZIP_STORED) as openzip:
+                filelist = openzip.infolist()
+                for f in filelist:
+                    if '.xml' in f.filename and '__MACOSX' not in f.filename:
+                        self.data.append({'_xml': openzip.read(f).decode(), '_filename': f.filename})
```
macintoshpie: @adrian-lara @nllong 
Looking for feedback on this section of code. Currently it's reading _all files_ into memory (yikes) to represent a 'row' of data (ie property instance). I'm looking for better ideas for what to do here. These are the two options I can think of:
1. turn self.data into a generator within the context of the zip file. Not sure what the implications are here.
2. Read from the zip for every property when mapping. This would require adding a field to each data entry, '_zip', where when true indicates it's from a zip file. Then when we go to read/map the data, we know we need to open the import_file.file to get the actual file info. I'm concerned about concurrent reads from the workers, but it should probably be fine.

macintoshpie: After initial research I've decided to just stick with reading the files into memory, as is currently implemented. In the future will will probably want to implement a generator class for the `data` property.

seed/data_importer/tasks.py
```diff
@@ -151,7 +155,7 @@ def do_checks(org_id, propertystate_ids, taxlotstate_ids, import_file_id=None):
         # specify the chord as an immutable with .si
         chord(tasks, interval=15)(finish_checking.si(progress_data.key))
     else:
-        finish_checking.s(progress_data.key)
+        progress_data.finish_with_success()
```
macintoshpie: Could someone verify this is ok to do?

If all uploaded bsync files were bad, the front-end would spin forever on checking the progress of this because the backend would always say it was not started. I couldn't figure out why calling `finishe_checking.s(progress_data.key)` wasn't setting it to finished, so I just did it here.

macintoshpie:
As a note, I ran into issue #1928 while testing. It only occurred once when uploading a large (~250 buildings) file, and from the looks of it it doesn't seem to be specific to this refactoring, but more generally an issue with timescaledb.

seed/data_importer/tasks.py
nllong: Nothing seems wrong with what you have; however, I am concerned with 1) always eager vs not-always eager 2) making sure that this is not orthogonal to https://github.com/SEED-platform/seed/pull/2055

seed/models/models.py
```diff
@@ -17,13 +17,15 @@
...
 COMPOSITE_BS = 4
+BUILDINGSYNC_RAW = 5
 
 SEED_DATA_SOURCES = (
     (ASSESSED_RAW, 'Assessed Raw'),
     (ASSESSED_BS, 'Assessed'),
     (PORTFOLIO_RAW, 'Portfolio Raw'),
     (PORTFOLIO_BS, 'Portfolio'),
     (COMPOSITE_BS, 'BuildingSnapshot'),  # I don't think we need this, but I am leaving it for now.
+    (BUILDINGSYNC_RAW, 'BuildingSync Raw'),
```
adrian-lara: Still reviewing this PR, but this change should also come with a new migration.

seed/data_importer/tasks.py
macintoshpie: I just ran some tests locally:

Set `CELERY_TASK_ALWAYS_EAGER = True
CELERY_TASK_EAGER_PROPAGATES = True` and not running celery then uploaded a single buildingsync file that was invalid and it still worked as expected.

Using celery it also worked. Let me know if these tests aren't sufficient

nllong: excellent. Thanks for verifying!

seed/lib/xml_mapping/reader.py
```diff
@@ -0,0 +1,39 @@
...
+        filename = file_.name
+        _, file_extension = os.path.splitext(filename)
+
+        # grab the data from the zip or xml file
+        self.data = []
+        if file_extension == '.zip':
+            with zipfile.ZipFile(file_, 'r', zipfile.ZIP_STORED) as openzip:
+                filelist = openzip.infolist()
+                for f in filelist:
+                    if '.xml' in f.filename and '__MACOSX' not in f.filename:
```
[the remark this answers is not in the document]
macintoshpie: I'm not sure what other OS-specific files there might be. I think th I copy/pasted this from another place where we were parsing BSync zips so I guess it's working so far. I can look further into it if you'd like

seed/data_importer/tasks.py
```diff
@@ -629,7 +710,10 @@ def _save_raw_data_chunk(chunk, file_pk, progress_key):
                     if key == "bounding_box":  # capture bounding_box GIS field on raw record
                         raw_property.bounding_box = v
                     elif isinstance(v, basestring):
-                        new_chunk[key] = unidecode(v)
+                        try:
+                            new_chunk[key] = unidecode(v)
+                        except Exception:
```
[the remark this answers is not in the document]
macintoshpie: I remember having some issues trying to decode the parsed xml at one point and it was throwing on the unidecode... It might have been when I was fooling with the encoding of the xml string. Maybe I fixed that when I added `.decode()` [here](https://github.com/SEED-platform/seed/pull/2144/files/cc520d27a09d7355a02d40c39484183adaf581a0#diff-3132c1cd243c9b15ced1a284b8dbb944R30). Or maybe one of the files I was testing with locally was ASCII encoded. It's probably safe to keep this here right?

seed/lib/xml_mapping/reader.py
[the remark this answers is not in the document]
adrian-lara: I think it's fine. At worst, this catches a corner case.

seed/static/seed/partials/mapping_xml.html
```diff
@@ -0,0 +1,225 @@
...
+                          <th><translate>Inventory Type</translate></th>
+                          <th class="mapping_field ellipsis-resizable" sd-resizable><translate>SEED Header</translate></th>
+                          <th><translate>Measurement Units</translate></th>
+                          <th class="import_column_name ellipsis-resizable" sd-resizable><translate>Data File Header</translate></th>
+                      </tr>
+                  </thead>
+                  <tbody id="mapped-table">
+                      <tr ng-repeat="col in mappings">
+                          <td style="text-align: right;" ng-class="{'danger': col.is_duplicate || col.suggestion === ''}" ng-attr-id="mapped-row-type-{$:: $index $}">
+                              <select ng-model="col.suggestion_table_name" ng-change="updateInventoryTypeDropdown(); change(col)" ng-disabled="true">
```
adrian-lara: I reviewed this thinking that this file is basically a copy of the existing mapping partial except the ng-disabled="true" was added on these table items. 

Given we're not sure what'll need later, so I think keeping all of this is fine.

seed/data_importer/models.py
```diff
@@ -779,6 +785,9 @@ def cache_first_rows(self):
 
     @property
     def first_row_columns(self):
```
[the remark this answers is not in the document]
macintoshpie: Yeah I think you're right. For now I'm ok with leaving as is since it's just iterating over a static dict

seed/data_importer/views.py
```diff
@@ -1280,6 +1284,8 @@ def mapping_suggestions(self, request, pk):
                 default_mappings=pm_mappings,
                 thresh=80
             )
+        elif data_source_map.get(import_file.source_type) == BUILDINGSYNC_RAW:
```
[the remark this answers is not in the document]
macintoshpie: fixed, thanks for the tip!

PR #2145, opened 2020-03-18
1946 pm meter import fix

adrian-lara:
From the here https://github.com/SEED-platform/seed/issues/1946#issuecomment-579642099
> ...
> There's a notable flaw in the way I built out the importing of PM meters. That is, the PM import feature for meters and meter readings was built on the assumption that PM Property ID would always be at least one of the matching fields. During a PM meter import, I assumed it would be enough to find just one record with a given PM Property ID and then associate meters to that lone record. The thought was that matching and linking would take care of associating the meters and meter readings to multiple records via that first found record.
> 
> Of course, as you both know, PM Property ID may or may not be considered "matching criteria" within any given organization. I think the ideal scenario is _still_ to rely on PM Property ID during PM meter imports, but instead, apply incoming meters and meter readings to **all** records within a given cycle that have the given PM Property ID. I don't think this would be a massive effort, but I wanted to get both of your reactions on this before thinking about implementation steps.
> 
> Note that it would still be the case that meters are cycle-agnostic. Meaning, if a property has meters and meter readings and is linked to a property in another cycle, that other cycle['s PropertyState] will be associated to the same meters and meter readings.

#### What's this PR do?
Updates the PM Meter import code to associate incoming meters and meter readings to **all**  records across an org's cycles with a given PM Property ID.

Updates the feedback provided to the user on import to list the Cycles for each record.

**Caveat**: In cases where a single cycle has multiple properties with the same PM Property ID that's having meters and meter readings imported, it could look like there are duplicate summary rows.
Screen Shot 2020-03-18 at 10 46 48 AM
This is a product of it being difficult to distinguish these affected records in this view. From the backend perspective, these are differentiated by property_id, but showing that to users wouldn't be very useful. 

For code review proposes, using the commits to step through the changes is a viable approach. The most dense commit is the first as it contains a refactor as well as a "feature implementation" (sorry 😅 ).

Import Portfolio Manager meters with "PM Property ID"s that are found among multiple linked and unlinked property records. Here are the scenarios that should be tested:
- PM Meter import target records have the same PM Property ID and have links across cycles.
- PM Meter import target records have the same PM Property ID and are within the same cycle.

#1946 

Here's a case of linked records being PM Meter import targets:
Screen Shot 2020-03-18 at 10 51 33 AM


seed/data_importer/meters_parser.py
```diff
@@ -201,44 +244,43 @@ def _parse_gb_meter_details(self):
 
     def _get_property_id(self, source_id, shared_details):
         """
-        Find and cache property_ids to avoid querying for the same property_id
```
adrian-lara: The biggest change for this method was the shift from finding one (the first) property_id to capturing all property_ids associated to PropertyStates with multiple property_ids. This shift affects downstream methods.

seed/data_importer/meters_parser.py
```diff
@@ -264,16 +306,22 @@ def _parse_meter_readings(self, details, meter_details, start_time, end_time):
             'conversion_factor': conversion_factor
         }
 
-        meter_identifier = '-'.join([str(v) for k, v in meter_details.items()])
```
adrian-lara: This was affected by upstream changes (property_id to property_ids). Specifically, the code is modified to run through _all_ property_ids, instead of just _one_ property_id.

The changes here are mostly just reordering and indentation.

seed/data_importer/meters_parser.py
```diff
@@ -297,16 +345,22 @@ def _parse_cost_meter_reading(self, details, meter_details, start_time, end_time
             'conversion_factor': 1
         }
 
-        meter_identifier = '-'.join([str(v) for k, v in meter_details.items()])
+        # Associate this meter reading to each meter and property_id combination
```
adrian-lara: as stated in another comment:
> This was affected by upstream changes (property_id to property_ids). Specifically, the code is modified to run through _all_ property_ids, instead of just _one_ property_id.
> 
> The changes here are mostly just reordering and indentation.

seed/data_importer/tasks.py
```diff
@@ -779,6 +779,12 @@ def _save_greenbutton_data_task(readings, meter_id, meter_usage_point_id, progre
     meter = Meter.objects.get(pk=meter_id)
 
     result = {}
+    result_summary_key = "{} - {} - {}".format(
```
adrian-lara: Builds the summary key once in the beginning vs twice in the two most likely cases below.

seed/data_importer/tasks.py
```diff
@@ -856,12 +860,20 @@ def _save_pm_meter_usage_data_task(meter_readings, file_pk, progress_key):
             )
             with connection.cursor() as cursor:
                 cursor.execute(sql)
-                key = "{} - {}".format(meter.source_id, meter.get_type_display())
+                key = "{} - {} - {}".format(
```
adrian-lara: This is slightly different from how the other key in this method is built, so a similar pattern to the [other method](https://github.com/SEED-platform/seed/pull/2145/files#r398051680) cannot be used.

seed/data_importer/meters_parser.py
```diff
@@ -378,7 +356,14 @@ def _parse_cost_meter_reading(self, raw_details, meter_details, start_time, end_
...
-        # Associate this meter reading to each meter and property_id combination
+        self.distribute_meter_reading(meter_reading, meter_details)
+
+    def distribute_meter_reading(self, meter_reading, meter_details):
+        """
+        If a meter has not yet been parsed, its first reading details are saved
+        in a list. If a meter was previously parsed and has readings already,
+        any new readings are appended to that list.
+        """
         for property_id in meter_details.get('property_ids', []):
```
adrian-lara: The diff isn't showing as I expected, but the DRY'ed code can be found in this new method.

PR #2148, opened 2020-03-20
BRICR Prod into BRICR Dev

nllong:
I made some tweaks to get BRICR Prod running again. @macintoshpie can you verify these and merge with bricr-dev if they make sense. You may have already updated bricr-dev... so let me know.

I don't think the functional tests are going to pass, but the others should.


nllong:
> This generally looks good to me.
> 
> The functional tests did not pass:`session not created: This version of ChromeDriver only supports Chrome version 75`. Do we want to comment out the functional tests from the travis config so it won't be failing every time, or should we go ahead and fix this now?
> 
> I also added a commit to fix a flake8 issue where an imported package wasn't being used. On a related note, we may want to lockdown bricr-prod branch so that I can't commit/merge without a pr approval.

Thanks for fixing the flake8 issue. I'll comment out the test for now.

Issue #2119, opened 2020-02-24
Re-enable created and updated fields for master records on the front end


adrian-lara:
### Background
Within SEED, users were presented with created and updated datetime stamps for property and tax lot records.

During the matching revamp work, it was discovered that the datetime stamp was actually coming from the canonical Property and canonical Tax Lot respectively. Ideally, these would be coming from the respective -States models since that is what actually holds the record information that users are seeing. Since linking was not in place, canonical records and connected -State records were 1-to-1, and it wasn't actually an issue.

In preparation for linking, where canonical records might be 1-to-many with -State records, there was an attempt to present users with the created and updated from the -State models instead. To do that, those two fields were both initially created for the -State models, and the canonical versions were disabled. In doing so, the any fields named created and updated were inadvertently prevented from appearing to users.

### Expected Behavior
Created and updated values exist for property and tax lot records and are viewable on the frontend.

### Actual Behavior
Created and updated values for property and tax lot records are no longer viewable.

### Steps to Reproduce
Visit any of the inventory list or inventory detail pages and find that Updated or Created columns are missing. View the column list settings page, and see that these columns are missing as options.

### Instance Information
Server Instance: `prod` and `develop`
Server SHA: `9105437`

adrian-lara:
For `prod`, it was enough to just re-reveal the (canonical) Property and Tax Lot `created` and `updated values. This was possible because, as mentioned above, at the time, that deployment didn't have linking and the updated and created  canonical times was a reasonable stand in for the actual -State values.

adrian-lara:
For `develop`, the fix for this had more complexity and research as we wanted to shift users from the canonical to the -State values. 

Actually, when matching was built out, [the -State values were actually taken from corresponding audit log objects](https://github.com/SEED-platform/seed/commit/efbff203616583d726a526838c5cfabb7df5c754#diff-44e89af066088d7be1e76bd679c1a247). 

Later, the SQL query conducting this copying was inadvertently moved into a different migration: `seed/migrations/0111_rehash.py`. Upon this discovery, it was noticed that this rehashing changed every updated value to whenever that migration was run. 

We [changed the order of when those migrations were run](https://github.com/SEED-platform/seed/pull/2149/files/746c83dcff2c40805abfeb106a4f67107c7a1fed#diff-41b300152661b0f46301c2177e654337).
Also, we decided to add another migration similar to the first (that copied values from audit logs) just for the updated times.
[...]

PR #2149, opened 2020-03-20
2119 created updated

adrian-lara:
---------This PR contains a migration---------

See #2118 for the 2.6.1 release version of this fix - that was more of a hot-fix.

This is presumably a more thorough fix.

From that PR:
> Within SEED, users were presented with created and updated datetime stamps for property and tax lot records.
> 
> During the matching revamp work, it was discovered that the datetime stamp was actually coming from the canonical Property and canonical Tax Lot respectively. Ideally, these would be coming from the respective -States models since that is what actually holds the record information that users are seeing. Since linking was not in place, canonical records and connected -State records were 1-to-1, and it wasn't actually an issue.
>...

#### What's this PR do?
This PR shifts users away from created and updated values coming from the canonical objects and uses the values coming from the -State objects, with the main focus being on what's provided in the inventory list and detail pages.

Also, on the inventory detail page, all the historical created and updated values are provided as well.

Import, merge, update, etc. records and sanity-check created and updated times on inventory list and detail pages. 

On a production database, cross-check unedited, unmerged records' created and updated times with the corresponding file import time. I've spot checked this myself, but more checks from others couldn't hurt. (see **Further considerations**)

As far as code review, there aren't many code changes, but commits can be used to help distinguish the "reason" behind changes.

#### Further considerations
- ~~I was expecting to have to backfill the created and updated -State values with -related audit logs, but that didn't seem to be the case. The -State records seem to have created values that would indicate they are the actual created times when looking at corresponding import file dates - the hour might be off but I think that might be caused by timezone differences. **Anyone know why the -State records looks to have valid values already?**~~ The migrations filling in these values can be found in 0111_rehash.py.

- Initial record edits trigger the creation of a new -State. As far as I know, users aren't relying on the created time for bulk downstream processes. For non-bulk created times analysis, since historical records now have these values, created times can be read off the "oldest" version of the record. (See screenshot below). I don't think needs to be changed but is worth noting.

#2119 

Screen Shot 2020-03-20 at 12 41 34 PM


seed/migrations/0121_update_updated_timestamps.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+from django.db import connection, migrations
+
+
+def forwards(apps, schema_editor):
+    property_sql = (
+        'UPDATE seed_propertystate '
+        'SET updated = (SELECT created '
```
adrian-lara: This and the tax lot version of this change looks good - thanks for making it Alex!

PR #2156, opened 2020-03-26
Update TravisCI postgres port to 5433

adrian-lara:
Cherry-picks #2124 into 2.7.0

See that PR for details.

For testing, we only need to verify the TravisCI run and pass.


adrian-lara:
Closed since ESPM patch needed to have test pass.

PR #2166, opened 2020-03-27
270 - Update 0119_ migration with a state_operation to align migration state

adrian-lara:
`recognize_empty` was added to the column table as a patch in a previous release, 2.6.1. A modified version of that migration needed to be included in later releases like this so that databases could be carried over easily between release upgrades.

#### What's this PR do?
This PR further modifies that migration to align Django's underlying migration state.

With a DB that already has `recognize_empty` on the column table, run this updated migration.
Then verify that running `./manage.py makemigrations` leads to `No changes detected`.

Ideally, check out `2.6.1-patch3`, create a new DB, run migrations, check out this branch and finishing running all the migrations. Finally verify with `makemigrations` as described above.


axelstudios:
Didn't test, but the changes look good

seed/migrations/0119_column_recognize_empty.py
```diff
@@ -15,6 +15,13 @@ class Migration(migrations.Migration):
         migrations.RunSQL(
             'ALTER TABLE seed_column'
             '    ADD COLUMN IF NOT EXISTS recognize_empty BOOLEAN NOT NULL DEFAULT FALSE;',
-            reverse_sql='ALTER TABLE seed_column DROP COLUMN recognize_empty'
+            reverse_sql='ALTER TABLE seed_column DROP COLUMN recognize_empty',
+            state_operations=[
+                migrations.AddField(
```
nllong: 👍

PR #2171, opened 2020-04-05
Initial AutoSchemaHelper example

adrian-lara:
With this PR, I'm looking to establish a pattern to build, maintain, and standardize the API definitions that will be used by Swagger. We'll likely iterate on this pattern as we come across different cases while building out v3.

**Note, the FE code doesn't reference the new data_quality_checks endpoints nor are they updated as required for v3. This is only to be used as a reference for using AutoSchemaHelper.**

The AutoSchemaHelper class will be used to create Schema classes that will be applied to ViewSets.

Given that some v1, v2, v2.1 endpoints will be updated heavily changed, we'll be copying/creating ViewSet classes to the `./seed/views/v3` directory as shown in this example.

### DIFF
--- a/seed/api/base/urls.py
+++ b/seed/api/base/urls.py
@@ -9,10 +9,12 @@ from django.conf.urls import url, include
 from seed.api.v1.urls import urlpatterns as apiv1
 from seed.api.v2.urls import urlpatterns as apiv2
 from seed.api.v2_1.urls import urlpatterns as apiv2_1
+from seed.api.v3.urls import urlpatterns as apiv3
 
 urlpatterns = [
     # add flat urls namespace for non-conforming endpoints, ugh
     url(r'^v1/', include((apiv1, 'seed'), namespace='v1')),
     url(r'^v2/', include((apiv2, 'seed'), namespace='v2')),
     url(r'^v2.1/', include((apiv2_1, 'seed'), namespace='v2.1')),
+    url(r'^v3/', include((apiv3, 'seed'), namespace='v3')),
 ]
--- a/seed/utils/api_schema.py
+++ b/seed/utils/api_schema.py
@@ -1,0 +1,75 @@
+# !/usr/bin/env python
+# encoding: utf-8
+from rest_framework.schemas import AutoSchema
+import coreapi
+
+
+class AutoSchemaHelper(AutoSchema):
+    def __init__(self):
+        super().__init__()
+
+    def org_id_field(self, location='query', required=True, type='integer', description='Organization ID'):
+        return coreapi.Field(
+            'organization_id',
+            location=location,
+            required=required,
+            type=type,
+            description=description
+        )
+
+    def path_id_field(self, description):
+        return coreapi.Field(
+            "id",  # matches reference name in uri path
+            location='path',
+            required=True,
+            type='integer',
+            description=description
+        )
+
+    def body_field(self, required, description):
+        return coreapi.Field(
+            'body',
+            location='body',
+            required=required,
+            type='object',
+            description=description
+        )
+
+    def form_field(self, name, required, description):
+        return coreapi.Field(
+            name,
+            location='form',
+            required=required,
+            type='object',
+            description=description
+        )
+
+    def test_field_options(self):
+        return [
+            coreapi.Field('remove_label_ids', location='form', required=False, type='array'),
+            coreapi.Field('inventory_ids', location='form', required=False, type='array', description="This is what the description looks like for an 'array'."),
+            coreapi.Field('string test', location='form', required=False, type='string'),
+            coreapi.Field('integer test', location='form', required=False, type='integer'),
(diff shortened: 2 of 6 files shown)


seed/api/base/urls.py
```diff
@@ -9,10 +9,12 @@
...
 from seed.api.v2.urls import urlpatterns as apiv2
 from seed.api.v2_1.urls import urlpatterns as apiv2_1
+from seed.api.v3.urls import urlpatterns as apiv3
 
 urlpatterns = [
     # add flat urls namespace for non-conforming endpoints, ugh
     url(r'^v1/', include((apiv1, 'seed'), namespace='v1')),
     url(r'^v2/', include((apiv2, 'seed'), namespace='v2')),
     url(r'^v2.1/', include((apiv2_1, 'seed'), namespace='v2.1')),
+    url(r'^v3/', include((apiv3, 'seed'), namespace='v3')),
```
nllong, line 19: yep, v3!

seed/utils/api_schema.py
```diff
@@ -0,0 +1,75 @@
+# !/usr/bin/env python
+# encoding: utf-8
+from rest_framework.schemas import AutoSchema
+import coreapi
+
+
+class AutoSchemaHelper(AutoSchema):
```
nllong, line 7: seems like a nice class

seed/views/data_quality.py
```diff
@@ -92,12 +93,51 @@ def _get_severity_from_js(severity):
     return d.get(severity)
 
 
+class DataQualitySchema(AutoSchemaHelper):
+    def __init__(self):
+        super().__init__()
+
+        self.manual_fields = {
```
nllong, line 100: I'm going to miss the magic that we had. But there is something nice to be said about being explicit.

seed/api/base/urls.py
adrian-lara: Ah, I kept meaning to ask you - Did you want to call this v2.2 or v3? I think you were originally thinking of calling it v2.2 and I might've accidentally called it v3 in one of our meetings and just kept calling it that. 😅

seed/views/v3/data_quality.py
```diff
@@ -0,0 +1,478 @@
...
+from seed.lib.superperms.orgs.decorators import has_perm_class
+from seed.lib.superperms.orgs.models import (
+    Organization,
+)
+from seed.models.data_quality import (
+    Rule,
+    DataQualityCheck,
+)
+from seed.utils.api import api_endpoint_class
+from seed.utils.api_schema import AutoSchemaHelper
```
adrian-lara, line 27: This line and others defining or referencing the DataQualitySchema class are the only notable changes. The rest of the lines were just copied over.

seed/utils/api_schema.py
```diff
@@ -0,0 +1,88 @@
...
+    def form_field(self, name, required, description):
+        return coreapi.Field(
+            name,
+            location='form',
+            required=required,
+            type='object',
+            description=description
+        )
+
+    def test_field_options(self):
```
adrian-lara, line 60: Leaving this here for reference:

Screen Shot 2020-04-11 at 1 58 36 PM

seed/views/v3/data_quality.py
```diff
@@ -0,0 +1,478 @@
...
+    d = {v: k for k, v in dict(Rule.SEVERITY).items()}
+    return d.get(severity)
+
+
+class DataQualitySchema(AutoSchemaHelper):
+    def __init__(self):
+        super().__init__()
+
+        self.manual_fields = {
+            ('POST', 'create'): [
```
adrian-lara, line 101: Can anyone think of a case where (`HTTP method`, `ViewSet method/function name`) would not be enough to identify a ViewSet method?

seed/views/v3/data_quality.py
```diff
@@ -0,0 +1,478 @@
...
+class DataQualityViews(viewsets.ViewSet):
+    """
+    Handles Data Quality API operations within Inventory backend.
+    (1) Post, wait, get…
+    (2) Respond with what changed
+    """
+    schema = DataQualitySchema()
+
+    @action(detail=False, methods=['GET'])
+    def testing_core_api(self, request):
```
adrian-lara, line 136: Created to capture other, likely less common schema field types.

seed/api/base/urls.py
nllong: I know we have been saying v2.2. I'm fine with v3. It seems like this is going to be a big update so maybe a big version bump is worth it.

seed/views/v3/data_quality.py
nllong: not that I can think of with this level of information. Can you give an example?

adrian-lara: I agree and think a ViewSet method has a unique pair of values for HTTP method and ViewSet function name. I was thinking that ViewSet function name would be enough, but I figured that having both couldn't hurt.

PR #2173, opened 2020-04-08
fix(buildingsync): defer creation of bsync models until last task

macintoshpie:
After moving BuildingSync import into the normal import/mapping flow, I realized there were a few issues:
1. the mapping task was not idempotent - multiple building files, property states, etc were duplicated each time
2. we created the Scenarios, Meters, and other models associated with the property too soon (during the mapping), and should defer until the user is done and wants to finish the import.
#### What's this PR do?
Addresses the issues above:
1. ensures mapping can get called multiple times
2. defers the creation of additional models by creating a new task that BuildingSync imports will take, rather than the property merging/matching step.

NOTE: as mentioned above, BuildingSync imports don't go through the merging step. This is how the import of BuildingSync files worked originally before moving it into the tasks backend, so we decided this was an ok option for now. However, we need to have discussions down the road about merging BuildingSync properties as there are a few small complications.
Use the data from the building_sync/tests/data directory and upload a couple different files (at a minimum the bricr_workflow file, the zip file). You should be directed through the mapping flow, and by the end scenarios and meters should be created for them (if they have them, which the bricr_workflow file does).

Since this change partially touches the code which "Update with BuildingSync" uses, it's worth additionally updating a property with a buildingsync file and verifying that it worked.

### DIFF
--- a/seed/data_importer/tasks.py
+++ b/seed/data_importer/tasks.py
@@ -446,7 +446,7 @@ def map_xml_chunk(ids, file_pk, file_type, prog_key, **kwargs):
 
     :param ids: list of PropertyState IDs to map.
     :param file_pk: int, the PK for an ImportFile obj.
-    :param file_type: int, represented by either BUILDINGSYNC or HPXML
+    :param file_type: int, represented by either BuildingFile.BUILDINGSYNC or BuildingFile.HPXML
     :param prog_key: string, key of the progress key
     :param increment: double, value by which to increment progress key
     """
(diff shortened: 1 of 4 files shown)


seed/data_importer/tasks.py
```diff
@@ -464,18 +464,35 @@ def map_xml_chunk(ids, file_pk, file_type, prog_key, **kwargs):
...
+                    # This buildingfile already exists
+                    building_file = building_files[0]
+                else:
+                    # Uh oh, apparently our requirements for "uniqueness" is insufficient
+                    raise Exception('Found more than one building file and expected 1 or 0')
+
+                # create or update the PropertyState linked to the building file
+                p_status, property_state, messages = building_file.process_property_state(org.id)
 
-                p_status, property_state, property_view, messages = building_file.process(org.id, import_file.cycle)
```
adrian-lara, line 478: 💯

seed/models/building_file.py
```diff
@@ -120,33 +129,73 @@ def process(self, organization_id, cycle, property_view=None):
...
+                state_id=property_state.id,
+                name='Import Creation',
+                description='Creation from Import file.',
+                import_filename=self.file.path,
+                record_type=AUDIT_IMPORT
+            )
+        else:
+            property_state = self.property_state
+            property_state.__dict__.update(create_data)
+            property_state.extra_data = extra_data
```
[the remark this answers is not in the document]
macintoshpie: We also call `_create_property_state` in `process_property_state`, which can be called more than once (create the first time, update the following times). This happens when the user goes back and forth between the mapping pages (reapplying the mapping)

seed/models/building_file.py
```diff
@@ -143,10 +145,50 @@ def process(self, organization_id, cycle, property_view=None):
             record_type=AUDIT_IMPORT
         )
 
+        property_state.save()
+
```
adrian-lara, line 149:
```suggestion

```
I think we can entirely remove this now.

PR #2174, opened 2020-04-08
Fixed unnecessary URL params

axelstudios:
Since late February URLs have had a ton of unnecessary params added on:
`https://dev1.seed-platform.org/app/#/properties?=%2F%7Binventory_type:properties%7Ctaxlots%7D`
instead of
`https://dev1.seed-platform.org/app/#/properties`

#### What's this PR do?
Fixes it.  The main issue was calling $location.search.

Related bug: https://github.com/angular-ui/ui-router/issues/3416

Look at the address bar

### DIFF
--- a/seed/static/seed/js/controllers/menu_controller.js
+++ b/seed/static/seed/js/controllers/menu_controller.js
@@ -67,17 +67,7 @@ angular.module('BE.seed.controller.menu', [])
       });
 
       $scope.is_active = function (menu_item) {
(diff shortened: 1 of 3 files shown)


axelstudios:
Yeah, to get rid of the `$location.search` usage I reorganized some of the logic around changing the URL when the user stays on the same page.  It made sense to put it in the ui-router transition error handler

PR #2187, opened 2020-04-16
Chore/merge bricr dev

macintoshpie:
Brings recent additions for BuildingSync files from bricr-dev into develop


seed/tests/test_api.py
```diff
@@ -171,7 +169,7 @@ def test_organization(self):
         self.assertEqual(r['organizations'][0]['owners'][0]['first_name'], 'Jaqen')
         self.assertEqual(r['organizations'][0]['cycles'], [
             {
-                'name': str(date.today().year - 1) + ' Calendar Year',
```
adrian-lara, line 174: I think we want deletions vs the additions here.

macintoshpie: I think you're right, I'll change that

PR #2188, opened 2020-04-23
Shift Swagger Schema Generator to drf-yasg

adrian-lara:
We needed to move away from the, now deprecated, django-rest-swagger as a schema generator/renderer. (Not to be confused with django-rest-**framework**, which is still being used 😄 .)

#### What's this PR do?
Swagger page now uses drf_yasg.

**Reminder to refresh installed packages.**

Check the Swagger page still renders. See that the v3 DQ endpoints are documented and "trying" each of the endpoints should work.

For code review, each commit can be used to break up and partition review. The last commit is a small clean up item.

For the https://github.com/SEED-platform/seed/pull/2188/commits/78fb1a849009978b549265b5076970d992b48e88 commit, it might be easier to review the resulting code as opposed to code diffs as the diffs shown by GH are not as readable in my opinion.

Also note that there are AssertionErrors being thrown across a number of ViewSets as there are some additional setup steps likely needed. Things seem to work to the extent they did before - v2 endpoints are still visible, so I think we can address those errors later.

Screen Shot 2020-04-23 at 2 03 41 PM

### DIFF
--- a/seed/utils/api_schema.py
+++ b/seed/utils/api_schema.py
@@ -36,11 +36,11 @@ class AutoSchemaHelper(SwaggerAutoSchema):
             type=openapi.TYPE_INTEGER
         )
 
-    def query_integer_field(self, name, description, required):
+    def query_integer_field(self, name, required, description):
         return openapi.Parameter(
             name,
             openapi.IN_QUERY,
-            description="Task ID created when DataQuality task is created.",
+            description=description,
             required=True,
             type=openapi.TYPE_INTEGER
         )
--- a/seed/utils/api_schema.py
+++ b/seed/utils/api_schema.py
@@ -1,90 +1,80 @@
 # !/usr/bin/env python
 # encoding: utf-8
-from rest_framework.schemas import AutoSchema
-import coreapi
+from drf_yasg.inspectors import SwaggerAutoSchema
+from drf_yasg import openapi
 
 
-class AutoSchemaHelper(AutoSchema):
-    def __init__(self):
-        super().__init__()
+class AutoSchemaHelper(SwaggerAutoSchema):
 
-    def base_field(self, name, location, required, type, description):
+    # Used to easily build out example values displayed on Swagger page.
+    body_parameter_formats = {
+        'interger_list': openapi.Schema(
+            type=openapi.TYPE_ARRAY,
+            items=openapi.Schema(type=openapi.TYPE_INTEGER)
+        )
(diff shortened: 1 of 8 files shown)


seed/views/v3/data_quality.py
```diff
@@ -94,33 +94,45 @@ def _get_severity_from_js(severity):
...
         self.manual_fields = {
             ('POST', 'create'): [
                 self.org_id_field(),
-                self.form_field(
+                self.body_field(
                     name='data_quality_ids',
                     required=True,
-                    description="An object containing IDs of the records to perform data quality checks on. Should contain two keys- property_state_ids and taxlot_state_ids, each of which is an array of appropriate IDs."
+                    description="An object containing IDs of the records to perform data quality checks on. Should contain two keys- property_state_ids and taxlot_state_ids, each of which is an array of appropriate IDs.",
+                    params_to_formats={
```
adrian-lara, line 107: This provides a nice "default value" on the Swagger pager, but I was struggling to find a pattern for us developers to use and reuse.

Screen Shot 2020-04-23 at 2 26 36 PM

seed/utils/api_schema.py
```diff
@@ -1,90 +1,80 @@
...
+    def body_field(self, required, description, name='body', params_to_formats={}):
+        return openapi.Parameter(
             name,
-            location='form',
+            openapi.IN_BODY,
+            description=description,
             required=required,
-            type='object',
-            description=description
+            schema=self._build_body_schema(params_to_formats)
```
adrian-lara, line 63: The starting point for building out the  "default value" on the Swagger pager as mentioned in [the other comment](https://github.com/SEED-platform/seed/pull/2188/files#r414104355). 

This will likely evolve/change as we try to apply the same type of "default value" to other endpoints with body params. In someway, we might also just give ourselves direct access to the openapi library to be able to build these out directly in the -Schema classes.

Screen Shot 2020-04-23 at 2 26 36 PM

aviveiros11:
@adrian-lara should we merge this? Or wait for @axelstudios and @nllong

adrian-lara:
@aviveiros11 I think this is fine to merge - _Friday's are for merging bugs anyway 😉_ 

The use and tweaking of the AutoSchemaHelper was meant to be iterative. Also, the v3 endpoints aren't being used on the frontend (minus the Swagger page), so I think this is pretty low risk.

PR #2189, opened 2020-04-27
refactor(bsync): fully integrate into import flow

macintoshpie:
This pr further integrates BuildingSync import with the regular import flow.
#### What's this PR do?
- parses buildingsync files immediately during 'saving the raw data' task, rather than waiting until the 'mapping' task
- removes the xml specific mapping task, now uses the map_row_chunk
- removes mapping_xml partial for showing the bsync mapping page, now uses the regular mapping partial

Previously, importing BuildingSync files would fail if there were any 'required' fields it found missing. This PR makes a change that ignores any fields it fails to parse for the PropertyState during the import flow. It is now the data quality checker's job to warn users when their properties are missing info (address line 1, PM property id, etc). This change _does_ warn users about bad scenarios/measures/meter data, which is provided after the last step of import ("save mapping") in the data import modal.

Import and export a variety of buildingsync files (`seed/building_sync/tests/data`)
- `buildingsync_ATT_export.xml`: import OK (it should give some warnings about scenarios not having names -- this is due to the way ATT uses Scenarios); should have scenarios and property info
- `buildingsync_ex01_measures_bad_names.xml`: import should be normal until you click "save mappings". It should give a warning that some measure names were bad and some scenarios were not linked to measures as a result.
- `buildingsync_v2_0_bricr_workflow.xml`: import OK; should have scenarios, measures, and Meter data
- `ex_1_no_schemaLocation.xml`: should immediately fail when you try to import it with an error about the schemaLocation.
- `ex_1_and_buildingsync_ex01_measures.zip`: this should show TWO properties on the mapping page. One should show measures and scenario info on the property detail page, while the other should not. Verify that "export as buildingsync" works for each property (e.g. just check it does/doesn't have the measures and scenarios)


seed/data_importer/tasks.py
```diff
@@ -253,11 +254,15 @@ def map_row_chunk(ids, file_pk, source_type, prog_key, **kwargs):
...
         save_type = ASSESSED_BS
+    elif source_type == BUILDINGSYNC_RAW:
+        save_type = BUILDINGSYNC_RAW
 
     org = Organization.objects.get(pk=import_file.import_record.super_organization.pk)
 
     # get all the table_mappings that exist for the organization
     table_mappings = ColumnMapping.get_column_mappings_by_table_name(org)
+    if source_type == BUILDINGSYNC_RAW:
+        table_mappings['PropertyState']['_source_filename'] = ('PropertyState', '_source_filename', '', True)
```
macintoshpie, line 265: When later mapping additional models for bsync files in map_additional_models (BuildingFile, Scenario, Meters, ...) we need to re-read the file. If the imported file was a zip (ie multiple xml files), we thus need to know which file a PropertyState came from.

It's unfortunate to have to include this but I couldn't think of a better way. I decided to basically 'hardcode' this into all buildingsync table mappings here because it felt too strange for whoever was generating column mappings (presets or some buildingsync function) to be responsible for this weird import quirk. Looking for feedback and ideas

seed/data_importer/tasks.py
```diff
@@ -1363,12 +1277,40 @@ def _map_additional_models(ids, file_pk, progress_key):
...
+
     # grab all property states linked to the import file and finish processing the data
     property_states = PropertyState.objects.filter(id__in=ids)
     for property_state in property_states:
         if source_type == BUILDINGSYNC_RAW:
-            # we expect only one buildingfile to be associated with the property
-            building_file = BuildingFile.objects.get(property_state=property_state.id)
+            # create a BuildingFile
+            if from_zipfile:
+                filename = property_state.extra_data.get('_source_filename')
```
macintoshpie, line 1288: see the comment I made in map_row_chunk

seed/data_importer/tasks.py
```diff
@@ -1363,12 +1277,40 @@ def _map_additional_models(ids, file_pk, progress_key):
...
+                        content_type='application/xml')
+            else:
+                filename = import_file.uploaded_filename
+                new_file = SimpleUploadedFile(
+                    name=filename,
+                    content=import_file.file.read(),
+                    content_type='application/xml'
+                )
+
+            if filename == '':
```
macintoshpie, line 1304: Added this check because I realized creating a file without a filename would result in the building file's file to be `None`, so when trying to parse the file it'd complain about trying to read a non-existent file

seed/static/seed/partials/mapping.html
```diff
@@ -148,15 +148,15 @@
                         </tr>
                     </thead>
                     <tbody id="mapped-table">
-                        <tr ng-repeat="col in mappings">
+                        <tr ng-repeat="col in mappings | filter:{suggestion: '!_source_filename'}">
```
macintoshpie, line 151: since `_source_filename` ends up a column mapping (see my other comments), we need to filter it out here so the user doesn't see it

seed/static/seed/js/controllers/data_upload_modal_controller.js
```diff
@@ -578,6 +574,13 @@ angular.module('BE.seed.controller.data_upload_modal', [])
                 $scope.uploader.in_progress = false;
                 $scope.uploader.progress = 0;
                 $scope.uploader.status_message = '';
+                debugger
```
adrian-lara, line 577: debugger 😬

adrian-lara:
> - ex_1_no_schemaLocation.xml: should immediately fail when you try to import it with an error about the schemaLocation.

Running with celery causes this to hang as opposed to showing the error. Running without celery does trigger an explicit fail - showing the error message.

adrian-lara:
> - buildingsync_ATT_export.xml: import OK; should have scenarios and property info

This import shows the following warnings/errors about the scenarios.
Screen Shot 2020-04-28 at 12 55 05 PM

macintoshpie:
> > * buildingsync_ATT_export.xml: import OK; should have scenarios and property info
> 
> This import shows the following warnings/errors about the scenarios.
> Screen Shot 2020-04-28 at 12 55 05 PM

Sorry, wrote the wrong thing for the test documentation. That's the expected behavior

seed/data_importer/tasks.py
```diff
@@ -743,6 +697,13 @@ def _save_raw_data_chunk(chunk, file_pk, progress_key):
                 raw_property.data_state = DATA_STATE_IMPORT
                 raw_property.organization = import_file.import_record.super_organization
                 raw_property.save()
+
+                if source_filename is not None:
+                    raw_property_state_to_filename[str(raw_property.id)] = source_filename
+
+            import_file.raw_property_state_to_filename.update(raw_property_state_to_filename)
```
adrian-lara, line 704: The way this is updated, I would think this **_avoids**_ a race condition if we have multiple tasks. But I may either be testing this weird or the race condition does actually happen.

Screen Shot 2020-04-29 at 4 41 03 PM

I uploaded the test zip file twice. For the first time, I allowed the 2 raw records to be created in one batch and it looks fine. For the second, I made the batch size 1 and saw that updating the zip only returned one key, value pair in the `raw_property_state_to_filename` field.

adrian-lara:
Nevermind, the bug was unrelated. This is good to go 👍

Issue #2180, opened 2020-04-12
API V3 - Dataset


adrian-lara:
Build out the following v2 endpoints into v3, update fields shown in Swagger, and also make changes as listed:
- [x] **/api/v2/datasets/ - get**: move
- [x] **/api/v2/datasets/ - post**: move
- [x] **/api/v2/datasets/{id}/ - delete**: move - and verify it deactivates instead
- [x] **/api/v2/datasets/{id}/ - get**: move
- [x] **/api/v2/datasets/{id}/ - put**: move
- [x] **/api/v2/datasets/count/ - get**: move

Additionally, review the following:
- [x] use of DRF ViewSets.
- [x] use of decorators is reasonable/
- [x] comments include at least a description. Details about incoming or outgoing fields are nice-to-haves as well.

Manual Testing:
Visit the /app/#/api/swagger page and "Execute" the endpoints listed above.

RDmitchell:
@nllong -- I don't believe this is anything I can test.

PR #2191, opened 2020-04-30
2180 apiv3 dataset

aviveiros11:
The effort to move to API v3 - specifical datasets
#### What's this PR do?
translates API v2 into v3 with auto schema helper 
Use the `Try It` functionality rendered by the API page on SEED for v3
#2180


aviveiros11:
@adrian-lara, @axelstudios this is just about there.  I wanted to get some more clarification on how we are defining "deactivate" in task 3 of #2180.  If we are just refactoring to deactivate I believe we have to touch the Model? Or are we looking to refactor the delete functionality to keep the dataset but flag it as deactivated?

axelstudios:
@aviveiros11 I'm pretty sure that means don't physically delete the row when the DELETE method is called, but instead toggle the `deleted` boolean column to true:

image

adrian-lara:
> @adrian-lara, @axelstudios this is just about there. I wanted to get some more clarification on how we are defining "deactivate" in task 3 of #2180. If we are just refactoring to deactivate I believe we have to touch the Model? Or are we looking to refactor the delete functionality to keep the dataset but flag it as deactivated?

Definitely +1 to what Alex said. It might be the case that this is already happening ("deactivate" flagging is happening instead of actual DB row deletions), so definitely double check that.

As far as potentially making changes, I think you're asking if the "deactivate" logic requires changing logic at the Model level or the ViewSet level? If I'm understanding that question correctly, I guess we'd have to answer the question of "Do we want this to happen only when this endpoint is hit or whenever anything (other models, utils, etc) tries to delete a record?".

aviveiros11:
Confirmed with Alex that the column in fact was being updated.  Ready for review.

seed/views/v3/datasets.py
```diff
@@ -0,0 +1,374 @@
...
+
+        return JsonResponse({
+            'status': 'success',
+            'dataset': dataset,
+        })
+
+    @api_endpoint_class
+    @ajax_request_class
+    @has_perm_class('requires_member')
+    def destroy(self, request, pk=None):
```
adrian-lara, line 257: Deactivates as expected 👍 
Screen Shot 2020-05-06 at 4 31 16 PM

seed/utils/api_schema.py
```diff
@@ -11,7 +11,8 @@ class AutoSchemaHelper(SwaggerAutoSchema):
         'interger_list': openapi.Schema(
             type=openapi.TYPE_ARRAY,
             items=openapi.Schema(type=openapi.TYPE_INTEGER)
-        )
+        ),
+        'string': openapi.Schema(type=openapi.TYPE_STRING)
```
adrian-lara, line 15: 👍

seed/views/v3/datasets.py
```diff
@@ -0,0 +1,374 @@
...
+from seed.models import obj_to_dict
+from seed.utils.api import api_endpoint_class
+from seed.utils.time import convert_to_js_timestamp
+
+from seed.utils.api_schema import AutoSchemaHelper
+
+_log = logging.getLogger(__name__)
+
+
+class DatasetSchema(AutoSchemaHelper):
```
adrian-lara, line 28: Let me know if there's anything notable changed outside of this class, the AutoSchemaHelper import, and the application of this DatasetSchema class to the -ViewSet (`swagger_schema=...`). 

I assumed these were the only "real" changes this file and mostly everything else was just copied over.

aviveiros11: Yes exactly, just wanted to follow the AutoSchmaHelper pattern but mostly everything was just copied over 👍 .

PR #2211, opened 2020-05-11
patch(meters.py): handle mistyping

macintoshpie:
Small patch for mistyped query parameter. This isn't a current issue (ie it should not actually cause runtime exceptions) but is an error nonetheless.

### DIFF
--- a/seed/utils/meters.py
+++ b/seed/utils/meters.py
@@ -41,6 +41,7 @@ class PropertyMeterReadingsExporter():
         self._cache_factors = None
         self._cache_org_country = None
 
+        scenario_ids = scenario_ids if scenario_ids is not None else []
         self.meters = Meter.objects.filter(
             Q(property_id=property_id) | Q(scenario_id__in=scenario_ids)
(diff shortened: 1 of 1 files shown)


seed/utils/meters.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, property_id, org_id, excluded_meter_ids, scenario_ids=None):
         self._cache_factors = None
         self._cache_org_country = None
 
+        scenario_ids = scenario_ids if scenario_ids is not None else []
```
adrian-lara, line 44: I think we could also set the default for `scenario_ids` as `[]` when defining the parameters.

macintoshpie: I used this pattern because python's default parameters are only evaluated once when the function is defined, so dicts and lists can cause some surprising results if modified. Check this out here: https://docs.python-guide.org/writing/gotchas/

adrian-lara: Ah ok - didn't realize this could have been an issue! 👍

Issue #2176, opened 2020-04-12
API V3 - Users


adrian-lara:
Build out the following v2 endpoints into v3, update fields shown in Swagger, and also make changes as listed:
- [x] **/api/v2/users/ - get**: move
- [x] **/api/v2/users/ - post**: move
- [x] **/api/v2/users/{id}/ - get**: move
- [x] **/api/v2/users/{id}/ - put**: move
- [x] **/api/v2/users/{id}/ - delete**: make new - deactivate user
- [x] **/api/v2/users/{id}/default_organization/ - put**: move
- [x] **/api/v2/users/{id}/generate_api_key/ - get**: change to post
- [x] **/api/v2/users/{id}/is_authorized/ - post**: move
- [x] **/api/v2/users/{id}/set_password/ - put**: move
- [x] **/api/v2/users/{id}/shared_buildings/ - get**: investigate - bpd link
- [x] **/api/v2/users/{id}/update_role/ - put**: rename to role (vs update_role)
- [x] **/api/v2/users/current_user_id/ - get**: rename to current (vs current_user_id)

Additionally, review the following:
- [x] use of DRF ViewSets.
- [x] use of decorators is reasonable/
- [x] comments include at least a description. Details about incoming or outgoing fields are nice-to-haves as well.

Manual Testing:
Visit the /app/#/api/swagger page and "Execute" the endpoints listed above.

RDmitchell:
@nllong -- I don't believe this is anything I can test.

PR #2212, opened 2020-05-11
2176 apiv3 users

aviveiros11:
The effort to move to API v3 - specifically users
#### What's this PR do?
translates API v2 into v3 with auto schema helper 
Use the `Try It` functionality rendered by the API page on SEED for v3
#2176


seed/views/v3/users.py
```diff
@@ -0,0 +1,887 @@
...
+                description: error message, if any
+                required: false
+        """
+        body = request.data
+        first_name = body['first_name']
+        last_name = body['last_name']
+
+        print(first_name, last_name)
+        # check if user exists
+        user = SEEDUser.objects.filter(
```
adrian-lara, line 876: I would've expected we want to use the id of the user vs `first_name` and `last_name`. Any reason to take this approach?

seed/views/v3/users.py
```diff
@@ -0,0 +1,887 @@
...
+        except OrganizationUser.DoesNotExist:
+            return None
+
+        return {
+            action: PERMS['requires_owner'](ou) for action in actions
+        }
+
+    @ajax_request_class
+    @action(detail=True, methods=['GET'])
+    def shared_buildings(self, request, pk=None):
```
adrian-lara, line 771: Was there anything that changed with this? I saw that we had a note down to investigate.

seed/utils/api_schema.py
```diff
@@ -12,7 +12,12 @@ class AutoSchemaHelper(SwaggerAutoSchema):
             type=openapi.TYPE_ARRAY,
             items=openapi.Schema(type=openapi.TYPE_INTEGER)
         ),
-        'string': openapi.Schema(type=openapi.TYPE_STRING)
+        'string': openapi.Schema(type=openapi.TYPE_STRING),
+
+        'string_array': openapi.Schema(
```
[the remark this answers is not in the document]
aviveiros11: Yeah! I can do that.

seed/views/v3/users.py
aviveiros11: In its current state, the GET successfully returns if the user has shared buildings enabled or not.  Did we want to change this?

adrian-lara: @nllong or @axelstudios This is the note we had `/api/v2/users/{id}/shared_buildings/ - get: investigate - bpd link`

Do either of you remember what the intent was behind that? (I'm not sure what bpd is.)

seed/views/v3/users.py
```diff
@@ -0,0 +1,887 @@
...
+            user = content
+        else:
+            return content
+        user.default_organization_id = request.query_params.get('organization_id', None)
+        user.save()
+        return {'status': 'success'}
+
+    @ajax_request_class
+    @action(detail=True, methods=['DELETE'])
+    def deactivate(self, request, pk=None):
```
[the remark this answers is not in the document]
adrian-lara: Thanks for making the PUT change, Austin.

@nllong @axelstudios thoughts on the above comment?

I wouldn't envision needing to build out the whole "deactivated user" feature here, but having an idea of how that'll be built out would help the answer above question. If we're not sure, we can just leave it as is for now and update it later.

seed/views/v3/users.py
```diff
@@ -0,0 +1,883 @@
...
+        user = SEEDUser.objects.filter(
+            id=user_id
+        )
+        if not user.exists():
+            return JsonResponse({
+                'status': 'error',
+                'message': 'user does not exist',
+            }, status=status.HTTP_403_FORBIDDEN)
+
+        user[0].deactivate_user()
```
adrian-lara, line 882: You could also use the try-except pattern with a `SEEDUser.objects.get()` vs the `.filter()`. I'd imagine we'd account for 2 exceptions: 
- [DoesNotExist](https://docs.djangoproject.com/en/2.2/ref/models/instances/#doesnotexist)
- everything else (`except Exception as e:`)

In each case, returning a 400-level response - like you have already for when the user does not exist and like [this](https://github.com/SEED-platform/seed/blob/develop/seed/views/column_mapping_presets.py#L44) for everything else.

seed/views/v3/users.py
```diff
@@ -0,0 +1,887 @@
...
+                        'first_name': 'string',
+                        'last_name': 'string'
+                    })
+
+            ]
+
+        }
+
+
+class UserViewSet(viewsets.ViewSet):
```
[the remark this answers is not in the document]
aviveiros11: @adrian-lara I do not think so, with the base ViewSet I think the @action decorators are used in order to specify the request type 

I changed
`@action(detail=True, methods=['DELETE'])`
to 
`@action(detail=True, methods=['PUT'])`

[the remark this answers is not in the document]
adrian-lara: I was referring to the `destroy` method: https://www.django-rest-framework.org/api-guide/viewsets/#viewset-actions

It doesn't seem to be present when inspecting the class via `dir(UserViewSet)`, so I think we're good 👍

seed/views/v3/users.py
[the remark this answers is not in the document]
adrian-lara: I think I misunderstood the state of this feature - It looks like a deactivated user can't log in anymore, and the base User class is based on `AbstractBaseUser, PermissionsMixin`. I'm not sure what we "get automatically" from that, but sounds like we weren't envisioning more being done for this "feature".

All that to say, this is good as is IMO. 👍

seed/views/v3/users.py
```diff
@@ -0,0 +1,886 @@
...
+
+        self.manual_fields = {
+            ('GET', 'list'): [],
+            ('POST', 'create'): [
+                self.org_id_field(),
+                self.body_field(
+                    name='New User Fields',
+                    required=True,
+                    description="An object containing meta data for a new user: "
+                                "required - [last_name, role(viewer/owner/member), email]",
```
adrian-lara, line 126: I think role is optional

seed/views/v3/users.py
```diff
@@ -0,0 +1,886 @@
...
+
+        if is_last_owner:
+            return JsonResponse({
+                'status': 'error',
+                'message': 'an organization must have at least one owner level member'
+            }, status=status.HTTP_409_CONFLICT)
+
+        OrganizationUser.objects.filter(
+            user_id=user_id,
+            organization_id=body['organization_id']
```
adrian-lara, line 435: I get this error. Should this use the `organization_id` in L410 instead?
```
File "[...]/seed/views/v3/users.py", line 435, in role
    organization_id=body['organization_id']
KeyError: 'organization_id'
```

seed/views/v3/users.py
aviveiros11: Screen Shot 2020-05-19 at 7 49 00 AM
I was just going off of the comments within the action.  If it is documented here as required, should we keep it that way?

aviveiros11:
@adrian-lara - it turns out for the UI to render the body field for "role" needs to say "string".  My solution to keep the param optional was to case switch on "string" and if unchanged default the user to be a viewer following least privilege.  Let me know if this works.

seed/views/v3/users.py
adrian-lara: Per discussions - this functionality hasn't been built out yet and copying over is enough for now.

seed/views/v3/users.py
```diff
@@ -300,6 +301,8 @@ def create(self, request):
             role = body['role']
             if isinstance(role, dict):
                 role = role['value']
+            elif role == 'string':
```
adrian-lara, line 304: I could go one of two ways here:
1. Add a default for any value that is received here that we don't recognize. Maybe within the `_get_role_from_js` method?
2.  Improve this check to throw a 400 error that lists the valid role string values. Also, maybe within the `_get_role_from_js` method.

seed/views/v3/users.py
```diff
@@ -123,7 +123,8 @@ def __init__(self, *args):
                     name='New User Fields',
                     required=True,
                     description="An object containing meta data for a new user: "
-                                "required - [last_name, role(viewer/owner/member), email]",
+                                "- Required - first_name, last_name, email "
+                                "- Optional - role viewer(default), member, owner",
```
adrian-lara, line 127: Let's make this more clear that the field is `role`, and we're listing the valid role values next to it. Just a suggestion, so feel free to put whatever makes sense to you here.
```suggestion
                                "- Optional - role ['viewer'(default), 'member', or 'owner']",
```

seed/views/v3/users.py
```diff
@@ -139,7 +140,7 @@ def __init__(self, *args):
                     name='Updated User Fields',
                     required=True,
                     description="An object containing meta data for a updated user: "
-                                "required - [first_name, last_name, email]",
+                                "- Required - first_name, last_name, email",
```
adrian-lara: see "\n" comment

seed/views/v3/users.py
```diff
@@ -123,7 +123,8 @@ def __init__(self, *args):
                     name='New User Fields',
                     required=True,
                     description="An object containing meta data for a new user: "
-                                "required - [last_name, role(viewer/owner/member), email]",
+                                "- Required - first_name, last_name, email "
```
adrian-lara: I prefer the "\n" explicitly added in. Any reason you took it out?

aviveiros11: just having so many issues with Travis today I thought it was causing my build to fail.  I will change it back.

seed/views/v3/users.py
aviveiros11: I think option 2 just mimics the functionality of a required field because we would throw a 400 error every time a user doesn't explicitly define viewer, member, or owner.

adrian-lara: I was thinking it'd slightly be different in that it'd still be optional and they can exclude it from the request. But if an API user does provide a role, we do verify it's validity.

seed/views/v3/users.py
adrian-lara: Works as expected now.

seed/views/v3/users.py
adrian-lara: This is basically good to go, but think you can add the `\n` in this description? as well as the ones on L142? Sorry 🙏

aviveiros11: Ah, no that is my bad I thought I added them back.

Issue #2024, opened 2019-12-02
Mapping: allow user to set up a mapping profile for a BldgSync file


RDmitchell:
### Expected Behavior
It would be nice to be able to map all the fields in a BuildingSync file 
### Actual Behavior 
Right now, when importing a BuildingSync file, the program doesn't go through the mapping steps.

If the user was able to set up a mapping profile for a buildingsync file, that would solve this problem. 
### Steps to Reproduce
see issue #1740 -- the comments at the end reflect the fact that not all the fields are currently being mapped into the SEED db
### Instance Information
Instance: dev1
SHA: 6dad9f7

macintoshpie:
Hi @RDmitchell ,

We have added the ability for BuildingSync files to use column mapping presets, like other file types, when importing and exporting. This will allow users to change what SEED header an xpath is mapped to or from. This does not however allow users to write their own xpaths - we decided this would be a bit too unruly. Below are some notes on how the presets work and how you might want to test them.

All organizations should now have a column mapping preset called 'BuildingSync v2.0 Defaults'. This preset cannot be edited, but you can save copies of it which are editable. These copies allow the user to specify the SEED header and units (if it's a column recognized to have units). Unlike the already existing column mapping presets, the columns can only be mapped to Properties (not Tax Lots). Lastly, you cannot remove rows from any BuildingSync presets.

With this feature comes the requirement that users specify a preset when exporting a property's BuildingSync file. Thus when exporting a property's BuildingSync file, the user should be prompted to select which preset to use.

Thanks!

RDmitchell:
instance; dev1
SHA: 8a303af2
Org: LBNL Test 200
Data Set: BuildingSyn V2 Import
buildingsync_V2_ex01_measures.xml
Location: see the file in this folder
https://drive.google.com/drive/folders/1vVBYnKaz0p6YQJw64wJPmatICcj0WM3E?usp=sharing

I tried importing what I thought would be a file that would import, which I believe came from the repo (and which I previously imported and it worked)
* buildingsync_V2_ex01_measures.xml

But I got a 403 error
![image](https://user-images.githubusercontent.com/6314950/85475857-b3269600-b56b-11ea-84e8-9b28261b5b57.png)

macintoshpie:
Good catch @RDmitchell , thanks! Once the PRs above get merged the issue should be fixed.

RDmitchell:
[...]

PR #2213, opened 2020-05-11
Feat/bsync mapping presets

macintoshpie:
Another step in aligning BuildingSync imports with other file types
#### What's this PR do?
Allows column mapping presets to be used for buildingsync imports
- creates a default column mapping preset for all organizations
- allows users to create custom presets off of the default one
- allows users to use these presets on the mapping page and when exporting buildingsync files

NOTE: this PR does NOT allow users to select a preset when doing an "Update with BuildingSync"

##### BuildingSync Mapping Page Basic tests
Setup: import a buildingsync file and go to the mapping page
- when no preset applied or if one has been edited without saving, then “Map your data” button should be disabled
- preset options should only show bsync presets (ie no Portfolio Manager Default preset)
- preset options should include Default Bsync preset
- selecting/editing the default buildingsync preset disables the “save” button, but the “pencil” button (create a new preset) is still enabled
##### Default BuildingSync Preset test
Setup: import bsync file and go to mapping page
- select and apply default preset. Should populate all fields. Follow through with mapping and it should work normally
##### Making custom preset from default on mapping page test
Setup: import bsync file, go to mapping page
- select and apply the default preset (all fields should be populated and “map your data” button should be enabled)
- edit one or more of the fields
- click the pencil button to create a new preset. Save it, “map your data” button should become enabled
- wrap up the mapping process and you should see the change in your preset in the result
##### Making custom preset with no preset selected on mapping page test
Setup: import bsync file go to mapping page
- Fill in the fields without selecting and applying a preset. Set all inventory type to Property
- save the preset, “map your data” button should become enabled
- finish the mapping flow and everything should work as expected
##### Making custom preset with custom preset selected on mapping page test
Setup: import bsync file go to mapping page
- select and apply an existing custom bsync preset (ie not the default one)
- edit the preset and save it
- finish the mapping flow and everything should work as expected
##### Updating custom preset on mapping page test
Setup: import bsync file go to mapping page
- select and apply existing custom bsync preset (ie not the default one)
- edit a field (the “save” preset button should be enabled)
- click the save preset button and it should save it successfully
- finish the mapping flow
##### Making custom bsync preset from Orgs page then using on mapping page test
- navigate  to the column mapping page
- make a new buildingsync mapping
- import a bsync file
- should be able to see and use that custom mapping
##### PM (and other file types) Mapping Page Basic Tests
Setup: import a non-buildingsync file and go to the mapping page (make sure a custom buildingsync mapping is created already, see above)
- you should not see any buildingsync mappings (default or custom)
- “map your data” button should NOT be dependent on presets (ie if you have a valid mapping defined without using a preset it should let you continue)
##### PM (and other file types) making presets test
Setup: import non-buildingsync file and go to mapping page
- you should be able to successfully create a new preset when one isn’t applied (verify on orgs page)
- you should be able to successfully apply a preset, edit it, and save as a new preset (verify on orgs page)
- you should be able to successfully apply a preset, edit it, and update it by clicking the save button (verify on orgs page)
- mapping flow should work like normal

##### Default bsync mapping test
- should see default bsync preset
- should not be able to do the following to the default bsync preset
  - delete the preset
  - edit to or from column names
  - delete rows
  - add new row
##### Custom bsync mapping test
- should be able to create a new custom preset from the default one
- should NOT be able to do the following to the preset
  - edit the from fields
  - add new row
- should be able to do the following to the preset
  - delete the preset
  - delete rows
  - edit the to fields
##### Non-bsync mapping tests
- should be able to do the following (pretty much anything)
  - delete preset
  - delete row
  - edit to and from fields
  - add new row

##### Export with default preset
Setup: import a buildingsync file through the full flow
- go to the property details page
- click "export buildingsync", should see a modal prompting you to select a preset
- you should only see buildingsync presets (ie no PM presets)
- choose the default preset and export
- the exported file should be NO DIFFERENT from your uploaded file

##### Export with custom preset
Setup: create a custom buildingsync preset that changes around a SEED mapped property (e.g. switch latitude and longitude) and import a buildingsync file through the full flow USING THAT CUSTOM PRESET
- go to the property details page, you should see that your custom mapping was used (e.g. if swapped lat lon then they should be swapped here)
- click "export buildingsync", should see a modal prompting you to select a preset
- you should only see buildingsync presets (ie no PM presets)
- choose your custom preset and export
- the exported file should be NO DIFFERENT from your original file, since using the same custom mapping maps those value back to their original location in the xml file (e.g. lat and lon would be the same as the were in the uploaded file)

#2024


seed/static/seed/js/controllers/mapping_controller.js
```diff
@@ -219,10 +270,13 @@ angular.module('BE.seed.controller.mapping', [])
...
-      }, {
-        name: 'Tax Lot',
-        value: 'TaxLotState'
       }];
+      if (!$scope.mappingBuildingSync) {
+        $scope.setAllFieldsOptions.push({
+          name: 'Tax Lot',
+          value: 'TaxLotState'
+        })
+      }
```
macintoshpie, line 279: Rather than disabling the select, we manage the select options here.
This is necessary because a user might be mapping a bsync file, and NOT be working off of an existing preset, and needs to set the inventory type

seed/static/seed/js/controllers/mapping_controller.js
```diff
@@ -332,22 +386,13 @@ angular.module('BE.seed.controller.mapping', [])
...
-        if ($scope.import_file.source_type === "BuildingSync Raw") {
-          const suggestion = $scope.suggested_mappings[name];
-
-          return {
-            name: name,
-            suggestion_column_name: suggestion[1],
-            suggestion_table_name: suggestion[0],
-            raw_data: _.map(first_five_rows_payload.first_five_rows, name)
-          };
-        }
```
macintoshpie, line 344: I don't _think_ this is required anymore...

seed/static/seed/js/seed.js
```diff
@@ -529,11 +530,21 @@ SEED_app.config(['stateHelperProvider', '$urlRouterProvider', '$locationProvider
         templateUrl: static_url + 'seed/partials/mapping.html',
         controller: 'mapping_controller',
         resolve: {
-          column_mapping_presets_payload: ['column_mappings_service', 'user_service', function (column_mappings_service, user_service) {
+          column_mapping_presets_payload: ['column_mappings_service', 'user_service', 'seedConstants', 'import_file_payload', function (column_mappings_service, user_service, seedConstants, import_file_payload) {
```
macintoshpie, line 533: we're only fetching preset types relevant to our import file source type

seed/static/seed/partials/mapping.html
```diff
@@ -150,13 +150,13 @@
...
+                                <input type="text" uib-typeahead="column.display_name for column in (col.suggestion_table_name === 'PropertyState' ? mappable_property_columns : mappable_taxlot_columns) | filter:{display_name:$viewValue} | limitTo:20" typeahead-template-url="typeahead.html" ng-model="col.suggestion" class="form-control input-sm tcm_field" ng-disabled="import_file.matching_done" ng-change="change(col)" typeahead-on-select='change(col)' ng-attr-id="mapped-row-input-box-{$ $index $}" autocomplete="off" autocorrect="off" autocapitalize="off" spellcheck="false">
```
macintoshpie, line 159: users can now edit the seed headers

seed/views/column_mapping_presets.py
```diff
@@ -109,13 +138,26 @@ def create(self, request, pk=None):
         """
```
adrian-lara, line 138: This might be a frontend issue instead of a create issue, but I can't edit the SEED headers when I make a copy of the default BS preset on the settings page.

_**Note: I can make edits to BS presets that are made from the import mapping page though.**_

Screen Shot 2020-05-13 at 2 09 31 PM

seed/static/seed/js/controllers/column_mappings_controller.js
```diff
@@ -317,4 +319,29 @@ angular.module('BE.seed.controller.column_mappings', [])
...
+        }
+
+        if ($scope.current_preset.preset_type === seedConstants.PRESET_TYPE_BUILDINGSYNC_CUSTOM) {
+          const allowed_actions = [
+            'update',
+            'rename',
+            'delete',
+            'change_to_field',
+            'change_from_units',
+            'remove_column',
```
adrian-lara, line 339: I might be misremembering our conversation, but I thought we wanted to disallow the removal of columns for non-default BS presets?

Screen Shot 2020-05-13 at 2 21 33 PM

seed/static/seed/partials/column_mappings.html
```diff
@@ -31,13 +31,13 @@ <h2><i class="fa fa-sitemap"></i> <span translate>Column Mappings</span></h2>
           <span>{$:: 'Column Mapping Preset' | translate $}: </span>
```
adrian-lara, line 31: Not exactly sure where, but changes would likely involve this file...
We should disable the units choice too (for Gross Floor Area right now, but the fix should be for any "unit-aware" field).

Screen Shot 2020-05-13 at 2 26 24 PM

seed/api/v2_1/views.py
```diff
@@ -169,6 +170,23 @@ def building_sync(self, request, pk):
               required: true
               paramType: query
         """
```
adrian-lara, line 172: Not sure where any potential code changes would end up, but relevant to this endpoint...

Did some monkey-testing with exporting and saw some weird stuff happening. See the image below (left: exported xml, top-right: preset chosen during export and value of **Address Line 2**, bottom-right: preset setting with xpath for **Address Line 2**):
Screen Shot 2020-05-13 at 2 40 11 PM

I would've expected `StreetAddress` xml node to be empty. But not only was it populated, it was populated with a previous Address Line **1** value (instead of an Address Line 2 value).

seed/static/seed/js/seed.js
```diff
@@ -1420,3 +1431,9 @@ SEED_app.constant('naturalSort', function (a, b) {
     else if (oFxNcL > oFyNcL) return 1;
   }
 });
+
+SEED_app.constant('seedConstants', {
```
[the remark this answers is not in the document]
adrian-lara: @axelstudios Can you chime in on this pattern?

seed/api/v2_1/views.py
adrian-lara: Per our conversation, this is expected. The process for building the BuildingSync XML export takes the previous import (with all it's values) and updates the values of that file using presets and the latest -State. 

@macintoshpie, thinking about this more, this pattern (of starting with the original xml upload and building off that) is great for keeping nodes/values that we never even looked at during import. But should "empty" or None values replace the original values for nodes that we did read from on import? In this case, would it be appropriate to make the StreetAddress node empty? 

Definitely just thinking out loud - I'm not sure if this makes sense in the larger BuildingSync context.

seed/lib/xml_mapping/mapper.py
```diff
@@ -23,3 +23,23 @@ def build_column_mapping(base_mapping=None, custom_mapping=None):
         xpath: ('PropertyState', db_column, 100)
         for xpath, db_column in column_mapping.items()
     }
+
+
+def default_buildingsync_preset_mappings():
```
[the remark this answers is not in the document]
macintoshpie: thoughts @nllong ?

seed/static/seed/js/seed.js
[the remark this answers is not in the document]
axelstudios: I'm good with this pattern (a constant that can only be accessed via dependency injection).  There are better patterns for this with TypeScript and Angular 2+, but it's way better than the global namespace pollution going on [here](https://github.com/SEED-platform/seed/blob/edd89fa9b126231e5fb50622b31c7b275ee8646d/seed/templates/seed/_js_urls.html#L5).

Technically it might be unsafe because I think you could override the original value, the way around that would be to make them functions that return a string.  You could also put these in a service.

seed/api/v2_1/views.py
macintoshpie: From an implementation standpoint that might complicate things. E.g. if we reomve auc:StreetAddress, should we prune back the xml tree and how far?

I think @nllong would have a better idea as to whether it would make sense to remove elements from the tree

seed/views/column_mapping_presets.py
macintoshpie: great catch 👍

seed/static/seed/js/controllers/mapping_controller.js
```diff
@@ -76,6 +78,16 @@ angular.module('BE.seed.controller.mapping', [])
         $scope.mappings_change_possible = true;
       };
 
+      $scope.is_buildingsync_and_preset_not_ok = function() {
```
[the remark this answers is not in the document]
macintoshpie: I like it, added

seed/static/seed/js/controllers/column_mappings_controller.js
macintoshpie: Right, I will remove this option

seed/static/seed/js/seed.js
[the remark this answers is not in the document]
macintoshpie: Refactored this pattern.
Constants are now in static/seed/js/constants.js, registered as a module, and are now injectable by constant identifier.

I opted for single value constants rather than a single injectable constant object in order to prevent accessing undefined values due to typos. I tested it and you CAN reassign values to constants, but hopefully future developers will recognize and respect the UPPERCASE_CONVENTION of constants.

I considered implementing a constants service, but decided against it b/c
1. this seems to be the prescribed way of using constants in angularjs
2. code completion is nice
3. slightly less code (e.g. MY_CONST vs seed_constants.get('MY_CONST')

seed/api/v2_1/views.py
adrian-lara: Sounds like BuildingSync doesn't have the notion of a node that has an empty value?

macintoshpie: Created issue #2224 to track this

seed/lib/xml_mapping/mapper.py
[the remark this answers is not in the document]
macintoshpie: using `ft**2`

seed/static/seed/js/controllers/mapping_controller.js
adrian-lara: I think so too.

seed/static/seed/partials/mapping.html
```diff
@@ -83,23 +83,23 @@
...
             <div class="left section_action_container">
-                <button type="button" class="btn btn-primary" ng-hide="import_file.matching_done && isValidCycle" ng-disabled="check_reset_mappings() || import_file.source_type == 'BuildingSync Raw'" ng-click="reset_mappings()" translate>Copy Data File Headers directly to SEED Headers</button>
+                <button type="button" class="btn btn-primary" ng-hide="import_file.matching_done && isValidCycle" ng-disabled="check_reset_mappings() || mappingBuildingSync" ng-click="reset_mappings()" translate>Copy Data File Headers directly to SEED Headers</button>
             </div>
             <div class="right section_action_container section_action_btn">
-                <button type="button" class="pull-right btn btn-primary mapping-button" ng-disabled="check_fields()" ng-click="remap_buildings()" ng-hide="import_file.matching_done" translate>Map Your Data</button>
+                <button type="button" class="pull-right btn btn-primary mapping-button" ng-disabled="check_fields() || is_buildingsync_and_preset_not_ok()" ng-click="remap_buildings()" ng-hide="import_file.matching_done" translate>Map Your Data</button>
```
adrian-lara, line 89: Do the same for the bottom button:
Screen Shot 2020-05-18 at 2 26 43 AM

seed/static/seed/js/controllers/mapping_controller.js
adrian-lara: Should we do something similar for the admin page? As of now, the BE doesn't actually let the inventory type be changed, but the FE displays both as options.

seed/static/seed/js/constants.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * :copyright (c) 2014 - 2020, The Regents of the University of California, through Lawrence Berkeley National Laboratory (subject to receipt of any required approvals from the U.S. Department of Energy) and contributors. All rights reserved.
+ * :author
+ */
+angular.module('BE.seed.constants', [])
```
adrian-lara, line 5: 👍

seed/static/seed/js/controllers/inventory_detail_controller.js
```diff
@@ -500,16 +500,32 @@ angular.module('BE.seed.controller.inventory_detail', [])
       };
 
       $scope.export_building_sync = function () {
```
adrian-lara, line 502: Not necessarily a code change for this spot, but...
Should we disable BS exports for non-BS imports? When there is not a "building_file" associated to a -State, the format isn't guaranteed to be right anyway... I think? 

Thinking about this more, maybe this doesn't need to be addressed now, but let me know what you think.

seed/static/seed/js/controllers/inventory_detail_controller.js
```diff
@@ -500,16 +500,32 @@ angular.module('BE.seed.controller.inventory_detail', [])
       };
```
adrian-lara, line 501: General comment,
Here's the diff I got from importing and exporting the same file - I don't know enough about the BS schema to know if this is expected, so feel free to "resolve" this comment if you think it looks good:
Screen Shot 2020-05-16 at 9 18 02 PM

seed/views/column_mapping_presets.py
```diff
@@ -67,24 +70,50 @@ def update(self, request, pk=None):
...
+
+        # update the name
+        if updated_name is not None:
+            preset.name = updated_name
+
+        # update the mappings according to the preset type
+        if updated_mappings is not None:
+            if preset.preset_type == ColumnMappingPreset.BUILDINGSYNC_CUSTOM:
+                # only allow these updates to the mappings
+                # - changing the to_field
```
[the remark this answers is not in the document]
macintoshpie: Since we only allow creating Properties from Bsync files, I'd say no to `to_table_name`.
I'm not sure if we want to allow users to change the `from_units`. Thoughts @nllong ?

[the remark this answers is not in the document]
adrian-lara: Makes sense. If we do end up not allowing these changes, I'd vote to disable changing this options in the FE for any BS preset (default or custom).

seed/static/seed/js/controllers/inventory_detail_controller.js
macintoshpie: It is able to export a file that should be valid, it's just likely to be veeery minimal. I'm sure if we should disable it or not

adrian-lara: Sorry - should have specified what I meant by "right" in my first comment. Would the export be a valid BuildingSync file or just valid xml if the -State didn't have a `building_file`?

macintoshpie: It should be a valid BuildingSync file, it's just going to have very little data probably

seed/views/column_mapping_presets.py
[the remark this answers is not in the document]
macintoshpie: Decided that users must be able to change the from_units for custom presets (still not allowed for default bsync presets). This is necessary because users might want to change a mapping to a SEED column that uses different units (maybe strange case but possible).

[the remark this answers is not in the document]
macintoshpie: Added this in 917384e

Test by creating a custom preset, changing the units, saving it, then REFRESHING the page and viewing the preset again (force it to reload the presets from the backend). The units should have been updated.

Issue #2202, opened 2020-05-08
Organization's DQ Rules can be inadvertently deleted


adrian-lara:
### Expected Behavior
When saving a DQ Rule as Valid Data but not applying a Label, an error is shown, and nothing is saved.

### Actual Behavior
The error-causing DQ Rule and all DQ rules listed after it will be deleted.

If the first DQ rule is the reason the error is thrown, all DQ Rules are deleted. This leads to the DQ Rules being completely reset to the default set.

### Steps to Reproduce
1. Try to save a Rule as Valid Data and don't apply a Label.
<img width="1144" alt="Screen Shot 2020-05-08 at 12 03 42 AM" src="https://user-images.githubusercontent.com/30608004/81376245-5c9af080-90c0-11ea-8255-6365a6431f1e.png">
2. See error as expected:
<img width="1144" alt="Screen Shot 2020-05-08 at 12 04 13 AM" src="https://user-images.githubusercontent.com/30608004/81376282-6cb2d000-90c0-11ea-9a9e-8fc96eebc184.png">
3. Refresh page - all rules are reset.
<img width="1144" alt="Screen Shot 2020-05-08 at 12 04 32 AM" src="https://user-images.githubusercontent.com/30608004/81376617-29a52c80-90c1-11ea-915b-5acd8d77d93e.png">

### Instance Information
`develop` `2.7.0` and `2.7.1`

adrian-lara:
@nllong , @axelstudios , @lainsworth8801 Here are my thoughts on this DQ bug. 

**TL;DR**: Long term, no major extra steps needed as this would've been fixed anyway. Short-term, we should prioritize a patch to desired releases with a fix.

### Long term:
Ideally, the long term solution would be to avoid destroying and recreating the Rules whenever "Save Changes" is clicked or, more specifically, whenever the `save_data_quality_rules` endpoint is hit. According to #2179 , we'll be deprecating this single endpoint in favor of individual POST endpoints for a single Rule. This individual POST will be non-destructive and would naturally inherit the "checks" performed by the old `save_data_quality_rules` endpoint.

As discussed separately, the frontend could also be updated to disable the "Save Changes" button when a "check" would eventually fail. This particular interaction needs further design before implementation.

This update would be included in future releases. Previous releases wouldn't need to be patched.

### Short term:
[...]

PR #2218, opened 2020-05-13
2.7.0 patch1 - for review only

adrian-lara:
DQ Rules were being inadvertently deleted.

#### What's this PR do?
In the save_data_quality_rules POST endpoint, move any "checks" that trigger a 400 to before the records are deleted. Also, for that same endpoint, update the try-except logic to "...except and continue rebuilding Rules", returning a JSON response at the end that includes a message that Rules were lost/deleted.

#2202 

The error message still appears as before.
Screen Shot 2020-05-13 at 2 45 43 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# SEED Version 2.7.0-patch1
+
+- Fixed [#2202]( https://github.com/SEED-platform/seed/issues/2202 ), Organization's DQ Rules can be inadvertently deleted
+
 # SEED Version 2.7.0
 
 SEED Version 2.7.0 includes several significant updates that need to be thoroughly tested on production data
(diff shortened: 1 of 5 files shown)


adrian-lara:
For some reason, the Travis build isn't showing as having run and passed.
https://travis-ci.org/github/SEED-platform/seed/builds/686696944

lainsworth8801:
Works!
1. Add a valid data without giving it a label and save changes. Error is reported as expected:
Screen Shot 2020-05-26 at 2 54 10 PM

2. refresh page. Initial rules persist:
Screen Shot 2020-05-26 at 2 54 21 PM

adrian-lara:
@nllong @axelstudios Given you all don't manage any deployments using on 2.7.0, what's the appropriate next step for 2.7.0 now that this bug fix has an approval?

Side note: I'll be cherry-picking this into `develop` and likely `2.7.1`.

RDmitchell:
@adrian-lara / @nllong / @axelstudios -- actually Nick and Alex manage the deployments on the LBNL instances. So we need to figure out what the process will be to update them. I suggest starting with staging, and have users test there, then move to production after users report in

adrian-lara:
> @adrian-lara / @nllong / @axelstudios -- actually Nick and Alex manage the deployments on the LBNL instances. So we need to figure out what the process will be to update them. I suggest starting with staging, and have users test there, then move to production after users report in

@RDmitchell This particular bug doesn't exist on the LBNL instance as that is on 2.6.1(-patch3).

RDmitchell:
ah, right ...

adrian-lara:
@RDmitchell - apologies from my end - I forgot to update the tagged issue.

It previously mentioned prod/2.6.1 as "maybe" having the bug . I just updated the issue text to reflect that 2.6.1 doesn't have the bug. 👍

nllong:
tagged and released (https://github.com/SEED-platform/seed/releases/tag/v2.7.0-patch1)

Issue #2216, opened 2020-05-13
StopIteration Error 


nllong:
![image](https://user-images.githubusercontent.com/1907354/81827260-d2a2bb80-94f5-11ea-9392-cd41e3eb77f5.png)

This is most likely related to Python 3.7. (https://www.python.org/dev/peps/pep-0479/)

PR #2219, opened 2020-05-13
support for python 3.7

nllong:
Python 3.7 does not work with SEED 

#### What's this PR do?
* Fixes the yield commands to rescue on StopIteration

* I've updated travis to Python 3.7
* To test locally:
  * install a python 3.7 (I used 3.7.7) and pip install the dependencies
  * Try to import the sample data 
  * before the fix it would have stopped with the image that is in the issue #2216 

#2216

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ cache:
 os: linux
 language: python
 python:
-  - "3.6"
+  - "3.7"
 addons:
   chrome: stable
   postgresql: "11.2"
--- a/seed/management/commands/create_geojson_test_data.py
+++ b/seed/management/commands/create_geojson_test_data.py
@@ -30,7 +30,7 @@ class Command(BaseCommand):
         :return:
         """
         if geom['type'] == "Polygon":
-            coords = [f"{l[0]} {l[1]}" for l in geom['coordinates'][0]]
+            coords = [f"{coord[0]} {coord[1]}" for coord in geom['coordinates'][0]]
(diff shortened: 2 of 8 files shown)


seed/management/commands/create_geojson_test_data.py
```diff
@@ -30,7 +30,7 @@ def convert_list_to_wkt(self, geom):
         :return:
         """
         if geom['type'] == "Polygon":
-            coords = [f"{l[0]} {l[1]}" for l in geom['coordinates'][0]]
+            coords = [f"{coord[0]} {coord[1]}" for coord in geom['coordinates'][0]]
```
nllong, line 33: flake8 was complaining about ambiguous variables.

seed/models/column_list_settings_columns.py
```diff
@@ -22,4 +22,4 @@ class ColumnListSettingColumn(models.Model):
     pinned = models.BooleanField(default=False)
 
     def __str__(self):
-        return self.column_list_setting.name + " %s %s".format(self.order, self.pinned)
+        return f"{self.column_list_setting.name} {self.order} {self.pinned}"
```
nllong, line 25: flake8 fix

.travis.yml
```diff
@@ -9,7 +9,7 @@ cache:
 os: linux
 language: python
 python:
-  - "3.6"
+  - "3.7"
```
nllong, line 12: We can create this as a matrix (that is support both 3.6 and 3.7) but I think supporting 3.7 should suffice.

PR #2234, opened 2020-05-27
Using selection-tool for bsync validation

macintoshpie:
#### What's this PR do?
Adds an extra step to validate schema and use case when importing BuildingSync files
Also removes some vestigial front-end code for the old buildingsync import flow.

These tests should probably be run with and without celery workers, ie in local config include/exclude these lines
```bash
CELERY_TASK_ALWAYS_EAGER = True
CELERY_TASK_EAGER_PROPAGATES = True
```
When testing celery workers, it might be worth setting the log level to debug so messages can be seen: `celery ... -l debug`. Note that you WON'T see any messages when not running workers

Files for testing:
[use_case_data.zip](https://github.com/SEED-platform/seed/files/4704360/use_case_data.zip)

##### single valid file
- navigate to the data upload modal, click upload buildingsync
- select the file `buildingsync_v2_0_bricr_workflow.xml`
- if looking at the network tab of debug tools, it hits `/validate_use_cases`
- eventually it should succeed and you should be allowed to go to the mapping page and continue like normal

##### single valid file with warnings
- navigate to the data upload modal, click upload buildingsync
- select this file: `buildingsync_v2_0_bricr_workflow_WARNINGS.xml`
- eventually it should SUCCEED and you should be allowed to go to the mapping page and continue like normal. You should ALSO see the warnings listed for the file

##### single file bad use case
- first, note the number of imported files for the data set you're going to upload to
- navigate to the data upload modal, click upload buildingsync
- select the file `ex_1.xml`
- eventually it should say it failed to pass validation, and list use case errors and use case warnings, but no schema errors
- you should only be allowed to dismiss the dialog (not continue to mapping) and if you refresh the data page you should NOT see the number of files increase.

##### sing file bad schema
- first, note the number of imported files for the data set you're going to upload to
- navigate to the data upload modal, click upload buildingsync
- select the file `ex_1_bad_schema.xml`
- eventually it should say it failed to pass validation, and only show schema errors
- you should only be allowed to dismiss the dialog (not continue to mapping) and if you refresh the data page you should NOT see the number of files increase.

##### sing file that can't be parsed
- first, note the number of imported files for the data set you're going to upload to
- navigate to the data upload modal, click upload buildingsync
- select the file `ex_1_bad.xml`
- eventually it should say it error out saying the request was unsuccessful (the file isn't parsable)
- you should only be allowed to dismiss the dialog (not continue to mapping) and if you refresh the data page you should NOT see the number of files increase.

##### zip file all good
- navigate to the data upload modal, click upload buildingsync
- select the file `many_all_good.zip`
- eventually it should succeed and you should be allowed to go to the mapping page and continue like normal

##### zip file one bad
- first, note the number of imported files for the data set you're going to upload to
- navigate to the data upload modal, click upload buildingsync
- select the file `many_one_bad.zip`
- eventually it should say it failed to pass validation, and only show use case warnings and errors for the bad file
- you should only be allowed to dismiss the dialog (not continue to mapping) and if you refresh the data page you should NOT see the number of files increase.

##### random api issues
These are some things to test if the validation api is down or we get an unexpected response.
1. test timeout by editing the timeout in `building_sync/validation_client.py` where the request is being made. Changing it to something like `1` should be sufficient (make sure you restart celery!). When uploading a buildingsync file it should fail saying it timed out
1. again in `validation_client.py` test a bad request by tweaking the API url, the default schema version or use case name (again remember to restart celery). These should all result in some sort of failure, and should be reported when trying to upload a buildingsync file

Example failure
Screen Shot 2020-05-29 at 4 50 41 PM

Example success
Screen Shot 2020-05-29 at 5 11 12 PM

### DIFF
--- a/seed/static/seed/js/controllers/data_upload_modal_controller.js
+++ b/seed/static/seed/js/controllers/data_upload_modal_controller.js
@@ -21,7 +21,6 @@
  * ng-switch-when="11" == Confirm Save Mappings?
  * ng-switch-when="12" == Error Processing Data
  * ng-switch-when="13" == Portfolio Manager Import
- * ng-switch-when="14" == Successful upload! [BuildingSync]
  */
 angular.module('BE.seed.controller.data_upload_modal', [])
   .controller('data_upload_modal_controller', [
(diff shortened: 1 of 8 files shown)


seed/static/seed/js/controllers/data_upload_modal_controller.js
```diff
@@ -21,7 +21,6 @@
  * ng-switch-when="11" == Confirm Save Mappings?
  * ng-switch-when="12" == Error Processing Data
  * ng-switch-when="13" == Portfolio Manager Import
- * ng-switch-when="14" == Successful upload! [BuildingSync]
```
macintoshpie, line 24: I removed the front-end logic for the old BuildingSync import flow

Issue #2226, opened 2020-05-22
Losing unit-aware data on import after initial column mapping


adrian-lara:
### Expected/Actual Behavior
When mapping columns to unit-aware fields, the values are lost after clicking Map Your Data. As expected, the values don't make it to the eventually imported records. See steps to reproduce below.

**Examples are nonsensical (mapping ENERGY STAR Score to an area field), but help illustrate the bug:**
<img width="1189" alt="Screen Shot 2020-05-08 at 2 38 10 PM" src="https://user-images.githubusercontent.com/30608004/82707113-ab0bcb80-9c38-11ea-981c-b427369b0fc1.png">
<img width="1189" alt="Screen Shot 2020-05-08 at 2 38 29 PM" src="https://user-images.githubusercontent.com/30608004/82707121-ad6e2580-9c38-11ea-9371-3a130b2e97ec.png">

### Steps to Reproduce
Import a spreadsheet. Go back and forth between the two mapping pages - actual mapping columns page and mapping confirmation page. 

On the first time visiting the first column mapping page, set mappings to non-unit-aware columns and "Map Your Data". Go back to the first column mapping page. Change the column mappings to include unit-aware fields, and click "Map Your Data".

You should see that this the unit-aware columns has last data. This doesn't seem to happen to non-unit-aware columns.

### Instance Information
Server Instance: `dev1`, `prod`

adrian-lara:
Documenting research:
When "Map Your Data" gets clicked on the initial mapping page, `save_column_mappings` is triggered and gets Columns objects for both the "raw" import file header columns, the corresponding "mapped" SEED columns, and the MappingColumn objects used to associate them. A `get_or_create` query is used to avoid creating multiple Column objects and subsequently ColumnMapping objects. 

Unfortunately, by following the steps detailed above, those additional Column and ColumnMappings objects lead to an ambiguous situation during the `perform_mapping` step where the wrong ColumnMapping object is used to try to build the incoming -State records.

During this, the raw columns were considering an unnecessary factor to decide whether to "get" or "create" a column object - `units_pint`.

adrian-lara:
Proposed solution:
When creating a Column record to represent the raw import file header column, the only meaningful identifying factor that should be taken into account is the column_name value (along with associated org of course). The value of `units_pint` shouldn't even be considered (and should probably be `None`) when `get_or_create`'ing these raw column objects.

[...]

PR #2235, opened 2020-05-28
2226 lost data on import

adrian-lara:
Data could potentially be lost on import (see attached issue).

#### What's this PR do?
Fixes a bug where multiple Column objects were being created for the same raw import file header. 

This caused multiple ColumnMapping objects to be created when only one was expected in order to initially create incoming import records.

Import a file twice. The first time, **don't** map incoming columns to unit-aware SEED columns. The second time, map to unit-aware columns. See that the data persists throughout import.

Import a file once, but map twice. The first time mapping, **don't** map incoming columns to unit-aware SEED columns. The second time (by clicking "Back to Mapping"), map to unit-aware columns. See that the data persists throughout import.

#2226


adrian-lara:
Leaving a note here - we should discuss if we should patch this to prod.

seed/models/columns.py
```diff
@@ -982,22 +982,30 @@ def select_col_obj(column_name, table_name, organization_column):
...
+
+                ColumnMapping.objects.filter(column_raw__id__in=models.Subquery(all_from_cols.values('id'))).delete()
+                all_from_cols.delete()
+
+                from_org_col, _ = Column.objects.get_or_create(
+                    organization=organization,
+                    table_name__in=[None, ''],
+                    column_name=field['from_field'],
+                    is_extra_data=False  # data from header rows in the files are NEVER extra data
+                )
```
macintoshpie, line 1007: I think some comments in this area would be very helpful at a higher level. e.g. "Since multiple were found, we're deleting these for such and such reason"

seed/models/columns.py
```diff
@@ -982,22 +982,30 @@ def select_col_obj(column_name, table_name, organization_column):
...
+
+                ColumnMapping.objects.filter(column_raw__id__in=models.Subquery(all_from_cols.values('id'))).delete()
+                all_from_cols.delete()
+
+                from_org_col, _ = Column.objects.get_or_create(
+                    organization=organization,
+                    table_name__in=[None, ''],
+                    column_name=field['from_field'],
+                    is_extra_data=False  # data from header rows in the files are NEVER extra data
```
macintoshpie, line 1006: Is it true that data from header rows are never extra data? I thought I could make up an arbitrary column in my csv/excel file and map it as extra data

seed/models/columns.py
```diff
@@ -982,22 +982,30 @@ def select_col_obj(column_name, table_name, organization_column):
...
+                    organization=organization,
+                    table_name__in=[None, ''],
+                    column_name=field['from_field'],
+                    is_extra_data=False
+                )
+
+                ColumnMapping.objects.filter(column_raw__id__in=models.Subquery(all_from_cols.values('id'))).delete()
+                all_from_cols.delete()
+
+                from_org_col, _ = Column.objects.get_or_create(
```
macintoshpie, line 1002: For clarity's sake, should this just be `Column.objects.create` since we deleted above?

adrian-lara: Yup! Nice catch!

seed/models/columns.py
adrian-lara: The type of Column object this is referring to is a representation of the "raw" header row we use for the ColumnMapping model. The ColumnMapping model takes a "PropertyState" or "TaxLotState" Column object (see `table_name` attribute), and associates that to one of these "raw" Column objects to create the mapping.

The comment could be improved, so I'll do that for this line and the original line in the `try` block above here.

seed/models/columns.py
adrian-lara: Agree - I'll do that.

adrian-lara:
> I tested and it worked as expected, though I'm not totally clear on how this resolved the issue

Short version: The `units_pint` Column attribute triggered an unexpected "create" when `get_or_create`'ing. This impacted downstream processes involving ColumnMappings.
See my comment here for the slightly longer version: https://github.com/SEED-platform/seed/issues/2226#issuecomment-634935150

adrian-lara:
Per our conversation, I'll be building a test to add more confirmation that this is working.

Issue #2207, opened 2020-05-08
Data Quality admin page only reports the first error message when multiple errors are detected


lainsworth8801:
### Expected Behavior
Data quality admin page should alert users of all existing errors if multiple errors are detected.

### Actual Behavior
It only reports message of the first error detected.

### Steps to Reproduce
Select any rule, change severity to 'valid' without applying a label and hit save changes -> page prompts 'must apply label' error message;
Without correcting the first label error, create or change a rule that include/exclude an empty string, now hit save;
Error message still shows 'must apply label' only.

### Instance Information
<img width="1187" alt="Screen Shot 2020-05-08 at 11 56 45 AM" src="https://user-images.githubusercontent.com/49968203/81434251-1c208e80-9123-11ea-9411-6f214f04a5ba.png">

Link: #2202 

Multiple error displayed when "save changes" button is clicked once after all errors have been generated. "Save changed" in between errors without clearing them will only report the first one.

RDmitchell:
instance: dev1
SHA: 38fb067
Org: LBNL Test 200

Seems to be checking for multiple errors, even before saving
![image](https://user-images.githubusercontent.com/6314950/94735831-2df80d80-0320-11eb-9677-ad641f24014f.png)

Also on save, detected multiple errors
![image](https://user-images.githubusercontent.com/6314950/94736105-8fb87780-0320-11eb-95cf-895a6520bdfb.png)

PR #2236, opened 2020-05-28
2202 deleted dq rules

adrian-lara:
DQ Rules could inadvertently be deleted. See tagged issue for more details.

#### What's this PR do?
Cherry-pick changes from 2.7.0-patch1.
- Performs validity checks on incoming DQ Rule updates _before_ the teardown and rebuild of DQ Rules (a pattern that is slated to be removed).
- Skips over DQ Rules that could not be recreated instead of stopping the recreation process entirely.

Try to save a Rule as Must Contain or Must Not Contain while keeping the text field blank. See that nothing gets saved, and we get the same 400 we got from before. Refresh the page and see that nothing was lost either.

#2202

### DIFF
--- a/seed/views/data_quality.py
+++ b/seed/views/data_quality.py
@@ -8,6 +8,7 @@
 import csv
 
 from celery.utils.log import get_task_logger
+from django.db import transaction
 from django.http import JsonResponse, HttpResponse
 from rest_framework import viewsets, serializers, status
 from rest_framework.decorators import action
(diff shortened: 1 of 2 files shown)


seed/views/data_quality.py
```diff
@@ -405,29 +428,27 @@ def save_data_quality_rules(self, request, pk=None):
...
-                    'status': 'error',
-                    'message': e,
-                }, status=status.HTTP_400_BAD_REQUEST)
-
-        return self.data_quality_rules(request)
+            with transaction.atomic():
+                try:
+                    dq.add_rule(rule)
+                except Exception as e:
+                    error_messages.add('Rule could not be recreated: ' + str(e))
```
lainsworth8801, line 441: Verified:
Screen Shot 2020-06-01 at 5 19 50 PM

adrian-lara:
> On the UI front it still shows just the first error captured:
> Screen Shot 2020-06-02 at 8 38 00 PM
> 
> There might be some additional changes required in menu_controller.js to show all errors on the admin page.

I can get the multiple error messages, though it shows up in one line (vs the intended multiple lines). 
Screen Shot 2020-06-03 at 5 42 46 PM

There does seem to be an issue with throwing errors on subsequent "Save Changes". The text doesn't seem to update at all on the 2nd+ save attempt unless you refresh the page first. I'm not sure if the `$emit('app_error')` pattern was meant for this type of error handling. Either way, I think we'll eventually want to transition away from using that pattern and possibly use Notifications, but that effort isn't in scope here.

Given the primary goal of this PR is to temporarily address the inadvertently deleted DQ Rules, we snowballed a little when included handling multiple errors on the frontend, so I'd like to avoid snowballing this PR anymore. 

Let me know what you think.

lainsworth8801:
@adrian-lara I think it makes sense. Let’s at least keep #2207 open for tracking purpose?

lainsworth8801:
Yes I was able to see both errors with saving changes once. Like you said, consistently saving changes in between errors only display the first error. I will update #2207 with this info thanks!

PR #2241, opened 2020-06-03
2.7.1 patch1 - for review only

adrian-lara:
Cherry-picked from the 2.7.0-patch1 with 1 additional commit related to the Django version difference between the two versions.

The details below was copied from that branch's PR.

DQ Rules were being inadvertently deleted.

#### What's this PR do?
In the save_data_quality_rules POST endpoint, move any "checks" that trigger a 400 to before the records are deleted. Also, for that same endpoint, update the try-except logic to "...except and continue rebuilding Rules", returning a JSON response at the end that includes a message that Rules were lost/deleted.

#2202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# SEED Version 2.7.1-patch1
+
+- Fixed [#2202]( https://github.com/SEED-platform/seed/issues/2202 ), Organization's DQ Rules can be inadvertently deleted
+
 # SEED Version 2.7.1
 
 Date Range: 12/21/19 - 03/27/20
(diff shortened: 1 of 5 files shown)


lainsworth8801:
Works expected on both property and taxlot view:
Screen Shot 2020-06-10 at 8 23 00 PM
Screen Shot 2020-06-10 at 8 22 49 PM
Screen Shot 2020-06-10 at 8 22 35 PM
Screen Shot 2020-06-10 at 8 22 22 PM

nllong:
I assume that we need to release this, correct? @adrian-lara @axelstudios

axelstudios:
> I assume that we need to release this, correct? @adrian-lara @axelstudios

I'm good with releasing

adrian-lara:
> I assume that we need to release this, correct? @adrian-lara @axelstudios

Yes 🙏

nllong:
tagged and released (https://github.com/SEED-platform/seed/releases/tag/2.7.1-patch1)

Issue #2242, opened 2020-06-04
ComStock Support


axelstudios:
#### Feature request

- Add opt-in at the organizational level to enable ComStock functionality
- Add ability to map existing columns to ComStock fields
- Add protected list settings profile for ComStock fields

PR #2243, opened 2020-06-04
Comstock support

axelstudios:
Please review the code changes to views/column_list_settings.py, there may be a more elegant way to inject the ComStock list setting

#### What's this PR do?
- Adds an opt-in at the organization level to Enable ComStock Functionality, with ComStock column mappings and a protected list settings view
- Switched the UI from using the deprecated PATCH to PUT for /api/v2/columns/X/
- Updates AngularJS to v1.8.0

- `npm install` and `manage.py migrate` 
- Check the `Enable ComStock Functionality` checkbox on your organization settings page
- Go to your organization’s Column Settings page and make sure there’s a new column called ComStock Mapping
- Map some columns to ComStock columns.  Attempt to set two columns to the same ComStock column (duplicates are not allowed).  Save, and then switch to the other property/taxlot tab and map a column to one that you already mapped and then save (it should succeed and remove the mapping from the other table).
- When viewing your inventory check that the list settings profile “ComStock” exists
- Attempt to change/rename/delete the ComStock list setting, there should be a UI notification explaining that it’s protected.  This can also be done by attempting to drag columns on the inventory list view
- Test `GET /api/v2/column_list_settings/` and ensure that the ComStock list setting was injected at the end with the ID null.
- Test `GET /api/v2/column_list_settings/null/` and ensure the ComStock list setting is returned.
- Disable the ComStock functionality and make sure that the opt-in features are no longer available

#2242


axelstudios:
@adrian-lara I considered some way of making the grid readonly, but I don't think it would be easy using the actual grid options.  The only easy solution would be to have a conditional overlay on the grid to block click events and make it a bit more opaque, I'll see if that's low effort

seed/views/columns.py
```diff
@@ -149,6 +149,20 @@ def retrieve(self, request, pk=None):
...
+    def update(self, request, pk=None):
+        organization_id = request.query_params.get('organization_id', None)
+
+        request.data['shared_field_type'] = request.data['sharedFieldType']
+        del request.data['sharedFieldType']
+
+        # Ensure ComStock uniqueness across properties and taxlots together
+        if request.data['comstock_mapping'] is not None:
+            Column.objects.filter(organization_id=organization_id, comstock_mapping=request.data['comstock_mapping']) \
+                .update(comstock_mapping=None)
```
adrian-lara, line 163: Ah I see - I was looking for this dedup on the frontend.  My only hesitation with this approach is that we don't indicate what's happening on the frontend. 

Brainstorming some:
- What if we updated the frontend to expand the scope of `$scope.columns` to include both inventories? We'd be able to track and dedup the comstock_mapping value across the whole org there. We'd probably have to update the html partial to only display columns of a the given inventory type. Also, we'd have to update the confirm save modal to show inventory type since we could be on the property page but change a tax lot column.
- Would it be enough to modify the saved changes notifications to reflect that another column was changed? This approach would probably require us to edit the response coming out of this method. Initial feeling is that this wouldn't be a great approach.

seed/views/columns.py
```diff
@@ -149,6 +149,20 @@ def retrieve(self, request, pk=None):
             'column': ColumnSerializer(c).data
         })
 
+    @ajax_request_class
+    @has_perm_class('can_modify_data')
+    def update(self, request, pk=None):
+        organization_id = request.query_params.get('organization_id', None)
+
+        request.data['shared_field_type'] = request.data['sharedFieldType']
+        del request.data['sharedFieldType']
```
adrian-lara, line 158: This could definitely be done via serializer, but I'd think that we should solve the dedup problem first in case this method gets heavily refactored.

seed/views/column_list_settings.py
```diff
@@ -46,3 +55,92 @@ class ColumnListingViewSet(OrgValidateMixin, SEEDOrgCreateUpdateModelViewSet):
     pagination_class = None
     # force_parent = True  # Ideally the column list settings would inherit from the parent,
     # but not yet.
+
```
adrian-lara, line 58: General comments about these two methods and the customizations happening within them. 

My initial thought was to use a serializer, but before I even looked at if that was possible, I changed my mind as this injection of a ColumnListSetting shouldn't be tucked away into the serializer. In other words, it should be explicitly pointed out in these methods like you have them already.

seed/views/column_list_settings.py
```diff
@@ -46,3 +55,92 @@ class ColumnListingViewSet(OrgValidateMixin, SEEDOrgCreateUpdateModelViewSet):
...
+        for index, column in enumerate(comstock_columns):
+            result['data']['columns'].append({
+                "id": column.id,
+                "pinned": False,
+                "order": index + 1,
+                "column_name": column.column_name,
+                "table_name": column.table_name,
+                "comstock_mapping": column.comstock_mapping
+            })
```
adrian-lara, line 96: I'm wondering if we could abstract out the logic of building out this ComStock setting record. I was envisioning that it could be reusable, but I could be missing something about how this is built in the `list` endpoint. 

My initial thought is that we could delegate it to the the ColumnListSetting model? But maybe it's not the model's "responsiblity". So maybe, in a utils file? Or maybe even just abstract it into a "_"/private method in this class?

What are your thoughts on this?

seed/models/columns.py
```diff
@@ -594,6 +594,13 @@ class Column(models.Model):
 
     recognize_empty = models.BooleanField(default=False)
 
+    comstock_mapping = models.CharField(max_length=64, null=True, blank=True, default=None)
```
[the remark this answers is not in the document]
adrian-lara: As discussed, these are still being lockeddown on ComStock-end.

seed/views/columns.py
adrian-lara: Documented as needing to change in the future.

seed/views/column_list_settings.py
adrian-lara: After discussing, it was both easier and preferably more obvious to implement these injections explicitly.

seed/static/seed/js/controllers/column_settings_controller.js
```diff
@@ -67,9 +67,30 @@ angular.module('BE.seed.controller.column_settings', [])
...
+        {id: 'weekday_duration', label: $translate.instant('comstock.weekday_duration')},
+        {id: 'building_shape', label: $translate.instant('comstock.building_shape')},
+        {id: 'built_code', label: $translate.instant('comstock.built_code')},
+        {id: 'rotation', label: $translate.instant('comstock.rotation')},
+        {id: 'aspect_ratio', label: $translate.instant('comstock.aspect_ratio')},
+        {id: 'building_type', label: $translate.instant('comstock.building_type')},
+        {id: 'state', label: $translate.instant('comstock.state')},
+        {id: 'county', label: $translate.instant('comstock.county')},
+        {id: 'climate_zone', label: $translate.instant('comstock.climate_zone')}
+      ];
```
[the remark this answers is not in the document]
axelstudios: I attempted this, no dice since constants can't inject provider dependencies.  I used [this trick](https://stackoverflow.com/a/31434481) to inject $translate, but then it couldn't access the $translateProvider.

seed/views/columns.py
axelstudios, line 158: @nllong Why is it necessary to do this?  Seems like it comes from here:
https://github.com/SEED-platform/seed/blob/7c539d722609b296a676953f48a9f96633a936cd/seed/models/columns.py#L1247-L1248

seed/views/column_list_settings.py
axelstudios: Refactored into a single function 👍

seed/views/columns.py
axelstudios: The confirm modal now shows `(removed)` for comstock mappings that already existed on the other table 👍

seed/static/seed/js/seed.js
```diff
@@ -830,33 +830,25 @@ SEED_app.config(['stateHelperProvider', '$urlRouterProvider', '$locationProvider
...
-                });
-                columns.sort(function (a, b) {
-                  return naturalSort(a.displayName, b.displayName);
-                });
-                return columns;
-              });
+              return inventory_service.get_taxlot_columns_for_org(organization_id);
             }
           }],
+          columns: ['all_columns', 'naturalSort', function (all_columns, naturalSort) {
```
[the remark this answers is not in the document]
axelstudios: Yeah, this is something sneaky I figured out a long time ago that can make your resolves depend on other resolves 🤓

Issue #2175, opened 2020-04-12
API V3 - Import Files


adrian-lara:
Build out the following v2 endpoints into v3, update fields shown in Swagger, and also make changes as listed:
- [x] **/api/v2/import_files/{id}/ - delete**: move
- [x] **/api/v2/import_files/{id}/ - get**: move
- [x] **/api/v2/import_files/{id}/data_quality_progress/ - get**: deprecate
- [x] **/api/v2/import_files/{id}/filtered_mapping_results/ - post**: name /mapping_results/ (vs filtered_mapping...)
- [x] **/api/v2/import_files/{id}/first_five_rows/ - get**: move
- [x] **/api/v2/import_files/{id}/mapping_done/ - put**: move to post
- [x] **/api/v2/import_files/{id}/mapping_suggestions/ - get**: investigate whether it's used. maybe should be on columns
- [x] **/api/v2/import_files/{id}/matching_and_geocoding_results/ - get**: move
- [x] **/api/v2/import_files/{id}/perform_mapping/ - post**: name /map/ (vs perform_mapping)
- [x] **/api/v2/import_files/{id}/raw_column_names/ - get**: move
- [x] **/api/v2/import_files/{id}/save_column_mappings/ - post**: move to organizations/{id}/column_mappings
- [x] **/api/v2/import_files/{id}/save_raw_data/ - post**: name /start_save_data/
- [x] **/api/v2/import_files/{id}/start_data_quality_checks/ - post**: move
- [x] **/api/v2/import_files/{id}/start_system_matching_and_geocoding/ - post**: move
- [x] **/api/v2/import_files/{id}/validate_use_cases/ - post**: move

Consider splitting this into multiple PRs as code changes could be substantial and difficult to review and test if one PR covers all these bullets.

Additionally, review the following:
- [ ] use of DRF ViewSets.
- [ ] use of decorators is reasonable/
- [ ] comments include at least a description. Details about incoming or outgoing fields are nice-to-haves as well.

Manual Testing:
Visit the /app/#/api/swagger page and "Execute" the endpoints listed above.

macintoshpie:
Note the endpoint `/api/v2/import_files/{id}/data_quality_progress/` is no longer used.  it appears `/api/v2/progress/:1:SEED:check_data:PROG:5/` is the location that does this described work.

macintoshpie:
`/api/v2/import_files/{id}/mapping_suggestions` is still being used in the mapping_controller

[...]

PR #2244, opened 2020-06-04
(2175) Feat/api v3 import files

macintoshpie:
#### What's this PR do?
Bump up the `import_files` endpoint to v3, tweak some names, and make swagger docs. Also, removed a dead endpoint, `data_quality_progress`.

Test each of the endpoints with `/import_files/` prefix
- probably a good idea to import at least two files for testing (one for deleting and another for running the other endpoints on)

Also test `organizations/{id}/column_mappings` endpoint.
- Just fill the template data with something reasonable, then you can check in the shell with `ImportFile.objects.get(id=<your id here>).cached_mapped_columns`
#2175


seed/views/v3/import_files.py
```diff
@@ -540,34 +473,9 @@ def start_data_quality_checks(self, request, pk=None):
             'progress': return_value,
         })
 
-    @api_endpoint_class
-    @ajax_request_class
-    @action(detail=True, methods=['GET'])
-    def data_quality_progress(self, request, pk=None):
```
macintoshpie: This endpoint is no longer used

seed/views/v3/import_files.py
```diff
@@ -669,7 +538,14 @@ def mapping_done(self, request, pk=None):
...
 
-        import_file = ImportFile.objects.get(pk=import_file_id)
+        try:
+            import_file = ImportFile.objects.get(pk=import_file_id)
+        except ImportFile.DoesNotExist:
+            return JsonResponse({
+                'status': 'error',
+                'message': 'no import file with given id'
+            }, status=status.HTTP_404_NOT_FOUND)
+
```
macintoshpie, line 548: Wrapped in try except just because I noticed it wasn't already

seed/utils/api_schema.py
```diff
@@ -80,20 +86,27 @@ def path_id_field(self, description):
...
             openapi.IN_BODY,
             description=description,
             required=required,
-            schema=self._build_body_schema(params_to_formats)
+            schema=cls.schema_factory(params_to_formats)
         )
 
-    def _build_body_schema(self, params_to_formats):
+    @classmethod
+    def schema_factory(cls, params_to_formats):
```
adrian-lara, line 100: 👍 I like this name better!

seed/views/v3/organizations.py
```diff
@@ -15,10 +18,46 @@
...
+                'to_field': 'gross_floor_area',
+                'to_field_display_name': 'Gross Floor Area',
+                'to_table_name': 'PropertyState',
+            }
+        ]
+    }
+    """
+    mappings = serializers.ListField(child=MappingSerializer())
+
+
```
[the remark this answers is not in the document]
macintoshpie: I feel kind of indifferent about it. It's only use is for swagger documentation currently, so to me it feels a little weird to "legitimize" the serializer when it's not really used. If we refactor the endpoints to use it then I'd probably be for it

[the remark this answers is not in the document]
adrian-lara: I was thinking about how surprising it was to find ColumnMapping-specific Serializers in the Organization Views file, especially once the organization endpoints continue to be built out further. In a sense, I was kind of "distracted" by it when opening this file.

I do get what you mean though - their existence is completely dependent on it being needed for an organization Swagger response. 

What do you think about moving it and adding some comments as to the purpose of each class (and how they're only used for a Swagger-specific response)?

seed/views/v3/import_files.py
```diff
@@ -0,0 +1,808 @@
...
+                    'status': 'error',
+                    'message': 'Internal problem occurred, more than one merge record found'
+                },
+                status=status.HTTP_500_INTERNAL_SERVER_ERROR
+            )
+
+        return audit_entry[0]
+
+    @swagger_auto_schema(
+        request_body=AutoSchemaHelper.schema_factory({
```
adrian-lara, line 370: Seeing it in _action_ - what do you think about renaming this to "body_schema_factory"? If you agree with the change, we could also just do it in the next PR, where the other v3 endpoints are built out.

seed/views/v3/organizations.py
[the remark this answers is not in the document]
macintoshpie: That works for me 👍

seed/views/v3/import_files.py
macintoshpie: I might prefer to keep it generic, as there's nothing relating this directly to the body. In #2247 I also use it to create _response_ schemas as well (rather than a serializer).

adrian-lara: Ah ok, didn't realize it could be used in that way. Agree - let's keep it as is. 👍

Issue #2250, opened 2020-06-09
API V3 - Properties


adrian-lara:
Build out the following v2 endpoints into v3, update fields shown in Swagger, and also make changes as listed:

### V2 (separate PR?)
- [x] **/api/v2/properties/ - get**: move
- [x] **/api/v2/properties/{id}/ - get**: move
- [x] **/api/v2/properties/{id}/ - put**: move
- [x] **/api/v2/properties/{id}/delete_measures/ - deprecate
- [x] **/api/v2/properties/{id}/delete/ - delete**: deprecate - can potentially leave "hanging" -State object
- [x] **/api/v2/properties/{id}/links/ - post**: Should be a get
- [x] **/api/v2/properties/{id}/match_merge_link/ - post**: move
- [x] **/api/v2/properties/{id}/measures/ - : deprecate
- [x] **/api/v2/properties/{id}/pair/ - put**: move
- [x] **/api/v2/properties/{id}/taxlots/ - get**: deprecate
- [x] **/api/v2/properties/{id}/unmerge/ - post**: change to put
- [x] **/api/v2/properties/{id}/unpair/ - put**: move
- [x] **/api/v2/properties/{id}/update_measures/ - put**: deprecate
- [x] **/api/v2/properties/batch_delete/ - delete**: move
- [x] **/api/v2/properties/columns/ - get**: move
- [x] **/api/v2/properties/cycles/ - post**: Rename to "filter_by_cycle"
- [x] **/api/v2/properties/filter/ - post**: move
- [x] **/api/v2/properties/mappable_columns/ - get**: move
- [x] **/api/v2/properties/merge/ - post**: move
- [x] **/api/v2/properties/meters_exist/ - post**: move

### V2.1 (separate PR?)
- [x] **/api/v2.1/properties/ - get**: needs to be investigated - Check the BRICR style workflow. I think this one matches the green building registry and uses more Django-y methods.
- [x] **/api/v2.1/properties/{id}/ - get**: deprecate
- [x] **/api/v2.1/properties/{id}/building_sync/ - get**: move
- [x] **/api/v2.1/properties/{id}/hpxml/ - get**: move
- [x] **/api/v2.1/properties/{id}/update_with_building_sync/ - put**: move
- [x] **/api/v2.1/properties/{properties_pk}/notes/ - get**: move (keep nested router pattern)
- [x] **/api/v2.1/properties/{properties_pk}/notes/ - post**: move (keep nested router pattern)
[...]

PR #2255, opened 2020-06-11
2250 refactor/api v3 properties

macintoshpie:
#### What's this PR do?
Moves the api v2 properties endpoints to v3. Includes some refactoring and swagger docs.
Test all `/properties/*` endpoints on the swagger page
#2250


seed/views/v3/properties.py
```diff
@@ -45,3 +208,769 @@ def labels(self, request):
...
+            return JsonResponse(
+                {'status': 'error', 'message': 'Need to pass organization_id as query parameter'},
+                status=status.HTTP_400_BAD_REQUEST)
+
+        response = properties_across_cycles(org_id, profile_id, cycle_ids)
+
+        return JsonResponse(response)
+
+    @swagger_auto_schema(
+        manual_parameters=[
```
aviveiros11, line 268: This `Add Item` option is slick

seed/views/v3/properties.py
```diff
@@ -45,3 +208,769 @@ def labels(self, request):
...
+            AutoSchemaHelper.query_org_id_field(),
+            AutoSchemaHelper.query_integer_field(
+                'cycle',
+                required=False,
+                description='The ID of the cycle to get properties'),
```
aviveiros11, line 273: This `Add Item` option is slick

seed/views/v3/properties.py
```diff
@@ -45,3 +208,769 @@ def labels(self, request):
...
+            del h['state_id']
+
+        return history, master
+
+    @swagger_auto_schema_org_query_param
+    @api_endpoint_class
+    @ajax_request_class
+    def retrieve(self, request, pk=None):
+        """
+        Get property details
```
aviveiros11, line 769: Should we let the user know to use the `property_view_id` here to avoid confusion with the `property_state_id`?

seed/views/v3/properties.py
```diff
@@ -45,3 +208,769 @@ def labels(self, request):
...
+        - get the property audit log for this property state
+        - if the new property state has extra_data, the original extra_data is update'd
+        - and then whoa stuff about the audit log?
+        - I'm going to assume 'Import Creation' is the key I'm looking for
+        - create a serializer for the new property state
+        - if it's valid, save this new serialized data to the db
+        - assign it to the original property view and save the property view
+        - create a new property audit log for this change
+        - return a 200 if created
+        """
```
aviveiros11, line 820: We should probably revise the roadmap above to exclude mental notes.  Specifically on lines 810, 813, 814.

seed/views/v3/properties.py
macintoshpie: That's a great question, @adrian-lara what are your thoughts on this? Looking at the detail views (those that use `id` in the path), they all seem to be using property view ids. Does the user of the api know/care about property states?

adrian-lara: From what I've seen, frontend users have no notion of inventory -Views or -States.

Are all the property endpoints using View ID? If so, I'd definitely feel OK using just `id` and **not clarifying** it's actually a View ID.

seed/views/v3/properties.py
```diff
@@ -45,3 +208,767 @@ def labels(self, request):
...
+    @has_perm_class('can_modify_data')
+    @action(detail=False, methods=['POST'])
+    def merge(self, request):
+        """
+        Merge multiple property records into a single new record, and run this
+        new record through a match and merge round within it's current Cycle.
+        """
+        body = request.data
+
+        state_ids = body.get('state_ids', [])
```
macintoshpie, line 351: @adrian-lara @aviveiros11 I missed this, it looks like we _do_ use property state ids in a body for this request. Maybe we should clarify that path ids are property view ids. Thoughts?

seed/views/v3/properties.py
```diff
@@ -45,3 +208,767 @@ def labels(self, request):
...
+    @ajax_request_class
+    @has_perm_class('requires_viewer')
+    @action(detail=False, methods=['GET'])
+    def filter_by_cycle(self, request):
+        """
+        List all the properties	with all columns
+        """
+        org_id = request.query_params.get('organization_id', None)
+        profile_id = request.query_params.get('profile_id', -1)
+        cycle_ids = request.query_params.get('cycle_ids', [])
```
macintoshpie, line 256: This is broken, drf-yasg gives query array as abc=1,2,3 while django expects abc=1&abc=2&abc=3. On my list to come back and fix this by just splitting on commas

macintoshpie: more context: https://github.com/axnsan12/drf-yasg/issues/317#issuecomment-486104837

seed/views/v3/properties.py
adrian-lara: I think we should convert these to view_ids. Check out https://github.com/SEED-platform/seed/pull/2257/commits/cb17d404b62964667702712699d70e02d4165f85

seed/views/v3/properties.py
macintoshpie: After a discussion, we decided to change the view http method to POST. Reasoning is that it's messy to split on commas (swagger method), we want to be able to do requests from swagger (django method prevents) and its overcomplicated to allow both.

seed/views/v3/properties.py
```diff
@@ -45,3 +208,771 @@ def labels(self, request):
...
+                required=False,
+                description='Determine whether or not to show only the used fields. Ones that have been mapped'
+            )
+        ]
+    )
+    @api_endpoint_class
+    @ajax_request_class
+    @has_perm_class('requires_viewer')
+    @action(detail=False, methods=['GET'])
+    def columns(self, request):
```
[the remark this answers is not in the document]
macintoshpie: That makes sense, I'll make that change

seed/views/v3/properties.py
```diff
@@ -45,3 +208,771 @@ def labels(self, request):
...
+        unitted_columns = [add_pint_unit_suffix(organization, x) for x in columns]
+
+        return JsonResponse({'status': 'success', 'columns': unitted_columns})
+
+    @swagger_auto_schema_org_query_param
+    @api_endpoint_class
+    @ajax_request_class
+    @has_perm_class('requires_viewer')
+    @action(detail=False, methods=['GET'])
+    def mappable_columns(self, request):
```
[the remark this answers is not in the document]
macintoshpie: I like it, will change it

seed/views/v3/properties.py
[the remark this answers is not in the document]
macintoshpie: After looking into it, all we needed to do was deprecate this view. I also added the `add_pint_unit_suffix` to the `list` view for columns. See 02dec60a4ee2b3a0423ee38485c6ca5596c80ff0

seed/views/v3/properties.py
```diff
@@ -2,28 +2,191 @@
...
+                        'total': 0
+                    },
+                    'cycle_id': None,
+                    'results': []
+                })
+
+        # Return property views limited to the 'inventory_ids' list.  Otherwise, if selected is empty, return all
+        if 'inventory_ids' in request.data and request.data['inventory_ids']:
+            property_views_list = PropertyView.objects.select_related('property', 'state', 'cycle') \
+                .filter(property_id__in=request.data['inventory_ids'],
```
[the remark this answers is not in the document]
macintoshpie: (note to self) I will research if there are existing relevant tests which I can move to v3. If so, I might go ahead and refactor this function to use non-canonical ids.

seed/views/v3/columns.py
```diff
@@ -96,7 +98,7 @@ def list(self, request):
         """
         organization_id = self.get_organization(self.request)
         inventory_type = request.query_params.get('inventory_type', 'property')
-        only_used = request.query_params.get('only_used', False)
+        only_used = json.loads(request.query_params.get('only_used', 'false'))
```
macintoshpie, line 101: Note that we were previously getting the query param as a string, which means when doing conditionals with it later it would always be true whenever this param was set.

Issue #2179, opened 2020-04-12
API V3 - Data Quality


adrian-lara:
Build out the following v2 endpoints into v3, update fields shown in Swagger, and also make changes as listed. Note that the path ID is Organization ID. See comments below for more details:
- [x] **/api/v2/data_quality_checks/ - post**: ~~investigate - maybe nested rules under data_quality_checks~~ move to /data_quality_checks/{organization_id}/start - has nothing to do with Rules, instead, this starts a DQ check.
- [x] **/api/v2/data_quality_checks/results/ - get**: move
- [x] **/api/v2/data_quality_checks/{id}/csv/ - get**: investigate - id is "importfile or cache key" - move to /data_quality_checks/results_csv
- [x] **/api/v2/data_quality_checks/data_quality_rules/ - get**: move to /data_quality_checks/{organization_id}/rules - update response format
- [x] **/api/v2/data_quality_checks/reset_all_data_quality_rules/ - put**: deprecate
- [x] **/api/v2/data_quality_checks/reset_default_data_quality_rules/ - put**: move to /data_quality_checks/{organization_id}/rules/reset - update response format
- [x] **/api/v2/data_quality_checks/save_data_quality_rules/ - post**: ~~slated for refactor - possibly move to /data_quality_checks/{organization_id}/rules/batch_save~~ replace with /data_quality_checks/{organization_id}/rules/{id} PUT for individual rules

Additionally, review the following:
- [ ] use of DRF ViewSets.
- [ ] use of decorators is reasonable/
- [ ] comments include at least a description. Details about incoming or outgoing fields are nice-to-haves as well.

Manual Testing:
Visit the /app/#/api/swagger page and "Execute" the endpoints listed above.

adrian-lara:
Ideally, these endpoints would become more "RESTful" by doing the following. A DataQualityCheck object would be a nested resource under an Organization, and Rules would be a nested resource under DataQualityChecks.

Unfortunately, the underlying models would need to be refactored - expanding this work beyond the scope of the API V3 effort. Currently, each Organization only has one DataQualityCheck object, and custom logic is built into the DataQualityCheck model to ensure an organization only has one DataQualityCheck object. Each of the endpoints above use an Organization ID to "retrieve" an Organization's one and only DataQualityCheck object - treating the Organization ID as if it's a DataQualityCheck ID.  In those cases, it's passed in as a query parameter.

[...]

PR #2257, opened 2020-06-15
2179 api v3 dq

adrian-lara:
The previous Data Quality Check endpoints worked but could be improved upon in terms of "RESTful-ness" and DRF serializer usage.

#### What's this PR do?
Build out V3 Data Quality Check and Data Quality Check Rule endpoints per the attached issue. 

Start using nested_routers to separate Rule resource. Start using Rule serializer to allow the usage of auto-generated Django Rest Framework endpoints.

"Try" the API V3 Data Quality Check and Data Quality Check Rule endpoints. 

Code review can be done by viewing commits in order.

#2179


seed/views/v3/rules.py
```diff
@@ -0,0 +1,281 @@
...
+        request_body=SaveDataQualityRulesPayloadSerializer,
+        responses={
+            200: DataQualityRulesResponseSerializer
+        }
+    )
+    @api_endpoint_class
+    @ajax_request_class
+    @has_perm_class('requires_parent_org_owner')
+    @action(detail=False, methods=['POST'])
+    def batch_save(self, request, nested_1_pk=None):
```
adrian-lara, line 182: No need to review as this is still being updated.

seed/views/v3/data_quality.py
```diff
@@ -148,23 +148,22 @@ class DataQualityViews(viewsets.ViewSet):
             })
         }
     )
-    def create(self, request):
+    @api_endpoint_class
+    @ajax_request_class
+    @has_perm_class('requires_member')
+    @action(detail=True, methods=['POST'])
+    def start(self, request, pk):
```
macintoshpie, line 68: Since this breaks convention with the PK being Organization not DataQuality, what do you think about using `organization_pk` rather than just `pk` here?

seed/serializers/rules.py
```diff
@@ -0,0 +1,105 @@
...
+        from the same org.
+        """
+        if label is not None and label.super_organization_id != self.instance.data_quality_check.organization_id:
+            raise serializers.ValidationError(
+                f'Label with ID {label.id} not found in organization, {self.instance.data_quality_check.organization.name}.'
+            )
+        else:
+            return label
+
+    def validate(self, data):
```
adrian-lara, line 56: This method was an attempt to replicate the manual validation checks happening in the previous DQ `save_data_quality_rules` endpoint.

seed/views/v3/data_quality_check_rules.py
```diff
@@ -0,0 +1,79 @@
...
+    name='update',
+    decorator=swagger_auto_schema(manual_parameters=[nested_org_id_path_field])
+)
+class DataQualityCheckRuleViewSet(viewsets.GenericViewSet, ListModelMixin, UpdateWithoutPatchModelMixin):
+    serializer_class = RuleSerializer
+    model = Rule
+    pagination_class = None
+
+    def get_queryset(self):
+        # Handle the anonymous case (e.g. Swagger page load)
```
adrian-lara, line 50: Swagger page was loading with an error/warning, and this was the fix: https://github.com/carltongibson/django-filter/issues/966

seed/views/v3/data_quality.py
```diff
@@ -10,124 +10,40 @@
 from celery.utils.log import get_task_logger
```
macintoshpie, line 10: What do you think about renaming this file to data_quality_checks.py?
I think thus far our files have been following the plural form of the model.

seed/serializers/rules.py
adrian-lara: Hoping for feedback on the "two" validations performed here.

Could we make it easier to follow? Is it correct, or are there any logical holes or cases not covered? Is there a better way to implement this?

adrian-lara: @lainsworth8801 Might be worth your second opinion since you implemented the checks initially.

seed/serializers/rules.py
```diff
@@ -0,0 +1,105 @@
...
+        contained in 'validate_{field_name}' methods which are only checked when
+        that field is in 'data'.
+        """
+        data_invalid = False
+        validation_messages = []
+
+        # Rule with SEVERITY setting of "valid" should have a Label.
+        severity_is_valid = self.instance.severity == Rule.SEVERITY_VALID
+        severity_unchanged = 'get_severity_display' not in data
+        severity_will_be_valid = data.get('get_severity_display') == "valid"
```
macintoshpie, line 70: Not sure if this is any better, but maybe we could do `dict([(display, val) for val, display in Rule.SEVERITY])[Rule.SEVERITY_VALID]` instead of comparing directly to a string literal

seed/views/v3/data_quality.py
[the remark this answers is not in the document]
adrian-lara: Good catch! 👍

seed/serializers/rules.py
```diff
@@ -0,0 +1,105 @@
...
+            'required',
+            'rule_type',
+            'severity',
+            'status_label',
+            'table_name',
+            'text_match',
+            'units',
+        ]
+
+    def validate_label(self, label):
```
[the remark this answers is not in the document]
adrian-lara: Yes and yes! Nice catch 💯

seed/views/v3/data_quality_checks.py
```diff
@@ -0,0 +1,168 @@
...
+    @has_perm_class('requires_member')
+    @action(detail=True, methods=['POST'])
+    def start(self, request, organization_id):
+        """
+        This API endpoint will create a new data_quality check process in the background,
+        on potentially a subset of properties/taxlots, and return back a query key
+        """
+        body = request.data
+        property_state_ids = body['property_state_ids']
+        taxlot_state_ids = body['taxlot_state_ids']
```
adrian-lara, line 78: @macintoshpie Do you think these should be `_view_ids` instead? I'm leaning that way but wanted your thoughts.

macintoshpie: I'm honestly not sure. I took a look at how the api is used on the front end ([here](https://github.com/SEED-platform/seed/blob/4829ad94af7b7391e6ce28f8873c6d0b39bcdc94/seed/static/seed/js/controllers/inventory_list_controller.js#L404)) and it seems kind of trivial to make the change there. I'm not totally sure how, on the backend, you find the correct property states given a list of property views, but if that's easy as well then 🤷 .

I guess this goes back to our discussion of whether an api user is aware of and cares about <Inventory>States.

adrian-lara: Agree - both wouldn't take too much effort to change. I'll tag you for re-review.

Issue #2183, opened 2020-04-12
API V3 - Column Mapping Profiles


adrian-lara:
Build out the following v2 endpoints into v3, update fields shown in Swagger, and also make changes as listed:
- [x] **/api/v2/column_mapping_presets/ - get**: rename "filter" and use POST for 'preset_type' array param passed via body
- [x] **/api/v2/column_mapping_presets/ - post**: column_mapping_profiles
- [x] **/api/v2/column_mapping_presets/{id}/ - put**: column_mapping_profiles
- [x] **/api/v2/column_mapping_presets/{id}/ - delete**: column_mapping_profiles (make visible on Swagger page)
- [x] **/api/v2/column_mapping_presets/suggestions/ - post**: column_mapping_profiles

Additionally, review the following:
- [ ] use of DRF ViewSets.
- [ ] use of decorators is reasonable/
- [ ] comments include at least a description. Details about incoming or outgoing fields are nice-to-haves as well.

Manual Testing:
Visit the /app/#/api/swagger page and "Execute" the endpoints listed above.

PR #2262, opened 2020-06-23
2183 api v3 column mapping profile

adrian-lara:
Test Column Mapping Profile endpoints on the Swagger page.

#2183


seed/views/v3/column_mapping_profiles.py
```diff
@@ -0,0 +1,313 @@
...
+              required: true
+              paramType: path
+            - name: name
+              description: Name of profile
+              required: false
+              paramType: body
+            - name: mappings
+              description: List of dictionaries
+              required: false
+              paramType: body
```
macintoshpie, line 104: Do you think we can delete these old swagger docs here as well as in the other views?

Issue #2259, opened 2020-06-22
API V3 - LabelInventory (property or tax lot)


adrian-lara:
Build out the following v2 endpoints into v3, update fields shown in Swagger, and also make changes as listed:

- [ ] **/api/v2/labels-taxlot/ - put**: move to LabelInventoryViewSet and labels_inventories.py
- [ ] **/api/v2/labels-property/ - put**: move to LabelInventoryViewSet and labels_inventories.py

NOTE: Use a similar entry in `urlpatterns` in the v3/urls.py file as in v2/urls.

Additionally, review the following:
- [ ] use of DRF ViewSets.
- [ ] use of decorators is reasonable/
- [ ] comments include at least a description. Details about incoming or outgoing fields are nice-to-haves as well.

Manual Testing:
Visit the /app/#/api/swagger page and "Execute" the endpoints listed above.

macintoshpie:
Note that it seems any authenticated user can currently add and remove labels to any property in SEED (eg cross org). Will discuss this further with Adrian

PR #2264, opened 2020-06-23
2259 refactor/api v3 label inventory

macintoshpie:
#### What's this PR do? 
Moves v2 /labels-taxlot and /labels-property views to v3 
Test `api/v3/labels_taxlot/` and `api/v3/labels_property/` on swagger page
#2259


macintoshpie:
> Empty inventory_ids and populated remove_label_ids does it to everything

That's appears to be the intended effect, though it _should_ limit itself to the removal of labels that are within the org. If this is not what we want then we can change that. I'll make this PR a draft until we decide what to do here.

adrian-lara:
> > Empty inventory_ids and populated remove_label_ids does it to everything
> 
> That's appears to be the intended effect, though it _should_ limit itself to the removal of labels that are within the org. If this is not what we want then we can change that. I'll make this PR a draft until we decide what to do here.

That's a good point - I could definitely imagine users wanting to just completely remove a label from their entire org's properties or tax lots.

Looking into the way it's used on the FE, inventory_ids are basically required since the Update Labels button isn't "activated" until records are checked on the inventory list page. Given that, I'd vote to keep the changes you have currently (also requiring it for the API endpoint). What do you think?

macintoshpie:
I think that sounds good 👍

Issue #2182, opened 2020-04-12
API V3 - Column List Profiles


adrian-lara:
Build out the following v2 endpoints into v3, update fields shown in Swagger, and also make changes as listed:
- [x] **/api/v2/column_list_settings/ - get**: name column_list_profiles
- [x] **/api/v2/column_list_settings/ - post**: name column_list_profiles
- [x] **/api/v2/column_list_settings/{id}/ - delete**: name column_list_profiles
- [x] **/api/v2/column_list_settings/{id}/ - get**: name column_list_profiles
- [x] **/api/v2/column_list_settings/{id}/ - patch**: deprecate
- [x] **/api/v2/column_list_settings/{id}/ - put**: name column_list_profiles

Additionally, review the following:
- [ ] use of DRF ViewSets.
- [ ] use of decorators is reasonable/
- [ ] comments include at least a description. Details about incoming or outgoing fields are nice-to-haves as well.

Manual Testing:
Visit the /app/#/api/swagger page and "Execute" the endpoints listed above.

Issue #2254, opened 2020-06-10
API V3 - Geocode, UBID, and Misc.


adrian-lara:
Build out the following v2 endpoints into v3, update fields shown in Swagger, and also make changes as listed:

Geocode:
- [x] **/api/v2/geocode/confidence_summary/ - post**: move. Keep as POST for array
- [x] **/api/v2/geocode/geocode_by_ids/ - post**: move. Keep as POST for array

UBID:
- [x] **/api/v2/ubid/decode_by_ids/ - post**: move. Keep as POST for array
- [x] **/api/v2/ubid/decode_results/ - post**: move. Keep as POST for array

Misc 2.0:
- [x] **/api/v2/version/ - get**: move to /api/version
- [x] **/api/v2/schema/ - get**: investigate - search FE - likely deprecate
- [x] **/api/v2/progress/{id}/ - get**: move
- [x] **/api/v2/get_upload_details/ - get**: investigate - search FE - likely deprecate

Misc 2.1:
- [x] **/api/v2.1/portfolio_manager/report/ - post**: move
- [x] **/api/v2.1/portfolio_manager/template_list/ - post**: move
- [x] **/api/v2.1/tax_lot_properties/export/ - post**: move

Additionally, review the following:
- [ ] use of DRF ViewSets.
- [ ] use of decorators is reasonable/
- [ ] comments include at least a description. Details about incoming or outgoing fields are nice-to-haves as well.

Manual Testing:
Visit the /app/#/api/swagger page and "Execute" the endpoints listed above.

PR #2269, opened 2020-06-24
2182 api v3 column list profiles

adrian-lara:
#2182 

Test listed endpoints from the associated GH Issue within SEED's Swagger page.


seed/tests/test_api.py
```diff
@@ -32,32 +32,6 @@ def test_get_api_endpoints_utils(self):
         for url, fn in res.items():
             self.assertTrue(fn.is_api_endpoint)
             self.assertTrue(url.startswith('/'))
-            # these urls are generated by the DRF router and with /?
```
adrian-lara, line 35: @nllong and/or @axelstudios 
I removed this thinking it was stale. ([It's from the initial commit of SEED.](https://github.com/SEED-platform/seed/blame/2d924db8d16c2052034d5b0e3889c53b5b2b3f24/seed/tests/test_api.py#L94)).

Before this removal, we just kept adding URIs to the tuple below. It seemed unnecessary, but let me know if you all have more historical background on this and why we've kept it.

axelstudios: I'll be glad to see that gone if it's no longer needed

macintoshpie: After taking a look myself, I'm not sure we really need it anymore. it's used by the `get_api_schema` view, at `/api/v2/schema`. If we're deprecating that view it might be nice to clean all of that stuff up

adrian-lara: > After taking a look myself, I'm not sure we really need it anymore. it's used by the `get_api_schema` view, at `/api/v2/schema`. If we're deprecating that view it might be nice to clean all of that stuff up

That endpoint might be cleaned up (/deprecated) with #2254.

PR #2270, opened 2020-06-25
Fix/validate use cases org

macintoshpie:
#### What's this PR do?
Fixes a bug where anyone who is not a superuser is not able to upload a buildingsync file. Issue was the view had `has_perm_class('can_modify_data')` but the service on the frontend was not adding `organization_id` as a query parameter, so authz fails.
Log in as an Owner or Member, you should be able to navigate to the data upload modal, select buildingsync, then get a validation response.
Addresses Robin's comment from #2024 where the page returned a 403


aviveiros11:
@macintoshpie this may be an issue on my end but after pulling your branch I am getting this error.  
Screen Shot 2020-06-25 at 7 56 53 PM
I attempted migrations but I'm gonna look into it a bit more.

nllong:
Looks good.

PR #2273, opened 2020-06-29
2250 refactor/api v3 properties v21

macintoshpie:
#### What's this PR do? 
Moves and refactors v2.1 properties endpoints to v3
Test the endpoints mentioned in the issue within the v2.1 section on the swagger page.
#2250

### DIFF
--- a/seed/views/v3/properties.py
+++ b/seed/views/v3/properties.py
@@ -2,26 +2,29 @@
 :copyright (c) 2014 - 2020, The Regents of the University of California, through Lawrence Berkeley National Laboratory (subject to receipt of any required approvals from the U.S. Department of Energy) and contributors. All rights reserved.  # NOQA
 :author
 """
+import os
 from collections import namedtuple
 
 from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
-from django.db.models import Q
-from django.db.models import Subquery
-from django.http import JsonResponse
+from django.db.models import Q, Subquery
+from django.http import HttpResponse, JsonResponse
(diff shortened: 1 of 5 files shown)


seed/views/v3/properties.py
```diff
@@ -1010,6 +1014,244 @@ def update(self, request, pk=None):
...
+
+        if not cycle_pk:
+            return JsonResponse({
+                'success': False,
+                'message': "Cycle ID is not defined"
+            })
+        else:
+            cycle = Cycle.objects.get(pk=cycle_pk)
+
+        result = self._get_property_view_for_property(pk, cycle_pk)
```
macintoshpie, line 1216: I'd like to refactor this view to just accept a pk that's a property_view, not a canonical property id. Then we're much more consistent and we don't have to fool with this method. Thoughts @adrian-lara ?

adrian-lara: Agree - that's a good path to take.

seed/views/v3/property_scenarios.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class PropertyScenarioViewSet(SEEDOrgReadOnlyModelViewSet):
+    """
+    API View for Scenarios. This only includes retrieve and list for now.
+    """
+    serializer_class = ScenarioSerializer
+    parser_classes = (JSONParser, FormParser,)
+    renderer_classes = (JSONRenderer,)
+    queryset = Scenario.objects.all()
```
adrian-lara, line 26: I think we can remove this since `get_queryset` is being overridden.

PR #2274, opened 2020-06-29
2254 api v3 geocode ubid

adrian-lara:
Test geocode and UBID endpoints on the Swagger page.

This PR address half of the listed endpoints in #2254

### DIFF
--- a/seed/views/v3/geocode.py
+++ b/seed/views/v3/geocode.py
@@ -16,7 +16,7 @@ from seed.utils.api import api_endpoint_class
 from seed.utils.geocode import geocode_buildings
 
 
-class GeocodeViews(viewsets.ViewSet):
+class GeocodeViewSet(viewsets.ViewSet):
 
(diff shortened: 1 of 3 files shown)


seed/views/v3/geocode.py
```diff
@@ -0,0 +1,145 @@
...
+                'taxlot_view_ids': ['integer'],
+            },
+            description='IDs by inventory type for records to be geocoded.'
+        )
+    )
+    @api_endpoint_class
+    @ajax_request_class
+    @has_perm_class('can_modify_data')
+    @action(detail=False, methods=['POST'])
+    def geocode_by_ids(self, request):
```
macintoshpie, line 39: Could we get a short docstring for the view here? Same for the other views added in this PR

adrian-lara: Nice catch! Added them in, let me know if you had something else in mind.

Issue #2253, opened 2020-06-10
API V3 - Measures, building file, and upload


adrian-lara:
Build out the following v2 endpoints into v3, update fields shown in Swagger, and also make changes as listed:

Measures:
- [x] **/api/v2/measures/ - get**: move
- [x] **/api/v2/measures/{id}/ - get**: move
- [x] **/api/v2/measures/reset/ - post**: move

Building File:
- [x] **/api/v2/building_file/ - get**: ~~investigate - search FE - likely deprecate~~ move to /building_files/ (with an "s")
- [x] **/api/v2/building_file/ - post**: move to /building_files/ (with an "s")
- [x] **/api/v2/building_file/{id}/ - get**: ~~investigate - search FE - likely deprecate~~ move to /building_files/ (with an "s")

Upload:
- [x] **/api/v2/upload/ - post**: move
- [x] **/api/v2/upload/create_from_pm_import/ - post**: move

Additionally, review the following:
- [x] use of DRF ViewSets.
- [x] use of decorators is reasonable/
- [x] comments include at least a description. Details about incoming or outgoing fields are nice-to-haves as well.

Manual Testing:
Visit the /app/#/api/swagger page and "Execute" the endpoints listed above.

macintoshpie:
After chatting with Austin, I've decided that we can deprecate `POST /building_file/`. It was previously the entry for uploading a new property from a buildingsync file, but we now use the `POST /upload/` endpoint like all of the other files.

I'm not sure where to note this, but when we come back to make the changes on the front end, we should remove the default buildingsync uploader since it's not used (see here https://github.com/SEED-platform/seed/blob/9e23e67342cf33c156dfe7b8b4983c4b4919b022/seed/static/seed/js/directives/sdUploader.js#L164)

PR #2275, opened 2020-06-30
2253 refactor/api v3 building file

macintoshpie:
#### What's this PR do? 
Adds the building_files endpoint to v3 api, and fixes an issue with the hpxml export
Test the `api/v3/building_files/*` endpoints. Additionally, test `api/v3/properties/{id}/hpxml` for a view that _was_ imported using hpxml, and for a view that was not (ie made from or updated with buildingsync file). For the second case it should still succeed, whereas previously this would fail b/c it was trying to parse the BuildingSync file as an hpxml file.
#2253

### DIFF
--- a/seed/views/v3/building_files.py
+++ b/seed/views/v3/building_files.py
@@ -30,10 +30,6 @@ class BuildingFileViewSet(SEEDOrgReadOnlyModelViewSet):
     parser_classes = (MultiPartParser,)
 
     def get_serializer_class(self):
-        # Super hacky, but this allows us to have separate serializers for different
-        # views
-        # This is necessary b/c bad swagger docs were being generated due to the
-        # serializer for the `create` view
         if self.request.method.lower() == 'post':
             return None
         return BuildingFileSerializer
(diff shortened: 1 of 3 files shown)


seed/views/v3/building_files.py
```diff
@@ -0,0 +1,174 @@
...
+    parser_classes = (MultiPartParser,)
+
+    def get_serializer_class(self):
+        # Super hacky, but this allows us to have separate serializers for different
+        # views
+        # This is necessary b/c bad swagger docs were being generated due to the
+        # serializer for the `create` view
+        if self.request.method.lower() == 'post':
+            return None
+        return BuildingFileSerializer
```
macintoshpie, line 39: I'm not happy with this, but it was better than the other options:
- apply serializer to entire class, then have some unwanted params in the swagger docs for the create view
- apply serializer on swagger_auto_schema_decorator for the list and retrieve endpoints. This seemed less preferable b/c then DRF wouldn't know about the serializer and use it.

adrian-lara: I think this isn't bad/hacky. Actually, according to the DRF docs, this sounds exactly like what overriding the get_serializer_class was meant for:  

>May be overridden to provide dynamic behavior, such as using different serializers for read and write operations...
https://www.django-rest-framework.org/api-guide/generic-views/#get_serializer_classself

I'd vote to remove all or most of those comments actually.

Issue #2281, opened 2020-07-02
Making new DQ rule, set Condition check, then set Field, condition check get set to blank


RDmitchell:
### Expected Behavior
Make new DQ rule, set the first criteria, which is Condition Check, then set Field. Program should keep the original setting for the Condition Check
### Actual Behavior
After selecting the field, the program sets the Condition Check to blank
### Steps to Reproduce
in DQ Admin
* Click Create a new rule button -- the Condition Check and Field pulldown lists are empty
![image](https://user-images.githubusercontent.com/6314950/86304961-4fcce180-bbc5-11ea-9e85-4015a3857bc9.png)
* Set Condition Check to anything, in this example, to Not Null
![image](https://user-images.githubusercontent.com/6314950/86305010-7559eb00-bbc5-11ea-8eae-1a09ceac3abd.png)
* Set the field to "Address Line 1" and the program resets the Condition Check to "Must Contain" which makes no sense because each address will be different (could this be because it is a matching field? Even so, the program behavior makes no sense).
![image](https://user-images.githubusercontent.com/6314950/86305169-e00b2680-bbc5-11ea-8086-ac431db06c78.png)
* Starting again with a new rule, setting Condition Check to Not Null, then set Field to blank -- this happens to most fields (except maybe matching fields?)
![image](https://user-images.githubusercontent.com/6314950/86305376-54de6080-bbc6-11ea-971e-91220bd9c897.png)

### Instance Information
instance: dev1
SHA: 40ca210b

RDmitchell:
Not sure what priority this should have. 

The work around is for the user to set the Field first, then the Condition Check.

RDmitchell:
Instance: dev1
SHA: fd75dea4
Org: LBNL Test 200

The program now keeps the selected Condition Check after a Field name is selected.

PR #2287, opened 2020-07-06
DQ Admin page - doesn't guess condition when field changes

adrian-lara:
See issue

#### What's this PR do?
The DQ Admin page doesn't guess condition when Field values is changed for one of the rules.

In addition, validations have been added to `condition` and `data_type`. These are triggered whenever `condition`, `field`, and `data_type` values are changed. Warning messages are provided to help users "fix" any invalid rules.

Visit DQ admin page, and change the Field value for one of the rules. See that the Condition doesn't change.

In addition, monkey test changing the `condition`, `field`, and `data_type` values for rules.

#2281


axelstudios:
Huge improvement with the error message and red outlines:

> image

adrian-lara:
Final changes look good to me 👍

PR #2288, opened 2020-07-06
set updated and created to read_only to unbreak write actions

fablet:
The addition of 'created' and 'updated' to the PropertySerializer makes those fields required on write operations even though they are auto fields in the model.  Also in PropertyStateWritableSerializer, allow_null=True still leaves the fields required even though they are not.
#### What's this PR do?
Changes 'created' and 'updated' to read_only fields in the PropertySerializer.  Adds required=False to PropertyStateWritableSerializer non-required date and time fields.

### DIFF
--- a/seed/serializers/properties.py
+++ b/seed/serializers/properties.py
@@ -230,11 +230,11 @@ class PropertyStateWritableSerializer(serializers.ModelSerializer):
 
     # support naive datetime objects
     # support naive datetime objects
-    generation_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True)
-    recent_sale_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True)
-    release_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True)
-    analysis_start_time = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True)
-    analysis_end_time = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True)
+    generation_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True, required=False)
+    recent_sale_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True, required=False)
+    release_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True, required=False)
+    analysis_start_time = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True, required=False)
+    analysis_end_time = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True, required=False)
 
     # support the pint objects
     conditioned_floor_area = PintQuantitySerializerField(allow_null=True, required=False)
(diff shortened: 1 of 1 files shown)


seed/serializers/properties.py
```diff
@@ -230,11 +230,11 @@ class PropertyStateWritableSerializer(serializers.ModelSerializer):
...
-    generation_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True)
-    recent_sale_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True)
-    release_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True)
-    analysis_start_time = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True)
-    analysis_end_time = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True)
+    generation_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True, required=False)
+    recent_sale_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True, required=False)
+    release_date = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True, required=False)
+    analysis_start_time = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True, required=False)
+    analysis_end_time = serializers.DateTimeField('%Y-%m-%dT%H:%M:%S', allow_null=True, required=False)
```
adrian-lara, line 237: The only set of endpoints that I see using this for write purposes are `/api/v2/property_views/` - which seems to be doing so indirectly via `PropertyViewAsStateSerializer`. Are you still using those property_views endpoints? They were slated to be deprecated, but we can add them back in.

If you aren't using those, just for my own curiosity, which endpoint(s) are directly or indirectly using `PropertyStateWritableSerializer`?

fablet: It looks like the api/v2/property_view were not among the ones you asked me about in the other thread so I wasn't aware you were planning to deprecate those, soI may still need them.  

As far as I can tell, PropertyStateWriteableSerializer is only being used by PropertyViewAsStateSerializer which is being used by v2 PropertyViewViewSet and v2_1 PropertyViewSetV21.  Paul Munday and I set up the v2 PropertyViewViewSet.  

I have not had a chance to review v2.1  PropertyViewSetV21 so it's possible I don't need the v2 property_view endpoint if v2.1 properties endpoint is being ported into 3 and has all the same functionality

adrian-lara: > It looks like the api/v2/property_view were not among the ones you asked me about in the other thread so I wasn't aware you were planning to deprecate those, soI may still need them.

Sorry about that, I think we might have misunderstood which those endpoints actually were when we initially scoped out endpoints to deprecate, so I never included it in the list of endpoints to ask you about.

> As far as I can tell, PropertyStateWriteableSerializer is only being used by PropertyViewAsStateSerializer which is being used by v2 PropertyViewViewSet and v2_1 PropertyViewSetV21. Paul Munday and I set up the v2 PropertyViewViewSet.
> 
> I have not had a chance to review v2.1 PropertyViewSetV21 so it's possible I don't need the v2 property_view endpoint if v2.1 properties endpoint is being ported into 3 and has all the same functionality

Nice catch! I did look over `PropertyViewSetV21` and it doesn't provide the same functionality that `PropertyViewViewSet` does. We'll add `PropertyViewViewSet` back in! 


------------------------------
Finally, unless I hear otherwise, I'll go ahead and open another PR with these changes cherry-picked on top of latest develop. I'll close this PR once I do that.

adrian-lara:
Changes moved to PR #2298

Issue #2290, opened 2020-07-07
DQ Admin: Valid Data Rule run from Inventory seems to "hang"


RDmitchell:
### Expected Behavior
I created a DQ rule to label records where ENERGY STAR Score is between 10 and 100 as "DQ: Valid ENERGY STAR Score". 

I expected this rule to generate "Valid ENERGY STAR Score" labels for all the records that had scored between 10 and 100.
![image](https://user-images.githubusercontent.com/6314950/86682759-be3ee480-bfb5-11ea-893c-e955bcf36658.png)

In Inventory, I selected one record, which I knew had valid ENERGY STAR Score data, and ran the DQ results on that one record.
![image](https://user-images.githubusercontent.com/6314950/86682987-f80feb00-bfb5-11ea-926f-74c190d9e074.png)

I expected that record to have the label applied to it.
### Actual Behavior
The program seemed to "hang" -- the spinning ball just kept spinning
![image](https://user-images.githubusercontent.com/6314950/86683170-24c40280-bfb6-11ea-9d92-b8fd9d0163cd.png)

But no errors in either the Network or Console tabs of the Developer tools.
### Steps to Reproduce
Create a Valid Data rule in DQ Admin.
Run it on a set of data in the Inventory List
### Instance Information
instance: dev1
SHA: 40ca210b

RDmitchell:
@adrian-lara / @axelstudios -- I have had this same experience trying to run a Valid Data check on multiple records so decided to make it simple by running it on just one record. But same issue, program seems to hang, without errors in the developer tools.

RDmitchell:
After letting it run for about 10 minutes, and the spinner still there, I refreshed the page, and it seems that it did apply the label to the correct record, Custom ID 1 == 107, but also an incorrect record, Custom ID 1 = 103 (Energy Star Score = 1)

??

![image](https://user-images.githubusercontent.com/6314950/86684324-4a9dd700-bfb7-11ea-8380-af57aa1394ed.png)

RDmitchell:
Instance: dev1
SHA: fd75dea4
Org: LBNL Test 200
Cycle: DQ testing

This appears to be working now

I set up the Valid data rule
![image](https://user-images.githubusercontent.com/6314950/88122071-71920680-cb7c-11ea-8986-60f1516aa615.png)

Then ran it in the Inventory List for Cycle = DQ Testing
Program labeled the appropriate records
[...]

PR #2291, opened 2020-07-07
2290 dq valid label results fix

adrian-lara:
In cases that a Rule with "valid" severity and a "range" condition is triggered, a line is **not** included in the DQ Check results. Because of this, it's possible for SEED to hang when trying to display the label that is accompanied with that triggered rule.

See the attached issue to see how this was discovered.

#### What's this PR do?
This PR implements a flag to allow for the application of the label to the record while avoiding the attempt to display the applied label in the result.

In addition, the following inventory list label clean ups are included here as well:
With labels applied, the following actions would remove selected and valid (is_applied) labels.
- Run DQ check - Now selected labels should remain.
- Switch cycles - Now, only no-longer-applicable labels are filtered out.
- Remove selected label from some records - Now, selected labels can remain only if they're applied to records. Also, the filtered records are refreshed.

Run DQ checks from the inventory list page while forcing valid data rules to trigger. Check that any of these valid data rules had their labels applied to **only** the records that should have triggered these rules. 

Also, specifically test that a single Rule can successfully be triggered on a single record without "hanging".

Lastly, monkey test labels, and specifically test that with labels applied, the following actions have a reasonable UX (see changes above):
- Run DQ check
- Switch cycles
- Remove selected label from records

#2290

### DIFF
--- a/seed/models/data_quality.py
+++ b/seed/models/data_quality.py
@@ -787,7 +787,6 @@ class DataQualityCheck(models.Model):
                         if rule.severity == Rule.SEVERITY_VALID:
                             label_applied = self.update_status_label(label, rule, linked_id, row.id, False)
 
-
                 if not label_applied and rule.status_label_id in model_labels['label_ids']:
                     self.remove_status_label(label, rule, linked_id)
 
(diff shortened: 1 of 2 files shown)


adrian-lara:
Worth noting, there is an inconsistency in how the "valid" rules are displayed. Specifically, valid rules that have the condition of "range" (these are rules that check for min and max values) DO NOT appear in DQ check results when triggered. Alternatively, valid rules that are of any other condition DO appear in DQ check results when triggered.

Even though this is inconsistent, the good news is that labels seem to be applied correctly in all cases.

Issue #2289, opened 2020-07-07
DQ Admin: Can't Save New Rule after deleting all existing rules


RDmitchell:
### Expected Behavior
Deleted all rules, then made one new rule, and wanted to save it.
### Actual Behavior
The Save Changes button is grayed out so the new rule can't be saved
### Steps to Reproduce
Not quite sure how to exactly reproduce it, but in an existing org, in the DQ Admin page, I deleted all the rules, clicked Save Changes (which I could do), logged out, logged back in, went to the now empty DQ Admin page (for View by Property tab)
![image](https://user-images.githubusercontent.com/6314950/86673569-a9118800-bfac-11ea-831f-c14adbf97cdd.png)
Created a new rule, then wanted to save it but the Save Changes button was grayed out.
![image](https://user-images.githubusercontent.com/6314950/86673617-b3338680-bfac-11ea-8052-d2327815330b.png)
### Instance Information

Server Instance: dev1
Server SHA: 40ca210b
Org: LBNL 101

RDmitchell:
@adrian-lara -- created this new issue, and put it in the current project, awaiting triage. Seems like maybe someone else should try to replicate it and if they can't then move it to a lower priority (??). Anyway, not quite sure what to do with it.

RDmitchell:
If I click Restore Default Rules, still can't Save Changes

But if I then delete 1 rule, the Save Changes button comes alive again.

RDmitchell:
Instance: dev1
SHA: fd75dea4
Org: LBNL Test 200

This is fixed. I deleted all the rules, logged out, logged back in, came to the empty DQ admin rules for Tax Lot, made a new rule, and was able to save it.

Issue #1848, opened 2019-03-27
DQ Rules -- Add warning to save rules before leaving screen


RDmitchell:
### Expected Behavior
It would be nice if the program would tell you that you haven't saved your Data Quality rule changes when you attempt to leave the screen, particularly when switching between the Property and Tax Lot tabs
### Actual Behavior 
Program lets you move between the two tabs without asking you to save, and you lose the changes that you made to the tab you just left if you don't remember to save. 
### Steps to Reproduce
In Property tab, make a new rule. Don't click the Save Changes button. Select the Tax Lot tab. Then select the Property Tab -- the rule you just made in the Property Tab is not there because you didn't remember to save it 
### Instance Information
Server Instance: dev1
Server SHA: 0469a1dc

RDmitchell:
@adrian-lara -- This works on Property tab going to Tax Lot tab, but not the other way around. So if I'm on the Tax Lot tab, make a change, and then click on the Property tab, the program doesn't put up the warning, goes to Property, then when I go back to Tax Lot the change I made was not saved. 

Am putting this back in "In Progress", and maybe you need to move it to the current March 2020 project.

RDmitchell:
Tested on

Instance: dev1
SHA: 6a6d4ba0

lainsworth8801:
<img width="896" alt="Screen Shot 2020-02-10 at 2 46 00 PM" src="https://user-images.githubusercontent.com/49968203/74193206-332f0d80-4c14-11ea-9930-27afd856029c.png">

lainsworth8801:
@robin
not sure why this is happening on your end but I don't seem to be able to recreate this on my front:
1. Restore and reset will automatically "save". Restore only reset the initial set of rule so your added rules will/should still be there. Only "reset" puts everything back to a complete clean slate;
2. Here i uncheck the first rule. "Save changes" button turn blue for click:
<img width="1115" alt="Screen Shot 2020-02-10 at 2 57 02 PM" src="https://user-images.githubusercontent.com/49968203/74194370-535fcc00-4c16-11ea-8ef3-1cde2477988d.png">
3. Then I saved it. It allows me to go to "tax lot" page and do modification:
<img width="1108" alt="Screen Shot 2020-02-10 at 2 57 21 PM" src="https://user-images.githubusercontent.com/49968203/74194438-7a1e0280-4c16-11ea-97a0-64a1da067304.png">
4. I clicked on the column, without actually changing anything, save button is activated:
[...]

PR #2292, opened 2020-07-08
2289 dq admin save button fix

adrian-lara:
On the DQ admin page, it was possible for the the "Save Changes" button to not be re-enabled when expected.

See attached issue for a specific case.

#### What's this PR do?
Removes extra logic around deciding whether a rule change has happened on the DQ admin page and clean up relevant methods (rename and delete no longer used variables).

Make changes on the DQ admin page, see that the Save Changes button is enabled.

Also, take the steps detailed in the attached issue to ensure that the original "steps" taken to discover this bug no longer yields this bug.

#2289 
Should also address #1848

### DIFF
--- a/seed/static/seed/js/controllers/data_quality_admin_controller.js
+++ b/seed/static/seed/js/controllers/data_quality_admin_controller.js
@@ -153,23 +153,12 @@ angular.module('BE.seed.controller.data_quality_admin', [])
       $scope.isModified = function () {
         return modified_service.isModified();
       };
-      var originalRules = angular.copy(data_quality_rules_payload.rules);
-      $scope.original = originalRules;
+
       $scope.change_rules = function () {
         $scope.defaults_restored = false;
         $scope.rules_reset = false;
         $scope.rules_updated = false;
-        $scope.setModified();
-      };
-      $scope.setModified = function () {
-        var cleanRules = angular.copy($scope.ruleGroups);
-        _.each(originalRules, function (rules, index) {
-          Object.keys(rules).forEach(function (key) {
-            _.reduce(cleanRules[index][rules[key].field], function (result, value) {
-              return !_.isEqual(value, rules[key]) && modified_service.setModified();
-            }, []);
-          });
-        });
+        modified_service.setModified();
       };
 
       // Restores the default rules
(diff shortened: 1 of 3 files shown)


seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -153,23 +153,12 @@ angular.module('BE.seed.controller.data_quality_admin', [])
...
-      $scope.original = originalRules;
-      $scope.change_rules = function () {
+
+      $scope.rules_changed = function () {
         $scope.defaults_restored = false;
         $scope.rules_reset = false;
         $scope.rules_updated = false;
-        $scope.setModified();
-      };
-      $scope.setModified = function () {
```
adrian-lara, line 164: Effectively, this only ever triggered the `modified_service.setModified()` method, so I simplified this logic.

From what I could tell, this method was attempting to check that the previous rules were still equal. I'm assuming that the intent was to eventually "re-disable" the "Save Changes" button if all rules were unchanged. But, I don't think this quite got there. 

Notably, there were some challenges in checking if rules have changed, including the potential for the inclusion of a $$hashkey to the changed rule as well as the potential to have a nested object (in the form of a label) for each rule object. 

I think it is possible to do, but it isn't as straight forward, so I took the easy path for now. Let me know if I should spend time on this.

seed/static/seed/partials/data_quality_admin.html
```diff
@@ -67,7 +67,7 @@ <h3 translate>Modifying Data Quality Rules</h3>
                         </tr>
                     </thead>
                     <tbody ng-repeat="field in sortedRuleGroups()">
-                        <tr ng-repeat="rule in ruleGroups[inventory_type][field]" ng-class="{ 'warning' : rule.new, 'disabled' : !rule.enabled }" ng-click="change_rules()">
+                        <tr ng-repeat="rule in ruleGroups[inventory_type][field]" ng-class="{ 'warning' : rule.new, 'disabled' : !rule.enabled }" ng-click="rules_changed()">
```
adrian-lara, line 70: This was triggered on any click (not just when the values are changed). Is this OK? Should we just apply this to every sub-html element with `ng-model`?

Issue #2293, opened 2020-07-09
DQ check - Not Null and Required rules for extra_data fields bug


adrian-lara:
### Expected Behavior
DQ rules for extra_data fields that were "required" or "not_null" should be able to flag any records with these extra_data fields missing (key-value in a record's extra_data field).

### Actual Behavior
DQ rules for extra_data fields that were "required" or "not_null" were not being triggered.

### Steps to Reproduce
Create a required or not_null DQ rule for an extra_data field. See that the rule is not triggered.

### Instance Information

Server Instance: `production` `develop`

RDmitchell:
Instance; Dev1
SHA:
Org: LBNL Test 200

* Set up DQ rule for Campus ID for required, which is an extra_data field
* Results came back that "campus ID is required but is None"
* Changed the rule to be "Not Null" and got the exact same results
I would have thought "required" would mean that the field had to be present, which it is; so these records were the ones that were there but empty; so in this case not really a difference between "Required" and "Not Null", but that's probably ok for now. 

The real test for Required in my mind is if the field is absent.
So a 2nd test
* Set up DQ rule for an extra_data field that I knew was not in the dataset, but which was mapped (for some import), ie: "Address 2"
* In this case, the program reported "Address 2 is required but is None", which again is not quite the desired message, but that's for a different issue.
* The program reported this error for all records in the dataset, which is correct.

Soooo, the program is definitely now doing DQ checking on extra_data fields for Required and Not Null.

Closing this issue.

Issue #2208, opened 2020-05-08
Data Quality tests


lainsworth8801:
### Expected Behavior
Write tests to validate each data quality check condition.

### Steps to Reproduce
Update test_data_quality_checks with a new class of condition tests.

### Instance Information

Server Instance: `dev1, dev2, production, staging, etc`
Server SHA: `abcd123`

PR #2294, opened 2020-07-09
2293 dq extra data bug

adrian-lara:
DQ rules for extra_data fields that were "required" or "not_null" were not being tagged.

#### What's this PR do?
Fixes that issue and add testing to extra_data field rules and non-extra_data field rules.

Monkey test DQ rules

#2293 
The new tests contribute to #2208 .

### DIFF
--- a/seed/models/data_quality.py
+++ b/seed/models/data_quality.py
@@ -726,7 +726,7 @@ class DataQualityCheck(models.Model):
 
             # get the status_labels for the linked properties and tax lots
             linked_id = model_labels['linked_id']
-            # import pdb; pdb.set_trace()
+
             if (rule.table_name, rule.field) not in self.column_lookup:
                 # If the rule is not in the column lookup, then it may have been a required
                 # field that wasn't mapped
--- a/seed/models/data_quality.py
+++ b/seed/models/data_quality.py
@@ -696,107 +696,102 @@ class DataQualityCheck(models.Model):
                 #                                             len(model_labels['label_ids'])))
 
         for rule in rules:
-            # create an extra data flag for the rule
-            is_extra_data = rule.field in row.extra_data
+            value = None
 
-            # check if the field exists
-            if hasattr(row, rule.field) or is_extra_data:
-                value = None
+            label_applied = False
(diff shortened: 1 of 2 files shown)


seed/models/data_quality.py
```diff
@@ -696,107 +696,102 @@ def _check(self, rules, row):
                 #                                             len(model_labels['label_ids'])))
 
         for rule in rules:
-            # create an extra data flag for the rule
-            is_extra_data = rule.field in row.extra_data
-
-            # check if the field exists
-            if hasattr(row, rule.field) or is_extra_data:
```
adrian-lara, line 703: CHANGE: `rule.field in row.extra_data` wasn't actually checking if the rule was for an extra_data column, since row.extra_data could just be missing keys. This check was removed, which lead to the removal of the top level `if` statement.

seed/models/data_quality.py
```diff
@@ -696,107 +696,102 @@ def _check(self, rules, row):
...
+            label_applied = False
+            display_name = rule.field
+
+            if hasattr(row, rule.field):
+                value = getattr(row, rule.field)
+                # TODO cleanup after the cleaner is better able to handle fields with units on import
+                # If the rule doesn't specify units only consider the value for the purposes of numerical comparison
+                if isinstance(value, ureg.Quantity) and rule.units == '':
+                    value = value.magnitude
+            else:  # rule is for extra_data
```
adrian-lara, line 710: CHANGE: This was changed to just assume that if "not" `hasattr(row, row.field)`, then the rule is for an extra data field.

seed/models/data_quality.py
```diff
@@ -696,107 +696,102 @@ def _check(self, rules, row):
...
+                value = row.extra_data.get(rule.field, None)
+
+                if ' (Invalid Footprint)' in rule.field and value is not None:
```
adrian-lara, line 713: CHANGE: ` and value is not None` was added since the previous iteration of this code already kicked out any rules with missing keys in the current record's extra_data. A test failed, and this change was made.

Issue #2286, opened 2020-07-02
Indicate mapped fields in DQ Admin definition or when running DQ Check in Inventory


RDmitchell:
### Expected Behavior
In the DQ Admin screen, when defining rules, it would be nice if the program would somehow indicate, in the list of fields in the Field pulldown, which fields were actually mapped in the existing data. 

There is the default list of fields (not sure how that is determined) and then there are the mapped fields. In my case, I had changed the default field name "Release Date" to "PM Release Date" when I mapped the data. 

However, I had forgotten that I mapped the fields that way, and so in DQ Admin, I selected the field "Release Date" for the DQ checking, rather than "PM Release Date". 
![image](https://user-images.githubusercontent.com/6314950/86414999-857fd200-bc7a-11ea-9457-9c97d420c638.png)

Then when I ran the DQ check in the Inventory list, records that I expected to be labeled based on the new rule were not. No records were labeled "DQ: Bad PM Release Date" even though I knew there were some that should have been labeled as such.
![image](https://user-images.githubusercontent.com/6314950/86415033-9b8d9280-bc7a-11ea-9691-2fa090530bb4.png)

Then I realized that I might have selected the wrong field, which it turns out I did. I changed the field name to "PM Release Date" in the DQ Admin screen
![image](https://user-images.githubusercontent.com/6314950/86414661-764c5480-bc79-11ea-890a-3889e528639e.png)

And then re-ran the DQ Check in Inventory, and the program generated the labels appropriately
![image](https://user-images.githubusercontent.com/6314950/86414714-a562c600-bc79-11ea-8bae-bbc84300d5cb.png)

This is somewhat important because it is possible to create DQ rules that you think are correct, when they actually are not, and thus a user could get faulty results when they run those DQ rules in the Inventory List. 

*******
It would be useful if somewhere in the process the program would have indicated to me that I had defined a rule for a field that wasn't in my mapped/imported data. 

A few options come to mind:
* Some sort of indication (a symbol to the left of the field name, colored background of field name ??) which fields are actually mapped when looking at the Field pulldown list of fields
* An information message when the DQ rules are saved that there are fields defined which are not in the current data, and list the fields. This would not be an error, just info, because a user might want to predefine the rules before having imported/mapped data. 
* An information message when the DQ rules are "run" in the Inventory list, indicating that some fields (and list the fields) that are defined in the DQ Admin but don't exist in the inventory. Again, not an error, just a message, which the user can determine whether they care about or not. 
[...]

PR #2297, opened 2020-07-15
2286 dq indicate mapped fields

adrian-lara:
When defining DQ Rules, we realized it would be helpful to flag fields that were not mapped by an organization. 

#### What's this PR do?
This PR introduces an indicator on the DQ admin page that shows which fields have been mapped to during file imports. In the backend, the mapped, or "used", columns are flagged by checking if `ColumnMapping.objects.filter(column_mapped=[col_object]).exists()`. [This check existed already and didn't need to be written.](https://github.com/SEED-platform/seed/blob/4baeb3441f92a4e886d2d28c515dde39289f05c7/seed/models/columns.py#L1321-L1323)

Note that no checks are performed on the actual inventory records as that operation could potentially be very expensive.

Visit the DQ admin page, and see that a ** is added to indicate fields that have not been mapped. Spot check to confirm fields are correctly flagged.

#2286 

Screen Shot 2020-07-15 at 1 21 30 PM

### DIFF
--- a/seed/static/seed/js/seed.js
+++ b/seed/static/seed/js/seed.js
@@ -941,6 +941,26 @@ SEED_app.config(['stateHelperProvider', '$urlRouterProvider', '$locationProvider
               });
             }
           }],
+          used_columns: ['$stateParams', 'inventory_service', function ($stateParams, inventory_service) {
+            var organization_id = $stateParams.organization_id;
+            if ($stateParams.inventory_type === 'properties') {
+              return inventory_service.get_property_columns_for_org(organization_id, true).then(function (columns) {
+                columns = _.reject(columns, 'related');
+                columns = _.map(columns, function (col) {
+                  return _.omit(col, ['pinnedLeft', 'related']);
+                });
+                return columns;
+              });
+            } else if ($stateParams.inventory_type === 'taxlots') {
+              return inventory_service.get_taxlot_columns_for_org(organization_id, true).then(function (columns) {
+                columns = _.reject(columns, 'related');
+                columns = _.map(columns, function (col) {
+                  return _.omit(col, ['pinnedLeft', 'related']);
+                });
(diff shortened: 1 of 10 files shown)


seed/views/properties.py
```diff
@@ -814,7 +816,7 @@ def columns(self, request):
               paramType: query
         """
         organization_id = int(request.query_params.get('organization_id'))
-        only_used = request.query_params.get('only_used', False)
+        only_used = json.loads(request.query_params.get('only_used', 'false'))
```
adrian-lara, line 819: Before, sending `"false"` to this endpoint actually ended up making `only_used` be a truthy, non-empty string. Note, this fix happened for the corresponding endpoint in v3, so no changes needed there.

seed/static/seed/js/seed.js
```diff
@@ -941,6 +941,26 @@ SEED_app.config(['stateHelperProvider', '$urlRouterProvider', '$locationProvider
               });
             }
           }],
+          used_columns: ['$stateParams', 'inventory_service', function ($stateParams, inventory_service) {
```
adrian-lara, line 944: `used_columns` is the same as `columns` just above this except the `get_[inv_type]_columns_for_org` methods take a `true` arg and sorting isn't performed.

axelstudios:
Updated screenshot with grouping:
image

RDmitchell:
@axelstudios / @adrian-lara -- wow, this looks great !! Thank you !!

RDmitchell:
Instance: dev1
SHA: fd75dea4
Org: LBNL Test 200

This is working as described in the PR, ie, grouped by Mapped and Not Mapped. I think this is a decent way to group the fields, but when there are a lot of fields that are mapped, it's hard to see which category a field that you are looking for belongs to.

But it's fine for now. 

image

Issue #2164, opened 2020-03-27
Update label get_is_applied to support OEP


nllong:
This method fails at the moment and is needed. Also, write a test!

https://github.com/SEED-platform/seed/blob/34fca03ccbed74f9b17641705c209422f2fa0b68/seed/serializers/labels.py#L50-L56

nllong:
I confirmed that the code is okay, the only issue is lack of test.

adrian-lara:
@nllong Agree with the test being needed. Also, I think it might make sense to not even have an `is_applied` key if the `self.inventory` is falsy  

In these cases, filtering isn't desired, so it's misleading to see `{... is_applied: []}` - seemingly displaying that no record IDs are returned.

I'll write the test and remove the `is_applied` key entirely when filtering isn't resulted then tag you in the PR since I don't expect it would be difficult to implement.

adrian-lara:
@RDmitchell Can you verify that applying and filtering labels on the inventory list page still works for properties and tax lots? I did some manual testing myself when I initially submitted this #2299 and this other PR, #2291 , but it's definitely worth your second pass.

RDmitchell:
@adrian-lara -- this is on dev1, which is the release from yesterday?

RDmitchell:
@adrian-lara -- just doing a few quick tests on dev1, seems like labels and filters are both working on both tax lot and property tabs in the Inventory List view.

PR #2299, opened 2020-07-16
2164 label is applied

adrian-lara:
For label read endpoints, we were missing some tests to ensure a vital aspect of these endpoints worked. That is, the addition of an "is_applied" key to serialized label records.

As a followup point, it was decided that the "is_applied" key should only really be present when the search for records actually happened. When the search did not happen, SEED would still provide the key without any record IDs, giving the impression that a search did actually happen, but no record IDs were found.

#### What's this PR do?
Add testing for the is_applied field.

Also, make updates to the serializer to remove the "is_applied" key when the search for is_applied records did not occur. 

Ensure that the API v3 label read endpoints still work on the Swagger page.

Test that the inventory list page still can fetch labels that are applied to records, and the label admin page still works.

#2164

### DIFF
--- a/seed/serializers/labels.py
+++ b/seed/serializers/labels.py
@@ -47,6 +47,15 @@ class LabelSerializer(serializers.ModelSerializer):
         }
         model = Label
 
+    def to_representation(self, instance):
+        ret = super().to_representation(instance)
+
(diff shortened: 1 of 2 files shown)


nllong:
@macintoshpie after your review, feel free to merge and move the linked issue to ready to deploy. Thanks!

Issue #2309, opened 2020-07-20
Frontend uses API v3 - ColumnMappings


adrian-lara:
For each v3 endpoint listed below,
- [x] check for recent changes in the <= v2.1 version and update v3 endpoint if necessary.
- [x] use it to replace the <= v2.1 version within tests. (deprecate tests)
- [x] use it to replace the <= v2.1 version on the frontend. This may be as straightforward as replacing "v2" with "v3" in the `[resource]_service.js`, or it can be as involved as refactoring the frontend code receiving the response. (deprecate logic)

Once that's complete, 
- [x] verify that the relevant pages still work as expected - these being the pages where the updated endpoints are used.
- [x] update this issue with items to test from the user perspective

### Endpoints 
NOTE: Each is listed with the change made in order to create the v3 version.

- **/api/v2/column_mappings/ - get**: - deprecate - remove from the frontend
- **/api/v2/column_mappings/{id}/ - delete**: - deprecate - remove from the frontend
- **/api/v2/column_mappings/delete_all/ - post**: - deprecate - remove from the frontend

### Manual Testing
- [x] check that the column mapping tab works as expected

PR #2328, opened 2020-07-22
2309 apiv3 frontend column mappings

aviveiros11:
Part of the effort to convert the frontend to use API v3

#### What's this PR do?
See attached issue.

Verify checks in attached issue are warranted.

Also check the following:
- [ ] Org column mappings page should work as before

#2309


adrian-lara:
Flagging as ready for review since we released 2.7.2

PR #2337, opened 2020-07-27
Refactor/building files transition

macintoshpie:
Refactor FE and tests for v3 building_files
#### What's this PR do?
- removes the sdUploader BuildingSync source type (no longer used)
- updates tests for v3 api
No manual tests are needed. The existing FE didn't use the `v2/building_file` endpoint already, so nothing has changed.
#2317


seed/static/seed/js/directives/sdUploader.js
```diff
@@ -161,146 +161,6 @@ var makeFileSystemUploader = function (scope, element, allowed_extensions) {
   return uploader;
 };
 
-var makeBuildingSyncUploader = function (scope, element, allowed_extensions) {
```
adrian-lara, line 164: Nice! We remove this because BuildingSync files now use the "normal" upload flow right?

macintoshpie: That's correct 👍

Issue #2313, opened 2020-07-20
Frontend uses API v3 - Meters


adrian-lara:
For each v3 endpoint listed below,
- [x] check for recent changes in the <= v2.1 version and update v3 endpoint if necessary.
- [x] use it to replace the <= v2.1 version within tests.
- [x] use it to replace the <= v2.1 version on the frontend. This may be as straightforward as replacing "v2" with "v3" in the `[resource]_service.js`, or it can be as involved as refactoring the frontend code receiving the response.

Once that's complete, 
- [x] verify that the relevant pages still work as expected - these being the pages where the updated endpoints are used.
- [x] update this issue with items to test from the user perspective

### Endpoints 
NOTE: Each is listed with the change made in order to create the v3 version.

- **/api/v2/meters/greenbutton_parsed_meters_confirmation/ - post**: - move to V3 ImportFileViewSet greenbutton_meters_preview and GET
- **/api/v2/meters/parsed_meters_confirmation/ - post**: - move to V3 ImportFileViewSet to pm_meters_preview and GET
- **/api/v2/meters/property_meter_usage/ - post**: - move to /properties/{id}/meter_usage. Keep as POST for array.
- **/api/v2/meters/property_meters/ - post**: - move to /properties/{id}/meters and convert to GET
- **/api/v2/meters/valid_types_units/ - get**: - move - add comment about using "US" conversion factors and how these are the same as CAN conversion factors so it doesn't matter

### Manual Testing
`v3/import_files/<id>/pm_meters_preview`: 
- upload example-pm-properties.xlsx (make sure PM ID is set)
- upload example-pm-monthly-meter-usage.xlsx
- you should see a "preview" of the meters to be imported before confirming the import

`v3/import_files/<id>/greenbutton_meters_preview`:
- go to the meters page for a property
- click 'Upload Greenbutton meters', choose example-GreenButton-data.xml
- you should see a "preview" of the meters to be imported before confirming the import

`v3/properties/<id>/meters`:
- upload any kind of meters for a property
- navigate to the meters page of inventory detail and you should see the meter data there.

`v3/properties/<id>/meter_usage`:
- upload any kind of meters for a property
- navigate to the meters page of inventory detail and you should see the meter data there

PR #2339, opened 2020-07-28
Refactor/v3 meters transition

macintoshpie:
Part of task of moving to api v3 for tests and FE
#### What's this PR do?
Updates tests and FE for v3 endpoints that were originally part of v2 meters views
`v3/import_files/<id>/pm_meters_preview`: 
- upload example-pm-properties.xlsx (make sure PM ID is set)
- upload example-pm-monthly-meter-usage.xlsx
- you should see a "preview" of the meters to be imported before confirming the import

`v3/import_files/<id>/greenbutton_meters_preview`:
- go to the meters page for a property
- click 'Upload Greenbutton meters', choose example-GreenButton-data.xml
- you should see a "preview" of the meters to be imported before confirming the import

`v3/properties/<id>/meters`:
- upload any kind of meters for a property
- navigate to the meters page of inventory detail and you should see the meter data there.

`v3/properties/<id>/meter_usage`:
- upload any kind of meters for a property
- navigate to the meters page of inventory detail and you should see the meter data there

#2313


seed/static/seed/js/services/meter_service.js
```diff
@@ -4,10 +4,11 @@ angular.module('BE.seed.service.meter', [])
     function ($http) {
       var meter_factory = {};
 
-      meter_factory.get_meters = function (property_view_id) {
```
adrian-lara, line 7: Nice - I could see arguments for moving these to properties, but I think it's better to have a meters service. Otherwise the inventory service would basically have everything in there.

seed/static/seed/js/services/uploader_service.js
```diff
@@ -111,5 +111,24 @@ angular.module('BE.seed.service.uploader', []).factory('uploader_service', [
       }, failure_fn);
     };
 
+    uploader_factory.parsed_meters_confirmation = function (file_id, org_id) {
```
adrian-lara: 💯  good choice for moving this function

Issue #2318, opened 2020-07-20
Frontend uses API v3 - Upload and Progress


adrian-lara:
For each v3 endpoint listed below,
- [ ] check for recent changes in the <= v2.1 version and update v3 endpoint if necessary.
- [ ] use it to replace the <= v2.1 version within tests.
- [ ] use it to replace the <= v2.1 version on the frontend. This may be as straightforward as replacing "v2" with "v3" in the `[resource]_service.js`, or it can be as involved as refactoring the frontend code receiving the response.

Once that's complete, 
- [x] verify that the relevant pages still work as expected - these being the pages where the updated endpoints are used.
- [x] update this issue with items to test from the user perspective

### Endpoints 
NOTE: Each is listed with the change made in order to create the v3 version.

- **/api/v2/upload/ - post**: - move
- **/api/v2/upload/create_from_pm_import/ - post**: - move
- **/api/v2/progress/{id}/ - get**: - move

### Manual Testing
- [x] Make sure `file upload` progress bar behavior remains the same
- [x] Make sure `mapping` progress bar behavior remains the same
- [x] Make sure `confirm mapping & start matching` progress bar behavior remains the same

PR #2342, opened 2020-07-29
2318 apiv3 frontend upload progress

aviveiros11:
Part of the effort to convert the frontend to use API v3

#### What's this PR do?
See attached issue.

Verify checks in attached issue are warranted.

Also check the following:
- [x] Make sure `file upload` progress bar behavior remains the same
- [x] Make sure `mapping` progress bar behavior remains the same
- [x] Make sure `confirm mapping & start matching` progress bar behavior remains the same

#2318


seed/views/v3/uploads.py
```diff
@@ -59,7 +58,7 @@ class UploadViewSet(viewsets.ViewSet):
             ),
             AutoSchemaHelper.form_string_field(
                 name='source_type',
-                required=True,
+                required=False,
```
adrian-lara, line 61: I would've thought that source_type should still be required? Though admittedly, I don't know why we're handling the "missing" case when treating the request as POST by checking for source_type when treating the request as a GET.
`source_type = request.POST.get('source_type', request.GET.get('source_type'))`

seed/views/v3/uploads.py
```diff
@@ -46,7 +46,6 @@ class UploadViewSet(viewsets.ViewSet):
     @swagger_auto_schema(
         request_body=no_body,
         manual_parameters=[
-            AutoSchemaHelper.query_org_id_field(),
```
adrian-lara, line 49: I think we want to undo these changes and update the frontend to pass in an org_id for this request.

seed/static/seed/js/directives/sdUploader.js
```diff
@@ -17,10 +17,11 @@
  */
 
 var makeFileSystemUploader = function (scope, element, allowed_extensions) {
+  console.log(scope);
```
adrian-lara, line 20: 😬 Note to remove.

seed/views/v3/uploads.py
```diff
@@ -103,12 +103,11 @@ def create(self, request):
         with open(path, 'wb+') as temp_file:
             for chunk in the_file.chunks():
                 temp_file.write(chunk)
-        org_id = request.query_params.get('organization_id', None)
+
         import_record_pk = request.POST.get('import_record', request.GET.get('import_record'))
         try:
             record = ImportRecord.objects.get(
                 pk=import_record_pk,
-                super_organization_id=org_id
```
adrian-lara, line 111: We should undo these code changes too.

adrian-lara:
For some reason, I'm having issues Updating With BuildingSync. Can you try to reproduce on the inventory detail page? I have a feeling it might just be user error 😬

aviveiros11:
@adrian-lara I think this is an unrelated bug? I was able to reproduce this on both my feature branch as well as develop.  It looks like when updating with building sync on the inventory detail page it actually hits `http://0.0.0.0:8000/api/v2.1/properties/2122/update_with_building_sync/?cycle_id=29&organization_id=12` instead of the upload path.

PR #2346, opened 2020-07-31
fix(validation_client): fix default use case name

macintoshpie:
A recent update to the BSync selection tool broke our client, specifically some use case names were appended with "v2.0.0"
#### What's this PR do?
Fixes the name of the use case we're looking for
upload a buildingsync file (e.g. ex1.xml). It should run validation and return with warnings/errors from use case tool.


macintoshpie:
Name was fixed by selection-tool

Issue #2302, opened 2020-07-20
Frontend uses API v3 - Organizations


adrian-lara:
For each v3 endpoint listed below,
- [ ] check for recent changes in the <= v2.1 version and update v3 endpoint if necessary.
- [ ] use it to replace the <= v2.1 version within tests
- [ ] use it to replace the <= v2.1 version on the frontend. This may be as straightforward as replacing "v2" with "v3" in the `[resource]_service.js`, or it can be as involved as refactoring the frontend code receiving the response.

Once that's complete, 
- [ ] verify that the relevant pages still work as expected - these being the pages where the updated endpoints are used.
- [ ] update this issue with items to test from the user perspective

### Endpoints 
NOTE: Each is listed with the change made in order to create the v3 version.

- **/api/v1/delete_organization_inventory/ - delete**: - move to organizations/id/inventory with DELETE
- **/api/v2/organizations/ - get**: - move
- **/api/v2/organizations/ - post**: - move
- **/api/v2/organizations/{id}/ - delete**: - move
- **/api/v2/organizations/{id}/ - get**: - move
- **/api/v2/organizations/{id}/add_user/ - put**: - move /organizations/{id}/users/{id}/add
- **/api/v2/organizations/{id}/geocoding_columns/ - get**: - move
- **/api/v2/organizations/{id}/match_merge_link_preview/ - post**: - move
- **/api/v2/organizations/{id}/match_merge_link_result/ - get**: - move
- **/api/v2/organizations/{id}/match_merge_link/ - post**: - move
- **/api/v2/organizations/{id}/matching_criteria_columns/ - get**: - move
- **/api/v2/organizations/{id}/query_threshold/ - get**: - move
- **/api/v2/organizations/{id}/remove_user/ - delete**: - move /organizations/{id}/users/{id}/remove
- **/api/v2/organizations/{id}/save_settings/ - put**: - move
- **/api/v2/organizations/{id}/shared_fields/ - get**: - move
- **/api/v2/organizations/{id}/sub_org/ - post**: - move
- **/api/v2/organizations/{id}/users/ - get**: - move

### Manual Testing
`GET /api/v3/organizations/`
- Navigate to `/app/#/profile/admin`, this will use the api to load all orgs

`POST /api/v3/organizations/`
- Navigate to `/app/#/profile/admin` . create org

`GET /api/v3/organizations/{orgid}/users/`
- Navigate to the orgs member page.

[...]

PR #2351, opened 2020-08-03
2302 refactor/v3 organizations transition

macintoshpie:
Part of effort moving FE and tests to v3 api
#### What's this PR do?
Moves FE and tests to v3 organizations api
See issue
#2302


seed/tests/test_account_views.py
```diff
@@ -324,36 +318,26 @@ def test_remove_user_from_org_missing_org_id(self):
...
-        self.assertDictEqual(
-            json.loads(resp.content),
-            {
-                'status': 'error',
-                'message': 'missing the user_id'
-            })
+        with self.assertRaises(NoReverseMatch):
+            resp = self.client.delete(
+                reverse_lazy('api:v3:organization-users-remove', args=[self.org.id]),
+            )
```
adrian-lara: What do you think about removing these two tests entirely? My thought is that since we use the nested_router package to implement `/api/v3/organizations/{org_id}/users/{user_id}`, we don't necessarily need to test what happens when org_id or user_id isn't included.

adrian-lara: Per discussion, I'll go ahead and remove them

Issue #2344, opened 2020-07-29
Non-default measurement units for unit-aware columns are ignored during mapping.


adrian-lara:
### Expected Behavior
When importing records and mapping to unit-aware columns, specifying the incoming data with non-default units should trigger conversions of the incoming values to then be stored in the database.

### Actual Behavior
The specification of non-default units are ignored during mapping.

### Steps to Reproduce
On the organization settings page, use default display units for both area (sq. ft) and EUI (kbtu/sq. ft/year).
Import a file and map to either an area or EUI field while using a non-default unit.
Complete the import then view the recently imported records' values for the non-default unit.
See that the values have remained the same - the conversions never occurred. 

For example, after importing a record with a Site EUI of 100 being specified as GJ/m^2/year, viewing that value as kbtu/sq. ft/year, we should not see a value of 100 Site EUI as the conversion from GJ/m^2/year and kbtu/sq. ft/year is not 1:1.

### Fix
The fix will likely include performing the conversion when creating the raw, PropertyState-extra_data-only version of the records. These are the ones that are displayed on the mapping preview page, and these are used to eventually build the PropertyStates and TaxLotStates associated to Property- and TaxLotViews.

### Instance Information
Server Instance: `dev1`, `production`, `develop`

RDmitchell:
instance: dev1
SHA: 38fb067
Org: LBNL Test 200
Cycle: 2015 Compliance Cycle

It doesn't seem like the conversion is happening correctly, but maybe I am doing something wrong? 

The default settings are IP units
![image](https://user-images.githubusercontent.com/6314950/94327804-32a87480-ff62-11ea-8484-6976372c897c.png)

On initial mapping, units are set to defaults
![image](https://user-images.githubusercontent.com/6314950/94327813-47850800-ff62-11ea-9323-2fe441847318.png)

I change them to metric units in mapping
![image](https://user-images.githubusercontent.com/6314950/94327827-5b306e80-ff62-11ea-8a4c-a2fa08b37e2d.png)

After import, looking at the inventory view, the Gross Floor Area values have changed, but is the conversion correct?
![image](https://user-images.githubusercontent.com/6314950/94327842-7f8c4b00-ff62-11ea-9032-b1d346c7b3ff.png)

Should Custom ID 100, which had a Gross Floor Area of 12,555 sf be converted to a value of 135,140.90 ? This doesn't seem correct. 

Also, we don't show units in the headers, which seems very bad. 

[...]

PR #2355, opened 2020-08-04
Mapping on import takes into account selected units

adrian-lara:
During import, mapping with non-default units was ignored.

#### What's this PR do?
Fixes that problem by creating "raw" Column objects that have a `units_pint` value (previously, it was always `None`) during mapping save. It turns out that this is enough of a fix since, during `perform_mapping`, "cleaners" are built using those Column objects, and subsequently, they use the `units_pint` values for conversions.

Prior, with `units_pint` being None, incoming values defaulted to the base units of the respective fields.

Import records; map to unit-aware columns; choose non-default units; finally, see that the values have been converted according to your selection on the mapping confirmation page.

#2344


seed/models/columns.py
```diff
@@ -985,11 +985,12 @@ def select_col_obj(column_name, table_name, organization_column):
             try:
                 # the from column is the field in the import file, thus the table_name needs to be
                 # blank. Eventually need to handle passing in import_file_id
-                from_org_col, _ = Column.objects.get_or_create(
+                from_org_col, _ = Column.objects.update_or_create(
                     organization=organization,
-                    table_name__in=[None, ''],
+                    table_name='',
```
adrian-lara, line 990: This has been updated here and below for two reasons.
- There exists a constraint that doesn't allow `table_name` to be None (null in PG). 
- it seems that `update_or_create` doesn't take `__in`.

seed/models/columns.py
```diff
@@ -985,11 +985,12 @@ def select_col_obj(column_name, table_name, organization_column):
             try:
                 # the from column is the field in the import file, thus the table_name needs to be
                 # blank. Eventually need to handle passing in import_file_id
-                from_org_col, _ = Column.objects.get_or_create(
+                from_org_col, _ = Column.objects.update_or_create(
```
adrian-lara, line 988: The intent here is to always use the given "from_units" to populate or update `units_pint`.

seed/models/columns.py
```diff
@@ -1010,7 +1011,8 @@ def select_col_obj(column_name, table_name, organization_column):
 
                 from_org_col = Column.objects.create(
                     organization=organization,
-                    table_name__in=[None, ''],
+                    table_name='',
```
adrian-lara, line 1014: See comment above. And also, this would've thrown an error if it ever was hit since `create` can't take the `__in` kwarg.

Issue #2301, opened 2020-07-20
Frontend uses API v3 - Properties v2


adrian-lara:
For each v3 endpoint listed below,
- [ ] check for recent changes in the <= v2.1 version and update v3 endpoint if necessary.
- [ ] use it to replace the <= v2.1 version within tests
- [ ] use it to replace the <= v2.1 version on the frontend. This may be as straightforward as replacing "v2" with "v3" in the `[resource]_service.js`, or it can be as involved as refactoring the frontend code receiving the response.

Once that's complete, 
- [ ] verify that the relevant pages still work as expected - these being the pages where the updated endpoints are used.
- [ ] update this issue with items to test from the user perspective

### Endpoints 
NOTE: Each is listed with the change made in order to create the v3 version.

- **/api/v2/properties/ - get**: - move
- **/api/v2/properties/{id}/ - get**: - move
- **/api/v2/properties/{id}/ - put**: - move
- **/api/v2/properties/{id}/delete/ - delete**: - deprecate - redundant - property_views DELETE
- **/api/v2/properties/{id}/links/ - post**: - Should be a get
- **/api/v2/properties/{id}/match_merge_link/ - post**: - move
- **/api/v2/properties/{id}/pair/ - put**: - move
- **/api/v2/properties/{id}/unmerge/ - post**: - change to put
- **/api/v2/properties/{id}/unpair/ - put**: - move
- **/api/v2/properties/batch_delete/ - delete**: - refactor, replace body param `inventory_ids` with `property_view_ids`, an array of property view ids to delete
- **/api/v2/properties/columns/ - get**: - deprecate - redundant - api/v3/columns GET
- **/api/v2/properties/cycles/ - post**: - Rename to "filter_by_cycle"
- **/api/v2/properties/filter/ - post**: - refactor, rename `inventory_ids` body param to `property_view_ids`.
- **/api/v2/properties/mappable_columns/ - get**: - move to /api/v3/columns/mappable
- **/api/v2/properties/merge/ - post**: - refactor, replace `state_ids` with `property_view_ids`, an array of property view ids to merge
- **/api/v2/properties/meters_exist/ - post**: - move

### Manual Testing
`POST /api/v3/properties/filter/`
- Upload properties, then go to properties list page. Also used if you select some properties then choose the merge action.

`POST /api/v3/properties/filter_by_cycle/`
- Upload properties, then go to properties list page, then the cycles tab. Should be able to select different cycles to load data

[...]

PR #2356, opened 2020-08-04
2301 refactor/v3 properties transition

macintoshpie:
Part of effort moving FE and tests to v3 apis
#### What's this PR do?
Moves FE and tests to v3 properties apis
See issue
#2301


seed/static/seed/js/services/matching_service.js
```diff
@@ -27,10 +27,10 @@ angular.module('BE.seed.service.matching', []).factory('matching_service', [
       });
     };
 
-    matching_service.mergeProperties = function (state_ids) {
+    matching_service.mergeProperties = function (property_view_ids) {
       spinner_utility.show();
-      return $http.post('/api/v2/properties/merge/', {
-        state_ids: state_ids
+      return $http.post('/api/v3/properties/merge/', {
+        property_view_ids
```
adrian-lara, line 33: I saw this with tax lots and was surprised this worked. We don't have to define the key for the `property_view_ids` value?

seed/tests/test_property_views.py
```diff
@@ -947,8 +947,9 @@ def test_properties_merge_combining_bsync_and_pm_sources(self):
             filename=bs_filename,
             file_type=BuildingFile.BUILDINGSYNC,
         )
```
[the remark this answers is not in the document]
macintoshpie: I think this actually goes back to a bug we found a while back (February maybe?). We started to fix this but then decided not to, I can't remember why exactly. I think maybe we decided it was too much of an edge case.

seed/views/v3/properties.py
```diff
@@ -892,11 +892,11 @@ def batch_delete(self, request):
...
 
         property_view_ids = request.data.get('property_view_ids', [])
-        property_states = PropertyView.objects.filter(
+        property_state_ids = PropertyView.objects.filter(
             id__in=property_view_ids,
             cycle__organization_id=org_id
-        )
-        resp = PropertyState.objects.filter(pk__in=Subquery(property_states.values('id'))).delete()
+        ).values_list('state_id', flat=True)
+        resp = PropertyState.objects.filter(pk__in=Subquery(property_state_ids)).delete()
```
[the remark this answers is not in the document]
macintoshpie: I think the original could have worked (using .values inside of subquery), but it still is not written correctly. We were originally grabbing property _views_ (as `property_states` var) then looking for property _states_ whose id match the view ids. I fixed it and changed it to use values_list to be consistent with the equivalent taxlot view. Both values_list and values return querysets so they are both lazily evaluated, the difference is the structure of the data (dict vs tuples)

seed/static/seed/js/services/matching_service.js
macintoshpie: I spent some time looking at the data, and though I don't totally understand where it comes from it seems to always have it 🤷‍♂️

seed/tests/test_property_views.py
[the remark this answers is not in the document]
adrian-lara: Oh ok - that sounds familiar. Let's leave this skipped.

seed/views/v3/properties.py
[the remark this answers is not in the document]
adrian-lara: Ah ok - thanks for the clarification 👍

seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -326,8 +326,16 @@ angular.module('BE.seed.controller.inventory_list', [])
...
             has_meters: function () {
               if ($scope.inventory_type === 'properties') {
-                var inventory_ids = $scope.selectedOrder.slice().reverse();
-                return inventory_service.properties_meters_exist(inventory_ids).then(function (has_meters) {
+                const selected_property_view_ids = _.map(
+                  _.filter($scope.gridApi.selection.getSelectedRows(), function (row) {
+                    return row.$$treeLevel === 0;
+                  }),
+                  'property_view_id'
+                ).reverse();
```
macintoshpie, line 334: This was easier than refactoring $scope.selectedOrder, which is tied up with other uses. This is a pattern I basically copied from line 722

adrian-lara: I think this is definitely a reasonable approach. The amount of effort needed to refactor the grid selections to use view_ids is definitely non-negligible, and it might be up for debate as to whether it should happen in the first place.

seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -296,23 +296,30 @@ angular.module('BE.seed.controller.inventory_list', [])
...
                 });
               } else if ($scope.inventory_type === 'taxlots') {
-                return inventory_service.get_taxlots(1, undefined, undefined, -1, selectedOrder).then(function (inventory_data) {
-                  _.forEach(selectedOrder, function (id, index) {
-                    var match = _.find(inventory_data.results, {id: id});
+                return inventory_service.get_taxlots(1, undefined, undefined, -1, selectedViewIds).then(function (inventory_data) {
+                  _.forEach(selectedViewIds, function (id, index) {
+                    const matchPredicate = {}
+                    matchPredicate[viewIdProp] = id
+                    var match = _.find(inventory_data.results, matchPredicate);
```
adrian-lara, line 322: Works as is in both places, but it was a little confusing to read through on first pass.

We can definitely leave it as is, but something that might be a little more straightforward is to avoid creating the `matchPredicate` variable and insert the object directly into the `_.find()` method.  
```
var match = _.find(inventory_data.results, {viewIdProp: id});
```

Optionally, we could also make use of shorthand that comes with `_.find`. I could go either way on this one.
```
var match = _.find(inventory_data.results, [viewIdProp, id]);
```

Issue #2345, opened 2020-07-29
Hand-edit saves without changes results in a 500 being thrown


adrian-lara:
### Expected/Actual Behavior
Hand-edits during saves without changes should not yield a server error. Instead, we should the save should just work. This likely happens for both Properties and Tax Lots.

### Steps to Reproduce
On the inventory detail page, click "Edit" then immediately click "Save".

### Fix
The bug was introduced during the introduction of creating notes during hand edits (#1716).

### Instance Information
Server Instance: `develop`, `dev1` (not in `prod`)

adrian-lara:
Save button is now deactivated if no changes are found in the fields.

RDmitchell:
instance: dev1
SHA: 38fb06738
Org: LBNL Test 200
Cycle: 2015 Compliance Cycle

Right -- can't click the Save button if no changes have been made

PR #2359, opened 2020-08-04
2345 fix/manual edit saving

macintoshpie:
When editing inventory manually on details page, you could change nothing and still click save. Posting empty changes would make the server 500.
#### What's this PR do?
- Fixes 500 issue by returning correct number of values in utility func. Specifically, in `seed/utils/properties.py` function `get_changed_fields` was returning a single `None` if nothing was changed, while the caller (the update view) expected two returned values resulting in `TypeError: 'NoneType' object is not iterable`
- Fixes frontend issue of detecting changes when none really exist. Specifically, in `seed/static/seed/js/controllers/inventory_detail_controller.js`, the function `$scope.diff` was returning non-empty diff's when nothing was changed (it had props `files`, `measures`, and `scenarios` containing empty arrays). `$scope.modified` checks if the return value of `diff` is empty, which is used to disable the "Save" button when nothing has been edited.
- Upload taxlot/properties. click edit, you should not be able to click save until you change something. Click save and it should save properly
#2345


seed/static/seed/js/controllers/inventory_detail_controller.js
```diff
@@ -360,9 +360,10 @@ angular.module('BE.seed.controller.inventory_detail', [])
       $scope.diff = function () {
         if (_.isEmpty($scope.item_copy)) return {};
         // $scope.item_state, $scope.item_copy
+        const ignored_root_keys = ['extra_data', 'files', 'measures', 'scenarios']
         var result = {};
         _.forEach($scope.item_state, function (value, key) {
-          if (key === 'extra_data') return;
+          if (ignored_root_keys.includes(key)) return;
```
adrian-lara, line 366: Oh also, does this change anything practically speaking? or was this more of an enhancement (not building the key-values that we plan to just omit later)?

macintoshpie: This changes what data is returned in the diff. The only places this is used is for this functionality of updating states, so I don't think we have to worry about breaking anything.

PR #2360, opened 2020-08-05
2322/refactor/v3 taxlot notes transition

macintoshpie:
Part of effort moving FE and tests to v3 apis
#### What's this PR do?
Moves FE and tests to v3 notes api for properties and taxlots.
Also, fixes bug in `users_service`, where creating members from the org page was causing errors.
See issue
#2322 
#2323 (only addresses part of this issue)


seed/static/seed/js/controllers/new_member_modal_controller.js
```diff
@@ -36,7 +36,11 @@ angular.module('BE.seed.controller.new_member_modal', [])
        * adds a user to the org
        */
       $scope.submit_form = function () {
-        user_service.add($scope.user).then(function () {
+        // make `role` a string 
+        const u = _.cloneDeep($scope.user)
+        u.role = u.role.value
+
+        user_service.add(u).then(function () {
```
macintoshpie, line 43: @adrian-lara I found a bug in using the v3 api for user creation, this fixes that.

adrian-lara: User creation seems to work as expected. Just curious, what was the bug? and was it caused by v3 changes?

seed/static/seed/js/services/note_service.js
```diff
@@ -55,15 +55,15 @@ angular.module('BE.seed.service.note', []).factory('note_service', [
...
+      return $http.post('/api/v3/' + inventory_type + '/' + view_id + '/notes/', payload).then(function (response) {
         return response.data;
       });
     };
 
     note_factory.update_note = function (org_id, inventory_type, view_id, note_id, note_data) {
       var payload = note_data;
       payload.organization_id = org_id;
-      return $http.put('/api/v2.1/' + inventory_type + '/' + view_id + /notes/ + note_id + '/', payload).then(function (response) {
+      return $http.put('/api/v3/' + inventory_type + '/' + view_id + '/notes/' + note_id + '/', payload).then(function (response) {
```
adrian-lara, line 66: 🤷  ... surprised these worked without the quotes before

seed/static/seed/js/controllers/new_member_modal_controller.js
macintoshpie: The role on the FE is actually an object, which v2 used to handle by checking if it was a dict:
https://github.com/SEED-platform/seed/blob/f1d2347d146c9d4164752d8d699966f619464a96/seed/views/users.py#L218
This check was dropped in v3 api (which I think makes sense). So now we always expect role to just be a string. Creating users from the admin page would work previously b/c no `role` property was added to the payload.

seed/static/seed/js/services/note_service.js
macintoshpie: We got lucky haha
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/toString

seed/static/seed/js/controllers/new_member_modal_controller.js
adrian-lara: 👍 Thanks for the clarification!

Issue #2316, opened 2020-07-20
Frontend uses API v3 - Geocode


adrian-lara:
For each v3 endpoint listed below,
- [x] check for recent changes in the <= v2.1 version and update v3 endpoint if necessary.
- [x] use it to replace the <= v2.1 version within tests.
- [x] use it to replace the <= v2.1 version on the frontend. This may be as straightforward as replacing "v2" with "v3" in the `[resource]_service.js`, or it can be as involved as refactoring the frontend code receiving the response.

Once that's complete, 
- [x] verify that the relevant pages still work as expected - these being the pages where the updated endpoints are used.
- [x] update this issue with items to test from the user perspective

### Endpoints 
NOTE: Each is listed with the change made in order to create the v3 version.

- **/api/v2/geocode/api_key_exists/ - get**: - move to OrganizationViewSet and make it a detail view
- **/api/v2/geocode/confidence_summary/ - post**: - move. Keep as POST for array
- **/api/v2/geocode/geocode_by_ids/ - post**: - move. Keep as POST for array

### Manual Testing
`GET /api/v3/organizations/<id>/geocode_api_key_exists/`
- go to mapping page, should call on load

`POST /api/v3/geocode/confidence_summary/`
- on inventory list page, select a property, choose the Geocode Selected action (note you need to set the MapQuest API key in the org settings). You should be able to rerun geocoding and view confidence summary.

`POST /api/v3/geocode/geocode_by_ids/`
- on inventory list page, select a property that's _not_ geocoded, choose the Geocode Selected action (note you need to set the MapQuest API key in the org settings). You should be able to run geocoding successfully to update the property.

PR #2363, opened 2020-08-07
2316 API V3 Geocode View

lainsworth8801:
#### What's this PR do?
Updated endpoint version -> v3;
monkey test new endpoints;
Modified seed tests accordingly.

1. Copy mapquest api key in Organization -> Settings -> Mapquest API Key;
2. In inventory list, select properties or taxlots desired for geocoding:
Screen Shot 2020-08-26 at 4 26 08 PM
3. Geocode selected properties:
Screen Shot 2020-08-26 at 4 26 19 PM
4. Geocoded results:
Screen Shot 2020-08-26 at 4 26 25 PM
5. Same for taxlots:
Screen Shot 2020-08-26 at 4 27 52 PM

#2316


seed/tests/test_geocode_views.py
```diff
@@ -45,13 +49,14 @@ def test_geocode_endpoint_base_with_prepopulated_lat_long_no_api_request(self):
...
         property.save()
 
-        url = reverse('api:v2:geocode-geocode-by-ids')
+        self.property_view_factory.get_property_view(state=property)
+
         post_params = {
-            'organization_id': self.org.pk,
-            'property_ids': [property.id],
-            'taxlot_ids': []
+            'property_view_ids': [property.id],
```
adrian-lara, line 55: I think we want the -View ID from the PropertyView created in L52.

seed/tests/test_geocode_views.py
```diff
@@ -64,70 +69,90 @@ def test_geocode_confidence_summary_returns_summary_dictionary(self):
...
-            'property_ids': [
+            'property_view_ids': [
                 property_none.id,
                 property_high.id,
                 property_low.id,
                 property_manual.id,
                 property_missing.id
             ],
-            'tax_lot_ids': [
+            'taxlot_view_ids': [
```
adrian-lara, line 155: Similar to the other comment, I think we want the -View IDs for the -View objects created in the new lines above.

seed/views/v3/geocode.py
```diff
@@ -123,7 +123,7 @@ def confidence_summary(self, request):
 
         if taxlot_view_ids:
             taxlot_views = TaxLotView.objects.filter(
-                id__in=taxlot_view_ids,
+                state_id__in=taxlot_view_ids,
```
adrian-lara, line 126: Let's undo these changes. We don't want to make changes to v3 to make the tests pass. Instead, let's make more changes to the tests to accommodate how the API endpoints were changed from v2 to v3.

seed/static/seed/js/services/geocode_service.js
```diff
@@ -29,7 +29,7 @@ angular.module('BE.seed.service.geocode', [])
 
       geocode_factory.check_org_has_api_key = function (org_id) {
         var params = {organization_id: org_id};
-        return $http.get('/api/v2/geocode/api_key_exists', {
+        return $http.get('/api/v3/organizations/' + org_id + '/api_key_exists', {
```
macintoshpie, line 32: Getting a 404 with this service (tested loading the mapping page). I think we need to change this to `'/api/v3/organizations/' + org_id + '/geocode_api_key_exists/'`

seed/static/seed/js/services/geocode_service.js
```diff
@@ -19,7 +19,7 @@ angular.module('BE.seed.service.geocode', [])
       };
 
       geocode_factory.confidence_summary = function (property_state_ids, taxlot_state_ids) {
-        return $http.post('/api/v2/geocode/confidence_summary/', {
+        return $http.post('/api/v3/geocode/confidence_summary/', {
```
macintoshpie, line 22: Looks like we're missing the org query param here, which is used for auth. We should be able to add just like the `geocode_by_ids` call above

seed/static/seed/js/services/geocode_service.js
```diff
@@ -19,7 +19,7 @@ angular.module('BE.seed.service.geocode', [])
...
           property_ids: property_state_ids,
           tax_lot_ids: taxlot_state_ids
         }).then(function (response) {
```
macintoshpie, line 25: We changed the request payload to accept view ids rather than state ids to make the APIs more consistent. Happy to help with this if you have any questions as it requires a few changes on the FE

seed/static/seed/js/services/geocode_service.js
```diff
@@ -6,7 +6,7 @@ angular.module('BE.seed.service.geocode', [])
       var geocode_factory = {};
 
       geocode_factory.geocode_by_ids = function (property_state_ids, taxlot_state_ids) {
-        return $http.post('/api/v2/geocode/geocode_by_ids/', {
+        return $http.post('/api/v3/geocode/geocode_by_ids/', {
           property_ids: property_state_ids,
           taxlot_ids: taxlot_state_ids
         }, {
```
macintoshpie, line 12: We changed the request payload to accept view ids rather than state ids to make the APIs more consistent. Happy to help with this if you have any questions as it requires a few changes on the FE

seed/static/seed/js/services/geocode_service.js
lainsworth8801: So regarding this...I think I just changed all the "v2" in this .js file where in this case I wasn't working directly on check_org_has_api_key. Would you rather me change it back to v2 in case someone else is working the ticket related to this call?

seed/static/seed/js/services/geocode_service.js
macintoshpie: Looks like we still need to add the organization_id query param in this request. I'm testing this by going to the inventory list page, selecting a property, then clicking "Geocode selected" and checking the result in the network tab of devtools.

PR #2373, opened 2020-08-18
2308 dq v3 transition

adrian-lara:


adrian-lara:
Generally, I could go either way with including general DQ admin page improvements here or throwing them in a separate PR:

> * no check that min < max
There's some logic to automatically swaps the values in each if min < max. I don't love this, but not sure if this was a requested feature. Leaving as is but will maybe keep it in mind for potential changes.

> * user can add units to some fields where it doesn't make sense (data type date, year)
Existed before, but open to changing here. Will add to list for potential changes later at the very least.

> * user can choose a data type when the condition check is "not null"
I can see why this is weird, but given you can create multiple rules for the same field and mix "not null" condition with others, I think this is reasonable to keep as is.

> * might be nice to have all alerts in the same place, some are shown at top of the table while others are notifications
Agree - I'm leaning towards alert html errors (vs the notification/top right errors), but I'll tag Alex to get his opinion when I take care of some of the other points you brought up.

> * crazy edge case, but if someone creates a rule, then disables it, then saves the rule doesn't seem to be saved
👍  (Also in testing this one, a rule doesn't seem to be acknowledged for saving until it's associated to a field.)

> * create new rule, select "not null" (apparently this also happens when you select other conditions it seems), select "city" as field, then click save and server 400's without specific error other than `There was an error loading the page`. Request payload:
> 
> Response:
> 
> ```json
> {"data_type":["\"string\" is not a valid choice."]}
> ```
> 
> * might be nice to disable export of dq errors button if there are none
Weird... I can replicate 👍 I'll take a look.

RDmitchell:
@adrian-lara -- seems like those DQ issues should be put into a new issue ?? They all seem like they are worth fixing at some rate (not emergencies)

PR #2381, opened 2020-08-25
500 errors for /api/v2/properties/columns/

axelstudios:
Leaving out the `organization_id` parameter or providing an invalid one would result in a 500 error for `/api/v2/properties/columns/`

#### What's this PR do?
Returns proper 400 and 404 errors

Test the following URLs with valid and invalid org ids, make sure json is always returned:
/api/v2/properties/columns/?only_used=false&organization_id=1
/api/v2/properties/columns/?only_used=false&organization_id=9999
/api/v2/properties/columns/?only_used=false


axelstudios:
I would've preferred to use OrgMixin.getOrganization, but I don't think that works with PropertyViewSets

Issue #2384, opened 2020-08-27
PM import record of one property returns 400 status


lainsworth8801:
### Expected Behavior
A ESPM template with only one property should be returned successfully.

### Actual Behavior
A ESPM template containing 1 property returned as a dict object results in a 400 status as portfolio_manager.py expects a list instance of possible multiple properties.

### Steps to Reproduce
1. Create a ESPM template with 1 property;
2. Import this template from SEED via import portfolio manager data.

### Instance Information
<img width="651" alt="Screen Shot 2020-08-26 at 9 10 28 PM" src="https://user-images.githubusercontent.com/49968203/91379779-abd86e80-e7e0-11ea-8f70-df3952d36e54.png">

Server Instance: dev1

Server SHA: `abcd123`

RDmitchell:
instance: dev1
SHA: 38fb067
Org: LBNL Test 200

Imported spreadsheet with one record, no problems.

PR #2385, opened 2020-08-29
2384 espm import single property template

lainsworth8801:
When an espm template has only one property, it is returned as a dictionary object as SEED expected a list of records. Import espm with single property failed to proceed to data mapping.

#### What's this PR do?
Append single property espm template/record to a list object for upload.

Create an espm with 1 property. Import a espm template with one property and expect it to be successfully imported for mapping.

#2384 

1. Create an espm with 1 property:
Screen Shot 2020-08-29 at 3 11 37 PM

2. From SEED, import espm:
Screen Shot 2020-08-29 at 3 08 48 PM

3. Expect to see success:
Screen Shot 2020-08-29 at 3 10 54 PM

### DIFF
--- a/seed/building_sync/tests/test_buildingsync_views.py
+++ b/seed/building_sync/tests/test_buildingsync_views.py
@@ -75,12 +75,10 @@ class InventoryViewTests(DeleteModelsTestCase):
     def test_upload_and_get_building_sync(self):
         filename = path.join(path.dirname(__file__), 'data', 'ex_1.xml')
 
-        url = reverse('api:v2:building_file-list')
+        url = reverse('api:v3:building_files-list') + f'?organization_id={self.org.id}&cycle_id={self.cycle.id}'
         fsysparams = {
             'file': open(filename, 'rb'),
             'file_type': 'BuildingSync',
-            'organization_id': self.org.id,
-            'cycle_id': self.cycle.id
         }
 
         response = self.client.post(url, fsysparams)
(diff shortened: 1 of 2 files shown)


seed/views/v3/portfolio_manager.py
```diff
@@ -182,7 +182,7 @@ def report(self, request):
                     status=status.HTTP_400_BAD_REQUEST)
             try:
                 possible_properties = content_object['report']['informationAndMetrics']['row']
-                if isinstance(possible_properties, list):
+                if isinstance(possible_properties, list) or isinstance(possible_properties, dict):
                     properties = possible_properties
```
macintoshpie, line 186: Did you try something like this?
```
if it's a list:
  properties = possible_properties
else if it's a dict:
  properties = [possible_properties]
```
That might be nice so that down the line (ie in the view create_from_pm_import) we don't have to handle dicts and it would the api simpler

lainsworth8801:
> Awesome, it looks like this is working. Could we add a test to test_portfoliomanager.py? I'm thinking you could go into the SEED test portfolio manager account, create a template with only one property, then use that template to verify we can get a list of a single property (rather than a dict of a single property). If you need the username/password for creating the template in the test account let me know

Definitely and I think I should. Thanks for the reminder!

seed/views/v3/portfolio_manager.py
lainsworth8801: That's genius thanks!

PR #2389, opened 2020-09-04
ESPM single property report for upload and test

lainsworth8801:
When an espm template has only one property, it is returned as a dictionary object as SEED expected a list of records. Import espm with single property failed to proceed to data mapping.

#### What's this PR do?
Append single property espm template/record to a list object for upload.

Create an espm with 1 property. Import a espm template with one property and expect it to be successfully imported for mapping.

#2384 

Create an espm with 1 property:
image

From SEED, import espm:
Screen Shot 2020-07-29 at 8 11 46 PM

Expect to see report successfully added:
image

### DIFF
--- a/seed/tests/test_portfoliomanager.py
+++ b/seed/tests/test_portfoliomanager.py
@@ -365,3 +365,66 @@ class PortfolioManagerReportGenerationViewTestsSuccess(TestCase):
         # in this case, we expect a meaningful error message
         self.assertIn('message', body)
         self.assertIn('empty', body['message'])
+
+
+class PortfolioManagerReportSinglePropertyUploadTest(TestCase):
+
+    def setUp(self):
+        user_details = {
+            'username': 'test_user@demo.com',
+            'password': 'test_pass',
+        }
+        self.user = User.objects.create_superuser(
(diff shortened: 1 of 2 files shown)


seed/tests/test_portfoliomanager.py
```diff
@@ -365,3 +365,66 @@ def test_report_generation_empty_child_template(self):
...
+        )
+        dataset = response.json()
+        self.dataset_id = dataset['id']
+
+        self.pm_un = os.environ.get(PM_UN, False)
+        self.pm_pw = os.environ.get(PM_PW, False)
+        if not self.pm_un or not self.pm_pw:
+            self.fail('Somehow PM test was initiated without %s or %s in the environment' % (PM_UN, PM_PW))
+
+    def test_single_property_template_for_upload(self):
```
macintoshpie, line 398: I think we need to add the `pm_skip_test_check` decorator here like the other PM tests to make sure it's skipped if we're missing the env vars or if PM is unavailable

macintoshpie (changes requested):
Looks good! One small change to make

macintoshpie (approved):
Awesome, looks good! 👍

Issue #2374, opened 2020-08-19
Import spreadsheets without a raw header results in an error


adrian-lara:
### Expected Behavior
TBD: it's an open question regarding how this should be handled in SEED, if at all.

An easy approach is to _not_ handle it, and prevent users from continuing with an error message on the mapping page indicating that the file must have a header.

If we do want to be able to handle files with missing headers, more brainstorming is needed. 
For handling files like this:
- ~One approach we can take is to just take the mapped column name and use it as the raw column name. Downstream impacts will need to be considered further, but my initial thinking is this wouldn't cause any breaking issues. Further consideration might also be needed for mapping profiles - i.e consider what happens when a user tries to save a preset with a missing raw header.~ This wouldn't work as the raw PropertyState objects holding everything in extra_data would not have keys for these values.
- Could there be filler header names we could use in these cases?

### Actual Behavior
When importing a file, if one of the incoming raw headers is blank,
![Screen Shot 2020-08-04 at 11 19 03 AM](https://user-images.githubusercontent.com/30608004/90664320-0c2b4700-e208-11ea-80f8-46f7b8d734be.png)

SEED will throw an error.
```
DoesNotExist: Column matching query does not exist.
  File "django/db/models/query.py", line 464, in get_or_create
    return self.get(**lookup), False
  File "django/db/models/query.py", line 380, in get
    self.model._meta.object_name
 
ValidationError: {'column_name': ['This field cannot be blank.']}
(19 additional frame(s) were not displayed)
...
  File "seed/lib/superperms/orgs/decorators.py", line 244, in _wrapped
    return fn(self, request, *args, **kwargs)
  File "seed/data_importer/views.py", line 1043, in save_column_mappings
    result = Column.create_mappings(mappings, organization, request.user, import_file.id)
  File "seed/models/columns.py", line 814, in create_mappings
    mappings = Column._column_fields_to_columns(mappings, organization)
  File "seed/models/columns.py", line 982, in _column_fields_to_columns
    is_extra_data=False  # data from header rows in the files are NEVER extra data
  File "seed/models/columns.py", line 1397, in validate_model
    instance.full_clean()
```

### Instance Information
all

RDmitchell:
@adrian-lara -- I think we should put up a message to the user that they can't have blank headers, and that they need to add a header and reimport the file.

Somehow I thought this error checking already existed, but I guess not.

adrian-lara:
[...]

PR #2391, opened 2020-09-11
2374 feat/handle missing raw headers

macintoshpie:
Previously, when importing csvs:
- empty header fields were just kept empty as the raw column name
- if there were multiple empty header fields, the last one would essentially 'overwrite' the other empty fields.

#### What's this PR do?
When importing CSVs (not spreadsheets) if there are empty headers, SEED will generate headers for them. Then on the mapping page users are given notice of this with the option to delete the file.

Run the migration and refresh browser cache.
Before each of the following instructions, upload `portfolio-manager-sample-missing-headers.csv`, once on the mapping page you should see a modal warning you that it auto-generated some headers.  

Test 1: click "Continue" on the warning modal, you should see the generated headers as `SEED Generated Header #`. Finish mapping. If you go back to the mapping page you should _not_ see this modal warning.

Test 2: click "Delete File" on the warning modal, you should be redirected to the dataset page and the imported file should be deleted.

#2374


macintoshpie:
I did not implement this functionality for excel spreadsheets b/c the way the parser works makes it pretty unlikely for there to be an empty header. Basically, the header of a spreadsheet is determined by finding the first row that has no empty cells.

seed/lib/mcm/reader.py
```diff
@@ -405,40 +428,23 @@ class CSVParser(object):
             # rows.next() will return the first row
     """
 
-    def __init__(self, csvfile, *args, **kwargs):
```
macintoshpie, line 408: I found no instances of *args or **kwargs being used for initialization, so I removed them (and their downstream uses)

adrian-lara:
When running through manual testing - Things seem to work as expected. But during one of my tests, I triggered a 500 when hitting the `/start_` endpoint.

The relevant parts of the error message here:
```
Internal Server Error: /api/v3/import_files/5617/start_system_matching_and_geocoding/
  File "./seed/venv367-django-2-2-10-upgrade-20200220/lib/python3.6/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "seed_propertyview_property_id_f8bdf6c2_uniq"
DETAIL:  Key (property_id, cycle_id)=(1189362, 370) already exists
...
  File "./seed/seed/views/v3/import_files.py", line 457, in start_system_matching_and_geocoding
    return task_match_buildings(pk)
  File "./seed/seed/data_importer/tasks.py", line 1250, in match_buildings
    finish_matching.s(file_pk, progress_data.key))
...
 File "./seed/seed/data_importer/match.py", line 139, in match_and_link_incoming_properties_and_taxlots
    merged_linked_property_views = link_views(merged_property_views, PropertyView)
  File "./seed/seed/data_importer/match.py", line 416, in link_views
    _merge_count, _link_count, view_id = match_merge_link(view.id, state_class_name)
  File "./seed/seed/utils/match.py", line 238, in match_merge_link
    link_count = _link_matches(matching_views, org_id, target_view, ViewClass)
  File "./seed/seed/utils/match.py", line 169, in _link_matches
    view.save()

```

I'm having trouble reproducing, and the error message would indicate the error triggered at a point in the code you didn't change. My UI actions leading up to the error were as follows:
- import missing headers CSV from test directory
- test "delete" feature once prompted on mapping page
- import missing headers CSV from test directory
- successfully map records
- revisit mapping page and see that no prompt provided (👍 )
- import a different spreadsheet without missing headers
- after clicking the Save Mappings on the mapping confirmation screen, I get the 500.

Any ideas? I think this might not be an issue with your PR and likely an issue with the state of my DB during the testing, as I've not been able to reproduce. Still poking around a bit, but wanted to see what you think.

adrian-lara (approved):
Code review looks good! 👍

adrian-lara:
> When running through manual testing - Things seem to work as expected. But during one of my tests, I triggered a 500 when hitting the `/start_system_matching_and_geocoding` endpoint.

Did some more digging, and couldn't figure out what caused that error. I also wasn't able to reproduce it. I'm fine with merging and just seeing if it comes up again in later testing.

RDmitchell:
Is there a reason we are not implementing this also for spreadsheets (it seems it is only implemented for CSV files)? It is definitely possible for a spreadsheet to have a column with data without a column name at the top.  Maybe that is handled elsewhere for spreadsheets?

PR #2394, opened 2020-09-14
merge column list profile migrations

adrian-lara:
Column List Setting models were recently renamed to Column List Profile

#### What's this PR do?
merge associated migrations that I missed during my review of that PR

review migrations still work


aviveiros11 (approved):
migrations do in fact still work 👍

PR #2396, opened 2020-09-14
Force git SHA to 9 digits

axelstudios:
The `git rev-parse --short HEAD` command used on the Info screen could result in a random number of characters [based on the repo packed data](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreabbrev).
#### What's this PR do?
Forces 9 characters
Browse to `/app/#/about` and check that the SHA is 9 characters
Before and after on dev1 and seeddemostaging
image


adrian-lara (approved):
Works as expected 💯

Issue #2397, opened 2020-09-15
DQ - failed to update/save rules after deleting duplicate rules


lainsworth8801:
Duplicate rules can only be deleted once without error. If a user has more than 1 duplicate rule for the same field, error message is triggered when the user tried to delete again.

### Expected Behavior
A user should be allowed to delete duplicate rules as many times as desired.

### Actual Behavior
For instance, if a user has 3 same rules defined. User starts to delete the 2 duplicate ones. The first one will be deleted without error. However if user proceed to delete another one (basically from 2nd try and on) and save changes, 404 status is returned.

### Steps to Reproduce
Dq admin create 2 or more identical rules;
![Screen Shot 2020-09-15 at 9 22 30 AM](https://user-images.githubusercontent.com/49968203/93244613-3e688f80-f747-11ea-81bf-813a51166aef.png)

Start to delete any one of duplicate rules and save changes. The 1st try will succeed;
![Screen Shot 2020-09-15 at 9 22 44 AM](https://user-images.githubusercontent.com/49968203/93244663-50e2c900-f747-11ea-8f34-8be6d1d4f55d.png)

Continuously delete more duplicate rules (2nd try). Upon hitting save_chnages button, error message emerged:
![Screen Shot 2020-09-15 at 9 22 56 AM](https://user-images.githubusercontent.com/49968203/93244700-622bd580-f747-11ea-8121-2078d3322c73.png)
![Screen Shot 2020-09-15 at 9 23 06 AM](https://user-images.githubusercontent.com/49968203/93244712-6657f300-f747-11ea-8c97-37f923d19eee.png)

### Instance Information
develop

Server Instance: `dev1, dev2, production, staging, etc`
Server SHA: `abcd123`

adrian-lara:
@lainsworth8801 Can you update this ticket? From what I understand, your further investigation lead to a different understanding of the problem. 

Specifically, can you explicitly list what needs to be tested here to show your fix worked?

lainsworth8801:
> @lainsworth8801 Can you update this ticket? From what I understand, your further investigation lead to a different understanding of the problem.
> 
> Specifically, can you explicitly list what needs to be tested here to show your fix worked?

@adrian-lara I believe this issue was caused by [#2399](https://github.com/SEED-platform/seed/pull/2399)

RDmitchell:
instance: dev1
SHA: 38fb067
Org: LBNL Test 200

Program detects duplicate rules on Save
[...]

PR #2399, opened 2020-09-15
#2397 data_quality_admin_controller.js update_rules variable typo causing d…

lainsworth8801:
…up rules deletion error

Not much. Just typo.

#### What's this PR do?
Corrected typo.

Under DQ page, create 2 or more duplicate rules and start deleting them. Should expect no error message returned.

#2397


seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -380,8 +380,8 @@ angular.module('BE.seed.controller.data_quality_admin', [])
         spinner_utility.show();
         $q.all(promises).then(function () {
           data_quality_service.data_quality_rules($scope.org.id).then(function (update_rules){
```
adrian-lara, line 382: Good catch! I think we'd want to rename the parameter of the callback function to `updated_rules` instead of changing everything to be `update_rules`.

adrian-lara (approved):
Works as expected! 👍

Issue #2370, opened 2020-08-11
Fix property detail page export as BuildingSync Excel


macintoshpie:
### Expected Behavior
Exporting a property from the property detail page as BuildlingSync (excel) should work.

### Actual Behavior
This doesn't work. The excel file is empty.

### Steps to Reproduce
Assumption: your canonical property and property view IDs are "out of sync" ie not at the same count.
Go to property list page, select the property, export as excel. Then click the detail page and try to do the same and the excel is empty.

### Suspected Problem
Property list page uses a row item's `id` attribute (which if I remember correctly is a canonical id), while the detail page is using a view id.

### Instances
`prod` (2.6.1-patch3)
`seeddemostaging` (currently 2.7.2)
`develop`

adrian-lara:
I edited the "Actual Behavior". Can you confirm that an error isn't necessarily thrown and that "doesn't work" refers to the output file being empty?

Assuming that, I can confirm this happens on `prod` (2.6.1-patch3), `seeddemostaging` (currently 2.7.2), and `develop`.

macintoshpie:
Yes, sorry for the ambiguity, it issue is that the file is empty after exporting.

RDmitchell:
instance: dev1
SHA: 
Org: LBNL Test 200
Cycle: 2015 Compliance Cycle

In Inventory Detail View I exported record as BldgSync Excel. There were not many fields filled in, but the file had the column headers and at least some fields had data
![image](https://user-images.githubusercontent.com/6314950/94325702-8744f300-ff54-11ea-9bff-905781d03e4e.png)

PR #2400, opened 2020-09-16
2370 fix/bsync excel export

macintoshpie:
#### What's this PR do? 
Fixes problem where exporting from the property detail page resulted in empty files. This was fixed by refactoring the `taxlot_properties/export` endpoint to accept View IDs rather than canonical IDs. This change required updating the frontend for the list view, as it was using canonical IDs.

1. Upload properties and taxlots
1. From inventory list page, select multiple properties and choose the "export selected" action. Verify that:
  - all selected properties were exported
  - the properties are in the same order as they appeared in the table
1. On a property details page, select "export buildingsync (excel)". Verify it exported correctly
1. Verify other export options still work as expected (CSV, geojson, etc)
1. Repeat for Taxlots (note you can't export excel for taxlots, and cant export individual taxlots from detail page)

#2370

### DIFF
--- a/seed/views/v3/tax_lot_properties.py
+++ b/seed/views/v3/tax_lot_properties.py
@@ -120,7 +120,7 @@ class TaxLotPropertyViewSet(GenericViewSet):
             prefetch_related = ['labels']
             filter_str = {'property__organization_id': org_id}
             if ids:
-                filter_str['property__id__in'] = ids
+                filter_str['id__in'] = ids
             # always export the labels and notes
             column_name_mappings['property_notes'] = 'Property Notes'
             column_name_mappings['property_labels'] = 'Property Labels'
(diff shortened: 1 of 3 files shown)


seed/views/v3/tax_lot_properties.py
```diff
@@ -120,7 +120,7 @@ def export(self, request):
             prefetch_related = ['labels']
             filter_str = {'property__organization_id': org_id}
             if ids:
-                filter_str['property__id__in'] = ids
+                filter_str['id__in'] = ids
```
macintoshpie, line 123: Note that we are making the query on the *View models here, which is why this is changed to just `id`

seed/views/v3/tax_lot_properties.py
```diff
@@ -163,7 +163,11 @@ def export(self, request):
         # force the data into the same order as the IDs
         if ids:
             order_dict = {obj_id: index for index, obj_id in enumerate(ids)}
-            data.sort(key=lambda x: order_dict[x['id']])  # x is the property/taxlot object
+            if view_klass_str == 'properties':
+                view_id_str = 'property_view_id'
+            else:
+                view_id_str = 'taxlot_view_id'
+            data.sort(key=lambda inventory_obj: order_dict[inventory_obj[view_id_str]])
```
macintoshpie, line 170: I did a double check in `TaxLotProperty.get_related()`, which is where `data` is coming from, and it seems we're guaranteed these view IDs so this should be OK

adrian-lara: That sounds right - thanks for double checking 💯

adrian-lara (approved):
Code changes look good and this works as expected 💯

Issue #1978, opened 2019-10-07
MapQuest API key error UX


adrian-lara:
### Expected and Actual Behavior
When an organization has an invalid MapQuest API key, the program should notify the user in some way. As of now, geocoding during file import or manually on the inventory list page leads to the status bar appearing "stuck". 

These do trigger 500s on the backend, but given the loading/status bar doesn't go away, this could be misleading and maybe frustrating for users. Along with that, this invalid MapQuest API key would prevent any imports involving records with address information.

### Steps to Reproduce
Give an organization an invalid MapQuest API key. Try to import files with enough address information (address_line_1, city, state, etc.). Try to manually geocode records with enough address information on the inventory list page.

RDmitchell:
Instance: dev1
SHA: 
Org: LBNL Test 200
Cycle: 2020 Compliance

* In organizations, set MapQuest API key to an invalid value, but keep the Enable Geocoding on
![image](https://user-images.githubusercontent.com/6314950/102674768-f5f8ad80-414b-11eb-9afe-11b29b297bfb.png)
* Import a file that has good addresses that should be able to geocoded, get this message
![image](https://user-images.githubusercontent.com/6314950/102674795-0ad54100-414c-11eb-9153-a8bc889babfa.png)
* Reset the API key to a valid value and import the same file
* Program is able to generate Lat/Long for the records
![image](https://user-images.githubusercontent.com/6314950/102675325-c480e180-414d-11eb-8124-2a5477a3ca0a.png)

RDmitchell:
@adrian-lara -- I just noticed that this triggered a Sentry error ... just a heads up in case you want to do something so that this message, which seems completely appropriate to me, doesn't trigger a Sentry error.
![image](https://user-images.githubusercontent.com/6314950/102676288-b208a700-4151-11eb-844a-ad35ab88638f.png)

PR #2402, opened 2020-09-18
1978 mapquest api key error

lainsworth8801:
Alert user of geocoding error if no valid mapquest api key is saved.

#### What's this PR do?
Exit and print error message via geocode modal if no valid mapquest api key is detected in stead of hanging.

Under organization settings, input an invalid string for mapquest api key. Map and save your data:
Screen Shot 2020-09-17 at 8 34 42 PM

Error message is expected when a user tries to save the mapping:
Screen Shot 2020-09-17 at 8 45 05 PM

Error message is also expected whe user tries to manually geocode selected properties/taxlots:
Screen Shot 2020-09-17 at 9 16 48 PM

After using a valid mapquest api key, one should expect to see success status:
Screen Shot 2020-09-17 at 9 15 46 PM

Screen Shot 2020-09-17 at 9 16 10 PM


seed/static/seed/js/services/matching_service.js
```diff
@@ -24,6 +24,8 @@ angular.module('BE.seed.service.matching', []).factory('matching_service', [
         params: { organization_id: user_service.get_organization().id }
       }).then(function (response) {
         return response.data;
+      }).catch(function(e) {
+        if (e.status === 403) return e.data;
```
adrian-lara, line 28: Did you encounter an uncaught exception that made you make this change? I think your change is likely fine as it is, but am just hoping to get more background here.

seed/static/seed/partials/geocode_modal.html
```diff
@@ -85,10 +86,15 @@ <h4 class="modal-title" id="exportModalLabel" ng-show="geocode_state === 'result
...
+            <p>{$ error_message | translate $}</p>
+        </div>
+    </div>
 
 </div>
 <div class="modal-footer">
     <button type="button" class="btn btn-primary" ng-click="geocode_buildings()" ng-show="geocode_state === 'verify' && has_api_key && has_enough_geocoding_columns" translate>Geocode</button>
     <button type="button" class="btn btn-default" ng-show="geocode_state == 'verify'" ng-click="cancel()" translate>Cancel</button>
-    <button type="button" class="btn btn-default" ng-show="geocode_state === 'result'" ng-click="close()" translate>Close</button>
+    <button type="button" class="btn btn-default" ng-show="geocode_state === 'result'||'fail'" ng-click="close()" translate>Close</button>
```
adrian-lara, line 99: I see this when I first open the geocoding modal now. 

Screen Shot 2020-09-18 at 1 44 38 PM

The string 'fail' is truthy, so it ends up being `ng-show=<boolean> || <truthy>`.

adrian-lara (commented):
Can you also update (including tests) the `geocode_by_ids` v3 endpoint to not 500 on an invalid API key? You'll likely have to wrap much [of the logic](https://github.com/SEED-platform/seed/blob/79adb203804485537d06b5ee7e73989ba85e0d69/seed/views/v3/geocode.py#L48-L66) in a try/except MapQuestAPIKeyError/except block in order to prevent the 500. Likely, we should make the exceptions return a 400. I imagine you won't hvae to make many changes to what you already have to handle that 400 in the frontend.

seed/static/seed/js/services/matching_service.js
lainsworth8801: I did. And i recall this is after/skip importing file step and when a user attempts to manually geocode selected properties/taxlots on inventory list, error message wasn't caught and prompted prior to this change. So in situation where a user changes the api key after importing a file and tried to manually geocode, an error could be missed.

lainsworth8801: @adrian-lara this seems a bit of a tricky situation. Assuming a user imported a file using a valid key, between he/she manually geocodes selected properties/tax lots on inventory list he/she accidentally changed the api key to an invalid one. In this case, geocoding attempts have indeed been made (from successfully importing a file prior) but will fail this time when he/she hits "Geocode" button to geocode again. But this time, error message pops up.
I could have been over-thinking again I guess this comes down to how possible it is for users to alter their mapquest api keys.

seed/static/seed/partials/geocode_modal.html
lainsworth8801: @adrian this relates to your last comment. Giving the same assumption where a user alters his/her mapquest api key in between importing a file and manually geocoding inventories, the geocoding attempts seen here makes sense to me. If the user proceed without correcting his/her key, error pops up upon them hitting "Geocode" button, in this case, also means geocoding for the 2nd time but manually.

seed/static/seed/js/services/matching_service.js
adrian-lara: > I could be over-thinking here butI wonder how possible it is for users to alter their keys in between confirm and saving their mapping and manually geocode on inventory lists?

I think I'm missing something, but this scenario shouldn't present a problem? If a user ever has an invalid key, they should get the appropriate message whenever trying to geocode records using their address fields and MapQuest.

seed/static/seed/partials/geocode_modal.html
adrian-lara: I'm not sure I understand what you mean. Let's have a call about these comments

seed/static/seed/partials/geocode_modal.html
```diff
@@ -85,10 +86,16 @@ <h4 class="modal-title" id="exportModalLabel" ng-show="geocode_state === 'result
...
+            <p>{$ error_message | translate $}</p>
+        </div>
+    </div>
 
 </div>
 <div class="modal-footer">
     <button type="button" class="btn btn-primary" ng-click="geocode_buildings()" ng-show="geocode_state === 'verify' && has_api_key && has_enough_geocoding_columns" translate>Geocode</button>
     <button type="button" class="btn btn-default" ng-show="geocode_state == 'verify'" ng-click="cancel()" translate>Cancel</button>
     <button type="button" class="btn btn-default" ng-show="geocode_state === 'result'" ng-click="close()" translate>Close</button>
+    <button type="button" class="btn btn-default" ng-show="geocode_state === 'fail'" ng-click="close()" translate>Close</button>
```
adrian-lara, line 100: This works but we could probably DRY things up a bit with this.
```suggestion
    <button type="button" class="btn btn-default" ng-show="geocode_state === 'fail' || geocode_state === 'result'" ng-click="close()" translate>Close</button>
```

seed/static/seed/js/services/matching_service.js
```diff
@@ -24,6 +24,8 @@ angular.module('BE.seed.service.matching', []).factory('matching_service', [
         params: { organization_id: user_service.get_organization().id }
       }).then(function (response) {
         return response.data;
+      }).catch(function(e) {
+        if (e.data.message === 'MapQuest API key may be invalid or at its limit.') return e.data;
```
adrian-lara, line 28: I think it might be better to always return e.data here.

seed/static/seed/js/controllers/geocode_modal_controller.js
```diff
@@ -49,24 +49,30 @@ angular.module('BE.seed.controller.geocode_modal', [])
...
+            $scope.error_message = "MapQuest API key may be invalid or at its limit.";
 
-            if (result.tax_lots) {
-              $scope.tax_lots_geocoded_high_confidence = result.tax_lots.high_confidence;
-              $scope.tax_lots_geocoded_low_confidence = result.tax_lots.low_confidence;
-              $scope.tax_lots_geocoded_manually = result.tax_lots.manual;
-              $scope.tax_lots_geocode_not_possible = result.tax_lots.missing_address_components;
-            }
+            $scope.geocode_state = 'fail';
+          } else {
```
adrian-lara, line 57: I think there's likely a better general pattern using `catch` for handling the MapQuestAPIKeyError, but I'm not 100% what that would look like. This `else` works for now though.

adrian-lara (commented):
This is pretty close. See comments for some recommended changes.

Additionally, one change we should definitely make for the import case is to still show the results of the import (along with the error message that's shown). 
Screen Shot 2020-09-22 at 8 32 35 AM

The way this looks, it would make users believe that the import completely failed.

adrian-lara:
> Can you also update (including tests) the `geocode_by_ids` v3 endpoint to not 500 on an invalid API key? You'll likely have to wrap much [of the logic](https://github.com/SEED-platform/seed/blob/79adb203804485537d06b5ee7e73989ba85e0d69/seed/views/v3/geocode.py#L48-L66) in a try/except MapQuestAPIKeyError/except block in order to prevent the 500. Likely, we should make the exceptions return a 400. I imagine you won't hvae to make many changes to what you already have to handle that 400 in the frontend.

Bumping this comment as to make sure it doesn't get lost

seed/static/seed/partials/geocode_modal.html
lainsworth8801: thank you!

lainsworth8801:
> > Can you also update (including tests) the `geocode_by_ids` v3 endpoint to not 500 on an invalid API key? You'll likely have to wrap much [of the logic](https://github.com/SEED-platform/seed/blob/79adb203804485537d06b5ee7e73989ba85e0d69/seed/views/v3/geocode.py#L48-L66) in a try/except MapQuestAPIKeyError/except block in order to prevent the 500. Likely, we should make the exceptions return a 400. I imagine you won't hvae to make many changes to what you already have to handle that 400 in the frontend.
> 
> Bumping this comment as to make sure it doesn't get lost

@adrian-lara in order for geocode service to properly catch error I added "[throw e](https://github.com/SEED-platform/seed/blob/affc208f4107d98fb4393c7d6191505698f69422/seed/static/seed/js/services/geocode_service.js#L18-L20)". This throws specific error based on message and catches both 403 and 500 error. Let me know if this works.

adrian-lara:
> Additionally, one change we should definitely make for the import case is to still show the results of the import (along with the error message that's shown).
> Screen Shot 2020-09-22 at 8 32 35 AM
> 
> The way this looks, it would make users believe that the import completely failed.

I misunderstood what was happening in the case of the screenshot. The import did, in fact, fail. 

Can you change the title of the modal to indicate that the upload failed? We should also remove the "Add another file" and "View my properties" buttons. Finally, I think you mentioned this before, maybe let's add a button to link to the org settings page.

adrian-lara (commented):
Works as expected! The only definite change I suggest is fairly minor. See the "Reset" comment

.gitignore
```diff
@@ -46,6 +46,7 @@ lin/migrations/disabled
 htmlcov/
 src/
 node_modules/
+npm-shrinkwrap.json
```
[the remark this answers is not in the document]
lainsworth8801: When I run install_javascript_dependencies.sh seed's version will change from 2.7.1 to 2.8.0. Adding this file to .gitignore makes sure the version change doesn't get committed.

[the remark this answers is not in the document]
adrian-lara: Gotcha - what's the outcome of that changing? It sounds like npm-shrinkwrap.json is used when publishing applciations? I see that we have updated this file in the past, so I wonder if we do want those changes to come through. @axelstudios Thoughts?

seed/static/seed/partials/data_upload_modal.html
```diff
@@ -403,12 +403,15 @@ <h5>Use Case Warnings</h5>
...
-        <div class="row text-center">
+        <div class="row text-center" ng-hide="step_10_mapquest_api_error">
             <button type="button" class="btn btn-primary col-sm-6 center-block" ng-click="goto_step(2)" translate>Add another file</button>
         </div>
-        <div class="row text-center">
+        <div class="row text-center" ng-hide="step_10_mapquest_api_error">
             <button id="view-buildings" type="button" class="btn btn-primary col-sm-6 center-block" ng-click="view_my_properties()" translate>View my properties</button>
         </div>
+        <div class="row text-center" ng-if="step_10_mapquest_api_error">
+            <button type="button" class="btn btn-primary col-sm-6 center-block" ng-click="reset_mapquest_api_key()" translate>Update MapQuest API key</button>
```
adrian-lara, line 413: Screen Shot 2020-10-27 at 10 36 54 AM
👍

adrian-lara (approved):
Looks good to me. Just waiting for responses on the npm-shrinkwrap.json comments.

.gitignore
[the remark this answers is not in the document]
axelstudios: We don't want to ignore the two `npm-shrinkwrap.json` files, those lock down dependency versions to make sure developers/deployments all have the same code.  It's fine if they're updated through the `npm install` process, in fact my pre-commit hook PR updates the main one:
https://github.com/SEED-platform/seed/pull/2427/files#diff-44a7c69871c5958e657226d5552c9451606a778233fb824f68530d0cfd3f8994

adrian-lara:
#1978

PR #2403, opened 2020-09-18
2203 dq duplicate rules

lainsworth8801:
Dq should alert user if duplicate rules are created.

#### What's this PR do?
On u added alert message if user accidentally created 2 identical rules.

Under dq page, create an identical rule to any existing rule and hit save. Expect to see error message:
Screen Shot 2020-09-17 at 9 31 53 PM
Screen Shot 2020-09-17 at 9 32 06 PM

Once user clears duplicate rule(s), save changes proceeds:
Screen Shot 2020-09-17 at 9 32 13 PM

TaxLot view:
Screen Shot 2020-09-17 at 9 32 29 PM
Screen Shot 2020-09-17 at 9 32 39 PM

Highlight duplicate rules:
Screen Shot 2020-09-24 at 1 21 03 PM


seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -352,6 +353,33 @@ angular.module('BE.seed.controller.data_quality_admin', [])
...
+              return recurse(rule, criteria.slice(1));
+            });
+          } else {
+            if (last_rules.length > 0) $scope.is_duplicate = true;
+          }
+        };
+
+        var criteria = ['table_name', 'condition', 'field', 'data_type', 'min', 'max', 'units', 'severity', 'label'];
+        recurse(rules, criteria);
+        if ($scope.is_duplicate) return Notification.error({message: "Duplicate rules detected.", delay: 10000});
```
adrian-lara, line 381: I'm having a bit of a harder time tracking the logic through this, but from what I can tell, each value for each condition triggers their own recursive step at each recursive step. This seems super inefficient, and is a bit confusing. 

As much as I appreciate recursion, I think we may be able to accomplish this without it. 

What if we use the `_.group_by` and group by concatenated values of each of the criteria you listed? I'm imagining group_by returning something like:
```
`{rule['table_name']}-{rule['condition']}-`
```
Would that work? Let me know what you think.

adrian-lara (commented):
Suggested change to algorithm - though it does seem to work as expected, except for when labels? See screenshot. Also, can we highlight the rows in red that are duplicates?

Screen Shot 2020-09-18 at 12 41 17 PM

seed/static/seed/js/controllers/data_quality_admin_controller.js
lainsworth8801: My initial thought was to break down all the rules by each criteria and at last, when all criteria has been iterated, rules set is not 0 in length (the unique-ness of each rule should lead to a empty rule group), then it's determined that there exist duplicate rules. With using recursion here, and if in the future more criteria is requested, we will just need to put the addition criteria in the criteria array.

adrian-lara: > ... if in the future more criteria is requested, we will just need to put the addition criteria in the criteria array.

I was thinking that grouping by concatenated values would address this as well. I'm still not convinced we need to have a recursive solution unless there's something I'm missing about the group by concatenated values approach, but let's chat about via call.

seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -352,6 +353,14 @@ angular.module('BE.seed.controller.data_quality_admin', [])
           show_configuration_errors(misconfigured_rules);
         }
 
+        // Find duplicate rules and trigger warnings
+        $scope.is_duplicate = _.some(_.groupBy(rules, function(rule) {
+          return `${rule.table_name}-${rule.condition}-${rule.field}-${rule.data_type}-${rule.min}-${rule.max}-${rule.units}-${rule.severity}-${rule.status_label}`;
```
adrian-lara, line 358: Looks like we're missing the text field here.

adrian-lara (commented):
This works as expected, and I think this is really close! One minor change for sure (missing text field criteria).

And also, after more extensive testing, it could be beneficial to highlight the duplicate rows in red. 
Screen Shot 2020-09-22 at 7 49 40 AM
What do you think?

lainsworth8801:
Added highlighted duplicate rules: 
Screen Shot 2020-09-24 at 1 21 03 PM

adrian-lara (commented):
I think something happened to how error messages are displayed. Non-duplicate errors don't seem trigger row highlighting anymore (see the 3rd rule).
Screen Shot 2020-09-24 at 3 08 44 PM

seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -352,6 +373,10 @@ angular.module('BE.seed.controller.data_quality_admin', [])
           show_configuration_errors(misconfigured_rules);
         }
 
+        // Find duplicate rules and trigger warnings
+        $scope.is_duplicate = ($scope.duplicate_rule_keys.length > 1);
+        if ($scope.is_duplicate) return Notification.error({message: "Duplicate rules detected.", delay: 10000});
+
```
adrian-lara, line 379: I think you want to move the notification logic into the show configuration errors

seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -222,8 +222,28 @@ angular.module('BE.seed.controller.data_quality_admin', [])
       var get_configured_rules = function () {
         var rules = [];
         var misconfigured_rules = [];
+        var duplicate_rule_group = [];
+        $scope.duplicate_rule_keys = [];
```
adrian-lara, line 226: maybe make use of the misconfigured rules somehow? so you can easily make use of the `show_configuration_errors` method.

seed/static/seed/js/controllers/data_quality_admin_controller.js
```diff
@@ -222,8 +222,28 @@ angular.module('BE.seed.controller.data_quality_admin', [])
...
+            });
+            // Case 1: multiple keys with duplicate rules;
+            if (Object.keys(duplicate_rules).length > 1) {
+              _.forEach(duplicate_rules, function(group) {
+                if (group.length > 1) {
+                  _.forEach(group, function(rule) {
+                    $scope.duplicate_rule_keys.splice(0, 0, rule.$$hashKey);
+                  });
+                }
+              });
```
adrian-lara, line 240: I might be missing something, but why can't we use this in either of the two cases?

seed/static/seed/js/controllers/data_quality_admin_controller.js
adrian-lara: As discussed, the `show_configuration_errors` method is broken. Not incorporating into that right now makes the most sense. I'll be cleaning that up soon

adrian-lara (approved):
Works as expected and code looks good! 
Screen Shot 2020-09-24 at 5 18 01 PM

Issue #1904, opened 2019-06-20
Add ability to delete extra data columns in Column Settings


RDmitchell:
This is a placeholder issue -- it may need more discussion and design once we decide to deal with it. 
### Expected Behavior
Add the ability to delete columns (rows) in the Column Settings screen. This would only be applied to "extra data" fields. 

And if that field had data in it, when the field is deleted, any data associated with it would also be deleted (right?). So having a dialog box warning the user of that would be needed. 

That functionality already exists in the API and it wouldn't be much work to add it to the UI

That would solve the problem show in the screenshot below, which can occur when different files are imported with slightly different mapping. This is particularly an issue that happens when ESPM files are imported using the default SEED ESPM mapping vs using the button in mapping that sets the mapping to the actual ESPM field name. 
![image](https://user-images.githubusercontent.com/6314950/59879156-ecc43000-935e-11e9-8060-63e438b1122a.png)
### Actual Behavior 
n/a
### Steps to Reproduce
n/a
### Instance Information
instance: LBNL production
SHA: df842634

axelstudios:
As a related issue, the current API that allows users to delete columns **needs to be improved** because it currently allows users to delete non-extra-data columns.  That can cause all sorts of issues around data quality checking, mapping, and matching

RDmitchell:
@nllong / @axelstudios -- I just added this to the dec 2019 project and put it in Triage.

RDmitchell:
instance: dev1
SHA: 38fb067
Org: LBNL Test 200

I don't see a delete button in Column Settings for extra data fields? 
![image](https://user-images.githubusercontent.com/6314950/94729869-313acb80-0317-11eb-9225-e56620e71266.png)

adrian-lara:
Apologies - we didn't update dev1 in time before you had a chance to review.

This is on dev1 and ready for you to test.

RDmitchell:
instance: dev1
SHA: 1e0884dd3

Seems to be working

PR #2407, opened 2020-09-23
#1904 - extra_data columns can now be deleted

axelstudios:
#### What's this PR do?
- Limits `DELETE /api/v3/columns/XXX/` to only extra_data columns
- Adds a delete button to extra_data columns on the org's column_settings page

- Import a file with columns mapped to extra_data fields (new or existing)
- Go to the column_settings page for the org and click the delete button next to extra_data fields
- Monkey test with deleting columns, reimporting data and recreating the column, attempting to delete non-extra_data columns, etc
#1904

## CAVEATS
When a column is deleted the column mapping with the raw column_name also has to be deleted.  If you go back to the Data Mapping for an imported record from the dataset the original data will still appear in the preview, but if you click the button to review the mapping the deleted column will no longer appear (which is expected once the column mapping has been deleted)

It takes 6 seconds to update 512 records, and this comes entirely from recalculating the state hash_objects.  It would be advisable to move this logic to a celery task as a next step

### DIFF
--- a/seed/static/seed/js/services/columns_service.js
+++ b/seed/static/seed/js/services/columns_service.js
@@ -44,6 +44,16 @@ angular.module('BE.seed.service.columns', []).factory('columns_service', [
       });
     };
 
+    columns_service.delete_column = function (column_id) {
+      return columns_service.delete_column_for_org(user_service.get_organization().id, column_id);
+    };
+
+    columns_service.delete_column_for_org = function (org_id, column_id) {
+      return $http.delete('/api/v3/columns/' + column_id + '/', {
+        organization_id: org_id
+      });
+    };
+
     return columns_service;
 
   }]);
--- a/seed/static/seed/js/services/columns_service.js
+++ b/seed/static/seed/js/services/columns_service.js
@@ -50,7 +50,7 @@ angular.module('BE.seed.service.columns', []).factory('columns_service', [
 
     columns_service.delete_column_for_org = function (org_id, column_id) {
       return $http.delete('/api/v3/columns/' + column_id + '/', {
-        organization_id: org_id
+        params: {organization_id: org_id},
       });
     };
 
--- a/seed/views/v3/columns.py
+++ b/seed/views/v3/columns.py
@@ -13,12 +13,17 @@ from drf_yasg.utils import swagger_auto_schema
 from rest_framework import status
 from rest_framework.parsers import JSONParser, FormParser
 from rest_framework.renderers import JSONRenderer
+
 from seed.decorators import ajax_request_class
 from seed.lib.superperms.orgs.decorators import has_perm_class
 from rest_framework.decorators import action
 from seed.models import (
     Column,
+    ColumnMapping,
+    DATA_STATE_MATCHING,
     Organization,
+    PropertyState,
+    TaxLotState,
 )
 from seed.serializers.columns import ColumnSerializer
 from seed.serializers.pint import add_pint_unit_suffix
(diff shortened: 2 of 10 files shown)


seed/views/v3/columns.py
```diff
@@ -149,6 +154,68 @@ def update(self, request, pk=None):
...
+        if not column.is_extra_data:
+            return JsonResponse({
+                'success': False,
+                'message': 'Only extra_data columns can be deleted'
+            }, status=status.HTTP_400_BAD_REQUEST)
+
+        # Delete key from jsonb data
+        if column.table_name == 'PropertyState':
+            states = PropertyState.objects.filter(organization_id=org_id, data_state=DATA_STATE_MATCHING,
+                                                  extra_data__has_key=column.column_name)
```
adrian-lara, line 178: In most cases, this query worked and this feature seems to work as expected. 

But in testing with an old prod database that's been upgraded to latest develop, I ran into one issue with this query:
```
django.db.utils.ProgrammingError: operator does not exist: json ? unknown
LINE 1: ..._state" = 3 AND "seed_propertystate"."extra_data" ? 'Email' ...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```
The only relevant difference I found between this DB compared to others is that the `extra_data` field type is json (vs jsonb). Given the error, it seems that this would be the cause.

I'm still doing research into what would've caused my current DB to type this field as json vs jsonb as well as the extra_data field type within a non-testing copy of the current prod DB.

adrian-lara (commented):
TL;DR - EDIT: I wouldn't expect this bug to pop up on DBs built _after_ the `django-pgjson version 0.3.1` commit listed below. Unfortunately, it seems prod was built _before_ then.

The issue described in my previous comment does seem to only happen when the extra_data field type is json instead of jsonb in the database.

The code changes look good - like you mentioned, testing and celery-izing(?) the iterating `.save()`s are good next steps, but this definitely gets the job done.

It looks like the prod DB dump uses `json` as the type. This makes given previous iterations of SEED used [django-pgjson version 0.3.1](https://github.com/SEED-platform/seed/commit/e90fa03272748c4db18adb0588189ed4f5c0e113)). Looks like [django-pgjson used `json` in that version](https://github.com/djangonauts/django-pgjson/blob/0.3.1/django_pgjson/fields.py#L53).

My guess is that we'd have to update the type, but I'm not 100% sure if that's the only choice. More research is needed.

adrian-lara:
It seems like it could be as simple as including a migration that executes `ALTER TABLE seed_propertystate ALTER COLUMN extra_data TYPE JSONB USING extra_data::JSONB`. 

The command is taking some time so I'll edit this comment later with updates.

adrian-lara (approved):
Approved with my changes, but probably worth someone else doing a quick run through.

seed/static/seed/js/services/columns_service.js
```diff
@@ -44,6 +44,16 @@ angular.module('BE.seed.service.columns', []).factory('columns_service', [
       });
     };
 
+    columns_service.delete_column = function (column_id) {
+      return columns_service.delete_column_for_org(user_service.get_organization().id, column_id);
+    };
+
+    columns_service.delete_column_for_org = function (org_id, column_id) {
+      return $http.delete('/api/v3/columns/' + column_id + '/', {
+        params: {organization_id: org_id},
```
adrian-lara, line 53: I added a wrapping object with a params key here since params were not being sent with the request to the backend. 

[That particular backend endpoint does some "smart searching"](https://github.com/SEED-platform/seed/pull/2407/files#diff-f49932cc42b1c65c1fae44800221e8aeR160) for an `organization_id`, getting that ID from a logged in user's "active" (listed in the top right on the UI) organization. This caused the request to happen on the active org vs the org which had it's column settings edited by the user.

macintoshpie (approved):
Looks good to me 👍 Just to confirm, the desired behavior is to _not_ delete the column from the column mapping profiles?

seed/views/v3/columns.py
```diff
@@ -149,6 +154,68 @@ def update(self, request, pk=None):
...
+    @ajax_request_class
+    @has_perm_class('can_modify_data')
+    def destroy(self, request, pk=None):
+        org_id = self.get_organization(request)
+        try:
+            column = Column.objects.get(id=pk, organization_id=org_id)
+        except Column.DoesNotExist:
+            return JsonResponse({
+                'success': False,
+                'message': 'Cannot find column in org=%s with pk=%s' % (org_id, pk)
```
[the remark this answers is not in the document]
nllong: I would argue that these should be f-strings :p. But we can bikeshed later.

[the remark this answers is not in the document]
axelstudios: Thanks for the suggestions - I'm definitely liking f-strings for speed and readability: https://realpython.com/python-f-strings/#speed

seed/static/seed/js/services/columns_service.js
axelstudios: Thanks for catching this!

Issue #2371, opened 2020-08-11
Checking for duplicate fields is not working when only one Inventory Type is selected


RDmitchell:
### Expected Behavior
Program would display an error if the same field name was applied to two different fields for the same table (ie, Property or Tax lot)
### Actual Behavior
Program allowed me to map two fields to "Custom ID 1" in the Property table
### Steps to Reproduce
* Import ESPM file called 2 - example-data-ESPM-Alerts-UBID.xlsx
* Select preset called "2020-07-24 org-wide mapping..." and click Apply Preset
![image](https://user-images.githubusercontent.com/6314950/89959634-bfd17d00-dbf1-11ea-8053-6331e186fb2a.png)
* From that preset, SEED assigns "Custom ID 1" to both Building ID and Property ID, both set to "Property"
![image](https://user-images.githubusercontent.com/6314950/89959734-02935500-dbf2-11ea-94a0-95d03707b664.png)
* Click the Map Your Data button and the program doesn't complain about the duplicate field name, but goes to Mapping Review. In Mapping Review it appears that Custom ID 1 is only there once, and it has been assigned to the "Property ID" field. The "Building ID" field appears to not be imported, ie, it is lost
![image](https://user-images.githubusercontent.com/6314950/89959957-6ae23680-dbf2-11ea-8ba9-78dfd1d23510.png)
* In the Inventory View, if I click on Show Only Populated columns, there is only the one Custom ID 1 field, mapped to the Property ID field, and Building ID is not in the mapping, so presumably it is not in the database (?) 
![image](https://user-images.githubusercontent.com/6314950/89960212-39b63600-dbf3-11ea-8be4-66b31cbd9b52.png)

However, if in the mapping I assign one of the fields to the Tax Lot table, then the program's error checking kicks in and it indicates that there are two fields with the same name for the Property table.
![image](https://user-images.githubusercontent.com/6314950/89960422-d8429700-dbf3-11ea-95eb-c5694d34f5e8.png)

### Instance Information

Instance: seeddemostaging
SHA: 62ee5c4
Org: LBNL 200
Cycle: 2020 Compliance Cycle

adrian-lara:
Per discussion, we're planning to include this in 2.7.2.1.

RDmitchell:
Instance: seeddemostaging
SHA:
Org: LBNL Test 200
Cycle: 2020 Compliance

[...]

PR #2408, opened 2020-09-23
Import mapping - check for duplicates when applying profile

adrian-lara:
Cherry-picks changes made into `develop` into 2.7.2.1 (#2372)

The bug was the result of a lack of duplicate column mapping checking when applying a mapping profile.

See issue for more details on the bug.

#### What's this PR do?
Add dup check when applying a profile.

Test on 2.7.2 database.

1. Import a file. 
2. Map to the same column twice.
3. Save preset.
4. Refresh page.
5. Apply that previously saved preset.
6. See an error.

#2371

### DIFF
--- a/seed/static/seed/js/controllers/mapping_controller.js
+++ b/seed/static/seed/js/controllers/mapping_controller.js
@@ -113,6 +113,7 @@ angular.module('BE.seed.controller.mapping', [])
             $scope.current_preset = $scope.dropdown_selected_preset;
             $scope.initialize_mappings();
             analyze_chosen_inventory_types();
+            $scope.updateColDuplicateStatus();
           }).catch(function () {
             $scope.dropdown_selected_preset = $scope.current_preset;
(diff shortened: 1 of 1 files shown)


RDmitchell:
@adrian-lara / @axelstudios -- so should this change get onto the LBNL staging server for the current SEED user testing?

adrian-lara:
> @adrian-lara / @axelstudios -- so should this change get onto the LBNL staging server for the current SEED user testing?

Yes - sorry. This patch was something that fell through the cracks on my end and should've been done sooner. I only caught it earlier when reviewing tickets for the latest release.

RDmitchell:
@adrian-lara -- no worries. Maybe let me know when this is on staging and I will do a quick test before sending the email to current user for testing.

macintoshpie (approved):
Looks good to me 💯 . Tested according to instructions and it worked as expected.

PR #2412, opened 2020-09-25
API v3 - Append ComStock list profile after serialization

adrian-lara:
Before, when the ComStock list profile feature was built, it originally injected data before serialization of response data.

Since then, the serializer was updated to expect data in a different format.

#### What's this PR do?
As a quick fix, the "fake" ComStock list profile is injected after serialization.

Enable ComStock in org settings.
Navigate to Column Settings and set some ComStock Mappings
Visit the property inventory list page and see that the ComStock profile works
Visit the column list profile page and see the ComStock profile is uneditable.


macintoshpie (approved):
Looks good to me 👍

PR #2423, opened 2020-10-15
feat(docker): add docker for local development

macintoshpie:
Currently, there's no convenient way start developing with docker.
#### What's this PR do?
- Makes it easy for developers to develop locally using docker containers, providing live reload, debugging, and testing.
- changes our docker-compose files from 3.0 to 3.4. This is a potentially breaking change for some servers, as 3.4 requires requires docker engine 17.09.0 or higher. I checked our BRICR server to see what it's running and it's safe - we still need to confirm our other servers are compatible
```bash
$ docker --version && docker-compose --version
Docker version 19.03.6, build 369ce74a3c
docker-compose version 1.25.4, build 8d51620a
```
Follow the instructions lined out in setup_docker.rst. You should be able to:
- start the server and navigate the site
- run the debugger
- run tests
- use a different database volume other than seed_pgdata
- customize server settings with local_untracked.py


macintoshpie:
@nllong can you confirm we're running docker engine >= 17.09.0?

nllong:
locally:
```
➜  ~ docker --version
Docker version 19.03.12, build 48a66213fe
```
And the servers are on 19.03 as well.
so no, sorry :(

macintoshpie:
> And the servers are on 19.03 as well.

We're good then, we just need >= 19.03.6, so updating the docker-compose file version should be fine. Thanks!

adrian-lara:
Gave this a spin and it seems to work as expected from initial testing. A few minor tweaks we discussed that Ted has already or is planning to make changes:
- Update documentation to mention local_untracked.py file
- Celery live reload

2 other topics that came up were:
- mounting docker volume to an existing directory on the host (with the intent of using databases on the host machine)
- web container logs. During our debugging we expected to see some errors but didn't.

macintoshpie:
@adrian-lara Re creating volumes from existing host directories, you can do
```
docker volume create --name my_test_volume --opt type=none --opt device=/path/to/dir --opt o=bind
```
No guarantees that this will work with an existing database, but worth a shot.

adrian-lara (approved):
On the testing side, this works as expected. I'm able to navigate/use the site. I imported records via spreadsheet as well as via ESPM login. I was able to use debuggers and run the test suite as well.

I'm not as well versed in docker configurations and usage, so it could still be worthwhjile to have someone else review the actual code docker-related changes. I gave it a quick scan and looked through the test changes as normal.

Dockerfile-dev
```diff
@@ -0,0 +1,37 @@
+# Start with 2.6.0. note that the source code will be removed and re-copied to this container. The
+# version of SEED here is used to load in the core system packages and dependencies.
+FROM seedplatform/seed:2.6.0
```
nllong, line 3: As alex said, this should be fixed eventually. But won't hold up this feature.

docker-compose.dev.yml
```diff
@@ -0,0 +1,42 @@
...
+    image: seedplatform/seed:dev
+    build:
+      context: .
+      dockerfile: Dockerfile-dev
+    command: python3 manage.py runserver --settings=config.settings.docker_dev 0.0.0.0:80
+    environment:
+      - DJANGO_SETTINGS_MODULE=config.settings.docker_dev
+    volumes:
+      - ref_seed_media:/seed/media
+      - ./:/seed/
```
nllong, line 24: 👍

docker-compose.dev.yml
```diff
@@ -0,0 +1,42 @@
...
+      context: .
+      dockerfile: Dockerfile-dev
+    command: watchmedo auto-restart --directory=/seed/seed/ --pattern=*.py --recursive -- /seed/docker/start_celery_docker.sh
+    environment:
+      - DJANGO_SETTINGS_MODULE=config.settings.docker_dev
+    volumes:
+      - ref_seed_media:/seed/media
+      - ./seed:/seed/seed
+volumes:
+  ref_seed_pgdata:
```
nllong, line 37: Any reason this is called ref? I agree that it shouldn't be seed_pgdata as that would be what is on production. Perhaps `dev_seed_pgdata`?

docs/source/setup_docker.rst
```diff
@@ -59,6 +59,85 @@ Building and Configuring Containers
...
+and debugging. It builds off the base docker-compose.yml, so it's necessary
+to specify the files being used in docker-compose commands as seen below.
+
+Build
+^^^^^
+
+.. code-block:: bash
+
+    # create volumes for the database and media directory
+    docker volume create --name=seed_pgdata
```
nllong, line 75: based on the docker file shouldn't this be in the form of {name}-seed_pgdata

seed/tests/test_greenbutton_import.py
```diff
@@ -99,13 +99,13 @@ def test_green_button_import_base_case(self):
 
         self.assertEqual(meter_reading_10.start_time, make_aware(datetime(2011, 3, 5, 21, 0, 0), timezone=self.tz_obj))
         self.assertEqual(meter_reading_10.end_time, make_aware(datetime(2011, 3, 5, 21, 15, 0), timezone=self.tz_obj))
-        self.assertEqual(meter_reading_10.reading, 1790 * 3.41 / 1000)
+        self.assertAlmostEqual(meter_reading_10.reading, 1790 * 3.41 / 1000)
```
nllong, line 102: yay unit conversions.

nllong (approved):

axelstudios (approved):

docker-compose.dev.yml
macintoshpie: This is called ref because we are templating in the actual volume name, so this is just a "reference" to a volume (not the volume name itself), see here: 
https://github.com/SEED-platform/seed/pull/2423/files/b0fdbc1338b7af66d5f6075a919e8560185fc80a#diff-9542f82d64bbeebd91f6236324bfe199e9657e2cb1fd9779d5d6dcdcf9cd4de1R39

docs/source/setup_docker.rst
macintoshpie: Not necessarily, as mentioned above we're templating in the volume name now, with a default value of `seed_pgdata`

PR #2429, opened 2020-10-29
fix(docker): fixed a web volume setting

Ryoken:
#### What's this PR do?
Fixes the docker config so the web volume can find all necessary libraries.


macintoshpie (approved):
Thanks for making this fix, tested locally (in fresh repo clone) and it works as expected.

PR #2431, opened 2020-10-30
fix: updated scripts to use lokalise2

Ryoken:
Regarding lokalise v1:
"This version of CLI tool is depreciated. We will stop supporting requests sent using this version of CLI at November 1st 2020. Use Lokalise CLI v2 instead."
From:
https://docs.lokalise.com/en/articles/1400697-lokalise-cli-v1
#### What's this PR do?
Updates scripts and related files to use v2 instead of v1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,6 +76,7 @@ docs/scripts/.python-version
 docs/scripts/seed_issues.csv
 
 lokalise.cfg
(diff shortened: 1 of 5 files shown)


macintoshpie (approved):
Awesome, thanks for doing this 👍  Tested locally and works as expected

PR #2435, opened 2020-11-02
docs: restore site settings

axelstudios:
When restoring a dump to a different deployment the site's URL and name will likely no longer match (e.g. SEED v2 ->Dev1).  If this URL is incorrect then password reset emails will link to the wrong site (and won't function).

#### What's this PR do?
Adds instructions to `developer_resources.rst` with the steps to update the site name and URL.


nllong (approved):
cool, thanks for adding this in the dev resources.

Issue #2414, opened 2020-09-26
Organizations: Column Mapping: Add sort and filter functionality


RDmitchell:
### Expected Behavior
Would be nice to be able to sort and filter the SEED HEADER and DATA FILE HEADER columns alphabetically. Particularly useful when there are a lot of fields, which can happen with ESPM files. Like in the Column Settings screen
![image](https://user-images.githubusercontent.com/6314950/94329079-aa7b9c80-ff6c-11ea-951b-ca00c556fb80.png)

### Actual Behavior
Can't sort those columns
### Steps to Reproduce
![image](https://user-images.githubusercontent.com/6314950/94329013-38a35300-ff6c-11ea-9200-760e7f4d0f1c.png)

### Instance Information
instance: dev1
SHA: 38fb067
Org: LBNL Test 200

RDmitchell:
@adrian-lara / @nllong -- for you to prioritize, etc

Ryoken:
https://github.com/SEED-platform/seed/pull/2436

RDmitchell:
Instance: dev1
SHA: 58adbb1d2

Looks good
![image](https://user-images.githubusercontent.com/6314950/102261311-5a0a4000-3ec6-11eb-8b7a-068acca0af49.png)

PR #2436, opened 2020-11-03
feat: Added sorting to the column mapping table

Ryoken:
#### What's this PR do?
Adds sorting to the column mapping table.  Clicking on the currently sorted header will toggle the sort direction, while clicking on a different header will change which header is being sorted.
https://github.com/SEED-platform/seed/issues/2414


adrian-lara (approved):
Works as expected! I like the the use of different icons and text-muted!
Screen Shot 2020-11-04 at 1 49 30 PM

PR #2445, opened 2020-11-04
feat(ci): use github actions for ci

macintoshpie:
Travis is mean, and this is a temporary workaround.
EDIT:
Travis is forcing public repos to move from .org to .com, see here: https://docs.travis-ci.com/user/migrate/open-source-repository-migration
We moved seed's repos to the .com site, and then they announced a new pricing model: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
TLDR; is that you have to request credits now if you're OSS, I reached out to them about this but haven't heard back yet.

Important part:
> Building on a public repositories only
We love our OSS teams who choose to build and test using TravisCI and we fully want to support that community. However, in recent months we have encountered significant abuse of the intention of this offering (increased activity of cryptocurrency miners, TOR nodes operators etc.). Abusers have been tying up our build queues and causing performance reductions for everyone. In order to bring the rules back to fair playing grounds, we are implementing some changes for our public build repositories.

> For those of you who have been building on public repositories (on travis-ci.com, with no paid subscription), we will upgrade you to our trial (free) plan with a 10K credit allotment (which allows around 1000 minutes in a Linux environment).
You will not need to change your build definitions when you are pointed to the new plan
When your credit allotment runs out - we’d love for you to consider which of our plans will meet your needs.
We will be offering an allotment of OSS minutes that will be reviewed and allocated on a case by case basis. Should you want to apply for these credits please open a request with Travis CI support stating that you’d like to be considered for the OSS allotment. Please include:
Your account name and VCS provider (like travis-ci.com/github/[your account name] )
How many credits (build minutes) you’d like to request (should your run out of credits again you can repeat the process to request more or discuss a renewable amount)
Usage will be tracked under your account information so that you can better understand how many credits/minutes are being used
#### What's this PR do?
Runs django tests and flake8 using github actions instead of travis. This might not be our long term solution (using github actions), but it will at least keep us rolling for a bit.

What this does _not_ include is:
- functional tests
- tox api test
- tox docs test
- coveralls integration
- some django tests which hit the PM site, we need to add the SEED_PM_UN and SEED_PM_PW to our repo's secrets, see here: https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,14 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+
+jobs:
+  tests:
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build the stack
+        run: docker-compose -f docker-compose.yml -f docker-compose.dev.yml up -d
+      - name: Test
+        run: docker exec -it seed_web_1 python manage.py test --config.settings.docker_dev
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build the stack
+        run: docker volume create --name=seed_pgdata
+        run: docker volume create --name=seed_media
         run: docker-compose -f docker-compose.yml -f docker-compose.dev.yml up -d
       - name: Test
         run: docker exec -it seed_web_1 python manage.py test --config.settings.docker_dev
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
   push:
 
 jobs:
-  tests:
+  django_tests:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


nllong (commented):
So this will run in parallel with the other travis.ci tests? And this run entirely on GH?

nllong:
> So this will run in parallel with the other travis.ci tests? And this run entirely on GH?

And this seems like a direct competitor to Travis... which maybe is a good thing?

macintoshpie:
> > So this will run in parallel with the other travis.ci tests? And this run entirely on GH?
> 
> And this seems like a direct competitor to Travis... which maybe is a good thing?

Yes this will run in parallel with travis, for now. We should remove the protection requirement of Travis passing and add the requirement for Github actions to pass (until we get our CI sorted out)

macintoshpie:
Note that adding coveralls to the process would be pretty trivial, we'd just need to add the coveralls repo token to the secrets (which I don't have permission to do), see here: https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository

adrian-lara:
I'd be good with implementing this for now as to not hold up any ongoing development. 

We can figure out a long-term solution (maybe this ends up being it) in separate conversations.

adrian-lara (approved):

Issue #2215, opened 2020-05-13
Blank option for Inventory Type when SEED HEADER is "blank"


RDmitchell:
### Expected Behavior
In this case where the preset mappings don't have a mapping for a field, the program leaves the SEED HEADER field blank. When I fill in the value for SEED HEADER, and then set the Inventory Type, I expect to see only the options of Property or Tax Lot
### Actual Behavior
However, in this case, there is a third option at the top of the pulldown, which is "blank"
![image](https://user-images.githubusercontent.com/6314950/81773852-208bd500-949e-11ea-84be-e6141d3ead77.png)

Even after I fill in the SEED HEADER Field and then select the pulldown for Inventory Type, there is still the option of leaving it blank (ie, the "blank" option is still in the pulldown list), which doesn't seem very user friendly. Also, the pulldown list is a bit "twitchy" when presenting those options. 
![image](https://user-images.githubusercontent.com/6314950/81773787-fa663500-949d-11ea-9a03-a5949b4c7c8d.png)

### Steps to Reproduce
Import a file and apply a mapping where not all the fields are mapped
### Instance Information
Instance: dev1
SHA: 74b48be4

adrian-lara:
Comment from the PR:
> I think disabling the undefined option is a great fix!
>
> When a user chooses a mapping profile, we don't 'guess' or default to either inventory type, so we leave it blank. When a user goes to change it from blank to "Property" or "Tax Lot", they shouldn't be able to purposely choose 'blank' again.
> 
> New:
> > ![new inventory type select](https://user-images.githubusercontent.com/30608004/99444792-ec282400-28d9-11eb-9320-8a8d1006349f.gif)
> 
> Old: 
> > ![inventory ttype select OLD](https://user-images.githubusercontent.com/30608004/99444818-f2b69b80-28d9-11eb-867a-eaa03cc90536.gif)
>

RDmitchell:
@adrian-lara / @aviveiros11 -- why would there ever be a "blank" option in the Inventory Type pulldown -- I wouldn't think we would ever want that to be blank.

Can that option just be removed from the list?

aviveiros11:
@RDmitchell The blank option from the dropdown comes from an `undefined` value when a Column Mapping Profile is not selected, or when there is not a match between the SEED Header and a suggested inventory type.  If we remove the blank option we would have to set the default value of the dropdowns to either `Property` or `Tax Lot`.

RDmitchell:
[...]

PR #2446, opened 2020-11-05
fix: added a conditional to populate property when selecting a profile

aviveiros11:
Addresses #2215 to fix the blank option
#### What's this PR do?
Adds a conditional statement in the mapping controller to set the Inventory Type after selecting a profile to avoid "Blank" option confusion
Upload a file -> select a default profile -> Inventory Type will be auto-populated with Property
#2215


adrian-lara (commented):
I think we might to limit the cases when this happens. Specifically, this always seems to override the inventory type even if the profile has tax lot across the board.
Screen Shot 2020-11-05 at 4 28 44 PM

adrian-lara (approved):
I think disabling the undefined option is a great fix!

When a user chooses a mapping profile, we don't 'guess' or default to either inventory type, so we leave it blank. When a user goes to change it from blank to "Property" or "Tax Lot", they shouldn't be able to purposely choose 'blank' again.

New:
> new inventory type select

Old: 
> inventory ttype select OLD

PR #2447, opened 2020-11-05
fix: order exported results by property view id

macintoshpie:
The test `test_report_export_excel_workbook` randomly failed, and we were able to reproduce by changing the ordering of property views in `get_raw_report_data`.
#### What's this PR do?
Adds an order_by condition to deterministically order property views


adrian-lara (approved):
LGTM - Thanks for making this change! 👍

Issue #2438, opened 2020-11-03
Build AnalysisRun Model and Migration


adrian-lara:
<img width="202" alt="Screen Shot 2020-11-02 at 11 27 19 PM" src="https://user-images.githubusercontent.com/30608004/97949872-fd622400-1d62-11eb-8ef7-f104b51b72f9.png">

As hinted by the diagram, input_ and output_ files will be added later when the AnalysisFile model is created.

The following statuses should be defined. Note that **order is significant**, and we might want to use enum values that are spaced out in case we want to add statuses in the future (e.g. 10, 20, 30, 40, 50, 60, 70)
1. Creating - validating and preparing AnalysisRun attributes
2. Ready - ready to actually start analysis
3. Queued - Celery job to send analysis has been created but is not yet running
4. Running - Celery job has been executed and waiting for result (or new status)
5. Failed - A previous status transition is unsuccessful.
6. Stopped - user requested the analysis to be stopped before it completed
7. Completed - Successful analysis

Other considerations:
Consideration was given to putting a Cycle FK on the Analysis object instead. In that case, associations between “the same Analysis for the same record” across multiple cycles would be difficult or impossible to track (e.g. Comparing BETTER results for same property in different cycles). Leaving Cycle on this model gives us the flexibility to create an Analysis object for the same record across multiple cycles.

**No manual UI testing is needed for this ticket.**

adrian-lara:
Nothing to test on the UI

PR #2449, opened 2020-11-06
#2438 - feat(analysisrun): create model and migration

adrian-lara:
Part of the larger Analysis epic

#### What's this PR do?
Create the AnalysisRun model

Run migration

#2438


seed/models/analysis_runs.py
```diff
@@ -0,0 +1,38 @@
...
+        (RUNNING, 'Running'),
+        (FAILED, 'Failed'),
+        (STOPPED, 'Stopped'),
+        (COMPLETED, 'Completed'),
+    )
+
+    analysis = models.ForeignKey(Analysis, on_delete=models.CASCADE)
+    cycle = models.ForeignKey(Cycle, on_delete=models.CASCADE)
+    property = models.ForeignKey(Property, on_delete=models.CASCADE)
+    property_state = models.ForeignKey(PropertyState, on_delete=models.CASCADE)
```
adrian-lara, line 36: When deciding to use CASCADE, I went with the approach of completing the following sentence: "If I delete the related [insert FK model], AnalysisRun should... ".

I ended up with "... also be deleted" myself, but I'm open to opposing thoughts.

seed/models/analysis_runs.py
```diff
@@ -0,0 +1,38 @@
...
+
+class AnalysisRun(models.Model):
+
+    CREATING = 10
+    READY = 20
+    QUEUED = 30
+    RUNNING = 40
+    FAILED = 50
+    STOPPED = 60
+    COMPLETED = 70
```
adrian-lara, line 21: @macintoshpie I really like the idea of going with 10, 20, etc. future flexibility. Thanks for suggesting it!

macintoshpie (approved):
Just one comment, but I think this is good to go 👍

macintoshpie (approved):
Awesome 🐞 🐞

Issue #2439, opened 2020-11-03
Build AnalysisFile Model and Migration


adrian-lara:
<img width="648" alt="Screen Shot 2020-11-03 at 10 00 55 AM" src="https://user-images.githubusercontent.com/18518728/98016707-8ad16280-1dbb-11eb-8f7f-e69de7fcd7f3.png">

For now, the first any only file type should be "BuildingSync".

As indicated in the diagram, there should be two one-to-many associations between AnalysisRun and this AnalysisFile model. Specifically, there should be two separate join tables, both setting up the one-AnalysisRun-to-multiple-AnalysisFiles pattern. One would be referred to as "input_files" and the other as "output_files".

~~Note, during usage, we envisioned that the bsyncr case will copy the file field value from an existing BuildingFile object's file field of a BuildingSync media file. Future usage may vary with new analysis type implementations.~~ Copying the file location from a BuildingFile model might not make the most sense since we have to a specify a `upload_to` directory, and that will be different between the two models' fields.
 
**No manual UI testing is needed for this ticket.**

adrian-lara:
Nothing to test on the UI

PR #2450, opened 2020-11-06
feat(analysisfile): create model and migration

adrian-lara:
[Waiting for the AnalysisRun model to be merged to develop]

Part of the larger Analysis epic

#### What's this PR do?
Creates the AnalysisFile model and corresponding migration

Run migration

#2439


seed/models/analysis_files.py
```diff
@@ -0,0 +1,18 @@
...
+    BUILDINGSYNC = 1
+
+    FILE_TYPES = (
+        (BUILDINGSYNC, 'BuildingSync'),
+    )
+
+    file = models.FileField(upload_to="analysis_files", max_length=500)
+    file_type = models.IntegerField(choices=FILE_TYPES)
+    input_for_run = models.ForeignKey(AnalysisRun, on_delete=models.CASCADE, related_name='input_files', blank=True, null=True)
+    output_for_run = models.ForeignKey(AnalysisRun, on_delete=models.CASCADE, related_name='output_files', blank=True, null=True)
```
adrian-lara, line 18: Looking for feedback and or other ideas on the `_for_run` names. Technically, an individual AnalysisFile is an "input/output for a run".

adrian-lara:
I'm wondering if anyone has an opinion on whether we should have a follow up PR to this that combines the 3 Analysis migrations together into one migration. Thoughts??

macintoshpie:
> I'm wondering if anyone has an opinion on whether we should have a follow up PR to this that combines the 3 Analysis migrations together into one migration. Thoughts??

Hmm from my point of viewI don't see a great advantage of merging these. It's very database specific and automated, I view it kind of like the commits of the database, and each of these migrations is a meaningful and distinct "commit". The advantages I can think of from merging would be:
- fewer files (marginal advantage since this is in a directory of mostly auto-generated files)
- if rolling back, with desire to undo analysis stuff, only need to care about 1 migration

I could go either way I guess, but might prefer keeping them separate unless theres a really good reason

seed/models/analysis_files.py
macintoshpie: I can't think of any better ones, I like this 👍

macintoshpie (approved):
Looks great 👍

Issue #2327, opened 2020-07-22
Unit-aware columns had display names appended with display units


adrian-lara:
### TL;DR
 Should we bring back column display names being shown with pints_units appended?

### Background
There's evidence that in previous versions of SEED (maybe as far back as 2017), pints-unit-aware columns had their display names appended with  ` (<pints unit>)` for the "list" column endpoint.

Note, I haven't seen any issues or heard any conversations around this being a desired feature.

### Research - Technical details
https://github.com/SEED-platform/seed/blob/9e23e67342cf33c156dfe7b8b4983c4b4919b022/seed/serializers/pint.py#L81-L105

The column "list" endpoint uses this method, but provides it with a dictionary of column info with snake_case keys whereas the method expects camelCase keys.

### Some fix options
- If **there's no desire** to re-implement this feature,

I'd suggest we remove the use of `add_pint_unit_suffix` in the "list" endpoints as it's, at best, misleading, at worst, an easy entry point for bugs if later changes occur.

- If **there is a desire** to re-implement this feature,

I suggest this happens with the introduction of API v3, where a new query param is introduced in order to enable or disable the append logic. This would be needed since the unit should only be added as a suffix in read-only cases. 

This would include the inventory list page, inventory list profile page, and likely others. I would suggest that the column admin page would not be included since the display names can actually be changed here, and the "last-minute" addition of the suffix would add unnecessary complications. 

A caveat here would be that the confirm mapping page does not yet show the incoming data with appropriate pint conversions. We could either avoid append suffixes to columns, or change it so the incoming data shows the conversion at this point in the import process. For example, when importing Occupied Floor Area as square feet but displaying as meters squared, the confirm mapping page shows the areas as the original square feet value, without conversion, while the column would read "Occupied Floor Area (m[^2])".

### Instance Information
Server Instance: prod, dev1, any recent version of SEED

adrian-lara:
@nllong @axelstudios @RDmitchell 
Any historical context or other thoughts on previously being able to see display units for applicable columns - `Conditioned Floor Area (m²)`, `Site EUI (kBtu/a/ft²)`, etc.

RDmitchell:
@adrian-lara -- I was actually thinking about this when I was testing the units aware fields. It's hard to tell which are which, ie units non aware and units aware, when the units are not added to the field. 

[...]

PR #2453, opened 2020-11-10
Fix/restore column units

Ryoken:
#### What's this PR do?
Restores displaying units in column headers.
Check any inventory with units.
https://github.com/SEED-platform/seed/issues/2327


seed/serializers/pint.py
```diff
@@ -78,13 +78,14 @@ def pretty_units_from_spec(unit_spec):
...
 
-def add_pint_unit_suffix(organization, column):
+def add_pint_unit_suffix(organization, column, data_key="data_type", display_key="display_name"):
     """
     transforms the displayName coming from `Column.retrieve_all` to add known
     units where applicable,  eg. 'Gross Floor Area' to 'Gross Floor Area (sq.
     ft.)', using the organization's unit preferences.
     """
 
+    # add units to the end of the field
```
adrian-lara, line 88: Sorry to be picky - think we could remove the comments on lines 88, 97, 103, and 114? As much as we can, we've tried to lean towards descriptive method and variable names if possible in this codebase (i.e. if you thought `format_column_name` wasn't descriptive enough, I'd be cool if you changed it since it's an inner function - maybe `append_display_units` 🤷 .) Ideally, we'd reserve in-line comments for odd or one-off clarifications.

seed/tests/test_column_views.py
```diff
@@ -152,6 +152,28 @@ def test_get_all_columns(self):
...
+    def test_column_units(self):
+
+        # get the columns list without units
+        responseWithoutUnits = self.client.get(reverse('api:v3:columns-list'), {
+            'organization_id': self.org.id,
+            'display_units': 'false'
+        })
+
+        # get an entry with EUI units
+        columnWithoutUnits = next((x for x in json.loads(responseWithoutUnits.content)['columns'] if x['data_type'] == 'eui'), None)
```
adrian-lara, line 165: In this case, since the set of Column objects that are created on initial org creation is well defined, I'd vote to try to isolate a specific Column object (vs the first one that returns a 'data_type' of 'eui'). That way you can assert against a string (`assertEqual(..., "Site EUI")` in both the with and without cases.

IMO, that would add clarity and consistency to these tests.

seed/views/v3/columns.py
```diff
@@ -106,7 +103,7 @@ def list(self, request):
         only_used = json.loads(request.query_params.get('only_used', 'false'))
         columns = Column.retrieve_all(organization_id, inventory_type, only_used)
         organization = Organization.objects.get(pk=organization_id)
-        if json.loads(request.query_params.get('display_units', 'false')):
```
adrian-lara, line 109: 👍 
The only thing we'd have to change is on the frontend. Specifically on the organization column settings page, we shouldn't see units appended to display names as the display names are editable.
Screen Shot 2020-11-12 at 3 39 11 PM

adrian-lara (commented):
Works as expected! I have some small changes in the comments 😄

seed/tests/test_column_views.py
```diff
@@ -152,6 +152,28 @@ def test_get_all_columns(self):
         # randomly check a column
         self.assertIn(expected, data)
 
+    # tests that units get added to columns
+    def test_column_units(self):
```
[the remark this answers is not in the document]
adrian-lara: oops - this comment definitely only applies to tests 😅

seed/views/v3/columns.py
Ryoken: Screen Shot 2020-11-12 at 4 46 24 PM
Screen Shot 2020-11-12 at 4 47 08 PM
Note that display units is false for the settings page and true for the inventory list.

seed/tests/test_column_views.py
Ryoken: I've now selected a specific record and am testing off of that.

seed/serializers/pint.py
Ryoken: Sorry!  I've always been in the habbit of adding comments to all of my code, rather than simply to be more descriptive.

seed/tests/test_column_views.py
```diff
@@ -152,6 +152,23 @@ def test_get_all_columns(self):
         # randomly check a column
         self.assertIn(expected, data)
 
+    def test_column_units_added(self):
+
+        responseWithoutUnits = self.client.get(reverse('api:v3:columns-list'), {
+            'organization_id': self.org.id,
+            'display_units': 'false'
+        })
+        columnWithoutUnits = next((x for x in json.loads(responseWithoutUnits.content)['columns'] if x['id'] == 61), None)
```
adrian-lara, line 161: Sorry - I definitely led you down this path with my last comment. I was envisioning using something like, for the column you chose,  `x['column_name'] == 'site_eui_modeled'`. The ID won't be consistent because of all the tear downs that happen between tests when this test is run with the full test suite.

To clarify my previous "well-defined" comment: when a new Organization is created, it always comes with several Column objects that have the `units` field populated. These Column objects have `column_name` values of "site_eui", "site_eui_modeled", "source_eui" among a few others.

adrian-lara (commented):
Last item is the test - otherwise, this works as expected, and I can confirm that we no longer see units appended to display names on the org column settings page 💯

seed/tests/test_column_views.py
```diff
@@ -152,6 +152,22 @@ def test_get_all_columns(self):
...
+        })
+        columnWithoutUnits = next((x for x in json.loads(responseWithoutUnits.content)['columns'] if x['display_name'] == 'Source EUI Modeled'), None)
+        self.assertEqual(columnWithoutUnits['display_name'], 'Source EUI Modeled')
+
+        responseWithUnits = self.client.get(reverse('api:v3:columns-list'), {
+            'organization_id': self.org.id,
+            'display_units': 'true'
+        })
+        columnWithUnits = next((x for x in json.loads(responseWithUnits.content)['columns'] if x['id'] == columnWithoutUnits['id']), None)
+        self.assertEqual(columnWithUnits['display_name'], 'Source EUI Modeled (kBtu/ft²/year)')
```
adrian-lara, line 169: 👍

seed/tests/test_column_views.py
```diff
@@ -152,6 +152,22 @@ def test_get_all_columns(self):
...
         self.assertIn(expected, data)
 
+    def test_column_units_added(self):
+
+        responseWithoutUnits = self.client.get(reverse('api:v3:columns-list'), {
+            'organization_id': self.org.id,
+            'display_units': 'false'
+        })
+        columnWithoutUnits = next((x for x in json.loads(responseWithoutUnits.content)['columns'] if x['display_name'] == 'Source EUI Modeled'), None)
+        self.assertEqual(columnWithoutUnits['display_name'], 'Source EUI Modeled')
```
adrian-lara, line 162: If you're using `display_name` to capture the dictionary representation of a desired column object, I don't think it's advisable to make an assertion on the `display_name` value. In a sense, it's kind of like testing the `next()` Python method as well as the existence of a column with that `display_name` - that existence check is valuable though.

Maybe we can use `x['column_name'] == 'source_eui_modeled'` instead?

adrian-lara (commented):
Suggesting one really small change - thanks!

seed/tests/test_column_views.py
Ryoken: Good point, thanks for catching this.

adrian-lara (approved):
Thanks! LGTM!

Issue #1869, opened 2019-05-04
Geocoding -- allow user to "uncheck" geocoding without deleting API key


RDmitchell:
### Expected Behavior
It would be nice to be able to turn the geocoding (generating lat/long) functionality on and off without having to delete the API key. Maybe just a checkbox in Org/Setings, where the MapQuest API key is set, to indicate whether to do the geocoding or not. 
### Actual Behavior 
In order to turn off geocoding (which I want to do for some testing) I have to delete the API key. Which means when I want to turn it back on, I have to go find the key and put it back.
### Steps to Reproduce
n/a
### Instance Information
Instance: seeddemostaging
SHA: d2b5ef67 (2.5.2)

Ryoken:
https://github.com/SEED-platform/seed/pull/2454

RDmitchell:
instance: dev1
SHA; 
Org: LBNL Test 200
Cycle: Test MapQuest API

* good MapQuest API key
* uncheck Enable Geocoding
![image](https://user-images.githubusercontent.com/6314950/102675504-81733e00-414e-11eb-9afc-caedcc9ff70e.png)
* import a file that has good addresses and should be able to generate lat/long if geocoding is enabled (see #1978 ) but because Geocoding is disabled, the geocoding fields are empty.
![image](https://user-images.githubusercontent.com/6314950/102675626-fa729580-414e-11eb-9d34-4ff6562d4e70.png)

PR #2454, opened 2020-11-11
Feat/disable geocoding option

Ryoken:
#### What's this PR do?
Adds a setting to organizations that disables geocoding.
https://github.com/SEED-platform/seed/issues/1869


seed/static/seed/js/controllers/mapping_controller.js
```diff
@@ -502,16 +502,20 @@ angular.module('BE.seed.controller.mapping', [])
...
-          $scope.taxlot_geocoding_columns_array = _.map(geocoding_columns.TaxLotState, function (column_name) {
-            return _.find($scope.mappable_taxlot_columns, {column_name: column_name}).display_name;
-          });
-          $scope.taxlot_geocoding_columns = $scope.taxlot_geocoding_columns_array.join(', ');
+              $scope.taxlot_geocoding_columns_array = _.map(geocoding_columns.TaxLotState, function (column_name) {
+                return _.find($scope.mappable_taxlot_columns, {column_name: column_name}).display_name;
+              });
+              $scope.taxlot_geocoding_columns = $scope.taxlot_geocoding_columns_array.join(', ');
+            });
+          }
```
adrian-lara, line 518: Left is activated and right is deactivated. Nice catch to hide the relevant help text! 
Screen Shot 2020-11-11 at 5 29 35 PM

adrian-lara (commented):
This looks great so far! One case that wasn't mentioned to you previously: Users can geocode records on the inventory list page.
Screen Shot 2020-11-11 at 5 04 42 PM

Right now, if I have an API key but disable geocoding, I see this and can click "Geocode".
Screen Shot 2020-11-11 at 5 05 01 PM

Maybe it's worth doing the enabled/disabled check and adding to the message we show when there isn't an API key.
Screen Shot 2020-11-11 at 5 38 32 PM


Otherwise this looks good during the import flow! See code comment and the following screenshot showing the results of importing the same file with geocoding enabled (left) and disabled (right):
Screen Shot 2020-11-11 at 5 35 21 PM

seed/static/seed/js/controllers/mapping_controller.js
Ryoken: Added a message to the geocode modal if disabled.

adrian-lara (approved):
Looks good and works as expected! Thanks!

The last requested change:
Screen Shot 2020-11-12 at 4 08 15 PM

Issue #2079, opened 2020-01-18
Gateway timeout (504) importing a file with 120,000+ records


RDmitchell:
### Expected Behavior
Would like to be able to import large files
### Actual Behavior
Large file (122,000+ records, 4 fields) on the staging server got through the import and mapping review, but then timed out in the "Finding inventory matches" section. 
### Steps to Reproduce
See the doc in this folder, along with a file that shows the problem
https://drive.google.com/drive/folders/1rH8tmoVjA09GaSjLu2-k4Z-DBqbGk25m?usp=sharing
### Instance Information
Instance: LBNL staging
SHA: 395db33d

RDmitchell:
Instance: dev1
SHA: 58adbb1d2
Org: LBNL Test 200
Cycle: Import Large File

Get a 520 error when trying to get to the Mapping page with the file
![image](https://user-images.githubusercontent.com/6314950/102675952-3ce8a200-4150-11eb-9e3b-8cd78821cb7c.png)

@adrian-lara -- maybe this is an issue with the dev1 setup, and might not be a problem on staging?

macintoshpie:
Thanks Robin, this is technically a slightly different issue than the original issue mentioned here (happening at a different step in the import process). Maybe we could move this conversation to #2515  or #2516 ?

But to answer your question, I don't believe this would be an issue specific to dev1.

RDmitchell:
@adrian-lara / @nllong / @macintoshpie -- I guess I don't know how I am supposed to test this since I get a 520 error during the Mapping step, before the program gets to the Mapping Review screen.

macintoshpie:
@RDmitchell are you starting from the beginning (ie uploading the file) rather than trying to remap the file? If you are trying to "remap" the file, then you will probably run into 520's since we currently do something slow on the backend before returning a response (which is why you would see the 520).

I'm attempting to reupload and map the original file myself and will report my results. We will definitely need to fix the `/map/` endpoint soon though to fix this problem.

RDmitchell:
@macintoshpie 

I started over importing the file from scratch. I did it twice with different results. In both cases the import eventually failed, but in different ways.

See this doc
https://docs.google.com/document/d/1NDuoPVlckv5FF1hjCyIBFJrnVzw9wm49XeAVTePj_qM/edit?usp=sharing

macintoshpie:
[...]

PR #2457, opened 2020-11-12
#2079 refactor(tasks): run geocoding task in celery

macintoshpie:
When importing a large file, you can experience timeouts when starting the final saving process by clicking "Save Mappings". This was happening because in the `start_system_matching_and_geocoding` endpoint, we would start the geocoding task, then block until we got the result, before finally kicking off the matching task and returning the progress data. If the geocoding task took a while to run, or was blocked itself because the worker(s) were already occupied (data quality checks are probably still running for a large file) then it's going to block for a long time. In our case, ELB would timeout after 60s and return an empty HTTP response (which cloudflare doesn't like, and ends up giving a 520 response).

#### What's this PR do?
This address the timeout by not blocking on the geocoding task. Specifically, I chained together the geocoding and matching tasks because the should always happen together.

Because I created a new task for combining the previous tasks, you need to test a few things. It touches geocoding as well as the mapping of additional BuildingSync models (which occurs instead of the buildings matching).
Please test:
- upload a spreadsheet that won't be geocoded (example-pm-properties) and verify it was uploaded properly
- upload a spreadsheet that will be geocoded, and verify it was geocoded successfully (must have valid MapQuest API key in org settings)
- upload a spreadsheet that will be geocoded, but have an INVALID MapQuest API key in org settings; you should get an error reporting a bad or overused key
- upload a BSync file that has measure data (buildingsync_ex01_measures.xml), and verify the Measures were successfully created by checking the property detail page and scrolling to the Measures section

#2079

### DIFF
--- a/seed/data_importer/tasks.py
+++ b/seed/data_importer/tasks.py
@@ -22,6 +22,7 @@ from math import ceil
 import zipfile
 
 from celery import chord, shared_task
+from celery import chain as celery_chain
 from celery.utils.log import get_task_logger
 from django.contrib.gis.geos import GEOSGeometry
 from django.core.files.uploadedfile import SimpleUploadedFile
@@ -79,7 +80,7 @@ from seed.models.data_quality import (
     Rule,
 )
 from seed.utils.buildings import get_source_type
-from seed.utils.geocode import geocode_buildings
+from seed.utils.geocode import geocode_buildings, MapQuestAPIKeyError
 from seed.utils.ubid import decode_unique_ids
 
 # from seed.utils.cprofile import cprofile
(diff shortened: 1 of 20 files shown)


seed/data_importer/tasks.py
```diff
@@ -1119,30 +1120,101 @@ def save_raw_data(file_pk):
...
+
+    if not import_file.mapping_done:
+        _log.debug('Mapping is not done yet')
+        return progress_data.finish_with_error(
+            'Import file is not complete. Retry after mapping is complete', )
+
+    post_geocode_tasks = None
+    if import_file.from_buildingsync:
+        if import_file.cycle is None:
+            _log.warn("This should never happen in production")
```
adrian-lara, line 1141: Nbd, but from what I can tell in `match_buildings()` and `map_additional_models()` I think we want this check to happen for either case - BuildingSync and else.

adrian-lara (approved):
This works as expected! One small comment, but otherwise the code changes look good as well.

Tested the following:
- import without geocoding, matching, merging, linking
- import without geocoding & with matching, merging, linking
- import with geocoding & without matching, merging, linking
- import with geocoding, matching, merging, linking
- import a BuildingSync file with measures

seed/data_importer/tasks.py
macintoshpie: Nice catch

seed/data_importer/tasks.py
```diff
@@ -1119,30 +1120,106 @@ def save_raw_data(file_pk):
...
-    async_result = _geocode_properties_or_tax_lots.s(file_pk).apply_async()
+    """
+    NOTE: This is an older entrypoint into geocoding buildings and should no longer
+    be used. Use geocode_and_match_buildings_task instead.
+    TODO: remove this task once api v2 is removed
+    """
+    progress_data = ProgressData(func_name='geocode_buildings', unique_id=file_pk)
+    progress_data.delete()
+    progress_data.save()
+    async_result = _geocode_properties_or_tax_lots.s(file_pk, progress_data.key).apply_async()
```
adrian-lara, line 1188: 💯

adrian-lara (approved):
Thanks for making those changes! Feel free to merge once CI finishes.

PR #2458, opened 2020-11-13
Change ESPM status url for unit tests

nllong:
Previous domain is going to expire.

#### What's this PR do?
Use new URL.

Tests should pass, assuming we have credits left 😬

N/A


seed/tests/test_portfoliomanager.py
```diff
@@ -29,7 +29,7 @@
 
 # override this decorator for more pressing conditions
 try:
-    pm_avail_check = requests.get('http://isthewallbuilt.us/api.json', timeout=5)
+    pm_avail_check = requests.get('http://isthewallbuilt.inbelievable.com/api.json', timeout=5)
```
nllong, line 32: @Myoldmopar -- good?

Myoldmopar: I get the expected response when querying that url, looks like it should be good for a while.  :+1:

macintoshpie:
Waiting until I have fixed the CI to make sure we're properly running PM tests. Will merge after that

PR #2460, opened 2020-11-16
build: update GitHub actions

macintoshpie:
Previous GH Actions did not run all tests
#### What's this PR do?
Runs all tests and adds a build workflow for tagging and pushing dev and master branch to dockerhub when they're updated.

Note these differences from our previous Travis CI build:
- many of the tests now use docker-compose for running services (server, postgres, redis, etc). Github actions does not offer services in the same way travis does, it only offers running docker containers. We _could_ install postgres on every build but that might be pretty slow. So instead of having GH actions run a container for us we just run our own docker-compose. Thus most of the tests are now `docker exec`'d within the web container.

Known issues:
There are no testing issues, however the caching of docker layers isn't too efficient. For some reason, docker always wants to rebuild from the `COPY . /seed` step, even if there's a direct cache hit in GH actions (ie from same sha). This might be worth looking into eventually b/c it takes an extra minute or two to build because of this.


seed/static/seed/tests/mapping_controller.spec.js
```diff
@@ -37,6 +37,11 @@ describe('controller: mapping_controller', function () {
           });
         });
 
+      spyOn(mock_geocode_service, 'check_org_has_geocoding_enabled')
+        .andCallFake(function () {
+          return $q.resolve(true);
+      });
+
```
macintoshpie, line 44: I wanted to point out I had to add this b/c we recently merged in some changes to the mapping_controller that didn't have the functional tests applied in CI b/c only the django tests were running on github actions at the time. This mock fixes the test.

seed/static/seed/tests/protractor-tests/protractorConfig.js
```diff
@@ -33,20 +23,6 @@ exports.config = {
...
-    chromeOptions: {
-      args: [
-        '--headless',
-        '--disable-gpu',
-        '--no-sandbox',
-        '--window-size=800,600',
-        '--disable-extensions',
-        '--disable-dev-shm-usage'
-      ]
-    }
```
macintoshpie, line 46: We don't actually run our CI tests headless. On travis we ran protractorConfigCoverage.js, which is essentially this exact file but not headless. I deleted that file and updated this one to consolidate things.

aviveiros11 (approved):
lgtm!

.github/workflows/ci.yml
```diff
@@ -5,33 +5,102 @@ on:
...
+      -
+        uses: actions/setup-node@v1
+      -
+        name: Install dependencies
+        run: |
+          npm install
+          sudo apt-get update
+          sudo apt-get install -y xvfb
+          wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+          sudo apt install ./google-chrome-stable_current_amd64.deb
```
axelstudios, line 57: @macintoshpie Mixed `apt`/`apt-get` usage, I'd probably just go with all `apt` personally

nllong (approved):
ciao travis!

Good job getting this up-and-running. Only 15 more repos to go (joking/not-joking).

axelstudios (approved):

PR #2465, opened 2020-11-20
fix: add nginx/uwsgi protections for oserrors and ioerrors

axelstudios:
We've seen occasional error messages over time like `OSError: write error`.  uWSGI documentation classifies this primarily as an "annoyance", like if the user disconnects and nginx closes the socket but uWSGI is still working.  The following SO links suggest config improvements to handle these:
https://stackoverflow.com/a/45393743/2730450
https://stackoverflow.com/a/40375474/2730450

#### What's this PR do?
Adds the recommended settings based on SO results

### DIFF
--- a/docker/nginx-seed.conf
+++ b/docker/nginx-seed.conf
@@ -64,6 +64,10 @@ server {
         uwsgi_pass  seed_upsteam;
         uwsgi_read_timeout 600;
         uwsgi_send_timeout 600;
+
+        # https://stackoverflow.com/a/40375474/2730450
+        uwsgi_ignore_client_abort on;
+
         include     /etc/nginx/uwsgi_params;
     }
 }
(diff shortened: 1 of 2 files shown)


docker/nginx-seed.conf
```diff
@@ -64,6 +64,10 @@ server {
         uwsgi_pass  seed_upsteam;
         uwsgi_read_timeout 600;
         uwsgi_send_timeout 600;
+
+        # https://stackoverflow.com/a/40375474/2730450
+        uwsgi_ignore_client_abort on;
```
nllong, line 69: 👍

nllong (approved):
this will be a nice update! Good fix.

adrian-lara:
We should also throw this into develop (and 2.8.0?)

axelstudios:
Yep, I'll cherry-pick it

Issue #2466, opened 2020-11-23
Update Analysis-related models


adrian-lara:
Per separate conversations, we want to implement the following schema for Analysis-related models: (EDIT to update the image)
![Screen Shot 2020-11-25 at 11 58 40 AM](https://user-images.githubusercontent.com/30608004/100270729-bad0d900-2f15-11eb-96d5-dba448cdb83b.png)

The previous migrations can be scrapped, since as of this writing, those migrations were not run in any deployment.

Ryoken:
https://github.com/SEED-platform/seed/pull/2467

adrian-lara:
Note: For this ticket, there's nothing to explicitly test on the UI.

PR #2467, opened 2020-11-23
Feat/analyses models

Ryoken:
#### What's this PR do?
Updates and adds relevant analysis models.
https://github.com/SEED-platform/seed/issues/2466


seed/models/analysis_messages.py
```diff
@@ -0,0 +1,27 @@
...
+    during an analysis, like a breadcrumb trail.
+    """
+    # if the message is relevant to the entire analysis, ie not property specific,
+    # then it should be linked to an analysis and not linked to an AnalysisPropertyView
+    # e.g. the API for the analysis service returned a non 200 response
+    analysis = models.ForeignKey(Analysis, on_delete=models.CASCADE)
+    # if the message is relevant to a specific property then it should be linked
+    # to an AnalysisPropertyView
+    # e.g. if the AnalysisPropertyView is missing some required data
+    analysis_property_view = models.ForeignKey(AnalysisPropertyView, on_delete=models.CASCADE)
```
macintoshpie, line 22: I'm wondering what our requirements for this should be (we didn't really specify in our diagram).

Should a message always be linked to an analysis? Or should it be either-or with a linked analysis property view? I think we should always require analysis

I'm thinking analysis_property_view should be optional.

@adrian-lara @Ryoken Thoughts?

macintoshpie (changes requested):
This looks good! I have a few questions for yall and I think I'd prefer moving the `AnalysisTypes` enums into their respective classes.

seed/models/analysis_output_files.py
```diff
@@ -0,0 +1,13 @@
...
+
+from seed.models import (
+    AnalysisPropertyView,
+    AnalysisTypes
+)
+
+
+class AnalysisOutputFile(models.Model):
+
+    file = models.FileField(upload_to="analysis_input_files", max_length=500)
```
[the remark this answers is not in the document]
adrian-lara: I like them in separate sub directories, but I don't know the process of backing up these files and whether this would add work for others during backup. @axelstudios ?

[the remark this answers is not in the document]
Ryoken: If we decide to keep them in one directory I'll rename it to "analysis_files" as well.

seed/models/analysis_types.py
```diff
@@ -0,0 +1,43 @@
+# !/usr/bin/env python
+# encoding: utf-8
+"""
+:copyright (c) 2014 - 2020, The Regents of the University of California, through Lawrence Berkeley National Laboratory (subject to receipt of any required approvals from the U.S. Department of Energy) and contributors. All rights reserved.  # NOQA
+:author
+"""
+
+
+class AnalysisTypes:
```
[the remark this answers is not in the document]
Ryoken: Makes sense, I'll split them out and give that enum format a look-over.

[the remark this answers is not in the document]
adrian-lara: I don't think 2.2 has enumeration types 😬 

I'd also vote to keep the enum definitions into each specific Django model. There are cases in the past where we consolidate related enum definitions into a central location, and I've always just found it easier when they were managed within the individual classes as needed.

seed/models/analysis_messages.py
adrian-lara: AnalysisPropertyView relation should definitely be optional. I'd lean towards making the Analysis relation required just so we can't ever have a hanging Message?

Ryoken: In the past when there were mulitple types of parents like this we'd use a pair of fields like "parent" and "parent_type".  That way we could still enforce having to have a parent, but I'm not sure how all that works with Django.  I agree on keeping analysis required and making analasys_property_view optional, but keeping both delete triggers.

seed/models/analysis_types.py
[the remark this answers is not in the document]
macintoshpie: Ah, I thought we were on 3! I'm good with this

seed/models/analysis_messages.py
macintoshpie: Hmm, re: parent fields, there might be some type of polymorphism thing we could implement in Django, but might be too much work. I'm good with requiring the analysis link, and making the AnalysisPropertyView optional 👍

macintoshpie (approved):
Looks good to me 👍

seed/models/analyses.py
```diff
@@ -1,21 +1,46 @@
...
+    The Analysis represents an analysis performed on one or more properties.
+    """
     BSYNCR = 1
 
-    ANALYSIS_TYPES = (
+    SERVICE_TYPES = (
         (BSYNCR, 'BSyncr'),
     )
 
-    name = models.CharField(max_length=255, blank=False, default=None)
```
adrian-lara, line 16: Oops - We should have a `name` on this model as well. That was our bad missing it in the initial diagrams 😬

seed/models/analysis_input_files.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    The AnalysisInputFile is a file used as input for an analysis.
+
+    For example, if running an analysis on multiple properties, it might be a
+    CSV containing data collected from each property.
+    """
+    DEFAULT = 1
+
+    CONTENT_TYPES = (
+        (DEFAULT, 'default'),
```
adrian-lara, line 16: What's meant by `default`? I was expecting to see something BSyncr-specific. (Same question for AnalysisOutputFile)

adrian-lara (changes requested):
One small change and a question.

seed/models/analysis_input_files.py
macintoshpie: I'm not too worried about this, I was thinking I'd just replace this once I start working on integrating with BSyncr since I don't know what I want to call the files for now

seed/models/analyses.py
macintoshpie: good catch, I'll add that real quick

seed/models/analysis_messages.py
```diff
@@ -0,0 +1,31 @@
...
+class AnalysisMessage(models.Model):
+    """
+    The AnalysisMessage represents user-facing messages of events that occur
+    during an analysis, like a breadcrumb trail.
+    """
+    DEFAULT = 1
+
+    MESSAGE_TYPES = (
+        (DEFAULT, 'default'),
+    )
```
adrian-lara, line 18: Sorry - missed this on my first pass. It's unclear what `default` stands for here as well. I do understand us not necessarily knowing what the different types should be yet, so I'm thinking this will just be an additional migration once we get to implementing these messages. No suggested change for me now on this one.

(We could modify the 0131 migration later, but that would mean we have to perform rollbacks on migrations in develop and locally.)

seed/models/analysis_messages.py
```diff
@@ -0,0 +1,31 @@
...
+    # then it should be linked to an analysis and not linked to an AnalysisPropertyView.
+    # e.g. the API for the analysis service returned a non 200 response
+    analysis = models.ForeignKey(Analysis, on_delete=models.CASCADE)
+    # if the message is relevant to a specific property then it should be linked
+    # to an AnalysisPropertyView
+    # e.g. if the AnalysisPropertyView is missing some required data
+    analysis_property_view = models.ForeignKey(AnalysisPropertyView, on_delete=models.CASCADE, null=True, blank=True)
+    type = models.IntegerField(choices=MESSAGE_TYPES)
+    # human-readable message which is presented on the frontend
+    user_message = models.CharField(max_length=255, blank=False, default=None)
```
adrian-lara, line 29: I'm fine with this, but just wanted to make this note: I'm wondering if we'll ever need to have messages that are purely only meant for developers - would be easy enough to make a migration to make this to `blank=True` though.

adrian-lara (approved):
Changes look good! A few comments/notes, but I would be fine with this being merged in as is.

seed/models/analysis_messages.py
macintoshpie: I think that's when we should use logging. I am even considering removing the debug_message field, but I guess we'll see if it's useful

PR #2468, opened 2020-11-23
fix(views/labels): require non-empty list for inventory_ids

adrian-lara:
When providing an empty list to the `inventory_ids` body param, SEED would default to trying to apply labels to all PropertyViews across all Orgs. A validation was in place to prevent this application, but we'd like to remove this default behavior and check against it.

#### What's this PR do?
Fixes this in production.

### DIFF
--- a/seed/views/labels.py
+++ b/seed/views/labels.py
@@ -144,6 +144,10 @@ class UpdateInventoryLabelsAPIView(APIView):
         'missing_org': ErrorState(
             status.HTTP_422_UNPROCESSABLE_ENTITY,
             'missing organization_id'
+        ),
+        'missing_inventory_ids': ErrorState(
+            status.HTTP_422_UNPROCESSABLE_ENTITY,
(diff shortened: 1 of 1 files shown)


seed/views/labels.py
```diff
@@ -218,9 +222,6 @@ def add_labels(self, qs, inventory_type, inventory_ids, add_label_ids):
             model = self.models[inventory_type]
             inventory_model = self.inventory_models[inventory_type]
             exclude = self.exclude(qs, inventory_type, add_label_ids)
-            inventory_ids = inventory_ids if inventory_ids else [
```
adrian-lara, line 221: This was just a reassignment, so I just removed it. Otherwise, all code changes mirror this: https://github.com/SEED-platform/seed/commit/e5e149d4809e3870582a39b2a91906e848c09679

macintoshpie (approved):
Looks good to me 👍

Issue #2440, opened 2020-11-03
Add API v3 endpoints /analyses GET and /analyses/{id}/ GET


adrian-lara:
# /analyses
### Permissions
Org member

### Body
None

### Request Type
GET

### Params
- property_id (filter Analysis objects by this) - optional
- organization_id

### Incorporate logic
- Include summarizing each set of AnalysisRun statuses
    - This needs to be built out (likely a function on the Analysis model). For now, this can be as simple as returning the “lowest” enum status. (e.g. out of Completed, Running, Stopped, "Running" would be returned as it has a enum value of 4 against Completed, 7, and Stopped, 6)
- Append all cycle_ids to results (query -Runs and return unique cycle_ids) (likely a function on the Analysis model)
- Append count of associated AnalysisRuns (likely a function on the Analysis model)

### Returns
JSON with status and data (Analysis[])

# /analyses/{id}
### Permissions
Org member

### Body
none

### Request method
GET

### Query Params
analysis_id
organization_id (query param)

### Incorporate logic (same as /analyses - GET)
- Include summarizing each set of AnalysisRun statuses
    - This needs to be built out (likely a function on the Analysis model). For now, this can be as simple as returning the “lowest” enum status. (e.g. out of Completed, Running, Stopped, "Running" would be returned as it has a enum value of 4 against Completed, 7, and Stopped, 6)
- Append all cycle_ids to results (query -Runs and return unique cycle_ids) (likely a function on the Analysis model)
- Append count of associated AnalysisRuns (likely a function on the Analysis model)

### Returns
JSON with status and data (Analysis object)

Ryoken:
The Analysis API is supposed to return all the relevant cycle IDs, which are stored on the AnalysisPropertyViews.  The ticket would like to see this functionality live on the Analysis model, but AnalysisPropertyViews already inherits Analysis in the model, so a circular dependency is introduced.

Ryoken:
Previously, there was an AnalysisRun that had a status on it, so the Analysis endpoint was to collect and report the lowest status.  However, status is now on the Analysis model.  Is there no longer a need to aggregate status?

Ryoken:
Does the /analyses/{id} still need an organization id?  An organization is required on each Analysis.

adrian-lara:
[...]

PR #2473, opened 2020-11-30
Feat/analyses endpoints

Ryoken:
#### What's this PR do?
Adds endpoints for getting Analysis information.
Two endpoints should be tested.
https://github.com/SEED-platform/seed/issues/2440


seed/views/v3/analyses.py
```diff
@@ -0,0 +1,71 @@
...
+        return JsonResponse({
+            'status': 'success',
+            'analyses': analyses
+        })
+
+    @swagger_auto_schema(manual_parameters=[AutoSchemaHelper.query_org_id_field(True)])
+    @require_organization_id_class
+    @api_endpoint_class
+    @ajax_request_class
+    def retrieve(self, request, pk):
```
macintoshpie, line 55: I looks like this endpoint may still need a check to determine if the requesting user is part of the organization

seed/models/analyses.py
```diff
@@ -45,3 +51,14 @@ class Analysis(models.Model):
     # that are applicable to the entire analysis (ie all properties involved).
     # For property-specific results, use the AnalysisPropertyView's parsed_results
     parsed_results = JSONField(default=dict, blank=True)
+
+    def getPropertyViewInfo(self, property_id=None):
+        if property_id is None:
+            analysis_property_views = self.analysispropertyview_set.filter(analysis=self.id)
```
macintoshpie, line 57: I think `self.analysispropertyview_set.filter(analysis=self.id)` should return the same thing as `self.analysispropertyview_set`. Every AnalysisPropertyView can only be linked to a single Analysis

seed/views/v3/analyses.py
```diff
@@ -0,0 +1,71 @@
...
+        organization_id = request.query_params.get('organization_id', None)
+        property_id = request.query_params.get('property_id', None)
+        analyses = []
+        for analysis in Analysis.objects.filter(organization=organization_id):
+            property_view_info = analysis.getPropertyViewInfo(property_id)
+            if property_id is not None and property_view_info["number_of_analysis_property_views"] < 1:
+                continue
+            serialized_analysis = AnalysisSerializer(analysis).data
+            serialized_analysis.update(property_view_info)
+            analyses.append(serialized_analysis)
```
macintoshpie, line 44: I think there might be a slightly simpler way to do this. Django has some powerful filtering abilities built in, see: https://docs.djangoproject.com/en/2.2/topics/db/queries/#spanning-multi-valued-relationships

What if we did something like this (not sure if this whole thing works but you get the idea)
```python
if property_id is not None:
    analyses_qs = Analysis.objects.filter(organization_id=organization_id, analysispropertyview__property_id=property_id)
else:
    analyses_qs = Analysis.objects.filter(organization_id=organization_id)
for analysis in analyses_qs:
    serialized_analysis = AnalysisSerializer(analysis).data
    serialized_analysis.update({
        'number_of_analysis_property_views': analysis.analysispropertyview_set.count(),
        'cycles': list(analysis.analysispropertyview_set.values_list('cycle', flat=True).distinct())
    })
    analyses.append(serialized_analysis)
```

seed/serializers/analyses.py
```diff
@@ -0,0 +1,24 @@
...
+class AnalysisSerializer(serializers.ModelSerializer):
+    service = serializers.SerializerMethodField('get_readable_service')
+    status = serializers.SerializerMethodField('get_readable_status')
+
+    class Meta:
+        model = Analysis
+        fields = '__all__'
+
+    def get_readable_service(self, obj):
+        return Analysis.SERVICE_TYPES[next((i for i, v in enumerate(Analysis.SERVICE_TYPES) if v[0] == obj.service), None)][1]
```
macintoshpie, line 21: nice 👍

macintoshpie (changes requested):
Lookin good, just a few comments.

I know this is a super simple API, but could you add some tests which make sure we are filtering correctly and summarizing analysis property view and cycles correctly?

seed/views/v3/analyses.py
Ryoken: Ah, will do

seed/models/analyses.py
Ryoken: If that's the case I will simplify it.  Good to know!

seed/views/v3/analyses.py
Ryoken: That would work for sure but the ticket states the number of views and cycles should be analysis model methods.  I built "getPropertyViewInfo" with this in mind so the analysisPropertyViews table wouldn't be hit more than once, even with Django's lazy loading.  I can split the difference and generate the property filtered queryset so there's no continue.  Does that work?

macintoshpie: I like the split the difference idea 👍 👍  I guess that would also mean we don't need the `property_id` param for your method, which I like as well

Ryoken: I just realized (thanks tests!) that param is going to be required no matter how you look at it.  In order to get an accurate count of property views that method will need to know if you are filtering on a property id or not.

seed/models/analyses.py
```diff
@@ -45,3 +51,14 @@ class Analysis(models.Model):
     # that are applicable to the entire analysis (ie all properties involved).
     # For property-specific results, use the AnalysisPropertyView's parsed_results
     parsed_results = JSONField(default=dict, blank=True)
+
+    def getPropertyViewInfo(self, property_id=None):
```
macintoshpie, line 55: Missed this first time around, but SEED uses snake-case for all python variable and function names. If you have a minute, could you update our tox.ini for flake8 to install pep8-naming? https://stackoverflow.com/questions/28725116/flake8-doesnt-report-mixed-case-function-names

seed/views/v3/analyses.py
```diff
@@ -35,10 +35,12 @@ def list(self, request):
...
-        for analysis in Analysis.objects.filter(organization=organization_id):
-            property_view_info = analysis.getPropertyViewInfo(property_id)
-            if property_id is not None and property_view_info["number_of_analysis_property_views"] < 1:
-                continue
+        if property_id is not None:
+            analyses_queryset = Analysis.objects.filter(organization=organization_id, analysispropertyview__property=property_id).distinct()
+        else:
+            analyses_queryset = Analysis.objects.filter(organization=organization_id).distinct()
+        for analysis in analyses_queryset:
+            property_view_info = analysis.getPropertyViewInfo(property_id=property_id if property_id is not None else None)
```
macintoshpie, line 43: I think we can drop this ternary, it seems redundant (if property_id is None, pass None; if property_id is not None, pass property_id)

macintoshpie (changes requested):
Thanks for adding the tests! One small change for snake-case naming. The other comment isn't super important

seed/views/v3/analyses.py
Ryoken: I wrote that thinking it would remove the entire arg, like a silly person.  Thanks for pointing it out!

seed/models/analyses.py
Ryoken: AFAIK this should be working.

macintoshpie (approved):
Looks good 👍 👍 

Let's just have Adrian take a look and then it's good to merge

adrian-lara (commented):
Awesome this looks great! Just a few small comments and suggestions, otherwise this seems to be working as expected!

seed/views/v3/analyses.py
```diff
@@ -0,0 +1,74 @@
...
+    @ajax_request_class
+    @has_perm_class('requires_member')
+    def list(self, request):
+        organization_id = request.query_params.get('organization_id', None)
+        property_id = request.query_params.get('property_id', None)
+        analyses = []
+        if property_id is not None:
+            analyses_queryset = Analysis.objects.filter(organization=organization_id, analysispropertyview__property=property_id).distinct()
+        else:
+            analyses_queryset = Analysis.objects.filter(organization=organization_id).distinct()
```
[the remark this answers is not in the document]
Ryoken: I didn't have them at first but they were actually needed.  I can't remember the exact situation but one of the joins and filters resulted in there bing two identical analysis in the response.  I can go back and find the specific test that failed.

seed/tests/test_analyses.py
```diff
@@ -0,0 +1,176 @@
...
+        property_state_a = property_state_factory.get_property_state()
+        property_state_b = property_state_factory.get_property_state()
+        property_state_c = property_state_factory.get_property_state()
+        property_state_d = property_state_factory.get_property_state()
+
+        # create an analysis with two property views, each with the same property but a different cycle
+        self.analysis_a = Analysis.objects.create(
+            name='test a',
+            service=1,
+            status=10,
```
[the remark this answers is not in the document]
Ryoken: Ah, can do.  I was trying to use Analysis.STATUS_TYPES.BSYNCR and that wasn't working, so I hardcoded it.  This is better!

seed/views/v3/analyses.py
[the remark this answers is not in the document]
macintoshpie: I saw this note here: https://docs.djangoproject.com/en/3.1/ref/models/querysets/#django.db.models.query.QuerySet.distinct
> By default, a QuerySet will not eliminate duplicate rows. In practice, this is rarely a problem, because simple queries such as Blog.objects.all() don’t introduce the possibility of duplicate result rows. However, if your query spans multiple tables, it’s possible to get duplicate results when a QuerySet is evaluated. That’s when you’d use distinct()

seed/tests/test_analyses.py
```diff
@@ -0,0 +1,176 @@
...
+            property=self.property_a,
+            cycle=cycle_a,
+            property_state=property_state_a
+        )
+        AnalysisPropertyView.objects.create(
+            analysis=self.analysis_a,
+            property=self.property_a,
+            cycle=cycle_b,
+            property_state=property_state_b
+        )
```
adrian-lara, line 71: @Ryoken Re: https://github.com/SEED-platform/seed/pull/2473#discussion_r534513955 , 
Could it be related to these two -Views associated to the same Property? That is a case we should accommodate for (This can happen during merging 🤮 , but let's not think about that yet!).

seed/views/v3/analyses.py
```diff
@@ -0,0 +1,74 @@
...
+    @has_perm_class('requires_member')
+    def retrieve(self, request, pk):
+        organization_id = int(request.query_params.get('organization_id', 0))
+        analysis = Analysis.objects.get(id=pk)
+        if analysis.organization_id != organization_id:
+
+            return JsonResponse({
+                'status': 'error',
+                'message': "Requested analysis doesn't exist in this organization."
+            }, status=HTTP_409_CONFLICT)
```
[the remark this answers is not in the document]
Ryoken: Looks good, thanks

seed/views/v3/analyses.py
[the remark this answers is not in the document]
adrian-lara: Right, and duplicates appearing would happen since we could have 2 AnalysisPropertyViews with the same property_id.

I think `distinct()` here makes sense in the case we're referencing the `analysispropertyview` relation, but maybe it doesn't make as much sense in the case where we're just using `organization_id`?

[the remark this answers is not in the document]
Ryoken: True.  I'll take it off the second case.

adrian-lara (approved):
This looks great! Thanks for iterating on this! 💯

PR #2482, opened 2020-12-02
fix(intermittent failed test): implied ordering is explicit

adrian-lara:
Intermittent failed test re: merging properties with meters.

#### What's this PR do?
The testing logic implicitly relied on querying Properties in ID order. That is made explicit with this PR.

Nothing to do but wait and see if this intermittent failure happens again.
```
  File "/seed/seed/tests/test_match_existing.py", line 1342, in test_match_merge_link_for_properties_meters_persist_in_different_situations
    self.assertTrue(agg_meter.meter_readings.filter(reading=100).exists())
```


macintoshpie (approved):
V nice 👍

Issue #2456, opened 2020-11-12
Implement front-end for Analyses summary page for Organization


macintoshpie:
### Tasks
See the screenshot below for reference (note that the columns might not align exactly with the new Analysis data model, so take it with a grain of salt)

- Create a new "tab" on the organization page labeled "Analyses"
  - update the accounts_nav partial
- Create the analyses summary page for organization using the `/analyses/` endpoint
  - should present a table that summarizes all Analyses within the organization, including:
    - analysis name
    - analysis type
    - status(es), colored as you see fitting
    - number of properties
    - user
    - start time
    - end time
  - there should be no hyperlinks or actions, this will be added at a later time when API and other frontend pages are more fleshed out

<img width="989" alt="Screen Shot 2020-11-12 at 12 50 48 PM" src="https://user-images.githubusercontent.com/18518728/98989006-bb18af80-24e5-11eb-9791-f1e75f078952.png">

Ryoken:
I'm not sure if I got the status and colors correct and am open to changing them!

adrian-lara:
No manual testing needed.

PR #2488, opened 2020-12-07
Feat/analyses frontend

Ryoken:
#### What's this PR do?
Adds a front-end for analyses.

https://github.com/SEED-platform/seed/issues/2456

Screen Shot 2020-12-07 at 4 34 08 PM


seed/static/seed/js/services/analyses_service.js
```diff
@@ -0,0 +1,27 @@
...
+    '$http',
+    '$log',
+    'user_service',
+    function (
+      $http,
+      $log,
+      user_service
+    ) {
+
+      let get_labels = function() {
```
adrian-lara, line 12: Thinking this might be a typo, but I'd avoid the term `labels` since SEED has the widely-used [Status]Label model.

Actually, maybe we don't even need this method? For the analyses GET list endpoint, we should always use the explicitly provided org_id as opposed to a user's "currently chosen" organization as indicated at the top right of the page.

seed/static/seed/partials/analyses.html
```diff
@@ -0,0 +1,60 @@
...
+                    <tbody>
+                        <tr ng-repeat="analysis in analyses  | filter:filter_params:strict">
+                            <td>{$:: analysis.name $}</td>
+                            <td>
+                            	<i class="glyphicon glyphicon-pencil" aria-hidden="true"></i>
+                            	<i class="glyphicon glyphicon-trash" aria-hidden="true"></i>
+                            </td>
+                            <td>{$:: analysis.number_of_analysis_property_views $}</td>
+                            <td>{$:: analysis.service $}</td>
+                            <td class="analysis-status {$:: analysis.status.toLowerCase() $}">{$:: analysis.status $}</td>
```
adrian-lara, line 49: No strong opinions from me here. I personally like this approach vs the ng-class, but I was never really a fan of the object you have to build for ng-class. @axelstudios thoughts?

seed/static/seed/partials/analyses.html
```diff
@@ -0,0 +1,60 @@
...
+                            <td>{$:: analysis.start_time | date : 'MM-dd-yyyy' $}</td>
+                            <td>{$:: (analysis.end_time - analysis.start_time < 1 ? '' : analysis.end_time - analysis.start_time) | date : 'MM-dd-yyyy' $}</td>
+                            <td>{$:: analysis.user $}</td>
```
adrian-lara, line 52: Screen Shot 2020-12-08 at 12 19 59 AM
3 things:

- Can we add the time to the start time?
- I'm not sure what's happening with the duration value.
- Author is ID right now. We should probably make that first and last name. I'm thinking that we should GET the org users and use that response to get the name.

adrian-lara (commented):
Looks great so far! A few comments and a few definite changes. 

Along with the specific comments below, can you add the Analyses option to the organization list page:
Screen Shot 2020-12-08 at 12 01 35 AM

seed/static/seed/js/services/analyses_service.js
Ryoken: I honestly think I just forgot to remove this function from a reference copy-paste.  It's not being used atm!

seed/static/seed/partials/analyses.html
Ryoken: Added some filters that should cover the last two.

seed/static/seed/js/controllers/analyses_controller.js
```diff
@@ -0,0 +1,66 @@
...
+      auth_payload,
+      urls
+    ) {
+      $scope.org = organization_payload.organization;
+      $scope.auth = auth_payload.auth;
+      $scope.analyses = analyses_payload.analyses;
+      $scope.users = users_payload.users;
+    }
+  ])
+  .filter('get_run_duration', function() {
```
adrian-lara, line 28: Nice - I haven't seen this pattern in SEED, but it looks good to me! I'll again tag @axelstudios just in case we don't want to introduce a new pattern.

adrian-lara (approved):
Thanks for making those changes - this looks good to me!
Screen Shot 2020-12-08 at 12 24 20 PM

Organizations page:
Screen Shot 2020-12-08 at 12 24 52 PM

seed/static/seed/js/controllers/analyses_controller.js
axelstudios: Since these filters are so specific to the analysis views I'm fine with this.  For more general filters (like title-casing text) we do have a [centralized location](https://github.com/SEED-platform/seed/tree/develop/seed/static/seed/js/filters) and [filter declarations](https://github.com/SEED-platform/seed/blob/develop/seed/static/seed/js/seed.js#L98-L106)

PR #2495, opened 2020-12-12
2451 feat: add analysis pipelines and tasks for preparing bsyncr input

### LINKED ISSUE #2451: Define background task for creating Analysis, AnalysisPropertyView, etc.
## Context
Before an analysis can be run, we must validate and create the associated models (Analysis, AnalysisPropertyView, etc). This is (mostly) a background celery task that would get kicked off when a users POSTS to /analysis/ to create a new analysis.

One notable change from when this ticket was first written, SEED will make an Analysis model for each property selected for BSyncr analysis.

## Proposed Implementation
_within the view aka request process_
- Validate inputs and create the Analysis model
- kick off a celery task for validating/creating analysis runs
- return a progress ID which is associated with the Analysis

_within the celery task/chord callback_ (I'm imagining this is BSyncer specific.)
- For each property, create a set of Analysis- models, and for each set:
  - Validate BuildingSync files exist for associated PropertyView model)
    - If invalid, create Analysis and AnalysisPropertyView objects but create an AnalysisMessage detailing invalid data and make Analysis status "Failed"
  - Create Analysis
    - Create PropertyState “copy”
    - Make associations to Property and Cycle objects
    - Create AnalysisInputFiles
    - Set status to “Creating" (this will just be for convention's sake since we're only dealing with one AnalysisPropertyView for now)
- (in chord callback) Once everything has been validated and created, set their status to "Ready" in a batch update - this must happen all at once! (note that status should only be updated for those that did not fail) (Second note, this might not impact current implementation, but I'm envisioning this will be used by other analysis services' creation tasks.)

macintoshpie:
When an analysis is created, e.g. when hitting the POST /analyses/ endpoint, there is preparation that must be done before the analysis can actually be run.

#### What's this PR do?
- Defines the basic structure of an "analysis pipeline", shared functionality between pipelines
- Creates a pipeline step for preparing the bsyncr analysis by creating BuildingSync files from the selected properties

Run the automated tests. They will generate BuildingSync files at `MEDIA_ROOT/analysis_input_files/<analysis id>/`. You can check to make sure they look ok.

If you _really_ want to test things, you can test it with the current bsyncr server. Setup the server here and follow the setup instructions: https://github.com/BuildingSync/bsyncr-server
Note that building the image might take a while, installing the R devtools is reeeal slow.
Start the server like instructed in the readme
Then post one of the files like so
```bash
curl -F 'file=@<your test file here>' localhost:5000
```
You should get back your original BuildingSync file that's been updated to include model data (see the DerivedModel section). If it failed you'll get back some json describing the error.

Resolves #2451


seed/analysis_pipelines/bsyncr.py
```diff
@@ -0,0 +1,254 @@
...
+            analysis_input_file.file.save(f'{analysis_property_view.id}.xml', ContentFile(bsync_doc))
+            analysis_input_file.clean()
+            analysis_input_file.save()
+            input_file_paths.append(analysis_input_file.file.path)
+        except Exception as e:
+            AnalysisMessage.objects.create(
+                analysis=analysis,
+                analysis_property_view=analysis_property_view,
+                type=AnalysisMessage.DEFAULT,
+                user_message='Unexpected error',
```
macintoshpie, line 118: Not sure what a better user message for this might be. Looking for feedback

seed/analysis_pipelines/bsyncr.py
```diff
@@ -0,0 +1,254 @@
...
+        # Steps:
+        # 1) ...starting
+        # 2) create AnalysisPropertyViews
+        # 3) create input files for each property
+        progress_data.total = 3
+        progress_data.save()
+
+        chain(
+            task_create_analysis_property_views.si(analysis_id, property_view_ids, progress_data.key),
+            _prepare_all_properties.s(analysis_id, progress_data.key),
```
macintoshpie, line 49: we should parallelize this part at some point, not worried about it for now

seed/analysis_pipelines/pipeline.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    if progress_data_key is not None:
+        progress_data = ProgressData.from_key(progress_data_key)
+        progress_data.step('Copying property data')
+    analysis_view_ids, failures = AnalysisPropertyView.batch_create(analysis_id, property_view_ids)
+    for failure in failures:
+        AnalysisMessage.objects.create(
+            analysis_id=analysis_id,
+            type=AnalysisMessage.DEFAULT,
+            user_message=f'Failed to copy property data for PropertyView ID {failure.property_view_id}: {failure.message}',
```
macintoshpie, line 29: Not sure if this is an appropriate user-level message

seed/building_sync/building_sync.py
```diff
@@ -42,6 +42,7 @@ class ParsingError(Exception):
 
 class BuildingSync(object):
     BUILDINGSYNC_V2_0 = '2.0'
+    BUILDINGSYNC_V2_2_0 = '2.2.0'
```
macintoshpie, line 45: Had to add the 2.2.0 BSync schema for testing our generated documents were valid

seed/models/analyses.py
```diff
@@ -30,6 +31,7 @@ class Analysis(models.Model):
     COMPLETED = 70
 
     STATUS_TYPES = (
+        (PENDING_CREATION, 'Pending Creation'),
```
macintoshpie, line 34: Needed some way to indicate "the preparation of the analysis has not been started" to prevent more than one celery task to start creating the analysis.

seed/models/analyses.py
```diff
@@ -43,7 +45,7 @@ class Analysis(models.Model):
     service = models.IntegerField(choices=SERVICE_TYPES)
     start_time = models.DateTimeField(null=True, blank=True)
     end_time = models.DateTimeField(null=True, blank=True)
-    status = models.IntegerField(choices=STATUS_TYPES)
+    status = models.IntegerField(default=PENDING_CREATION, choices=STATUS_TYPES)
```
macintoshpie, line 48: Setting a default for creating an analysis seemed like something we should do b/c 99% of the time it should be in its initial "state"

seed/models/analysis_input_files.py
```diff
@@ -22,6 +29,6 @@ class AnalysisInputFile(models.Model):
         (BUILDINGSYNC, 'BuildingSync'),
     )
 
-    file = models.FileField(upload_to="analysis_input_files", max_length=500)
+    file = models.FileField(upload_to=analysis_input_path, max_length=500)
```
macintoshpie, line 32: I thought it might be nice to break out input files into subdirectories according to the analysis id, rather than having all input files for all analyses living in the same place. Not sure if this is necessary/that important, did it without thinking too much about it, welcoming feedback

seed/tests/test_analyses_views.py
```diff
@@ -20,7 +20,7 @@
 from seed.utils.organizations import create_organization
 
 
-class TestAnalyses(TestCase):
+class TestAnalysesView(TestCase):
```
macintoshpie, line 23: thought I'd move the file and rename the test b/c it's testing the views not the model

seed/analysis_pipelines/bsyncr.py
adrian-lara: I think this is fine. We'll let the frontend handle any potential displaying of the debug message.

seed/tests/test_analysis_pipelines.py
```diff
@@ -0,0 +1,364 @@
...
+
+        # Act / Assert
+        # shouldn't matter what values are passed for property_view_ids
+        with self.assertRaises(AnalysisPipelineException) as context:
+            pipeline.prepare_analysis([1, 2, 3])
+
+        self.assertTrue('Analysis has already been prepared or is currently being prepared' in str(context.exception))
+
+        # the status should not have changed
+        self.analysis.refresh_from_db()
```
adrian-lara, line 83: 😮 TIL about `refresh_from_db()`

adrian-lara (commented):
I only reviewed the logic - I'll test this once we get the corresponding analyses POST endpoint set up.

Overall looks good! Thanks for the tests, and I'm looking forward to seeing this in action. 👍 

The only potential change is to merge the migrations.

seed/migrations/0132_auto_20201211_1949_squashed_0134_auto_20201211_2053.py
```diff
@@ -0,0 +1,31 @@
...
+        migrations.AlterField(
+            model_name='analysisinputfile',
+            name='file',
+            field=models.FileField(max_length=500, upload_to=seed.models.analysis_input_files.analysis_input_path),
+        ),
+        migrations.AlterField(
+            model_name='analysis',
+            name='status',
+            field=models.IntegerField(choices=[(8, 'Pending Creation'), (10, 'Creating'), (20, 'Ready'), (30, 'Queued'), (40, 'Running'), (50, 'Failed'), (60, 'Stopped'), (70, 'Completed')]),
+        ),
```
adrian-lara, line 25: I think we can remove this one since we have the 3rd one that includes all the changes.

adrian-lara (approved):
Thanks for squashing the migrations. This looks good to me 👍

seed/models/analysis_input_files.py
axelstudios: I like the idea of analysis subdirectories by id, the `uploads` directory is currently a structureless mess with 2951 files all in the same directory and lots of "unique" strings appended to prevent overwriting older files.  The only uploads subdirectory is `pm_imports` with another structureless directory of 1074 files, that makes it a pain to find the actual file you're looking for when you need it.

I'm definitely in favor of a comment that warns against changes to that function that could break existing paths

Issue #2496, opened 2020-12-15
Duplicate states can be created if the unit selection is empty


axelstudios:
**Describe the bug**
During the mapping process if you map a raw column to an EUI or Area column you have to select the import units, and the units have defaulted values so it's not possible to submit `null` for the units.  However, if you created a column mapping profile from the Column Mappings settings page it is possible to save a mapping with empty units.  If you leave the units blank when mapping it will bypass the de-duplicate-record-checker process.

**Expected Behavior**
Units should not be allowed to be blank which prevents de-duplication

**Actual Behavior**
Units can be saved blank on the column mappings page

**Steps to Reproduce**
1. Go to the organization's column mappings page
2. Create or edit a profile by mapping a field to an EUI or Area column (e.g. `Gross Floor Area` or `Site EUI`)
3. Leave the units blank (the default) and save
4. Import a file using that mapping
5. Import the same file again using that mapping, note that you now have duplicate records

**Instance Information**
v2.8.1

adrian-lara:
Do we still want to cherry-pick some version of this fix into production?

axelstudios:
@adrian-lara Cherry picked into v2.7.5 (v2.7.6 now) and v2.8.1 (v2.8.2 now), and both merged back down to develop

adrian-lara:
Cherry-picks look good to me! 👍

PR #2497, opened 2020-12-15
2496 fix: prevent column mappings with blank units

axelstudios:
#### What's this PR do?
- Highlights empty unit fields in red and prevents mapping/saving column mappings

Follow the steps to reproduce in the issue and verify that nothing can be saved/mapped when units are empty

#2496


adrian-lara (approved):
Looks good!

On the org column mapping settings page, I like how we always just default to some units. Then, if somehow, the users have empty units say from a prior version of SEED, they can't edit unless the units are populated.
Screen Shot 2020-12-15 at 7 25 04 PM
Screen Shot 2020-12-15 at 7 26 05 PM

Issue #2477, opened 2020-12-01
Add API v3 endpoint /analyses/{analysis_id}/start - POST and create BSyncr start analysis task


adrian-lara:
The endpoint should only verify perms, trigger the appropriate task, and return the progress_id.

permissions | query params | body params | Notes
-- | -- | -- | --
Org Member | organization_id |   | Triggers separately built celery task

Since this is the initial iteration of this endpoint, a task should be created to kick off the BSyncr analysis. Specifics around the BSyncr API need to be researched in order to help define specific implementation details here.

This task would likely be built alongside other BSyncr-related tasks (create and stop) - maybe within as BSyncrTaskService class?

adrian-lara:
No manual testing needed

PR #2498, opened 2020-12-17
#2477 feat: run bsyncr analysis

macintoshpie:
SEED is working on adding ability for users to run analyses on properties. The first analysis available will be bsyncr.

#### What's this PR do?
- Adds pipeline steps for running bsyncr analysis (once it's been prepared)
- Adds endpoint `/analyses/<id>/start` to kickoff this process

To fully test this, you'll need to get bsyncr-server up and running locally.

1. get bsyncr running with `docker run --network="seed_default" --name="bsyncr-server" -e NOAA_TOKEN=<your noaa token> seedplatform/bsyncr-server:latest`. You can get a noaa token here: https://www.ncdc.noaa.gov/cdo-web/token
If you're not using the docker-compose.dev file for running seed, then leave out the `--network=...` argument and replace it with `-p 5000:5000`.
Test that things are working by execing into the web container and running `curl -X POST bsyncr-server:5000`, or `curl -X POST localhost:5000`. You should get a response telling you to include a file.

2. Configure your SEED instance. If using docker-compose, edit your .env file to include the following
```bash
BSYNCR_SERVER_HOST='bsyncr-server'
BSYNCR_SERVER_PORT='5000'
```
If you're running SEED without docker-compose, you need to export these two environment variables (replacing 'bsyncr-server' with 'localhost') yourself and restart your celery workers and server (the celery workers part is probably unnecessary if you're doing the next step)
Then configure celery in your local_untracked.py with this to make things simpler
```python
CELERY_BROKER_BACKEND = 'memory'
CELERY_TASK_ALWAYS_EAGER = True
CELERY_TASK_EAGER_PROPAGATES = True
```

3. Run a script to setup and start the analysis. Download this script here: https://gist.github.com/macintoshpie/6a74540ed37bb6df8ecf847109170e5e
Run it with `python manage.py shell < test_bsyncr.py` (obviously replace the filename path with whatever you name it). Of course you'll need to exec into the web container if using docker-compose. This script will create a new user, org, property views, meters, and analysis before running the analysis (so maybe do it on a db you don't care about). If everything ran correctly, you should be able to check that the analysis property views were updated and the output files were created
```python
a = Analysis.objects.all().last()
apv = AnalysisPropertyView.objects.filter(analysis=a)
print(apv[0].parsed_results)
```

#2477


seed/analysis_pipelines/tasks.py
```diff
@@ -0,0 +1,14 @@
+# !/usr/bin/env python
+# encoding: utf-8
+"""
+:copyright (c) 2014 - 2020, The Regents of the University of California, through Lawrence Berkeley National Laboratory (subject to receipt of any required approvals from the U.S. Department of Energy) and contributors. All rights reserved.  # NOQA
+:author
+"""
+
+from seed.analysis_pipelines.bsyncr import (  # noqa: F401
```
macintoshpie, line 8: I don't like this, but celery requires you put your tasks in a file called `tasks.py`. I thought I'd rather keep the tasks inside the pipeline's file then import them here, but open to thoughts.
An alternative would be to change bsyncr.py into a directory, then have a tasks.py in there. Or have a common tasks.py file all pipelines would use (I don't like this alternative)

config/settings/common.py
```diff
@@ -301,6 +301,9 @@
     'LOGOUT_URL': '/accounts/logout',
 }
 
+BSYNCR_SERVER_HOST = os.environ.get('BSYNCR_SERVER_HOST')
```
macintoshpie, line 304: NOTE: we need to add a check, probably at the `POST /analyses/` endpoint, that makes sure SEED is configured to use bsyncr

seed/analysis_pipelines/tasks.py
adrian-lara: I like this current set up the best. I'm envisioning that once this gets built out, we'll have a `from seed.analysis_pipelines.<analysis_service> import (...` block for each analysis service.

seed/analysis_pipelines/bsyncr.py
```diff
@@ -252,3 +286,192 @@ def _build_bsyncr_input(analysis_property_view, meter):
...
+        raise AnalysisPipelineException(message)
+
+    return output_file_ids
+
+
+@shared_task
+def _process_results(analysis_output_file_ids, analysis_id, progress_data_key):
+    analysis = Analysis.objects.get(id=analysis_id)
+    analysis.status = Analysis.RUNNING
+    analysis.save()
```
adrian-lara, line 364: Might be remnant from copying the first task, or are there cases where the status wouldn't already be `RUNNING`? Even if this were redundant, it's non-breaking so not blocking merge.

adrian-lara (approved):
This works as expected! 🚀 💯 🔥 👍 

Tried it without then with internet and the results came out as expected.
Screen Shot 2020-12-20 at 11 19 18 PM

Note that I had to update the script you linked and there's a non-blocking comment that might still need a follow up code change but this is good to go from what I can tell!

seed/analysis_pipelines/tasks.py
macintoshpie: That sounds good to me 👍

seed/analysis_pipelines/bsyncr.py
macintoshpie: Whoops, yes this is a non-critical bug. It shouldn't be trying to update the status, I'll remember to remove this later

Issue #2475, opened 2020-12-01
Add API v3 endpoint /analyses - POST


adrian-lara:
This serves as an entry point and should only verify perms, then kick off the correct task based on analysis type, and finally, return the celery progress_id. (It should be the responsibility of the task to actually validate and create Analysis-models.)

permissions | query params | body params | Notes
-- | -- | -- | --
Org Member | organization_id | name, type, configuration, [propertyview_ids] | Triggers separately built celery task

adrian-lara:
During initial implementation/testing, this may be make sense to be completed with #2451

adrian-lara:
No manual testing needed

PR #2505, opened 2020-12-19
#2475 feat: analyses create view

macintoshpie:
SEED is adding ability to run analyses on properties.

#### What's this PR do?
Adds endpoint to create an analysis and start its preparation. Also adds FE logic for the analysis modal so you can create an analysis.

Go to a property's analysis tab, click new analysis and create a bsyncr analysis. For it to actually work you'll need to have bsyncr configured (see #2477 ) and the property will need to have 12+ electricity meter readings.

#2475

### DIFF
--- a/seed/serializers/analyses.py
+++ b/seed/serializers/analyses.py
@@ -9,8 +9,24 @@ from rest_framework import serializers
 from seed.models import Analysis
 
 
+class CustomChoicesField(serializers.ChoiceField):
+    def to_representation(self, obj):
+        if obj == '' and self.allow_blank:
+            return obj
+        return self._choices[obj]
+
+    def to_internal_value(self, data):
+        if data == '' and self.allow_blank:
+            return ''
+
+        for key, val in self._choices.items():
+            if val == data:
+                return key
+        self.fail('invalid_choice', input=data)
+
+
 class AnalysisSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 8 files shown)


seed/serializers/analyses.py
```diff
@@ -9,8 +9,24 @@
...
+
+        for key, val in self._choices.items():
+            if val == data:
+                return key
+        self.fail('invalid_choice', input=data)
+
+
 class AnalysisSerializer(serializers.ModelSerializer):
-    service = serializers.CharField(source='get_service_display')
+    service = CustomChoicesField(Analysis.SERVICE_TYPES)
```
adrian-lara, line 29: I think previously, we let the frontend handle the translation from enum to string, but I’m fine with this. And if others disagree, it's not a significant effort to make the change at some point.

adrian-lara (approved):
I just pushed up one debugger removal, but this works as expected! Thanks for hooking up the frontend to use this new endpoint as well.

Screen Shot 2020-12-20 at 11 23 34 PM

One non-blocking comment, but I'd say this is good to merge.

seed/serializers/analyses.py
macintoshpie: As a quick comment to why I did this, it seemed like the easiest way to deserialize (ie from json to python) when a user posts from the FE, now the FE just needs to know the "display" names.

Issue #2480, opened 2020-12-01
Implement front-end for an individual Analysis page


adrian-lara:
### The following image is meant to serve as a guide, and the content and layout are not solidified.
**A design discussion should happen once this ticket is ready to be picked up**

For this version of the Analyses feature (with only BSyncr), we likely won't need aggregate results, but design decisions should allow for a reasonable addition of that information later.

![Screen Shot 2020-12-01 at 3 32 04 PM](https://user-images.githubusercontent.com/30608004/100804859-ce36e500-33ea-11eb-914e-5b19427eaff1.png)

Ryoken:
Do we want the analysis opening on it's own page or somehow inline, whether a modal pop-up or an expanding row in the analyses table?

Ryoken:
Should a user be able to edit the analysis or it's runs from this page?  If so, what fields are editable?

adrian-lara:
> Do we want the analysis opening on it's own page or somehow inline, whether a modal pop-up or an expanding row in the analyses table?

On it's own page.


> Should a user be able to edit the analysis or it's runs from this page? If so, what fields are editable?

No edits for now.

adrian-lara:
No manual testing

PR #2506, opened 2020-12-19
Feat/analyses page

Ryoken:
As it is currently understood this page lives in the organization's section rather than the property detail section, though the analyses list can be accessed from either.

#### What's this PR do?
Creates an individual analysis frontend page.

Navigate to the organization's page and click on "Analyses", then click on an individual analysis' name.

https://github.com/SEED-platform/seed/issues/2480


seed/serializers/analysis_property_views.py
```diff
@@ -8,10 +8,12 @@
 
 from seed.models import AnalysisPropertyView
 from seed.serializers.analysis_output_files import AnalysisOutputFileSerializer
+from seed.serializers.properties import PropertyStateSerializer
 
 
 class AnalysisPropertyViewSerializer(serializers.ModelSerializer):
     output_files = AnalysisOutputFileSerializer(source='analysisoutputfile_set', many=True)
+    property_state = PropertyStateSerializer()
```
adrian-lara, line 16: Nice! Great to see adding PropertyState info is as easy as doing this. Note that displaying this info on the frontend is beyond MVP scope, but it's definitely something we want at some point.

seed/views/v3/analysis_messages.py
```diff
@@ -0,0 +1,56 @@
...
+        return JsonResponse({
+            'status': 'success',
+            'messages': AnalysisMessageSerializer(messages_queryset, many=True).data
+        })
+
+    @require_organization_id_class
+    @api_endpoint_class
+    @ajax_request_class
+    @has_perm_class('requires_member')
+    def retrieve(self, request, pk, analysis_pk, views_pk=None):
```
adrian-lara, line 41: Somehow noticed this 🤷 - can we make `views_pk` singular `view_pk` to avoid potential confusion.

seed/views/v3/analysis_messages.py
```diff
@@ -0,0 +1,56 @@
...
+
+class AnalysisMessageViewSet(viewsets.ViewSet):
+    serializer_class = AnalysisMessageSerializer
+    model = AnalysisMessage
+
+    @require_organization_id_class
+    @api_endpoint_class
+    @ajax_request_class
+    @has_perm_class('requires_member')
+    def list(self, request, analysis_pk, views_pk=None):
```
adrian-lara, line 26: See other comment: `views_pk` -> `view_pk`

seed/serializers/pint.py
```diff
@@ -137,7 +137,10 @@ def to_representation(self, obj):
             try:
                 org = state.organization
             except AttributeError:
-                org = state.state.organization
+                try:
+                    org = state.state.organization
+                except AttributeError:
+                    org = state.property_state.organization
```
adrian-lara, line 143: What's the story behind this change? Is `state` in L143 an AnalysisPropertyView object? Maybe this could just be a short clarifying comment in that last except block.

adrian-lara (commented):
Sorry - looks like there's a conflict 😬 Think you could resolve?

Also, I did do a first pass review, and things look great overall! 👌 👍
Additionally, there's maybe 2-3 small changes 🙏 

Screen Shot 2020-12-21 at 12 10 10 AM
Screen Shot 2020-12-21 at 12 23 52 AM
Screen Shot 2020-12-21 at 12 20 18 AM
Screen Shot 2020-12-21 at 12 12 27 AM

seed/serializers/pint.py
Ryoken: Ted and I spent some time trying to figure this one out but yeah, the state is an AnalysisPropertyView object.  I can add a comment to denote this.

seed/serializers/analysis_property_views.py
Ryoken: This was added so the address could be displayed.  Is there a better way to grab that?

adrian-lara: ah - forgot about that. Nope this is ideal since, we'll end up needing everything later! 👌

seed/static/seed/js/services/analyses_service.js
```diff
@@ -22,12 +22,6 @@ angular.module('BE.seed.service.analyses', [])
         });
       };
 
-      let get_analysis_for_org = function (analysis_id, org_id) {
-        return $http.get('/api/v3/analyses/' + analysis_id + '?organization_id=' + org_id).then(function (response) {
-          return response.data;
-        });
-      };
-
```
adrian-lara: I think we wanted to actually keep this method:
Screen Shot 2020-12-21 at 1 59 54 PM
Screen Shot 2020-12-21 at 1 59 37 PM

adrian-lara (changes requested):

seed/static/seed/js/services/analyses_service.js
Ryoken: Analysis vs Analyses has gotten me so many times during all these tickets..

adrian-lara (approved):
Thanks for fixing! Works again 👍

Issue #2502, opened 2020-12-17
User can view a list of all analyses created for an organization


adrian-lara:
**Is your feature request related to a problem? Please describe.**

This is a new feature

**Describe the solution you'd like**

As a logged in organization user, when I visit the organization Analyses page (by navigating to the organization settings and clicking `Analyses`, I see:

- All the analyses run for that organization and details about each.
- Each name is clickable.
- For each analysis entry, I can see the start button if the analysis is ready to be run
- For each analysis entry, the start button is disabled if the analysis is not ready to be run. (Note: for this story, no other buttons need to be visible.)

**Additional context**

- Original mock up: 
<img width="965" alt="Screen Shot 2020-12-17 at 1 15 29 PM" src="https://user-images.githubusercontent.com/30608004/102538879-1f410d00-406a-11eb-9967-3c75ad7b23ca.png">

- Current WIP (maintain this styling): 
![Screen Shot 2020-12-17 at 1 15 50 PM](https://user-images.githubusercontent.com/30608004/102538895-2405c100-406a-11eb-9bc5-3e6bda0bce4c.png)

RDmitchell:
@adrian-lara / @nllong -- I don't really know how the analysis feature works, so I am not sure how to test this. 

Looking at the detailed view for one of my test records, if I click the New Analysis button and say I want to do a BSyncr analysis, which is the only option, Run Status is returned as Error

RDmitchell:
Instance; Dev1
SHA:
Org: LBNL Test 200

I can see all the completed and failed analyses for my org
![image](https://user-images.githubusercontent.com/6314950/105130414-d7176f00-5a9b-11eb-9b17-082d7c765081.png)

PR #2522, opened 2020-12-23
feat: disable analysis start button if not ready

macintoshpie:
I am bad at css and javascript

#### What's this PR do?
Adds disabled status for play button if analysis is being prepared.

Verify start button is disabled when creating analysis. Verify no play button when state is past creating.

#2502

Screen Shot 2020-12-22 at 6 50 40 PM

### DIFF
--- a/seed/static/seed/js/controllers/analyses_controller.js
+++ b/seed/static/seed/js/controllers/analyses_controller.js
@@ -38,6 +38,9 @@ angular.module('BE.seed.controller.analyses', [])
       }
 
       $scope.start_analysis = function(analysis_id) {
+        analysis = $scope.analyses.find(function(a) { return a.id === analysis_id })
+        analysis.status = 'Starting...'
+
         analyses_service.start_analysis(analysis_id)
         .then(function (result) {
           if (result.status === 'success') {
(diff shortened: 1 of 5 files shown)


seed/static/seed/scss/style.scss
```diff
@@ -3683,3 +3683,8 @@ $pairedCellWidth: 60px;
     pointer-events: none;
   }
 }
+
+.glyphicon[disabled] {
```
macintoshpie, line 3687: someone who is better at CSS than me will have to determine if this is correct

seed/static/seed/js/controllers/analyses_controller.js
```diff
@@ -38,6 +38,9 @@ angular.module('BE.seed.controller.analyses', [])
       }
 
       $scope.start_analysis = function(analysis_id) {
+        analysis = $scope.analyses.find(function(a) { return a.id === analysis_id })
+        analysis.status = 'Starting...'
```
macintoshpie, line 42: added this to have the side effect of removing the "start" button once the click it

adrian-lara (approved):
Works as expected when the create task takes longer. We probably could use an update to the property detail anaylses page to show the newly created record right away (it'd have the "Creating" status), but that doesn't have to be included here.
Screen Shot 2020-12-23 at 1 45 02 PM

PR #2527, opened 2020-12-30
refactor(workflows): publish image on push to develop or release

macintoshpie:
Currently we only publish to dockerhub on push to master or develop. Sometimes we create tags that aren't part of this flow.

#### What's this PR do?
Publishes images whenever there's a push to develop (which is tagged as `develop`, or if a release is published (tagged with the tag).

Post-merge, we should see the develop branch get published again. We should publish a temporary test release to verify it's also triggering the workflow and properly tags the image.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,7 +4,8 @@ on:
   push:
     branches:
       - develop
-      - main
+  release:
+    types: [published]
 
 jobs:
   publish:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -4,7 +4,8 @@ on:
   push:
     branches:
       - develop
-      - main
```
macintoshpie, line 7: don't need `main` anymore b/c those releases should be tagged

.github/workflows/publish.yml
```diff
@@ -35,16 +36,14 @@ jobs:
...
-          PACKAGE_VERSION=$( sed -n 's/.*"version": "\(.*\)",/\1/p' package.json )
-          echo "BRANCH_NAME: ${BRANCH_NAME}"
-          echo "PACKAGE_VERSION: ${PACKAGE_VERSION}"
-          if [[ ${BRANCH_NAME} == "main" ]]; then
-            SEED_TAG=${PACKAGE_VERSION}
-          elif [[ ${BRANCH_NAME} == "develop" ]]; then
+          echo "GITHUB_EVENT_NAME: ${GITHUB_EVENT_NAME}"
+          echo "GITHUB_REF: ${GITHUB_REF}"
+          if [[ ${GITHUB_EVENT_NAME} == "push" ]]; then
             SEED_TAG=develop
```
macintoshpie, line 42: pushes only happen for develop now so this is fine

macintoshpie, line 42: pushes only happen for develop now so this is fine

.github/workflows/publish.yml
nllong: Woot. You rock!

.github/workflows/publish.yml
nllong: Amazing 🤩

PR #2528, opened 2020-12-30
fix(workflows): fix release tag parsing for publish

macintoshpie:
Publish fails when triggered by release
https://github.com/SEED-platform/seed/runs/1627398200?check_suite_focus=true#step:7:16

#### What's this PR do?
fixes bug where tag had `refs/tags` prefix when triggered by a release.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -41,7 +41,7 @@ jobs:
           if [[ ${GITHUB_EVENT_NAME} == "push" ]]; then
             SEED_TAG=develop
           elif [[ ${GITHUB_EVENT_NAME} == "release" ]]; then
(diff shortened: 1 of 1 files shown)


nllong:
Thanks! Do we need to merge down to main in order for the images to be built?

macintoshpie:
> Thanks! Do we need to merge down to main in order for the images to be built?

👍 👍 Nope, now we just need to publish a release for the image to be built. It doesn't matter if it's merged into main or not. But we also track merges into develop, to publish the `develop` image.

macintoshpie:
Just confirmed this is now working for publishing releases by creating a temporary tag and release:
Screen Shot 2021-01-04 at 11 55 52 AM

I will delete this tag and release now and delete the image from dockerhub.

Issue #2478, opened 2020-12-01
Add API v3 endpoint /analyses/{analysis_id}/stop - POST and create BSyncr stop analysis task


adrian-lara:
The endpoint should only verify perms, trigger the appropriate task, and return the progress_id.

permissions | query params | body params | Notes
-- | -- | -- | --
Org Member | organization_id |   | Triggers separately built celery task

Since this is the initial iteration of this endpoint, a task should be created to stop any BSyncr -specific tasks. I'm envisioning that we could implement the same logic to stop creations or runnings, but I could be missing something there that would prevent that from happening.

This task would likely be built alongside other BSyncr-related tasks (create and start) - maybe within as BSyncrTaskService class?

## Manual Testing
A stop button should be visible and usable for any time the Analysis is doing or about to do something which is indicated by the statuses - `Creating`, `Queued`, and `Running`.

From the tagged PR: 
> ... create an analysis for a property with valid meters, then start the analysis. You should either see queued or running (depending on how fast it updated the status). If it shows queued you can refresh the page to enable the stop button (should show running). Click the stop button and it should stop the analysis.

adrian-lara:
See this comment: https://github.com/SEED-platform/seed/issues/2533#issuecomment-790779036

This particular issue can be tested by stopping analysis at different points of the analysis process. Essentially, nothing should "break", no 500s or 400s, and we should see the status of the analysis as "stopped".

adrian-lara:
Per discussion, for this release, no manual testing is needed. Broader analyses testing will be conducted later once BETTER integration is complete.

PR #2529, opened 2020-12-30
#2478 feat: stop analysis

macintoshpie:
#### What's this PR do?
Adds method for stopping a pipeline, and the analysis endpoint for triggering this stop.

Note that the implementation does **not** stop a running celery task, but prevents future tasks from running (assuming multiple tasks were chained for the pipeline and there are tasks remaining after it was stopped) - see the new `check_analysis_status` decorator used. When `stop` is called, it sets the status to `Stopped`, "ending" the analysis.

The biggest implication of the implementation is that the background task might still update the analysis/analysis property views even after it's been stopped.

This assumes you have bsyncr server setup.
Tweak bsyncr.py to take a long time after starting the analysis. e.g. on line 324 insert
```python
import time
time.sleep(30)
```
create an analysis for a property with valid meters, then start the analysis. You should either see queued or running (depending on how fast it updated the status). If it shows queued you can refresh the page to enable the stop button (should show running). Click the stop button and it should stop the analysis.

#2478 

Screen Shot 2020-12-30 at 2 58 18 PM
Screen Shot 2020-12-30 at 2 58 02 PM

### DIFF
--- a/seed/analysis_pipelines/bsyncr.py
+++ b/seed/analysis_pipelines/bsyncr.py
@@ -6,7 +6,12 @@
 """
 import logging
 
-from seed.analysis_pipelines.pipeline import AnalysisPipeline, AnalysisPipelineException, task_create_analysis_property_views
+from seed.analysis_pipelines.pipeline import (
+    AnalysisPipeline,
+    AnalysisPipelineException,
+    task_create_analysis_property_views,
+    check_analysis_status,
+)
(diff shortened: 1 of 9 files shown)


adrian-lara (approved):
Code changes look good and this works as expected! Tested (with a `sleep()` as suggested) on both the property detail and org-level analyses pages.

Screen Shot 2021-01-04 at 5 05 07 PM
Screen Shot 2021-01-04 at 5 05 11 PM

Issue #2476, opened 2020-12-01
Add API v3 endpoint /analyses/{analysis_id} - DELETE


adrian-lara:
**Blocked by the creation of the "stop" celery task**

This endpoint is not out of the ordinary except for incorporating a check to see if the Analysis has a status of Creating or Running. In either of those cases, a "stop" celery task should be run and completed before performing the actual deletion. The given Analysis' deletion should cascade into all the other, related Analysis- records (though this is probably more of a model/migration definition).

permissions | query params | body params | Notes
-- | -- | -- | --
Org Member | organization_id |   | May need to trigger separately built, stop celery task

macintoshpie:
If the analysis is already in a "resting" state (ie ready, stopped, failed, completed) then we can just delete it without worrying. There are two approaches I can think of for deleting an analysis that's not in a "resting" state (ie creating or running). 

1. Immediately delete the analysis from database and do nothing else

This will cause errors in the celery tasks (e.g. creating analysis property views, input/output files, etc).

2. Do a logical delete of the analysis, and actually delete it from the database at a later time.

This will allow the celery task that's running to not cause any exceptions, but it should stop the chain of tasks somewhat gracefully (this could be done by checking the analysis still exists and has an expected status before starting the next task). This would require doing something similar to what we have for canonical Properties with the `NotDeletedManager`.

The question remains _when_ to actually delete the analysis from the database. The easiest and naive, but workable, solution is to queue a task delayed by some amount of time we expect each task to finish within (e.g. 1 hour) which deletes the analysis. Another easy but not entirely correct option is to create a background celery task which runs at some interval finding and deleting these analyses. One problem with the first is that it assumes all celery tasks will finish within `x` amount of time. The former also has an additional problem that if the celery task queue gets cleared before the analysis is deleted, the analysis will just be there forever. The latter has the problem that needs to assume no tasks are running at some schedule.

The more correct solution is to somehow delete it once the running task actually finishes, but I'm not sure what the best approach for this would be yet.


I think a logical delete with either of the delayed deletion options are the easiest solutions for now. @adrian-lara what are your thoughts?

adrian-lara:
See this comment: https://github.com/SEED-platform/seed/issues/2533#issuecomment-790779036

This particular issue can be tested by deleting analyses during different times (while running, stopped, completed, etc)

adrian-lara:
Per discussion, for this release, no manual testing is needed. Broader analyses testing will be conducted later once BETTER integration is complete.

PR #2530, opened 2021-01-05
2476 feat: delete analysis view

macintoshpie:
SEED is adding ability to run analyses on properties.

#### What's this PR do?
- refactor bsyncr celery tasks (and task wrapper) to more gracefully handle unexpected issues (e.g. someone deleting the analysis). In addition, avoid as much as possible raising unhandled exceptions that are unnecessary
- add a view for deleting an analysis
- add icon for deleting analysis from the frontend
- fix cases where user message or debug message for AnalysisMessage is too long by truncating the message

Monkey test the analysis functionality, deleting analyses in different statuses and see if you can break anything. After deleting an analysis you should no longer see it.

#2476 

Screen Shot 2021-01-05 at 5 10 04 PM

### DIFF
--- a/seed/analysis_pipelines/bsyncr.py
+++ b/seed/analysis_pipelines/bsyncr.py
@@ -10,7 +10,7 @@ from seed.analysis_pipelines.pipeline import (
     AnalysisPipeline,
     AnalysisPipelineException,
     task_create_analysis_property_views,
-    check_analysis_status,
+    analysis_pipeline_task,
 )
 from seed.building_sync.mappings import BUILDINGSYNC_URI, NAMESPACES
 from seed.lib.progress_data.progress_data import ProgressData
@@ -91,9 +91,9 @@ class BsyncrPipeline(AnalysisPipeline):
         return progress_data.result()
 
 
-@shared_task
-@check_analysis_status(Analysis.CREATING)
-def _prepare_all_properties(analysis_property_view_ids, analysis_id, progress_data_key):
+@shared_task(bind=True)
+@analysis_pipeline_task(Analysis.CREATING)
+def _prepare_all_properties(self, analysis_property_view_ids, analysis_id, progress_data_key):
     """A Celery task which attempts to make BuildingSync files for all AnalysisPropertyViews.
 
     :param analysis_property_view_ids: list[int]
@@ -173,9 +173,9 @@ def _prepare_all_properties(analysis_property_view_ids, analysis_id, progress_da
         raise AnalysisPipelineException(message)
 
 
-@shared_task
-@check_analysis_status(Analysis.CREATING)
-def _finish_preparation(analysis_id, progress_data_key):
+@shared_task(bind=True)
+@analysis_pipeline_task(Analysis.CREATING)
+def _finish_preparation(self, analysis_id, progress_data_key):
     """A Celery task which finishes the preparation for bsyncr analysis
 
     :param analysis_id: int
@@ -310,9 +310,9 @@ def _parse_analysis_property_view_id(filepath):
     return int(analysis_property_view_id_elem[0].text)
 
 
-@shared_task
-@check_analysis_status(Analysis.QUEUED)
-def _start_analysis(analysis_id, progress_data_key):
+@shared_task(bind=True)
+@analysis_pipeline_task(Analysis.QUEUED)
+def _start_analysis(self, analysis_id, progress_data_key):
     """Start bsyncr analysis by making requests to the service
 
     """
@@ -371,9 +371,9 @@ def _start_analysis(analysis_id, progress_data_key):
     return output_file_ids
 
 
-@shared_task
-@check_analysis_status(Analysis.RUNNING)
-def _process_results(analysis_output_file_ids, analysis_id, progress_data_key):
+@shared_task(bind=True)
+@analysis_pipeline_task(Analysis.RUNNING)
+def _process_results(self, analysis_output_file_ids, analysis_id, progress_data_key):
     analysis = Analysis.objects.get(id=analysis_id)
     analysis.status = Analysis.RUNNING
     analysis.save()
@@ -401,9 +401,9 @@ def _process_results(analysis_output_file_ids, analysis_id, progress_data_key):
             continue
 
 
-@shared_task
-@check_analysis_status(Analysis.RUNNING)
-def _finish_analysis(analysis_id, progress_data_key):
+@shared_task(bind=True)
+@analysis_pipeline_task(Analysis.RUNNING)
+def _finish_analysis(self, analysis_id, progress_data_key):
     """A Celery task which finishes the analysis run
 
     :param analysis_id: int
--- a/seed/analysis_pipelines/bsyncr.py
+++ b/seed/analysis_pipelines/bsyncr.py
@@ -108,62 +108,50 @@ def _prepare_all_properties(self, analysis_property_view_ids, analysis_id, progr
     analysis_property_views = AnalysisPropertyView.objects.filter(id__in=analysis_property_view_ids)
     input_file_paths = []
     for analysis_property_view in analysis_property_views:
-        try:
-            meters = (
-                Meter.objects
-                .annotate(readings_count=Count('meter_readings'))
-                .filter(
-                    property=analysis_property_view.property,
-                    type__in=[Meter.ELECTRICITY_GRID, Meter.ELECTRICITY_SOLAR, Meter.ELECTRICITY_WIND],
-                    readings_count__gte=12,
-                )
+        meters = (
+            Meter.objects
+            .annotate(readings_count=Count('meter_readings'))
+            .filter(
+                property=analysis_property_view.property,
+                type__in=[Meter.ELECTRICITY_GRID, Meter.ELECTRICITY_SOLAR, Meter.ELECTRICITY_WIND],
+                readings_count__gte=12,
             )
-            if meters.count() == 0:
-                AnalysisMessage.log_and_create(
-                    logger=logger,
-                    type_=AnalysisMessage.INFO,
-                    analysis_id=analysis.id,
-                    analysis_property_view_id=analysis_property_view.id,
-                    user_message='Property not used in analysis: Property has no linked electricity meters with 12 or more readings',
-                    debug_message=''
-                )
-                continue
-
-            # arbitrarily choosing the first meter for now
-            meter = meters[0]
-
-            bsync_doc, errors = _build_bsyncr_input(analysis_property_view, meter)
-            if errors:
-                for error in errors:
-                    AnalysisMessage.log_and_create(
-                        logger=logger,
-                        type_=AnalysisMessage.ERROR,
-                        analysis_id=analysis.id,
-                        analysis_property_view_id=analysis_property_view.id,
-                        user_message=f'Error preparing bsyncr input: {error}',
-                        debug_message='',
-                    )
-                continue
-
-            analysis_input_file = AnalysisInputFile(
-                content_type=AnalysisInputFile.BUILDINGSYNC,
-                analysis=analysis
-            )
-            analysis_input_file.file.save(f'{analysis_property_view.id}.xml', ContentFile(bsync_doc))
-            analysis_input_file.clean()
-            analysis_input_file.save()
-            input_file_paths.append(analysis_input_file.file.path)
-        except Exception as e:
+        )
+        if meters.count() == 0:
             AnalysisMessage.log_and_create(
                 logger=logger,
-                type_=AnalysisMessage.ERROR,
+                type_=AnalysisMessage.INFO,
                 analysis_id=analysis.id,
                 analysis_property_view_id=analysis_property_view.id,
-                user_message='Unexpected error occurred while preparing input for property',
-                debug_message='',
-                exception=e,
+                user_message='Property not used in analysis: Property has no linked electricity meters with 12 or more readings',
+                debug_message=''
             )
-            pass
+            continue
+
+        # arbitrarily choosing the first meter for now
+        meter = meters[0]
+
+        bsync_doc, errors = _build_bsyncr_input(analysis_property_view, meter)
+        if errors:
+            for error in errors:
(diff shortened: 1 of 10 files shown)


seed/analysis_pipelines/bsyncr.py
```diff
@@ -108,74 +109,62 @@ def _prepare_all_properties(analysis_property_view_ids, analysis_id, progress_da
     analysis_property_views = AnalysisPropertyView.objects.filter(id__in=analysis_property_view_ids)
     input_file_paths = []
     for analysis_property_view in analysis_property_views:
-        try:
-            meters = (
```
macintoshpie, line 112: I removed the `try/except` clauses which previously caught unexpected errors (now handled by the `analysis_pipeline_task` decorator). Kind of a bummer it makes it look like everything was modified, sorry ☹️

seed/analysis_pipelines/bsyncr.py
```diff
@@ -108,74 +109,62 @@ def _prepare_all_properties(analysis_property_view_ids, analysis_id, progress_da
...
-        raise AnalysisPipelineException(message)
+        raise StopAnalysisTaskChain(message)
 
 
-@shared_task
-@check_analysis_status(Analysis.CREATING)
-def _finish_preparation(analysis_id, progress_data_key):
+@shared_task(bind=True)
+@analysis_pipeline_task(Analysis.CREATING)
+def _finish_preparation(self, analysis_id, progress_data_key):
```
macintoshpie, line 167: note that adding `bind=True` gives us access to the celery task instance in the function (self)

seed/analysis_pipelines/pipeline.py
```diff
@@ -31,49 +31,166 @@ def task_create_analysis_property_views(analysis_id, property_view_ids, progress
...
+    The decorator provides the following functionality:
+    - Checks if the analysis exists before running the task
+    - Checks if the analysis status is the expected status before running the task
+    - Catches any uncaught exceptions from the task, and handles them as we see fit
+
+    The benefit of this functionality is to
+    1. guard tasks from starting when an analysis has already been stopped, deleted, etc.
+    2. handling some unhandled exceptions gracefully (e.g. a database error when someone else deletes the analysis)
+
+    :param expected_status: int, one of Analysis.STATUS_TYPES
```
macintoshpie, line 70: Looking for feedback on the clarity of the documentation and the general understandability of this function

seed/static/seed/partials/analyses.html
```diff
@@ -33,6 +33,7 @@ <h2><i class="fa fa-bar-chart"></i> {$:: 'Analyses' | translate $}</h2>
                                 <i class="glyphicon glyphicon-trash" aria-hidden="true"></i> -->
                                 <i class="glyphicon glyphicon-play" title="Start Analysis" aria-hidden="true" ng-if="analysis.status === 'Ready'" ng-click="start_analysis(analysis.id)"></i>
                                 <i class="glyphicon glyphicon-stop" title="Stop Analysis" aria-hidden="true" ng-if="['Pending Creation', 'Creating', 'Queued', 'Running'].indexOf(analysis.status) >= 0" ng-disabled="['Pending Creation', 'Queued'].indexOf(analysis.status) >= 0" ng-click="stop_analysis(analysis.id)"></i>
+                                <i class="glyphicon glyphicon-trash" title="Delete Analysis" aria-hidden="true" ng-click="delete_analysis(analysis.id)"></i>
```
macintoshpie, line 36: Not sure when we want to hide/disable this icon. I just left it showing and enabled for all analysis statuses.

seed/analysis_pipelines/pipeline.py
```diff
@@ -31,49 +31,166 @@ def task_create_analysis_property_views(analysis_id, property_view_ids, progress
...
+                    # someone deleted the analysis, and the error was probably due to that
+                    log_message = {
+                        'analysis_id': _analysis_id,
+                        'debug_message': 'Task raised unhandled exception, but the analysis no longer exists. '
+                                         'Assuming the analysis was deleted and stopping the celery task chain.',
+                        'exception': repr(e)
+                    }
+                    logger.info(json.dumps(log_message))
+                    _stop_task_chain(_self)
+                    return
```
adrian-lara, line 179: I'm surprised to see this again as I would've thought the first existence check would be enough. Is this to accommodate for if the Analysis is deleted while the function was being run?

seed/analysis_pipelines/pipeline.py
```diff
@@ -188,6 +297,15 @@ def stop(self):
             locked_analysis.end_time = tz.now()
             locked_analysis.save()
 
+    def delete(self):
```
adrian-lara, line 300: Any reason for the AnalysisPipeline to perform the delete (vs the Analysis object directly)?

seed/analysis_pipelines/pipeline.py
```diff
@@ -132,13 +249,12 @@ def start_analysis(self):
 
         return self._start_analysis()
 
-    def fail(self, message, progress_data_key=None, logger=None):
-        """Fails the analysis. Creates an AnalysisMessage and optionally logs it
-        if a logger is provided.
+    def fail(self, message, logger, progress_data_key=None):
```
adrian-lara, line 252: Was curious about this: "Where's this used? And could that react badly with a non-existent Analysis?" I can answer the first part myself with some more research, but figured I'd pose it here in case it was an easy answer for you.

seed/analysis_pipelines/bsyncr.py
adrian-lara: Thanks for pointing it out - were there any changes at all other than removing the try-except blocks within these lines 111 - 155?

seed/analysis_pipelines/pipeline.py
```diff
@@ -31,49 +31,166 @@ def task_create_analysis_property_views(analysis_id, property_view_ids, progress
...
+    ```
+    @shared_task(bind=True)
+    @analysis_pipeline_task(Analysis.SOME_STATUS)
+    def my_task(self, analysis_id, ...):
+        ...
+    ```
+    Note:
+    - `bind=True` is provided for the shared_task decorator (necessary for stopping the celery chain, ie when we don't want to run the remaining tasks)
+    - `self` is the first argument of the function (the celery task instance is passed as first argument when `bind=True`)
+    - `analysis_id` is some argument (it can be in any position)
```
adrian-lara, line 59: I thought `analysis_id` would be required to verify existence as well as status.

seed/analysis_pipelines/pipeline.py
```diff
@@ -31,49 +31,166 @@ def task_create_analysis_property_views(analysis_id, property_view_ids, progress
...
+                try:
+                    Analysis.objects.get(id=_analysis_id)
+
+                    log_message = {
+                        'analysis_id': _analysis_id,
+                        'debug_message': 'Caught unexpected exception occurred during analysis task (and analysis still exists). Re-raising exception and continuing.',
+                        'exception': repr(e)
+                    }
+                    logger.error(json.dumps(log_message))
+                    raise e
```
adrian-lara, line 168: I'm likely misunderstanding something, but I would've expected us to create an AnalysisMessage in here somewhere. Does that happen else where?

adrian-lara (commented):
I did an initial pass mostly focusing on backend changes, and it looks good so far! I had a few thoughts and comments and figured it was enough to share for discussion sooner than later. 

Happy to discuss these initial ones over a call if that's easier for you.

seed/analysis_pipelines/pipeline.py
macintoshpie: My general idea for the pipeline is to abstract all modifications to the analysis because things can get messy real quick. By forcing all external usages of the analysis to use the AnalysisPipeline we have more control over the expected behavior.

seed/analysis_pipelines/bsyncr.py
macintoshpie: No, there weren't any other changes in that range

seed/analysis_pipelines/pipeline.py
macintoshpie: That's correct. Did you see something that contradicts that? Or should I add that note?

seed/analysis_pipelines/pipeline.py
```diff
@@ -31,49 +31,166 @@ def task_create_analysis_property_views(analysis_id, property_view_ids, progress
...
+
+    :param expected_status: int, one of Analysis.STATUS_TYPES
     :returns: function, a decorator
     """
 
-    def decorator_check_analysis_status(func):
+    def decorator_analysis_pipeline_task(func):
+        # add a property to indicate this function was wrapped
+        # the only purpose of this is so we can test that all tasks have been properly wrapped
+        func._analysis_pipeline_task = True
```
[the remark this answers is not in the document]
macintoshpie: I'm definitely open to other ideas for how to do this. The reason for it all is to make sure our tasks are all wrapped properly. See this new test
https://github.com/SEED-platform/seed/pull/2530/files/b94d75e0262877833bd03ae1c4d6530c9c5eeb4e#diff-cf77a0f3ea8d162f78cd8a494ea87e79c7cbe13bb08b2df5b66db60edcfb52ddR93

seed/analysis_pipelines/pipeline.py
```diff
@@ -31,49 +31,166 @@ def task_create_analysis_property_views(analysis_id, property_view_ids, progress
...
-        return _check
-    return decorator_check_analysis_status
+            # Check the analysis before running the task
+            # If the analysis exists and has the expected status, run the task
+            # If the analysis exists but the status isn't what we expected, and...
+            #   - It WAS stopped or failed, stop the task chain (someone somewhere else stopped/failed the task)
+            #   - It WAS NOT stopped or failed, raise an exception! (something unexpected has happened)
+            # If the analysis no longer exists, stop the task chain (someone somewhere deleted the analysis)
+            try:
+                analysis = Analysis.objects.get(id=_analysis_id)
```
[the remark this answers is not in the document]
macintoshpie: I think that's a good idea, I'll make that change

seed/analysis_pipelines/pipeline.py
macintoshpie: I think that's a good idea. I will add that with a generic user message saying something unexpected went wrong.

seed/analysis_pipelines/pipeline.py
macintoshpie: Yes, this is to handle the case where the task is still running, someone else deleted the analysis, and as a result some unexpected exception is raised in the task (maybe we tried to link an analysis message or file to the analysis). The first existence check is to avoid starting a task when the analysis was already deleted.

seed/analysis_pipelines/pipeline.py
macintoshpie: > Where's this used

This is used within pipeline tasks to "fail" the analysis. bsyncr pipeline uses it if it's not configured to run bsyncr, or for some of the steps it will fail if it is unable to do work for any of the property views (e.g. none have meter data/lat/lng)

> And could that react badly with a non-existent Analysis?

This could cause raise a Analysis.DoesNotExist exception if the analysis doesn't exist.
If called within a pipeline task, and it's not caught (which is how we should call it), then the exception will be logged and 'swallowed' by the analysis_pipeline_task wrapper.
If it's called outside of a pipeline task (like we do in `BsyncrPipeline._prepare_analysis`), then the caller will need to catch it.

I think we should chat about how we should handle this

seed/analysis_pipelines/pipeline.py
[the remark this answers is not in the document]
macintoshpie: TODO: change property to `_is_analysis_pipeline_task`

seed/analysis_pipelines/pipeline.py
```diff
@@ -156,28 +160,78 @@ def _stop_task_chain(task_instance):
...
+                        # don't wait for lock to avoid deadlock
+                        locked_analysis = (
+                            Analysis.objects
+                            .select_for_update(nowait=True)
+                            .get(id=_analysis_id)
+                        )
+
+                        # analysis still exists, and we were able to get a lock on it
+                        # something unexpected happened during the task
+                        if locked_analysis.in_terminal_state():
```
macintoshpie, line 176: a new case I added in this commit. If something goes bad, and someone already put the analysis in a terminal state, just log and exit (don't raise an exception or create an analysis message b/c we could have already ended up in some invalid state)

seed/analysis_pipelines/pipeline.py
```diff
@@ -156,28 +160,78 @@ def _stop_task_chain(task_instance):
...
+                            AnalysisMessage.log_and_create(
+                                logger=logger,
+                                type_=AnalysisMessage.ERROR,
+                                user_message='Unexpected error occurred.',
+                                debug_message='Caught unexpected exception occurred during analysis task and the analysis still exists. '
+                                              'Failing the analysis and re-raising the exception.',
+                                analysis_id=_analysis_id,
+                                analysis_property_view_id=None,
+                                exception=e,
+                            )
```
macintoshpie, line 200: not using `AnalysisPipeline.fail()` here b/c that method tries to grab a lock on the analysis as well. From my testing locally it shouldn't matter (a nested transaction can lock and access same row as a parent) but avoiding it still b/c it makes me uncomfortable.

seed/analysis_pipelines/pipeline.py
adrian-lara: I think this is good - it's fairly robust and is probably needed given all that this wrapper is doing.

adrian-lara (approved):
This works as expected while monkey testing deleting while the Analysis had different statuses: created, running, finished and stopped. I also conducted those tests using multiple windows to mimic multiple users interacting with the same analyses as the same time.

I think I likely didn't hit every exception (try/except) that SEED is now set up to handle, but I'm ok with seeing how those react once we build out the analyses even more.

seed/static/seed/partials/analyses.html
adrian-lara: I think it's fine to display at all times like you have it now.

Issue #2533, opened 2021-01-08
Allow configuration of the bsyncr analysis to handle different model types


adrian-lara:
**Is your feature request related to a problem? Please describe.**

No

**Describe the solution you'd like**

SEED should allow a user to select between different model types when creating BSyncr analysis.

The list of types need to be well-defined, but we should write in a way to handle changes (adding, changing, deleting are potentially on the table):
- Simple Linear Regression with Outside Air Temperature
- Three Parameter Linear Model Cooling
- Three Parameter Linear Model Heating
- Four Parameter Linear Model
- Five Parameter Linear Model
- (maybe others… TBD)

**Describe alternatives you've considered**

N/A

**Additional context**

This could involve some backend work to determine what configurations are allowed for an analysis type (json schema??).

RDmitchell:
@adrian-lara -- is this something I can test? If so, I don't know what to test (!)

adrian-lara:
@RDmitchell Note that testing this ticket would also close this: #2534 

Using this dataset: [Analyses Demo 20201222.zip](https://github.com/SEED-platform/seed/files/6085374/Analyses.Demo.20201222.zip)
Import the buildings first while ensuring you're mapping to `PM Property ID`, `Latitude`, and `Longitude`. Then import the meters. 

You can find instructions on how to create analyses for properties in SEED here: https://github.com/SEED-platform/seed/issues/2499 Only 1 of the records should be able to successfully run Analyses - that one has an address of "123 Location and Meters Lane" (A record needs both the lat/long location and 12 electricity meters). The other records are there to ensure that failure cases are handled well.

**Finally, this particular ticket can be tested by creating and running the different BSyncr analysis types.** Note that the BSyncr type "Three Parameter Linear Model Cooling" isn't working. This has been identified before and is more of a Bsyncr issue vs a SEED issue.

Note that the entire Analyses feature is still being worked on so things may change. But we haven't officially announced it, and we aren't explicitly promoting it to users. Before we do that, we probably want your more thorough review of the entire Analysis feature as a whole, as well as some official documentation on how to use this feature.

adrian-lara:
Per discussion, for this release, no manual testing is needed. Broader analyses testing will be conducted later once BETTER integration is complete.

PR #2537, opened 2021-01-11
2533 feat/modify analyses payload config

aviveiros11:
related to the SEED analyses work
#### What's this PR do?
provides config vars for model choice
open the analysis modal and select a model type to run.  
#2533


macintoshpie (commented):
This mostly works as expected. One issue is that we end up saving a string into the `configuration` field for the analysis when it should be a JSON object.

I think a fix is to bind the secondary dropdown to `new_analysis.configuration.model_type` (not sure if this is the correct way to do this) or alternatively bind it to something else in scope and turn it into a json object before calling the service.
I think the resulting configuration should look something like this 
```json
{
  "model_type": "<selected model here>"
}
```

aviveiros11:
@macintoshpie, let me know if this works! 
image

macintoshpie (approved):
Great, thanks! 🚀

PR #2547, opened 2021-01-13
refactor: update bsyncr pipeline

macintoshpie:
The bsyncr service has been updated to
1. allow configuration of the regression model type
2. return the bsync xml file as well as a plot of the fitted model in a zip

#### What's this PR do?
Modifies the bsyncr pipeline to handle these changes

Run the new migration

Pull the latest bsyncr-server image
```bash
docker pull seedplatform/bsyncr-server:latest
```

Run the container by following steps 1 and 2 in the "how to test manually" section here:
https://github.com/SEED-platform/seed/pull/2498

Note that you don't have to set the `CELERY_*` variables mentioned in step 2.

To test, create and run a bsyncr analysis using different model types: SLR, 3 param cooling, 3 param heating, and 4 param using the New Analysis modal for a property that's been properly setup for bsyncr: it must have latitude, longitude, and 12+ electricity meter readings.

All analyses should run successfully, and should show the derived model JSON when viewing the results detail page. In addition, you should see both a PNG and BuildingSync file in the Output Files column of the results table. Clicking on the PNG should show you the plot generated by bsyncr

Screen Shot 2021-01-12 at 7 34 58 PM


adrian-lara (approved):
Code changes look good, and this works for all cases (I can view the results as well as download the PNG) except for the 3 param cooling option. 

Here's a screenshot of some relevant info on that option:
Screen Shot 2021-01-14 at 12 13 11 PM

Seems more like a BSyncr issue vs a SEED PR issue? So, I'll just approve, feel free to re-request a review if needed.

macintoshpie:
Yeah this is a bsyncr issue. I've seen that error (and others) but I couldn't really figure out why it's happening in bsyncr. I think it has to do with failing to fit a model to the data we're providing, but I'm not sure how to detect and handle this better on bsyncr side.

PR #2556, opened 2021-01-19
feat: misc analysis tweaks

macintoshpie:
We are demoing the analysis functionality soon and would like to make some tweaks on the SEED side.
I would consider this a temporary method of displaying the configuration of an analysis, purely for the demo. We will probably want to find a better way to display the json config down the road.

#### What's this PR do?
- returns analyses from the list endpoint ordered by _reverse_ id (ie newest comes first)
- add the analysis configuration json to the tables that show analysis info

- check all views of analyses to verify configuration is properly being displayed (analysis detail, analyses (org level list), inventory detail analysis list)

Screen Shot 2021-01-19 at 12 28 55 PM

### DIFF
--- a/seed/static/seed/partials/analyses.html
+++ b/seed/static/seed/partials/analyses.html
@@ -18,6 +18,7 @@
                             <th translate>Actions</th>
                             <th translate>Number of Properties</th>
                             <th translate>Type</th>
+                            <th translate>Configuration</th>
                             <th translate>Run Status</th>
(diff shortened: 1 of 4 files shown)


adrian-lara (approved):
I see configuration added in multiple spots - a property's results page, an analysis page, and both the property analyses + org analyses pages. 

Code changes look good - thanks for reordering the analyses. 👍

Issue #2558, opened 2021-01-23
API returns DRF html pages if you're not logged in


axelstudios:
**Describe the bug**
If you try to use the API when logged out (or in an incognito tab) you get a DRF debug page with extra information about the API.

**Expected Behavior**
You should get just the json response:
```json
{
  "detail": "You do not have permission to perform this action."
}
```

**Actual Behavior**
You get an html debug page for endpoints that exist, and json for endpoints that don't exist.

**Steps to Reproduce**
Browse to https://dev1.seed-platform.org/api/version/ in an incognito tab

adrian-lara:
@RDmitchell Testing would just consist of confirming you don't see anything but a plain text when you 
> Browse to https://dev1.seed-platform.org/api/version/ in an incognito tab
<img width="1440" alt="Screen Shot 2021-03-04 at 9 45 21 AM" src="https://user-images.githubusercontent.com/30608004/109998328-5cfb1c80-7cce-11eb-93ea-63e6bf3f93c1.png">

RDmitchell:
instance: dev1

I do get that message when trying to access that URL from an incognito web browser (chrome).
![image](https://user-images.githubusercontent.com/6314950/110372560-4cff7780-8003-11eb-8923-c76c8066fecd.png)

PR #2559, opened 2021-01-25
fix: ensure drf always returns json

axelstudios:
See related issue
https://www.django-rest-framework.org/api-guide/renderers/#setting-the-renderers

#### What's this PR do?
Removes the DRF `BrowsableAPIRenderer`, leaving only `JSONRenderer`

While logged out browse to `/api/version`.  Ensure that json is returned rather than html

#2558


adrian-lara (approved):
👍 
Screen Shot 2021-01-27 at 12 37 36 PM

Issue #2515, opened 2020-12-21
Progress bar sits on 33% for a very long time so that users think the program has crashed


RDmitchell:
**Describe the bug**
When importing a file, the program's progress bar sits on 33% for a very long time (when mapping the data, which I think is when it is doing matching), to the point that users (including me) think the program has crashed, and they/I cancel the mapping. I think this then potentially causes a 504 error when that import is attempted again (?)

**Expected Behavior**
A progress bar that nicely continued to slowly increment as the mapping / matching is working, so that the user knows that the program is still running

**Actual Behavior**
See describe the bug

**Steps to Reproduce**
Import a file with approx 2000 records into an org that has at least 2 cycles.

**Instance Information**
Instance; LBNL production
SHA: e8d5cde48

**Additional Context**
Several cities have reported this, and I have also experienced this doing testing. 

I think it should have a relatively high priority in the new year

adrian-lara:
### To add more details:
From the UI/UX perspective, the main problem point is when the user clicks `Save Mappings` and hits the `start_system_matching_and_geocoding` endpoint while importing a non-BuildingSync file. Currently, we base the progress bar percentage on the fact that we have 3 main tasks - geocoding, matching-merging-linking, pairing. (We denote 4, but it's really 3.) The 2nd tasks, typically called "matching", can take significantly long.
<img width="658" alt="Screen Shot 2021-01-04 at 6 44 09 PM" src="https://user-images.githubusercontent.com/30608004/103591082-1e5b0880-4ebd-11eb-86c0-0be91ef9920a.png">
<img width="658" alt="Screen Shot 2021-01-04 at 6 44 46 PM" src="https://user-images.githubusercontent.com/30608004/103591087-20bd6280-4ebd-11eb-9ed0-c8eb1306526c.png">
<img width="658" alt="Screen Shot 2021-01-04 at 6 45 06 PM" src="https://user-images.githubusercontent.com/30608004/103591093-231fbc80-4ebd-11eb-8dba-6c985cf7505d.png">

adrian-lara:
### Solution Option
@nllong @axelstudios @macintoshpie @Ryoken Let me know what you all think about the following. TL;DR: We keep 3 main steps, but we share more granular steps during the longest step, matching, where it looks like things are stuck at 33%.

[...]

Issue #2550, opened 2021-01-13
Mapping/ Matching finished but status bar never got above 0% complete


RDmitchell:
**Describe the bug**
Imported a file but mapping status bar never got over 0% even though the file was actually properly imported and mapped and the newly imported records show up in the Inventory list.

**Expected Behavior**
Status bar should have reported progress as file was mapped

**Actual Behavior**
See bug description

This was originally reported in #2504, but it is not related to any problems in the file itself, so that issue has been closed.

**Steps to Reproduce**
See this doc in the issue folder 
https://docs.google.com/document/d/1K0a6wCUTRUXpFcUgRCZ8CvOhosXG_sDyO1STzyCnJxM/edit?usp=sharing

The doc above has steps where I tried to reproduce the bug on the LBNL staging server that is updated to 2.9.0 but I can't reproduce it there. 

Not sure we will be able to reproduce this problem.

**Instance Information**
Instance; LBNL production
SHA: e8d5cde48

**Additional Context**
none

RDmitchell:
#2549 is basically a duplicate of this issue, and not as well explained, so I am going to close it, and put remaining comments about this problem in this issue.

RDmitchell:
Instance: LBNL production server
SHA: 93e05f6c9

@axelstudios -- I was on a call with Hannah from the City of San Jose today, and we experienced this same problem. This time I recorded the part where we never saw the program report more than 0% complete for mapping/matching, but the records were imported. 

Here is a link to the folder containing the video
https://drive.google.com/drive/folders/1KRDeOWLCaAxCMEXCMEZhf9k2_gLHB-ux?usp=sharing

It is in 
SEED Shared Folder\Software Development\Issues\Issue 2549

macintoshpie:
@RDmitchell can you share the file you were testing with?

RDmitchell:
It happened with 2 different files, and I am going to get them from the City of San Jose, but haven't received them yet. When I do I will share them with you. Thx

RDmitchell:
@macintoshpie -- you should consult with @axelstudios because he was thinking it might have to do with Cloudflare on the LBNL servers, but not sure if that was ever confirmed.

RDmitchell:
The first file we tried is in the folder linked above with the date "7.20.2020" in the filename.

RDmitchell:
[...]

PR #2560, opened 2021-01-26
#2515 feat: track updates to progress data on frontend

macintoshpie:
When users are importing/mapping large files some users think SEED is broken b/c we aren't updating the progress bar.

#### What's this PR do?
Adds a note to the geocoding/mapping progress bar which will appear to users if the progress hasn't been updated recently.

Import a large file (or add a long time.sleep in seed.data_importer.match.match_and_link_incoming_properties_and_taxlots). and when mapping (which should take a while) you should eventually (after 1 minute of starting the matching task) see the notice shown in the screenshot below.

This is a bandaid for #2515 , doesn't fix it as well as we might want but is a nice thing to have for now.

Screen Shot 2021-01-26 at 4 17 49 PM

### DIFF
--- a/seed/static/seed/js/controllers/data_upload_modal_controller.js
+++ b/seed/static/seed/js/controllers/data_upload_modal_controller.js
@@ -104,6 +104,9 @@ angular.module('BE.seed.controller.data_upload_modal', [])
        *  button.
        * progress: int or float - the progress bar value, i.e. percentage complete
        * complete: bool - true when the upload has finished
+       * status_message: str - status of the task
+       * progress_last_updated: null | int - when not null it indicates the last time the progress bar changed (UNIX Epoch in ms)
+       * progress_last_checked: null | int - when not null it indicates the last time the progress was checked (UNIX Epoch in ms)
        */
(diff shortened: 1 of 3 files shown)


macintoshpie:
@RDmitchell @adrian-lara 
The gist of this change is that if the progress bar isn't updated after 1 minute, we show that message seen in the screenshot. Do either of you have a better feeling for how long we should wait to show the message, or if 1 minute is good enough?

RDmitchell:
See also #2550 which is a different problem but not sure what the status bar should do in that circumstance, because it will never display anything other than 0% (at least I think that is true).

RDmitchell:
@macintoshpie -- you might want to look at #2550 because I think that's a case where it will never actually show any other status than 0% because of timeout issues (possibly) on the server, even though the matching and mapping is still happening. 

You might want to consult with @axelstudios about this, as I think he was looking into that issue.

adrian-lara (approved):
Works as expected. I'll approve for now, as I think this is a reasonable "band-aid". Feel free to re-request if this temporary solution is updated in some way.

Issue #2512, opened 2020-12-19
Inventory Detail: Allow the user to set the field that is displayed for the record


RDmitchell:
**Is your feature request related to a problem? Please describe.**
The program by default shows the Address Line 1 field data at the top of the Detail View to represent the record. 
![image](https://user-images.githubusercontent.com/6314950/102679185-aa9dc980-4162-11eb-90bb-e9295a4da4df.png)

Many times, that field is not mapped. so nothing useful shows up there
![image](https://user-images.githubusercontent.com/6314950/102679208-d9b43b00-4162-11eb-9fb5-32646a91d654.png)

**Describe the solution you'd like**
Allow the user to set the field that should be displayed there.

**Describe alternatives you've considered**
n/a

**Additional context**
San Jose has requested this, as they don't map their various addresses to the Address Line 1 field.

Ryoken:
Suggestion:  add a setting to the Organization called "Property Display Field" with a default of "Address Line 1" that a user can change:
<img width="474" alt="Screen Shot 2021-01-26 at 12 37 23 PM" src="https://user-images.githubusercontent.com/564231/105896056-922b9500-5fd3-11eb-85a7-6c7c409e428f.png">

This will change what field is used to display on the property details page.  If the user inputs a field that doesn't exist it will default to "Address Line 1".

RDmitchell:
@Ryoken -- is this something that should be a "global" setting, or something to be set as part of the Details screen? Seems like it might be more appropriate to have it be at the Details screen level, since that is where we are talking about. 

If it is a global setting, I think the name needs to be much more descriptive, ie, "Property Detail View Display Field".

But I still think it is something that should be set in the Property Detail screen.

Ryoken:
@RDmitchell  While it would work to save this setting in a column list profile, that would only work for pages that use a list profile to display the property.  There are other locations where a property's "Address Line 1" is being used as a title that don't utilize a list profile.  As such, I would recommend keeping it on the org-level settings page, but I can rename it and add a little description so it's clear what this is.

RDmitchell:
@Ryoken -- I actually wasn't thinking it would be part of the list settings but a new special option somewhere on the detail page. 

But putting in the org-level settings is fine. Thanks !!

Ryoken:
Ah, I understand.  Based on that and internal feedback, here's the current idea:

A user can set an org-level setting, selecting from a list of their columns:
[...]

PR #2561, opened 2021-01-26
2512 feat/add property display setting

Ryoken:
#### What's this PR do?
Adds an organization setting to change which field is used to display a property's name.

See the org settings page, change the setting and then view a property detail page.

https://github.com/SEED-platform/seed/issues/2512

(see ticket)


seed/static/seed/js/controllers/inventory_detail_analyses_controller.js
```diff
@@ -103,6 +103,14 @@ angular.module('BE.seed.controller.inventory_detail_analyses', [])
         })
       }
 
+      $scope.inventory_display_name = function() {
+        let field = $scope.org.property_display_field;
+        if (!$scope.item_state[field]) {
+          field = 'address_line_1';
+        }
+        return $scope.item_state[field] ? $scope.item_state[field] : '(invalid display field)';
```
macintoshpie, line 111: I like what you've got here, very clean. I'm wondering if this message might confuse some folks who don't manage the org settings. What do you think about something like `(no ${field})`, or even better `(no ${display name for the field})`, kind of like SEED displayed before?
Also I'm wondering if there is a way we could help org managers fix this if they see this message, ie some help message or link to settings. Not sure if either option is good UI, but in general its useful to say how to fix things in addition to reporting a problem

macintoshpie (commented):
Very nice! I left a few comments/questions

seed/static/seed/js/services/inventory_service.js
```diff
@@ -498,6 +498,21 @@ angular.module('BE.seed.service.inventory', []).factory('inventory_service', [
...
+    inventory_service.get_property_column_names_for_org = function (org_id) {
+      return $http.get('/api/v3/columns/', {
+        params: {
+          inventory_type: 'property',
+          organization_id: org_id,
+          only_used: false,
+          display_units: true
+        }
+      }).then(function (response) {
+        let names = Array.from(new Set(response.data.columns.map(a => a.column_name)));
```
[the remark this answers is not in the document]
Ryoken: Yeah, here's an output of "response.data.columns.map(a => a.column_name)" showing duplicates.
Screen Shot 2021-01-27 at 4 44 17 PM

It looks like they're due to belonging to multiple tables.
Screen Shot 2021-01-27 at 4 46 46 PM

seed/static/seed/partials/inventory_detail.html
```diff
@@ -20,8 +20,8 @@ <h1>{$:: (inventory_type === 'taxlots' ? 'Tax Lot Detail' : 'Property Detail') |
             <div class="section_action_container left" style="width: 50%;">
                 <span>
                     <h2>
-                        <span ng-if="::inventory_type==='properties'"><i class="fa fa-building-o"></i><span translate>Property</span> : {$ item_state.address_line_1 ? item_state.address_line_1 : ('(no address 1)' | translate) $}</span>
-                        <span ng-if="::inventory_type==='taxlots'"><i class="fa fa-map-o"></i><span translate>Tax Lot</span> : {$ item_state.address_line_1 ? item_state.address_line_1 : ('(no address 1)' | translate) $}</span>
+                        <span ng-if="::inventory_type==='properties'"><i class="fa fa-building-o"></i><span translate>Property</span> : {$ inventory_display_name() | translate $}</span>
+                        <span ng-if="::inventory_type==='taxlots'"><i class="fa fa-map-o"></i><span translate>Tax Lot</span> : {$ inventory_display_name() | translate) $}</span>
```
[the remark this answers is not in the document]
Ryoken: The previous implementation uses "address_line_1" regardless of if it was taxlot or property, so I didn't think a way to specify was needed.  I can definitely put in a second setting for tax lots and pull them seperately though.

seed/lib/superperms/orgs/models.py
```diff
@@ -193,6 +193,8 @@ class Meta:
 
     geocoding_enabled = models.BooleanField(default=True)
 
+    property_display_field = models.CharField(max_length=32, blank=False, default="address_line_1")
```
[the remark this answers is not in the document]
Ryoken: Yeah, it felt simpler at the cost of very rarely not having a correct lookup string, at which point the default kicks in just fine.

seed/static/seed/js/controllers/inventory_detail_analyses_controller.js
Ryoken: I was actually thinking about putting in something along those lines so I'm glad you brought it up too.  I'll update with something more clear and possibly a tooltip to direct users to fix it.

seed/static/seed/partials/inventory_detail.html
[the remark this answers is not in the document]
adrian-lara: Ah good catch here @macintoshpie . Yeah, we should probably have a separate setting for each.

seed/static/seed/js/services/inventory_service.js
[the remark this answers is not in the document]
macintoshpie: Yikes, now that I think about this, we shouldn't be getting TaxLotState results if we provide 'property' as the inventory type. Would you mind digging into this to see if that's what's actually happening?

seed/static/seed/js/controllers/inventory_detail_analyses_controller.js
Ryoken: screencapture

seed/static/seed/js/services/inventory_service.js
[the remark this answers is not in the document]
macintoshpie: Ah ok then I guess the API is doing what it says it should, I suppose we will filter on the frontend. Yeah, glancing at that it does seem strange, but maybe I don't understand really what the `related` thing does?

seed/static/seed/js/controllers/inventory_detail_analyses_controller.js
macintoshpie: ohh looks very nice!

seed/static/seed/js/services/inventory_service.js
```diff
@@ -498,6 +498,37 @@ angular.module('BE.seed.service.inventory', []).factory('inventory_service', [
...
+        }
+      }).then(function (response) {
+        let property_columns = response.data.columns.filter(column => column.table_name == 'PropertyState');
+        return property_columns.map(a => {
+          return { 'column_name': a.column_name, 'display_name': property_columns.find(x => x.column_name == a.column_name).display_name };
+        });
+      });
+    };
+
+    inventory_service.get_taxlot_column_names_for_org = function (org_id) {
```
macintoshpie, line 517: I didn't realize this before, but these are very similar to the `get_property_columns`/`get_property_columns_for_org` (same for taxlot) functions, not sure what the best option is here. I guess we could modify the original func with `ui_grid_formatting = true` then conditionally format the display name? Or are there other differences I'm missing as well?

seed/static/seed/js/services/inventory_service.js
```diff
@@ -498,6 +498,37 @@ angular.module('BE.seed.service.inventory', []).factory('inventory_service', [
...
+        params: {
+          inventory_type: 'property',
+          organization_id: org_id,
+          only_used: false,
+          display_units: true
+        }
+      }).then(function (response) {
+        let property_columns = response.data.columns.filter(column => column.table_name == 'PropertyState');
+        return property_columns.map(a => {
+          return { 'column_name': a.column_name, 'display_name': property_columns.find(x => x.column_name == a.column_name).display_name };
```
macintoshpie, line 512:
```suggestion
          return { 'column_name': a.column_name, 'display_name': a.display_name };
```
Would this work? Same goes for taxlot func

seed/static/seed/js/controllers/inventory_detail_analyses_controller.js
```diff
@@ -103,6 +103,19 @@ angular.module('BE.seed.controller.inventory_detail_analyses', [])
...
       }
 
+      $scope.inventory_display_name = function(property_type) {
+        let error = '';
+        let field = property_type == "property" ? $scope.org.property_display_field : $scope.org.taxlot_display_field;
+        if (!(field in $scope.item_state)) {
+          error = field + ' does not exist';
+          field = 'address_line_1';
+        }
+        if (!$scope.item_state[field]) {
```
macintoshpie, line 113: Looks like I'm getting both messages:
Screen Shot 2021-01-29 at 1 51 58 PM

seed/static/seed/partials/inventory_detail.html
```diff
@@ -20,8 +20,13 @@ <h1>{$:: (inventory_type === 'taxlots' ? 'Tax Lot Detail' : 'Property Detail') |
...
                     <h2>
-                        <span ng-if="::inventory_type==='properties'"><i class="fa fa-building-o"></i><span translate>Property</span> : {$ item_state.address_line_1 ? item_state.address_line_1 : ('(no address 1)' | translate) $}</span>
-                        <span ng-if="::inventory_type==='taxlots'"><i class="fa fa-map-o"></i><span translate>Tax Lot</span> : {$ item_state.address_line_1 ? item_state.address_line_1 : ('(no address 1)' | translate) $}</span>
+                        <span ng-if="::inventory_type==='properties'">
+                            <i class="fa fa-building-o"></i><span translate>Property</span> : 
+                        </span>
+                        <span ng-if="::inventory_type==='taxlots'">
+                            <i class="fa fa-map-o"></i><span translate>Tax Lot</span> : 
+                        </span>
+                        <span ng-bind-html="inventory_name">{$ inventory_display_name(inventory_type==='properties' ? 'property' : 'taxlot') | translate $}</span>
```
macintoshpie, line 29: Looks like sanitization is working properly 😉 
Screen Shot 2021-01-29 at 2 21 07 PM
Screen Shot 2021-01-29 at 2 23 13 PM

seed/views/v3/organizations.py
```diff
@@ -516,6 +518,16 @@ def warn_bad_units(kind, unit_string):
         if geocoding_enabled != org.geocoding_enabled:
             org.geocoding_enabled = geocoding_enabled
 
+        # Update property_display_field option
+        property_display_field = posted_org.get('property_display_field', True)
```
macintoshpie, line 522:
```suggestion
        property_display_field = posted_org.get('property_display_field', '')
```
or maybe None, then don't update if None?

seed/views/v3/organizations.py
```diff
@@ -1310,3 +1322,25 @@ def geocoding_enabled(self, request, pk=None):
...
+        return org.property_display_field
+
+    @has_perm_class('requires_member')
+    @ajax_request_class
+    @action(detail=True, methods=['GET'])
+    def taxlot_display_field(self, request, pk=None):
+        """
+        Returns the organization's taxlot_display_field setting
+        """
+        org = Organization.objects.get(id=pk)
```
macintoshpie, line 1344: Are these being used anywhere? I didn't see it anywhere, but could have missed it. Wondering if these endpoints are necessary

macintoshpie (commented):
Tested locally and everything is working as expected except for the case where the inventory doesn't have the requested column and it's address_line_1 is empty (see comment). Few comments/questions on code but things are looking good

seed/static/seed/js/services/inventory_service.js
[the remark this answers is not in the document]
Ryoken: Confirmed that the tables aren't filtered out, rather tagged with the "related" column.  Working as expected!

seed/static/seed/js/services/inventory_service.js
Ryoken: Actually, now that we are intelligently removing the duplicates, yeah, that would work much nicer.

seed/views/v3/organizations.py
Ryoken: No, they are not being used, I just assumed that new members needed to be added.  I honestly don't know if they will be super usefull either so I can remove them.

seed/static/seed/js/controllers/inventory_detail_analyses_controller.js
Ryoken: Modified to accomodate.

seed/views/v3/organizations.py
Ryoken: Updated to have the correct default "address_line_1", thanks for the catch.

seed/static/seed/js/services/inventory_service.js
Ryoken: This endpoint is returning an array of tuples, basically, with only the column name and the display name.  This tuple array is used to populate the select from the org settings page.  Do you think the original functions could and should be used for this?

macintoshpie: I'm good with just keeping it how you have it now, easy enough and your functions are simple.

Issue #2519, opened 2020-12-21
Labels are not always exported from Inventory List


RDmitchell:
**Describe the bug**
Labels are not always exported properly from Inventory List

**Expected Behavior**
exported records should contain all the associated labels in the Property Labels field

**Actual Behavior**
From the user: 
The issue appears to be that some properties are showing as having no flags when they should have them. I haven’t encountered any instances where the flags are there but wrong, only this issue of having none display in the spreadsheet.

I haven’t been able to determine a pattern as to why that is, but maybe it is something that will be resolved with the cycle issue? (see #2517 )

**Steps to Reproduce**
See this doc
https://docs.google.com/document/d/1ka0YIC6vZxyVpE4BLjdae0M2LMZB2i5UcSgvT_H0MCs/edit?usp=sharing

**Instance Information**
Instance: SEED LBNL production / staging
SHA: e8d5cde

**Additional Context**
This may be linked to #2517 or the new cross cycles changes

RDmitchell:
So far, I have not been able to reproduce this problem. I have asked the user to tell me which record(s) she sees this issue with and will update this when I have that information. 

I will assign this to myself in the meantime, to be updated.

RDmitchell:
Now that I have an example from the user, I am removing myself from the issue

Ryoken:
I believe I have figured out why this export problem has been coming up, as well as why it's so random.  Microsoft Excel has a cell character limit of 32,767:

<img width="657" alt="Screen Shot 2021-01-25 at 4 06 38 PM" src="https://user-images.githubusercontent.com/564231/105889281-3f4ddf80-5fcb-11eb-83e6-4e2b835556d1.png">

Google sheets also has a cell character limit, but it's a bit more generous at 50,000:

<img width="668" alt="Screen Shot 2021-01-25 at 4 19 30 PM" src="https://user-images.githubusercontent.com/564231/105889260-3826d180-5fcb-11eb-83bb-d4372869ff02.png">

What this has effecetivly meant is once the notes cell exceeded this limit, the rest of the data is loaded as if it were new cells, only now we're in the middle of a pair of escaping quotes.  Everything that happens after that is messed up due to commas, quotes and endlines that aren't escaped as a part of the previous cell.

Ryoken:
We have a few options here, in order of simplicity and work needed:

* do nothing and communicate that any fields over that length will break most viewing software
[...]

PR #2563, opened 2021-01-29
feat: add option to exclude notes when exporting properties

Ryoken:
The property_notes field was large enough to break Microsoft Excel's cell character limit, causing overflow that would prevent the remaining data from being displayed correctly.

#### What's this PR do?
Adds an option to exclude property notes when exporting.

See the original ticket.

https://github.com/SEED-platform/seed/issues/2519

Screen Shot 2021-01-29 at 3 25 53 PM
<hr>
Screen Shot 2021-01-29 at 3 41 36 PM

### DIFF
--- a/seed/static/seed/partials/export_inventory_modal.html
+++ b/seed/static/seed/partials/export_inventory_modal.html
@@ -2,18 +2,40 @@
     <h4 class="modal-title" id="exportModalLabel" translate>Export your Properties and Tax Lots</h4>
 </div>
 <div class="modal-body form-horizontal" style="padding-bottom: 0;">
-    <div class="form-group">
-        <label class="control-label col-sm-3" for="fileName" translate>Export Name</label>
-        <div class="col-sm-8">
-          <input type="text" class="form-control" id="fileName" ng-model="export_name" placeholder="{$:: 'File Name' | translate $}" required>
+    <div class="section">
+
+        <div class="section_content_container">
+            <div class="section_content with_padding">
+                <label class="control-label col-sm-4" for="fileName" translate>Export Name</label>
+                <form class="form-horizontal" role="form">
+                    <div class="form-group">
+                        <div class="col-sm-6">
+                            <input type="text" class="form-control" id="fileName" ng-model="export_name" placeholder="{$:: 'File Name' | translate $}" required>
+                        </div>
+                    </div>
+                </form>
+            </div>
         </div>
+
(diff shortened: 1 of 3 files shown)


RDmitchell:
@Ryoken -- looks great !

adrian-lara (commented):
Works as expected and I like how we have `(excluded during export)` for when the box isn't checked. One very small change, otherwise this is good to go!

seed/static/seed/partials/export_inventory_modal.html
```diff
@@ -2,18 +2,40 @@
...
+                            <input type="text" class="form-control" id="fileName" ng-model="export_name" placeholder="{$:: 'File Name' | translate $}" required>
+                        </div>
+                    </div>
+                </form>
+            </div>
+        </div>
+
+        <div class="section_content_container">
+            <div class="section_content with_padding">
+                <label class="control-label col-sm-4" for="fileName" translate>Include Property Notes</label>
```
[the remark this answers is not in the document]
Ryoken: I thought about that.  The field on the spreadsheet is named "Property Notes" though, so I ended up going this way to be literal.  Would you prefer it just be "Include Notes"?

adrian-lara:
[image: Screen Shot 2021-02-01 at 1.42.45 PM.png]
I see Tax Lot Notes in a tax lot CSV export. But in general, yeah I'd vote
to just have it as "Include Notes" just to be safe.

seed/views/v3/tax_lot_properties.py
```diff
@@ -142,7 +142,9 @@ def export(self, request):
         data = TaxLotProperty.get_related(model_views, column_ids, columns_from_database)
 
         # add labels and notes
-        include_notes = request.data.get('include_notes')
+        include_notes = True
+        if 'include_notes' in request.data.keys():
+            include_notes = request.data.get('include_notes')
```
adrian-lara: This works as expected but `get` also has an optional second arg that specifies a default that you can throw `True` into and avoid the extra code.

adrian-lara (approved):
This works as expected 👍 

Approving with a small note that could be a quick refactor if you'd like to get that change in.

seed/views/v3/tax_lot_properties.py
Ryoken: I do like that more.  Pushing a new branch soon

Issue #1753, opened 2018-12-20
Add header file to show the filter criteria


RDmitchell:
### Expected Behavior
Add a header file to an export file to show the filter criteria set to generate the list for exporting -- even if no filters were set, have the header file say "no filter criteria".
From the November 2015 mid-month user call (added by LBNL from previous list)
### Actual Behavior 
n/a
### Steps to Reproduce
n/a
### Instance Information
n/a

Ryoken:
How does this look?  With no labels selected it will output a txt file with the contents "no filter criteria".  With any labels selected it will output a txt file with a comma separated list of label names and their ids, plus the method of filter ("and", "or", "exclude").

Exporting a file with no filters:
<img width="1153" alt="Screen Shot 2021-01-11 at 11 58 03 AM" src="https://user-images.githubusercontent.com/564231/104226087-7eddce80-5404-11eb-9bdd-f23aa9008bcc.png">

Exporting a file with a filter:
<img width="1153" alt="Screen Shot 2021-01-11 at 11 52 51 AM" src="https://user-images.githubusercontent.com/564231/104226093-81d8bf00-5404-11eb-9440-379c8790e947.png">

RDmitchell:
@Ryoken -- The original ticket was, I believe, to add a header line to a CSV file exported from SEED, to show what filters were applied when the data was exported.

![image](https://user-images.githubusercontent.com/6314950/104231207-70db7e00-5403-11eb-8e95-4dfe5709170f.png)

So the final exported CSV file would look like:
![image](https://user-images.githubusercontent.com/6314950/104231686-2dcdda80-5404-11eb-8764-0257f340adde.png)

@nllong  -- I am wondering whether we should run this by our current users to see what they think, since they may have systems in place to read the CSV files that this would break ... 

??

nllong:
I'm okay with exporting the filter methods and labels within the CSV. It will require consumers of the CSV file to change their "skip header rows count". It looks like there will only be two new rows added (plus a blank row) regardless of the number of filter labels/methods, right? If it is always the same (i.e., 2 rows), then I think a single file is best.

Ryoken:
Updated to include the header in the first row of only CSV files.  It will always inject 1 and only 1 row.

Exporting a CSV with no filters:
[...]

PR #2566, opened 2021-02-02
feat: added option to exclude label header

Ryoken:
#### What's this PR do?
Adds an option on exporting CSVs to include a label header that describes which labels were used to filter the list.

Filter the property list by a label and export a CSV.

https://github.com/SEED-platform/seed/issues/1753

Screen Shot 2021-02-02 at 12 05 54 PM
Screen Shot 2021-02-02 at 12 05 13 PM
Screen Shot 2021-02-02 at 12 05 39 PM


adrian-lara (commented):
Works as expected and code changes look good. 

A few broad, feature questions I have that may result in code changes:
- Should we default to _not_ include labels to protect CSV integrity.
- When the new option is checked but no labels are used to filter records, should we have some kind of placeholder to specify that like we do for notes?

Ryoken:
I'm a fan of defaulting to not exporting the headers and including "Filter Method: none" if it is included without filters.

RDmitchell:
If there are no filters but the box is checked, I would say that the header should indicate that.

Ryoken:
Pushed a quick change that makes the header disabled by default.  The header was already displaying "none" if no filters existed so no change was needed there.

PR #2567, opened 2021-02-03
check the file to determine an empty excel spreadsheet

aviveiros11:
Error hit during DC workflow when uploading a blank spreadsheet. 

#### What's this PR do?
Checks the file extension and determines whether its populated or not

upload an empty spreadsheet

#2565


axelstudios (approved):

adrian-lara (approved):
Works as expected:
Screen Shot 2021-02-10 at 12 00 59 AM

After a few moments, I also get this error message but it doesn't seem to be added by any of the code you added here, so I'm fine with it. Especially since this seems to be more of a frontend thing vs something hit via the API.
Screen Shot 2021-02-10 at 12 00 38 AM

Issue #1113, opened 2016-11-05
Delete Cycle


ghost:
Hi all,

I have recently started to write a new standard API test that would work with the new data model. The test is supposed to automatically go through all the API requests that a typical user would use and finish by resetting the entire account to a clean slate. This means deleting all buildings records, datasets, projects, labels and cycles that have been created during the test.

The issue I have is that I can't find a way to delete the cycle -if there is any- in the user interface. Whether the features doesn't exist or it exists but there is no button to call it, I believe that it is an important features to have for any users that would want to undo a cycle creation.

Thank you,
Baptiste

macintoshpie:
@adrian-lara @RDmitchell Could you comment with your thoughts and opinions on the design options 1 or 2 below? Adrian one additional question for you: users cannot link properties/taxlots (link in the sense of match_merge_link), or create TaxLotProperties across cycles right?

---

## Research Notes
There is already an endpoint for deleting cycles, and has been around since 42727700ebd92ddcff2d6bd4f0f746b6caa75e64 (about 4 years ago, 2016). This endpoint is not accessible from the frontend.

Cycles are used by the PropertyView, TaxLotView, TaxLotProperty, and AnalysisPropertyView models. The former two use a ForeignKey reference with `on_delete=models.PROTECT`, while the latter two cascade.
This means that if you try to delete a cycle with linked properties or taxlots using the current API, you get an error:
```json
{
  "status": "error",
  "message": "Cannot delete protected objects while related objects still exist"
}
```

The potential consequences for deleting a cycle and all of its *Views are pretty low. TLDR; we've been batch deleting *Views for years, which are really the only things that point to Cycles, so it's safe to delete Cycles.
More specifically, since 42727700ebd92ddcff2d6bd4f0f746b6caa75e64 (the same commit as above) the `state` for *Views has been `on_delete=CASCADE`. And since ae934b80d5934dc4a6cf8b0449f1c289ee9f1228 we've allowed batch deletion of States as an endpoint (not sure when this was added to frontend, but pretty sure it was from the same commit). This means we've been batch deleting *Views for a long time (the main pointers to Cycles).

We have two ways to implement the deletion on the backend. Also see the Design Options section below for frontend choices (maybe do that first...)
1. update the *View models to delete on cascade (convenient but potentially dangerous from the perspective of an API user, ie a non-frontend user)
2. implement logic in endpoint to delete connected *Views before deleting the Cycle (convenient, just as dangerous to API user)
[...]

PR #2568, opened 2021-02-04
feat: add ui for deleting cycles

macintoshpie:
Users cannot delete cycles from the frontend

#### What's this PR do?
Adds ability to delete cycles from the frontend.

Create cycles with and without properties/taxlots. Navigate to the org cycles page, try to delete each cycle. You should be shown buttons to view the properties or taxlots depending on what's in the cycle. If you delete all properties and taxlots you should be allowed to delete the cycle.

#1113 

Note that it now says "Delete Cycle?" instead of "Delete User?" but I haven't updated the screenshots.

Screen Shot 2021-02-03 at 5 29 24 PM
Screen Shot 2021-02-03 at 5 30 13 PM
Screen Shot 2021-02-03 at 5 30 56 PM

### DIFF
--- a/seed/static/seed/js/controllers/delete_cycle_modal_controller.js
+++ b/seed/static/seed/js/controllers/delete_cycle_modal_controller.js
@@ -1,0 +1,59 @@
+/**
+ * :copyright (c) 2014 - 2020, The Regents of the University of California, through Lawrence Berkeley National Laboratory (subject to receipt of any required approvals from the U.S. Department of Energy) and contributors. All rights reserved.
+ * :author
+ */
+angular.module('BE.seed.controller.delete_cycle_modal', [])
+  .controller('delete_cycle_modal_controller', [
+    '$scope',
+    '$window',
+    '$state',
+    '$q',
+    '$uibModalInstance',
+    'inventory_service',
+    'cycle_service',
+    'cycle_id',
+    'cycle_name',
+    function ($scope, $window, $state, $q, $uibModalInstance, inventory_service, cycle_service, cycle_id, cycle_name) {
+      $scope.cycle_id = cycle_id;
+      $scope.cycle_name = cycle_name;
+
+      $scope.cycle_has_properties = null;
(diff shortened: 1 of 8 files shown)


seed/static/seed/js/controllers/delete_cycle_modal_controller.js
```diff
@@ -0,0 +1,59 @@
...
+      $scope.cycle_has_properties = null;
+      $scope.cycle_has_taxlots = null;
+      $scope.cycle_has_inventory = null;
+      $scope.delete_cycle_success = null;
+
+      // determine if there are any properties or tax lots in the cycle
+      // when fetching inventory, to reduce overhead ask for only 1 inventory per page and first page
+      $q.all([
+        inventory_service.get_properties(1, 1, {id: $scope.cycle_id}, null, null, false),
+        inventory_service.get_taxlots(1, 1, {id: $scope.cycle_id}, null, null, false)
```
adrian-lara, line 29: 🧠 👈

seed/static/seed/js/controllers/delete_cycle_modal_controller.js
```diff
@@ -0,0 +1,59 @@
...
+      ]).then(function(responses) {
+        $scope.cycle_has_properties = responses[0].results.length > 0;
+        $scope.cycle_has_taxlots = responses[1].results.length > 0;
+        $scope.cycle_has_inventory = $scope.cycle_has_properties || $scope.cycle_has_taxlots;
+      });
+
+      // open an inventory list page in a new tab
+      $scope.goToInventoryList = function(inventory_type) {
+        inventory_service.save_last_cycle($scope.cycle_id);
```
adrian-lara, line 38: 💯

adrian-lara (approved):
Works as expected and code changes look good!

One thing worth noting is that I feel like we've always developed under the assumption that at least one cycle exists. I monkey tested a bunch of pages on an org without a cycle, and everything seems fine. Probably just something to keep in mind at this point.

Issue #2577, opened 2021-02-12
Add notes count to appear in rows


axelstudios:
**Describe the bug**
Child rows on the inventory list view are using the wrong IDs to lookup the note counts:
![image](https://user-images.githubusercontent.com/411466/107821844-641bb400-6d3a-11eb-8d7d-b1058dc67778.png)

**Expected Behavior**
The shown count should be accurate

**Actual Behavior**
Rather than looking up notes attached to the related PropertyView/TaxlotView it was using the pairing ID from the TaxLotProperty table.

**Steps to Reproduce**
1. Add a note to a child row
2. Refresh the inventory list view and note that the child row note count is not correct

RDmitchell:
Instance: dev1
SHA: b12fbd0

The number updates properly in the Inventory List view after adding a new note.

PR #2578, opened 2021-02-12
fix: use the correct id for related note count lookups

axelstudios:
See the related ticket

#### What's this PR do?
Uses the PropertyView/TaxlotView ID instead of the TaxLotProperty join ID

1. Add a note to child rows in both the property and taxlot tabs
2. Refresh the inventory list view and check that the note counts on child rows are correct

#2577


adrian-lara (approved):
Looks good and works as expected!
Screen Shot 2021-02-15 at 11 58 06 AM

Issue #2379, opened 2020-08-20
Refactor perms for several API v3 views


adrian-lara:
Currently, some API v3 views fetch the organization ID from the request, or fallback to the requester's default org, then use that ID when querying. This will successfully prevent users from querying column mapping profiles of organizations which they are not members.

Unfortunately, the use of this method has been largely deprecated in favor or `@has_perm_class` decorator. We should align all of the v3 endpoints to use this method/pattern for primary authorization.

Note that it might be appropriate to still use the `get_organization` method to easily pull the organization ID or organization resource from the request. The main goal here is to implement more fine-tuned authorization with `@has_perm_class`.

Additional notes here: https://drive.google.com/drive/u/0/folders/1Nc33EosEzBnLzSKoOPUIUSCTWRre8em4

## Note about manual testing
Reminder that testing for this should include general usage of the app while not using a super user account.

aviveiros11:
@adrian-lara or @macintoshpie do you think this issue is related to #2548 because of the organization_id?

macintoshpie:
@aviveiros11 I think these are kinda related. You shouldn't be able to reach the `.get` mentioned in the linked issue if you include the `@has_perm_class`, _but_ if you're testing as a superuser then you bypass the middleware provided by the decorator, ie you can get a 500 a normal user can't. But a normal user can still get a 500 on some endpoints that don't include the decorator, but do try to do a `.get` using a bad org id.

adrian-lara:
From the tagged PR: 

> How should this be manually tested?
> Data Quality Checks
> DQ results are cached with a key that includes org ID now. Verify this still works by running DQ checks from inventory list page and the import file flow. Also test with a suborganization to make sure it works as well.
> 
> General functional testing
> Poke around the entire site to test functionality is maintained. The largest concern is that a user that was previously authorized to perform an action no longer is. Note that you must test as a non-superuser for authorization to be properly tested

PR #2589, opened 2021-02-15
Refactor/authorization

macintoshpie:
#### What's this PR do?
Makes authorization more consistent

Data Quality Checks
DQ results are cached with a key that includes org ID now. Verify this still works by running DQ checks from inventory list page and the import file flow. Also test with a suborganization to make sure it works as well.

General functional testing
Poke around the entire site to test functionality is maintained. The largest concern is that a user that was previously authorized to perform an action no longer is. Note that you must test as a non-superuser for authorization to be properly tested

#2379


seed/lib/superperms/orgs/decorators.py
```diff
@@ -17,6 +17,7 @@
     Organization,
     OrganizationUser
 )
+from seed.lib.superperms.orgs.permissions import get_org_id
```
macintoshpie, line 20: same method used by the OrgMixin, which is used for getting the org id inside of a view. This should be used to ensure each place is using the same strategy for getting the org id

seed/lib/superperms/orgs/permissions.py
```diff
@@ -55,21 +55,43 @@ def get_org_or_id(dictlike):
 
 
 def get_org_id(request):
-    """Get org id from request"""
-    org_id = get_org_or_id(request.query_params)
-    if not org_id:
+    """Extract the ``organization_id`` regardless of HTTP method type."""
```
macintoshpie, line 58: updated method by merging with the _get_org_id func from decorators.py. Also tweaked the sequence of where we look for IDs.

seed/utils/api.py
```diff
@@ -319,15 +319,16 @@ def get_organization(self, request, return_obj=False):
             if not org_id:
                 org = get_user_org(request.user)
                 org_id = int(getattr(org, 'pk'))
+            if not org:
```
macintoshpie, line 322: I can comment on this in person

seed/tests/test_labels.py
```diff
@@ -106,10 +106,10 @@ def test_error_occurs_when_trying_to_apply_a_label_to_taxlotview_from_a_differen
         with transaction.atomic():
             with self.assertRaises(IntegrityError):
                 self.api_view.add_labels(
-                    self.api_view.models['taxlot'].objects.none(),
-                    'taxlot',
-                    [org_1_taxlot.id],
-                    [self.org_2_status_label.id]
+                    qs=self.api_view.models['taxlot'].objects.none(),
+                    inventory_type='taxlot',
+                    inventory_ids=[org_1_taxlotview.id],
```
macintoshpie, line 111: The one important change is this line, we were passing canonical IDs when it needed to be view IDs.

Test was failing for me b/c I was using `--keepdb` flag when running test. Previously this passed b/c the IDs of views and canonical inventory were "synched". Same goes for the other label test I modified

seed/views/data_quality.py
```diff
@@ -159,7 +159,7 @@ def csv(self, request, pk):
               required: true
               paramType: path
         """
-        data_quality_results = get_cache_raw(DataQualityCheck.cache_key(pk))
+        data_quality_results = get_cache_raw(DataQualityCheck.cache_key(pk, self.get_organization(request)))
```
macintoshpie, line 162: Added org id to dq cache key to protect against unauthorized access

seed/views/v3/label_inventories.py
```diff
@@ -128,15 +137,23 @@ def label_factory(self, inventory_type, label_id, inventory_id):
...
-                self.label_factory(inventory_type, label_id, pk)
-                for label_id in add_label_ids for pk in inventory_ids
-                if pk not in exclude[label_id]
-            ]
+            new_inventory_labels = []
+            for label_id in add_label_ids:
+                for pk in inventory_ids:
+                    if pk not in exclude[label_id]:
+                        new_inventory_label = self.label_factory(inventory_type, label_id, pk)
+                        new_inventory_labels.append(new_inventory_label)
```
macintoshpie, line 156: There should be no functional change here. Just reorganized when I was debugging a test to help better understand what was happening

seed/views/v3/import_files.py
```diff
@@ -157,15 +157,15 @@ def retrieve(self, request, pk=None):
         })
 
     @swagger_auto_schema_org_query_param
-    @has_perm_class('requires_viewer')
     @api_endpoint_class
     @ajax_request_class
+    @has_perm_class('requires_viewer')
```
macintoshpie, line 162: has_perm_class must come "after" the other decorators, ie it must be wrapped itself by the other decorators. This is necessary b/c the api_endpoint_class decorator needs to configure the request.user before we check authz

adrian-lara (commented):
Code changes look good - a few very small tweaks. Wanted to leave drop the comments while I try some manual tests.

adrian-lara (approved):
Manual testing looks good to me! We should get this into development and dev1 sooner so that overall app functionality can be tested while other tickets are tested.

Issue #2596, opened 2021-02-18
Add warning to swagger endpoints for *views and *states


macintoshpie:
These endpoints are "dangerous" b/c they do not follow SEEDs normal interface for creating/updating these resources and could lead to a bad state.

We should make a bold warning about these endpoints that they should not be used

macintoshpie:
No further testing needed

PR #2597, opened 2021-02-18
docs: add warnings to nontraditional property endpoints

macintoshpie:
The `property_views/` and `property_states/` endpoints are "dangerous" in that they can put users data in an unexpected state.

#### What's this PR do?
Adds warnings to these endpoints suggesting that they aren't used.

Verify docs on swagger page are updated for these endpoints.

#2596 

Screen Shot 2021-02-18 at 2 20 45 PM


adrian-lara (approved):
Warnings added to the the endpoints that can mutate data 👍

Issue #2603, opened 2021-02-25
Deleting a column can crash SEED due to the memory spiking


axelstudios:
**Describe the bug**
Deleting a column can cause the memory usage to max out and crash seed depending on the number of affected PropertyState/TaxLotState records.

**Expected Behavior**
It shouldn't crash

**Actual Behavior**
It crashes when the memory + swap reaches capacity

**Steps to Reproduce**
Delete a column on an organization with a large number of PropertyState/TaxLotState records (in this case 2.9M PropertyStates are more than enough to exceed 8GB on a VM)

**Instance Information**
Server Instance: `production`

macintoshpie:
Appears to be working on dev1
<img width="1435" alt="Screen Shot 2021-09-27 at 1 52 30 PM" src="https://user-images.githubusercontent.com/18518728/134967892-c17dc973-5b5a-4d4a-ac88-e6ba8bc993e8.png">

SHA: 2.11.0.6247533c9

PR #2609, opened 2021-03-04
2603 fix/column delete memory

axelstudios:
Deleting an extra_data column worked previously, but only for a relatively small number of records and it was synchronous and not chunked.  If you attempt to delete a column that affects 3M records it can easily exhaust the available memory.

Also see the related ticket

#### What's this PR do?
Moves the column deletion into a chunked celery task with asynchronous progress updates

- Attempt to delete a column associated with a large number of records from an organization's Column Settings page.
- Test with both property and taxlot columns
- Verify that only the affected records have been updated, that the extra_data json blob no longer contains the column, and that their hash values have changed

#2603 

image

### DIFF
--- a/seed/lib/progress_data/progress_data.py
+++ b/seed/lib/progress_data/progress_data.py
@@ -150,6 +150,26 @@ class ProgressData(object):
 
         return self.result()
 
+    def step_with_counter(self):
+        """Step the function by increment_value and save back to the cache with a count"""
+        # load the latest value out of the cache
+        self.load()
+
+        value = self.data['progress']
+        if value + self.increment_value() >= 100.0:
+            value = 100.0
+        else:
+            value += self.increment_value()
+
+        self.data['progress'] = value
+        self.data['status'] = 'running'
+        self.data['completed_records'] = round(value / 100.0 * self.data['total_records'])
+        self.data['status_message'] = f'{self.data["completed_records"]:,} / {self.data["total_records"]:,}'
+
+        self.save()
+
+        return self.result()
+
     def result(self):
         """
         Return the result from the cache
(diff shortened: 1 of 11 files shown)


seed/lib/progress_data/progress_data.py
```diff
@@ -150,6 +150,26 @@ def step(self, status_message=None, new_summary=None):
...
+        value = self.data['progress']
+        if value + self.increment_value() >= 100.0:
+            value = 100.0
+        else:
+            value += self.increment_value()
+
+        self.data['progress'] = value
+        self.data['status'] = 'running'
+        self.data['completed_records'] = round(value / 100.0 * self.data['total_records'])
+        self.data['status_message'] = f'{self.data["completed_records"]:,} / {self.data["total_records"]:,}'
```
macintoshpie, line 167: maybe we should add these new fields to `initialize` method?

seed/tasks.py
```diff
@@ -211,6 +229,67 @@ def delete_organization_inventory(org_pk, prog_key=None, chunk_size=100, *args,
...
+            TaxLotState.objects.filter(organization_id=org_pk, data_state=DATA_STATE_MATCHING,
+                                       extra_data__has_key=column.column_name).values_list('id', flat=True)
+        )
+
+    total = len(ids)
+
+    # total is the number of records divided by the chunk size
+    progress_data.total = total / float(chunk_size)
+    progress_data.data['completed_records'] = 0
+    progress_data.data['total_records'] = total
```
macintoshpie, line 260: see other comment about adding these fields to the `initialize` method for ProgressData

macintoshpie (approved):
Just one comment, but everything else looks great 😎 

Tested locally for properties and taxlots and it worked as expected.

PR #2618, opened 2021-03-15
Several miscellaneous fixes

axelstudios:
#### What's this PR do?
- Updates copyrights to 2021
- Adds optional logging of all un-translated strings (commented by default)
- Updates translations with many missing strings
- Fixed Django's `debug` setting so that it could be passed to the UI
- Supervisor tweak so that commands like `supervisorctl stop uwsgi` will work
- Fixed maintenance page email and favicon request, added automatic timestamp

- Check that window.BE.debug accurately reflects the debug setting
- Check that enabling the maintenance page shows the timestamp when it was enabled


adrian-lara (approved):
Reviewing each commit, code changes look reasonable and the two features that can be "tested" with tangible results seem to work as expected
- the maintenance page now shows the time when maintenance mode was turned on Screen Shot 2021-03-17 at 2 37 07 PM
- The debug flag can now be used in the frontend, and I can activate and see the `useMissingTranslationHandlerLog` function produce console logs. Screen Shot 2021-03-17 at 2 36 47 PM

Issue #2410, opened 2020-09-25
Merge and Link Matches -- didn't seem to merge records that should have merged


RDmitchell:
### Expected Behavior
I expected to have the program automatically merge together two records that I had unmerged by hand.
### Actual Behavior
Program didn't merge them back together again
### Steps to Reproduce
Org: LBNL Test 200
Cycle: Test ReMerging

Imported two different files
* 1 - example-data-ESPM-Alerts-CustomID.xlsx
* 2 - example-data-ESPM-Alerts-UBID.xlsx (has a slightly different Energy Star Score than the first file)
![image](https://user-images.githubusercontent.com/6314950/94212719-ceb27d00-fe89-11ea-91e2-9423a3530d7c.png)

Matching fields were Custom ID 1 and PM Property ID
![image](https://user-images.githubusercontent.com/6314950/94212773-f275c300-fe89-11ea-9566-e69805a77ac1.png)

In the Inventory List view, they are all shown as merged
![image](https://user-images.githubusercontent.com/6314950/94212848-1f29da80-fe8a-11ea-8764-5970f9e26b87.png)

Looking at the Inventory Detail View for one of the records, ie, Custom ID 1 == 102, the two records are merged -- the first file (far right column) shows an Energy Star Score of 63. The second file shows an Energy Star Score of 67.
![image](https://user-images.githubusercontent.com/6314950/94212986-8e073380-fe8a-11ea-8b13-93a0bf0f5e8f.png)

I use the "Unmerge Last" button to unmerge the two records.
![image](https://user-images.githubusercontent.com/6314950/94213008-a4ad8a80-fe8a-11ea-9d76-cfbfa6912185.png)

They are now 2 separate records, ie, 2 records with Custom ID 1 == 102.
![image](https://user-images.githubusercontent.com/6314950/94213132-f5bd7e80-fe8a-11ea-89d7-87596507176b.png)

Going back to the Detail view for the first record, click the **Merge and Link Matches** option from the **Action** pulldown
![image](https://user-images.githubusercontent.com/6314950/94213216-1e457880-fe8b-11ea-990f-af75b8abdaea.png)

Nothing happens -- the two records are still unmerged
![image](https://user-images.githubusercontent.com/6314950/94213304-4df48080-fe8b-11ea-90bc-2d36f07751a4.png)

### Instance Information
Instance; seeddemostaging
SHA: 4afb845b3 (2.7.2.1)

RDmitchell:
[...]

PR #2619, opened 2021-03-15
refactor: disabled merge option

Ryoken:
#### What's this PR do?
disables the option to trigger a merge and link on the detail property view dropdown

https://github.com/SEED-platform/seed/issues/2410


seed/static/seed/partials/inventory_detail.html
```diff
@@ -82,7 +82,8 @@ <h2>
                     </li>
 
                     <li role="menuitem">
-                        <a ng-click="open_match_merge_link_warning_modal(match_merge_link_record, 'manual')">Merge and Link Matches</a>
+                        <!-- ng-click="open_match_merge_link_warning_modal(match_merge_link_record, 'manual')" -->
+                        <a style="opacity:.5;">Merge and Link Matches</a>
```
adrian-lara, line 86: Opacity change is reasonable given we can't truly "disable" an anchor element 👍 

We could also override the cursor change with something like `cursor: not-allowed;` or `cursor: default;`:
```suggestion
                        <a style="opacity:.5; cursor: not-allowed;">Merge and Link Matches</a>
```

adrian-lara (approved):
Works as expected. I made a non-essential suggestion. I don't have a strong preference myself, so feel free to make the change or keep it as is now.

seed/static/seed/partials/inventory_detail.html
Ryoken: Good call, I'll update and merge.  Thanks!

Issue #2542, opened 2021-01-11
Organization owners can delete their own organization


adrian-lara:
**Is your feature request related to a problem? Please describe.**

Organization owners are not able to delete their own organization

**Describe the solution you'd like**

This ability should be made available (with proper warnings and confirmation dialogues). 

Note that users should not be able to delete their own base organization - see #2539. As an alternative, users should be able to "reset" their own base organization.

**Describe alternatives you've considered**

**Additional context**

RDmitchell:
Instance: dev1
SHA: b12fbd00f

Got a 403 when tried to delete an org that I was the only member of.

![image](https://user-images.githubusercontent.com/6314950/113190980-d9bbe080-9211-11eb-98d5-13f8912942f9.png)

It does seem like the org was deleted because I can't log into it now.

RDmitchell:
Instance: dev1
SHA: 55c9d2e12

This seems to be fixed. Was able to delete an organization I was the only member of, no 403 error, can't log into it now.

PR #2620, opened 2021-03-18
feat: added delete org functionality for non-admins #2542

axelstudios:
#### What's this PR do?
- Adds the ability for users to delete organizations and sub-organizations that they own
- Reorganized org settings into vertical tabs

- Attempt to delete an org that you own
  - Once it starts the modal cannot be dismissed
  - On success the page will fully reload to the home page (since the page you were on will no longer be accessible)
- Compare the text when deleting orgs vs sub-orgs, it should be specific to the current type
- Ensure that non-owners of orgs and non-owners of sub-orgs cannot delete orgs they belong to (either via UI or API)

#2542 

New, better-organized settings page:
image

Delete modal:
image


seed/static/seed/js/controllers/organization_settings_controller.js
```diff
@@ -119,10 +123,22 @@ angular.module('BE.seed.controller.organization_settings', []).controller('organ
       value: 2
     }];
 
-    const resize_textarea = function () {
+    $scope.confirm_delete = function (org) {
+      $uibModal.open({
+        templateUrl: urls.static_url + 'seed/partials/delete_org_modal.html',
+        controller: 'delete_org_modal_controller',
+        backdrop: 'static',
+        keyboard: false,
```
adrian-lara, line 131: Great idea 💯  Limiting opportunities to just randomly close the modal at weird/unexpected times.

seed/static/seed/partials/organization_settings.html
```diff
@@ -20,201 +20,220 @@ <h2><i class="fa fa-cogs"></i> {$:: 'Settings' | translate $}</h2>
             </div>
```
adrian-lara, line 20: Looks like there's a lot of diffs from elements being reorganized. Is it easy to pick out which diffs are "new" and worth a more thorough review?

adrian-lara (commented):
Just went through the code diff and had a follow up question. Will go through the testing in the meantime.

seed/static/seed/partials/organization_settings.html
axelstudios: Yeah, the diff is actually super small, I only added a little bit of text and a button:
https://github.com/SEED-platform/seed/pull/2620/files#diff-d72e48d3c27f01f02b8a4f80a211aed92bf92ae06272064861d331995bfa4918R224-R232

Any other changes were primarily adding missing translations

adrian-lara (commented):
Trying this via the UI and API, I'm getting a 302 after confirming the deletion, and the org doesn't get deleted:
Screen Shot 2021-03-19 at 1 00 16 PM

Screen Shot 2021-03-19 at 12 53 00 PM

Via the frontend, the progress bar check is hitting the following endpoint `api/v3/progress/undefined/`.

The user I'm signed in as is an owner of the org being deleted and is not a superuser. There any other criteria I need to be checking with the user/org association?

Note, this does seem to work for a different, superuser account.

axelstudios:
@adrian-lara Should be fixed new, two changes:
- org delete permissions were changed from superuser to org owner
- orphaned users are no longer deleted

adrian-lara (approved):
Code changes look good and things seem to work as expected.

As a non-superuser, I could:
- delete organizations and sub-orgs that I was an owner of
- delete an organization even if it was my last, and my account would still exist and be accessible

As a non-superuser, I could **not**:
- delete an org or sub-org I was not a part of (via API).
- delete an org or sub-org for which I was a "viewer" (via API).
- delete an org or sub-org for which I was a "member" (via API).
- delete sub-orgs for which I was the owner of the parent org (via API).

PR #2622, opened 2021-03-18
feat: Postgres 13

axelstudios:
TimescaleDB now supports Postgres 13.  We may want to wait on merging this PR until benchmarking season is over

#### What's this PR do?
Docker deployments of SEED will use the new version of Postgres

- Deploy a SEED instance without a pre-existing database and ensure that SEED functions properly

- Existing installations of SEED will require the database to be dumped while Postgres 12 is still running, removed, and restored to Postgres 13

#### Additional Steps
We also need to run `ALTER EXTENSION timescaledb UPDATE;` on all deployments


nllong:
let's deal with this once we have to.

Issue #2629, opened 2021-03-19
Delete cycles does not work for non-default org


macintoshpie:
**Describe the bug**
For admin view, clicking the delete cycle button causes 500s on the server.

**Steps to Reproduce**
     1. Go to `http://host/app/#/accounts/<org_id>/cycles`.
     2. Choose a different org at top of page as your default
     3. Click delete for one of your cycles
     4. See 500s

**Cause**
For many of the frontend requests, we get org ID from the user service's `get_organization`, which should track whatever the user's current default org is. In this case we don't want the default org, but the org id of whatever organization they're managing.

This is a deeper issue b/c part of the delete cycle flow allows the user to open a new tab viewing the inventory list they need to delete. But in order to do this now, we'd need to modify the user's default org which is probably very surprising for a user, and would probably break some things.

macintoshpie:
Options for fixing:
1. Just say "hey user you need to go do xyz yourself then come back to this page to delete the cycle". This is easy to implement, but not convenient for the user.
2. Provide a link to the relevant inventory page (setting correct cycle and org) which doesn't open a new tab, but redirects in the current tab. It would have to refresh the page as well to make sure the other components update properly (this is because we'd need to update the user's default org). Could be surprising for the user b/c it changes their default org, but is a little more convenient.
3. refactor how FE handles default organization for all pages. This is probably idea, users should be able to create a link directly to a page owned by an org they're part of (but isn't necessarily their default org). This is likely a lot of work

PR #2631, opened 2021-03-22
fix: fix delete cycles flow

macintoshpie:
Ability to delete cycles from org admin page is broken.

#### What's this PR do?
Fixes deletion of cycles.

1. Create two organizations
2. go through cycle deletion flow as described in [this PR](https://github.com/SEED-platform/seed/pull/2568) when your default org (ie the dropdown in upper right) is the same as the one for the cycle you're deleting (ie the cycle is part of your default org), and then again when deleting a cycle not part of your default org. It should work for both cases.

When deleting cycles from non-default org, clicking the "view properties" button in the deletion modal should open a new tab with the correct cycle and organization set. In addition, it should refresh the cycles settings page (ie the page the deletion modal is on), which should then have the default org (drop down in upper right) set to the organization of the cycle you're deleting.

Test celery by uploading a file that has 100s of lines in it (I just copy/pasted first few lines in the example-pm properties file) and verify that the progress bar works as expected.
 
#2629


axelstudios:
Rather than trying to fetch 1 property and 1 taxlot it might be cleaner to request `/api/v3/organizations/XXX` which has the property/taxlot totals by cycle (so you could optionally show the quantity in the dialog):
image


Alternatively, you could just take the code from that endpoint that determines num_properties & num_taxlots and add it to the `/api/v3/cycles/` response

macintoshpie:
@axelstudios @adrian-lara I updated the branch to include Alex's suggestions. Can you all take another look?

macintoshpie:
@axelstudios could you take a glance at the most recent additions? Also can you load test this with your large file again to make sure it doesn't crash?
@adrian-lara ready for your review 👍

seed/static/seed/partials/delete_cycle_modal.html
```diff
@@ -49,6 +53,6 @@ <h4 class="modal-title" translate>Delete Cycle?</h4>
     </div>
 </div>
 <div class="modal-footer">
-    <button type="button" class="btn btn-default" ng-click="cancel()">{$delete_cycle_success == null ? 'Cancel' : 'Close'$}</button>
-    <button ng-if="cycle_has_inventory == false && delete_cycle_success == null" type="button" class="btn btn-danger" ng-click="confirmDelete()" autofocus translate>Yes, Delete Cycle</button>
+    <button type="button" class="btn btn-default" ng-click="cancel()">{$ delete_cycle_status == 'success' ? 'Close' : 'Cancel' $}</button>
```
adrian-lara, line 56: I believe that hitting `Cancel` wouldn't actually Cancel the deletion. If that's right, then I'd vote to not even show the `Cancel` button once this has kicked off. If we do that, it might be worth to disabling the click-away or keyboard escape like we do for owner org deletion here: https://github.com/SEED-platform/seed/pull/2620/files#diff-2705f5db618213153e520ff3d70128459ca769b8c0034414a2b4522235fdf3b1R130-R131

adrian-lara (approved):
Code look good! Worth calling out that I have one comment that could lead to code changes.

I'll approve despite that since this does seem to work as expected. I'll defer to you or others on whether we want to make the suggested change (the Cancel button comment).

seed/static/seed/partials/delete_cycle_modal.html
macintoshpie: ah nice, I like it

adrian-lara (approved):
Once kicked off, the Cancel button is no longer shown and users can't escape via off-modal click or esc key 💯

Issue #2595, opened 2021-02-18
New User Improvements: Update Sample Data


laurenliz22:
Update current sample data to allow new users to go through SEED's more notable features as a supplement to the tutorial.  Include easy-to-use column mapping with the sample data

Data:
- Use Alex's advanced features list (see data from https://seed-platform.org/references/ under "Presentations".  There is a zip file with a directory for each feature to try out
- Maintain existing simple dataset
- Provide data sufficient to running analysis for Bsyncr (Adrian has an example dataset he has been using for the Bsyncr analysis) or BETTER

Pre-populate new user accounts with the data:
- If someone has no Properties then have them "click here" to populate with sample data
- Option to reset the current organization (this is currently captured in another ticket)

laurenliz22:
@RDmitchell - The following New User Improvement tickets are suggestions our team developed to enhance the experience for new SEED Platform users.  Before we start working through these tickets to update SEED we want to get your sign off on the changes.  Would you mind looking at the descriptions and letting me know if you have any questions or changes.  Thanks! Lauren

RDmitchell:
I don't quite understand all the nuances of the data suggestions, but I'm sure you will come up with a good solution.

RDmitchell:
@macintoshpie / @axelstudios -- should I test this on a "public instance" ? Is there such an instance? 

And if not, should I just test on dev1, with a new user account?

macintoshpie:
@RDmitchell testing on dev1 should be sufficient, you can create a new account or new organization to test.

RDmitchell:
instance: dev1
SHA: 
Org: LBNL 444
Cycle: 2020 Calendar Year (default)

I created a new user and org, then upon logging in I clicked on the "Auto-Populate Sample Data"

This screen appeared
![image](https://user-images.githubusercontent.com/6314950/128071866-2b13f116-b83c-41b1-bfba-2e78529a8419.png)

I clicked the Continue button and the program populated the org with the data

RDmitchell:
Here are my comments about the sample data -- I can make separate issues for these if that is useful, and close this issue.

* Geocoding
  * There are actually 3 properties that show up on the map -- the one left out of the description in the screenshot above is PM Property ID # 5766973; Seems like it should be added to the list
![image](https://user-images.githubusercontent.com/6314950/128093993-6739b8d3-3bdd-4c4c-ba0b-cb52850ece26.png)

* Merging
[...]

PR #2636, opened 2021-03-25
Feat/populate sample data

laurenliz22:
This is specifically for new users to allow them the opportunity to learn and use the SEED Platform

#### What's this PR do?
This PR specifically updates the Download Sample Data files to include Interval data.  This PR also creates an API that will be hit for new users to have sample data auto-populated in SEED if they do not have an organization.  The data includes:
- 1 record for Geo-coding
- 2 merged records
- 1 record with building footprints
- 1 record with an associated tax-lot
- 4 records with labels
- 2 records with interval data
- 1 record with green button imported data

Check the API creates the records above by entering your organization.  Soon a button will be created on the landing page to do this automatically for new users once they click it.

#2595


axelstudios:
Here's the modal you see after clicking `Auto-Populate Sample Data`:

seed lan_8000_app_

It uses the first cycle to evaluate for a lack of inventory - if you have property or taxlot records already you won't be able to proceed and you'll see this banner in the modal:
seed lan_8000_app_ (1)

On success it will automatically close the modal and navigate to the Inventory List view - it should have the correct cycle and column settings profile (`Auto-Populate`) selected, showing only populated columns

seed/static/seed/js/controllers/sample_data_modal_controller.js
```diff
@@ -0,0 +1,131 @@
...
+
+          return inventory_service.get_property_columns().then(columns => {
+            return organization_service.insert_sample_data(organization.org_id).then(() => {
+              return inventory_service.get_properties(1, undefined, cycle, -1).then(inventory => {
+                const visibleColumns = findPopulatedColumns(columns, inventory.results);
+                const profileId = profile.id;
+                profile = _.omit(profile, 'id');
+                profile.columns = visibleColumns;
+                return inventory_service.update_column_list_profile(profileId, profile);
+              });
```
macintoshpie, line 59: A high level comment describing the overview of this process might be helpful.

Also, seems like we're doing a lot of work on the frontend here which I would assume we could do in the endpoint?

seed/static/seed/js/controllers/sample_data_modal_controller.js
```diff
@@ -0,0 +1,131 @@
...
+            }).catch(response => {
+              Notification.error('Error: Failed to insert sample data');
+              console.error(response.data);
+            });
+          });
+        }).then(() => {
+          inventory_service.save
```
macintoshpie, line 66:
```suggestion
```

seed/views/v3/organizations.py
```diff
@@ -1392,3 +1403,353 @@ def reset_all_passwords(self, request, pk=None):
...
+        org = Organization.objects.get(id=pk)
+        cycles = Cycle.objects.filter(organization=org)
+        if cycles.count() == 0:
+            return JsonResponse({
+                'status': 'error',
+                'message': 'there must be at least 1 cycle'
+            }, status=status.HTTP_400_BAD_REQUEST)
+        cycle = cycles.first()
+
+        property_details = [{
```
macintoshpie, line 1423: not critical, but it might be nice to move this into a json file

seed/views/v3/organizations.py
```diff
@@ -1392,3 +1403,353 @@ def reset_all_passwords(self, request, pk=None):
...
+
+        import_meterdata = ImportFile.objects.create(
+            import_record=import_record,
+            source_type="PM Meter Usage",
+            uploaded_filename=filename,
+            file=SimpleUploadedFile(name=filename, content=open(filepath, 'rb').read()),
+            cycle=cycle
+        )
+
+        save_raw_data(import_meterdata.id)
```
macintoshpie, line 1736: This starts a background task which might not finish before we return 'success' to the client. Not sure how much of a concern that is for now. (same goes for later call to save_raw_data)

macintoshpie (commented):
Great job Lauren! Here's my code review, I'm starting to manually test this now and will report back if I have any additional recommended changes.

seed/views/v3/organizations.py
```diff
@@ -1392,3 +1403,353 @@ def reset_all_passwords(self, request, pk=None):
...
+        for dic in property_details:
+            state = PropertyState(**dic)
+            state.save()
+            ids.append(state.id)
+
+            # merge
+            for count in range(len(property_details)):
+                if count > 9:
+                    state_ids_to_merge = [state.id]
+                    state_ids_to_merge.append(state.id)
```
macintoshpie, line 1682: [non-critical] This is a bit confusing. Could we break this loop out of here so we construct state_ids_to_merge after creating all of the property states? It looks like we want to merge states 10 and onwards, so something like this might be a bit more clear
```
# after creating all states
state_ids_to_merge = [id for id in ids[9:]]
```

seed/views/v3/organizations.py
```diff
@@ -1392,3 +1403,353 @@ def reset_all_passwords(self, request, pk=None):
...
+        Create a button for new users to import data below if no data exists
+        """
+        org = Organization.objects.get(id=pk)
+        cycles = Cycle.objects.filter(organization=org)
+        if cycles.count() == 0:
+            return JsonResponse({
+                'status': 'error',
+                'message': 'there must be at least 1 cycle'
+            }, status=status.HTTP_400_BAD_REQUEST)
+        cycle = cycles.first()
```
macintoshpie, line 1421: I'm wondering if we should also add a check like we do on FE to make sure this cycle doesn't have any inventory?

macintoshpie (commented):
Just tested everything manually and it works as expected 👍 
Added a few more comments on backend code, but since things work as is they might not be too important to address right now

seed/static/seed/js/controllers/menu_controller.js
```diff
@@ -147,6 +150,39 @@ angular.module('BE.seed.controller.menu', [])
...
+                  let lastCycle;
+                  if (typeof lastCycleId === 'number') {
+                    lastCycle = response.organization.cycles.find(cycle => cycle.cycle_id === lastCycleId)
+                  }
+                  if ((lastCycleId === undefined || !lastCycle) && response.organization.cycles.length) {
+                    lastCycle = response.organization.cycles[0];
+                  }
+                  return lastCycle;
+                });
+            }],
```
adrian-lara, line 178: This assumes the existence of a cycle. The homepage's "Auto-Populate Sample Data" button doesn't work (the modal won't show) if we've deleted all of our cycles. This is more of an edge case, but we do display the warning message of "To continue you must first remove pre-existing data", which doesn't specify what all should be deleted. I doubt a new user would think to delete a cycle.

Anyway, knowing it's an edge case, I'll defer to others on whether we want to make change here, but here are a few considerations/options:
- If we want to allow the auto-populate feature even without a Cycle, the sample data modal would need to be updated to not need a cycle. And the insert_sample_data endpoint would would need to be updated to create a cycle.
- Alternatively, we could check for a cycle in this function before opening the modal and, if none exists, prompt the user to create a cycle via Notification.

Issue #2626, opened 2021-03-19
Authorize /media file requests


macintoshpie:

macintoshpie:
Verified to be working on dev1:
<img width="1132" alt="Screen Shot 2021-09-27 at 1 49 20 PM" src="https://user-images.githubusercontent.com/18518728/134967600-b7a38d62-bd4c-4fd6-8dc6-d5f05e64dff6.png">

SHA: 2.11.0.6247533c9

PR #2639, opened 2021-03-26
refactor: add media api for serving protected user files

macintoshpie:
#### What's this PR do?
Adds permissioning in front of serving media files (ie user uploaded files).
More specifically, creates `/api/v3/media/<filepath>` (and sets it as the MEDIA_URL) which verifies the requesting user has access to the file, before instructing Nginx to serve the file using the X-Accel-Redirect header. If the user doesn't have access or the file doesn't exist, it 404s.

If using docker-compose setup, you must be using the non-dev version (ie docker-compose.local). This is because this update requires Nginx to serve the files.
Create two separate orgs with different user accounts
Try to access different media files while logged in as different users (or not logged in). Media file locations:
- /api/v3/media/uploads
    - (used for most property imports)
    - create by importing from a csv
- /api/v3/media/buildingsync_files/
    - (used for bsync imports)
    - create by importing bsync file
- /api/v3/media/analysis_input_files/<analysis id>/
    - (where input files for an analysis are stored)
    - create by running bsyncr analysis. (for refresher, [ see this PR](https://github.com/SEED-platform/seed/pull/2547))
- /api/v3/media/analysis_output_files/
    - (where analysis output files are stored)
    - create by running bsyncr analysis

Try to access `<host>/media/<some file>` directly and it should fail (/media is no longer the MEDIA_URL). Try to access `<host>/protected/<some file>` and it should fail (the protected location is _internal_ to Nginx, meaning it's only accessible when not requested externally, ie in client request)

Try to break it by using it in unexpected ways...

#2626


seed/views/v3/uploads.py
```diff
@@ -29,6 +29,14 @@
 _log = logging.getLogger(__name__)
 
 
+def get_upload_path(filename):
```
macintoshpie, line 32: moved this out to use with testing

axelstudios:
Nice work - looks like file uploads to `/seed/media/uploads/*` don't get listed under files?  We probably want to add a download icon to each column in the history that specifies `ImportFile` (and probably also next to each upload on the datasets page)

macintoshpie:
TODO: when checking ImportFiles, make sure `deleted=False`

macintoshpie:
TODO: add download links for files (e.g. property detail, data set, etc)

macintoshpie:
@adrian-lara Should be ready for your review now. Let me know if you need help getting bsyncr analyses up and running again, I know it can be a pain

adrian-lara (approved):
Code changes look good, and happy + sad paths seem to be working as expected! 👍🏽

Issue #2634, opened 2021-03-24
Error on importing xlsx with empty first sheet


macintoshpie:
**Describe the bug**
Importing xlsx file with empty first sheet results in error

**Expected Behavior**
It should not error if data from other sheets will be used

**Steps to Reproduce**
1. upload PM meter example file (data_imports/tests/data/example-pm-monthly-meter-usage.xlsx)
2. It will fail claiming it's empty

macintoshpie:
Suggested fix:
on this line, set `sheet_name=None` to read all sheets (defaults to first) before checking if any sheets have data
https://github.com/SEED-platform/seed/blob/3547ce6e9012fcbb13e477ad5fbc6b0c2f648c37/seed/views/v3/uploads.py#L105

RDmitchell:
Instance; Dev1
SHA: b12fbd00f

@adrian-lara / @macintoshpie -- does this only apply to PM Meter spreadsheets? 

I tried uploading a normal ESPM spreadsheet that just has one tab, added a blank tab as the first tab, tried importing it, and get a 503 error. But maybe that's not the test I should be doing. 
![image](https://user-images.githubusercontent.com/6314950/113199735-53f16280-921c-11eb-9338-2d59a3b5d21f.png)

RDmitchell:
Instance; Dev1
SHA: b12fbd0

Loading the file "example-pm-monthly-meter-usage.xlsx" from the repo which does have a blank first tab, it seems like the program imported it
![image](https://user-images.githubusercontent.com/6314950/113200638-6f10a200-921d-11eb-9d66-2de8abd1451c.png)

But now it appears "stuck" here
![image](https://user-images.githubusercontent.com/6314950/113200668-7932a080-921d-11eb-9896-e73976f95046.png)

RDmitchell:
@adrian-lara -- see the previous comment to figure out if my assumption about making the parent records is correct or not.

macintoshpie:
@RDmitchell This should only be successful when importing meter data. ESPM defaults to first sheet when importing, so maybe we should add an additional check to discriminate between meter/ESPM. I would consider this sufficient to merge for now.

I've created this ticket to handle empty first sheet for ESPM: #2648

RDmitchell:
@macintoshpie -- except that I don't see any new records in the Inventory list -- is this because the parent records didn't already exist?

macintoshpie:
[...]

PR #2641, opened 2021-03-26
fix: allow pm meter xlsx to have empty first sheet

macintoshpie:
SEED has a check to determine if an Excel file is empty when uploading

#### What's this PR do?
Fixes issue where it would reject meter spreadsheets that have an empty first sheet (supposedly allowed according to example files).

Upload `seed/seed/data_importer/tests/data/example-pm-monthly-meter-usage.xlsx` and it should work (has an empty first sheet). Upload a spreadsheet where all sheets are empty and it should fail.

#2634

### DIFF
--- a/seed/views/v3/uploads.py
+++ b/seed/views/v3/uploads.py
@@ -100,10 +100,11 @@ class UploadViewSet(viewsets.ViewSet, OrgMixin):
         if not os.path.exists(os.path.dirname(path)):
             os.makedirs(os.path.dirname(path))
 
-        extension = the_file.name.split(".")[1]
+        extension = the_file.name.split(".")[-1]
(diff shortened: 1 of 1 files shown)


seed/views/v3/uploads.py
```diff
@@ -100,10 +100,11 @@ def create(self, request):
         if not os.path.exists(os.path.dirname(path)):
             os.makedirs(os.path.dirname(path))
 
-        extension = the_file.name.split(".")[1]
+        extension = the_file.name.split(".")[-1]
```
adrian-lara, line 103: 👍🏽 surprised we haven't had any issues with this yet.

adrian-lara (approved):
LGTM! Meter uploads with an empty first sheet are successful

PR #2600, opened 2021-02-23
2538 feat/self registration

aviveiros11:
#### Any background context you want to provide?
PR to allow self registration of users first logging on to the site.  Note: This does not include assigning an organization, that will be handled by #2539.  When testing, the tester will be able to reach the home page with a red banner saying the user is unaffiliated with an organization.  Due to the reCAPTCHA domain req's - this can only be tested on `localhost:8000` 

#### What's this PR do?
Includes a self registration form with google reCAPTCHA and email verification link.  

#### How should this be manually tested?
Please contact me for the reCAPTCHA secret key to be included in the tester's `local_untracked`.  Reconfigure the development settings by running `DJANGO_SETTINGS_MODULE=config.settings.dev`. After the dev server is running: 

Scenario 1
- Follow the create account workflow all the way through.  (the verification email will be sent to the console)
- Logout and login with the same credentials used to create the account. 

Scenario 2
- Verify error handling on the create an account form is comprehensive. 
 - un-check reCAPTCHA and try to log in.
 - try entering password values that don't meet authentication validation.
 - try to create an account that already exists.

Scenario 3
- After creating an account, try to login with the credentials not via the confirmation email. (should not be possible).

#### What are the relevant tickets?
#2538

PR #2644, opened 2021-03-29
feat: add self registration env toggle for deployments

aviveiros11:
We needed a way to toggle self-registration for deployments

#### What's this PR do?
creates an env variable to be set for self registration capability

add `INCLUDE_ACCT_REG` to local_untracked and set to False.  The create account button should not show on the FE.  

#2600

### DIFF
--- a/seed/landing/urls.py
+++ b/seed/landing/urls.py
@@ -5,6 +5,7 @@
 :author
 """
 
+from django.conf import settings
 from django.conf.urls import url
 from django.contrib.auth.views import (
     logout_then_login, PasswordChangeView, PasswordChangeDoneView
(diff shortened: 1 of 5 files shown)


seed/landing/urls.py
```diff
@@ -17,10 +18,6 @@
 
 urlpatterns = [
     url(r'^$', landing_page, name='landing_page'),
-    url(r'^accounts/create/$', create_account, name='create_account'),
-    url(r'^account_activation_sent/$', account_activation_sent, name='account_activation_sent'),
-    url(r'^activate/(?P<uidb64>[0-9A-Za-z_\-]+)/(?P<token>[0-9A-Za-z]{1,13}-[0-9A-Za-z]{1,20})/$',
```
adrian-lara, line 22: Still reviewing the code changes, and I haven't tested yet... but I think this would prevent _anyone_ from activating an account even if the invite/activation came from an existing user trying to invite a new user to join an existing org.

aviveiros11: I was able to add users to organizations locally.  This logic flow is actually separate from the invitation and account setup flow from adding users to organizations.  I could be missing something though!

adrian-lara: My bad! I didn't verify my assumption that it was the same flow! :smh:

Yeah this does seem to still allow invitations from existing admins to go through and sign ups to be successful 👍🏽

adrian-lara (approved):
I don't see the Create Account button on the UI, and I 404'd hitting `accounts/create/`,  `account_activation_sent`, and `activate` endpoints with the new flag set to `False` 👍🏽 

Code changes look good! Thanks for correcting my incorrect assumption there 😓

PR #2652, opened 2021-04-01
fix: various auto-populate improvements

axelstudios:
The auto-populate functionality has already been merged, but this PR adds a number of small improvements

#### What's this PR do?
- Backend verification that the cycle doesn't contain any properties or taxlots
- Fixed merge logic and removed duplicate from merge
- Frontend error notification if no cycles exist
- Call `insert_sample_data` before fetching columns since that process creates one `Note` extra_data column.  This is for the Show Only Populated Columns functionality
- Moves data to insert into separate json file
- Creates a TaxlotAuditLog record for the one taxlot record

- Delete all cycles and attempt to open the modal
- Test the `insert_sample_data` API call for an org that already contains data

#2595


seed/views/v3/organizations.py
```diff
@@ -1702,7 +1483,8 @@ def insert_sample_data(self, request, pk=None):
             geocode = PropertyState.objects.filter(id__in=ids)
             geocode_buildings(geocode)
 
-        # Create a merge of 2 properties
+        # Create a merge of the last 2 properties
+        state_ids_to_merge = ids[-2:]
```
macintoshpie, line 1487: 🆒

seed/views/v3/organizations.py
```diff
@@ -1722,12 +1504,12 @@ def insert_sample_data(self, request, pk=None):
         import_record = ImportRecord.objects.create(name='Auto-Populate', super_organization=org)
 
         # Interval Data
-        filename = "example-pm-monthly-meter-usage.xlsx"
-        filepath = os.path.dirname(os.path.abspath(__file__)) + "/data/" + filename
+        filename = 'example-pm-monthly-meter-usage.xlsx'
+        filepath = f"{Path(__file__).parent.absolute()}/data/{filename}"
```
macintoshpie, line 1508: you probably know this already, but I learned you could also use `/` and strings to create paths (no better, just thought I'd mention it)
```python
str((Path(__file__).parent / 'data' / filename).absolute())
```

macintoshpie (approved):
Looks good to me! Tested both with cycle with data and no cycles and got the expected errors.
Screen Shot 2021-04-08 at 9 12 05 AM

adrian-lara (approved):
I only tested the items I left comments about (tax lot audit and no cycle exists bug), but they work well!

Code changes related to those two look good, and the rest of the code changes look reasonable! 👍🏽

seed/views/v3/organizations.py
axelstudios: I haven't seen that, but that's a cool operator override

Issue #2666, opened 2021-04-08
Fix bsync measures import


macintoshpie:
When importing bsync files, the "recommended" status is not correctly set and the list of measures pointed to a scenario is not working (only grabs the first).

macintoshpie:
See testing notes in mentioned PR

PR #2667, opened 2021-04-08
fix(building_sync): get all linked measures for scenario

macintoshpie:
Also correctly parse "auc:Recommended" for measure

We weren't properly getting all linked measures for scenarios (bad xpath). Also our check for "recommended" was bad for measures.

#### What's this PR do?
Fixes bsync import

Import [this file](https://gist.github.com/macintoshpie/4ffe5e66d43aba9e4735b048dcb8c567) (an tweaked version of buildingsync_ex01_measures.xml -- added extra measures to one of the scenarios and set recommended = false for one measure). Just use the default bsync column mapping.

Verify that:
- the measure "Retrofit with light emitting diode technologies" has recommended == false and all other measures have recommended == true
- the scenario "LED Only" points to 3 different measures

#2666 

Screen Shot 2021-04-08 at 11 33 59 AM
Screen Shot 2021-04-08 at 11 33 46 AM

### DIFF
--- a/seed/building_sync/mappings.py
+++ b/seed/building_sync/mappings.py
@@ -654,11 +654,11 @@ BASE_MAPPING_V2_0 = {
                 'value': 'exist'
             },
             'measure_ids': {
-                'xpath': './auc:ScenarioType/auc:PackageOfMeasures/auc:MeasureIDs',
+                'xpath': './auc:ScenarioType/auc:PackageOfMeasures/auc:MeasureIDs/auc:MeasureID',
                 'type': 'list',
                 'items': {
                     'id': {
-                        'xpath': './auc:MeasureID',
+                        'xpath': '.',
                         'type': 'value',
                         'value': '@IDref'
                     }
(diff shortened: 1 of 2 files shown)


seed/models/building_file.py
```diff
@@ -180,14 +180,15 @@ def process(self, organization_id, cycle, property_view=None):
             implementation_status = m['implementation_status'] if m.get('implementation_status') else 'Proposed'
             application_scale = m['application_scale_of_application'] if m.get('application_scale_of_application') else PropertyMeasure.SCALE_ENTIRE_FACILITY
             category_affected = m['system_category_affected'] if m.get('system_category_affected') else PropertyMeasure.CATEGORY_OTHER
+            recommended = str(m.get('recommended', 'false')).lower() == 'true'
```
macintoshpie, line 183: if `m['recommended'] == True` then this check failed.

seed/building_sync/mappings.py
```diff
@@ -654,11 +654,11 @@ def update_tree(schema, tree, xpath, target, value, namespaces):
                 'value': 'exist'
             },
             'measure_ids': {
-                'xpath': './auc:ScenarioType/auc:PackageOfMeasures/auc:MeasureIDs',
+                'xpath': './auc:ScenarioType/auc:PackageOfMeasures/auc:MeasureIDs/auc:MeasureID',
```
macintoshpie, line 657: whoops, needed to go one more level down to actually get a "list" of ids

adrian-lara (approved):
Code changes look reasonable, and I was able to verify the two items listed under testing notes. 

Thanks for adding the two short comments in the diffs.

Issue #2677, opened 2021-04-16
Multiple visual org member permissions issues


axelstudios:
**Describe the bug**
- If your role is `member` you see a dropdown AND text for every user
- As a `member` you had an option to change your role to `owner` even though the attempt would fail with a 403.  You could also attempt to change other users which would all fail with a 403
- UI doesn't update if you demote yourself to reflect your new level of permissions (like hiding the `Remove` buttons next to members or the buttons to invite new users, etc)
- Removing a user with only one org prompted about deleting the user entirely, which doesn't actually happen any more
- Roles are shown in alphabetical order rather than hierarchical order (owner > member > viewer)

**Expected Behavior**
- Superusers can edit any user's role except for the org owner if there is only one (that said, superusers can't view an org's member settings if they are configured as a `viewer` or if they aren't included in the org)
- Owners can edit all roles except their own if they are the only owner
- Members can only edit their own role, and only to change it to "viewer"
- Non-superusers cannot promote their role

**Steps to Reproduce**
- In an org with multiple owners demote your role from `owner` and note that the UI buttons for owners are still all there until you refresh
- As a `member` note that it appears you can edit or promote any user's role

**Screenshots**
Members see both a dropdown and text in the `Member Role` column:
![image](https://user-images.githubusercontent.com/411466/115074742-5d82f780-9eb7-11eb-9fcf-bb61a726ba18.png)

RDmitchell:
Instance: dev1
SHA: 9d9368cbf

I think all these changes are working as expected, and they have been tested by Adrian in the PR, so I am closing this issue. 

I ran through the following:

* Check that removing a user with only one org doesn't show the modal about deleting the user\
  **YES**
Removing Member that is the only Member of another org doesn't show a modal about deleting the user (not sure this was the right test)
* Check that roles are now in hierarchical order (owner -> member -> viewer)
  **YES**
![image](https://user-images.githubusercontent.com/6314950/126576073-c81ad3db-529a-4baa-b558-fce0f2ce26a4.png)
* You must be either an owner or member to see an org's members page (this hasn't changed)
  **YES**
![image](https://user-images.githubusercontent.com/6314950/126576193-f62ef3c9-6840-4df4-ba42-f1ebfc81911a.png)
* Superusers can edit any user's role except for the org owner if there is only one
  * _**Not sure what a superuser is, so didn't test this, but seems like this was tested by Adrian, so I am not worried about testing it.**_
[...]

PR #2678, opened 2021-04-16
fix: resolve several member permission issues

axelstudios:
See the related issue

#### What's this PR do?
Improves the logic around whether someone can edit org member roles

- Check that removing a user with only one org doesn't show the modal about deleting the user
- Check that roles are now in hierarchical order (owner -> member -> viewer)
- Check that the following revised rules work as expected:
    - You must be either an `owner` or `member` to see an org's members page (this hasn't changed)
    - Superusers can edit any user's role except for the org owner if there is only one
    - Owners can edit all roles except their own if they are the only owner
    - Members can only edit their own role, and only to change it to "viewer"
    - Non-superusers cannot promote their role
    - Once you demote yourself you no longer see options to promote yourself and the UI reflects your new permissions

#2677


adrian-lara (commented):
Code changes look good and I tested the different cases mentioned here. Each screenshot shows the items being verified.

Verified without screenshot:
> You must be either an owner or member to see an org's members page (this hasn't changed)
> Check that removing a user with only one org doesn't show the modal about deleting the user


Screen Shot 2021-04-19 at 3 51 28 PM
> Check that roles are now in hierarchical order (owner -> member -> viewer)
> Superusers can edit any user's role except for the org owner if there is only one
> Owners can edit all roles except their own if they are the only owner


Screen Shot 2021-04-19 at 4 07 59 PM
Screen Shot 2021-04-19 at 4 09 05 PM
> Non-superusers cannot promote their role
> Once you demote yourself you no longer see options to promote yourself and the UI reflects your new permissions
> Members can only edit their own role, and only to change it to "viewer"

adrian-lara:
As discussed, I came across two potential bugs.

As a user being removed from my last org by someone else, I'm no longer able to log in. (I can still _delete my own_ last org and be able to log in.)  Until I was added to an **existing** org, I could not log in. I couldn't log in if someone created a new org for me.

Also, 
Screen Shot 2021-04-19 at 4 10 37 PM
per discussion, I couldn't demote myself from owner to viewer.

axelstudios:
@adrian-lara Just made the following changes:
- Allow members to change their role to viewer (enforced by the backend).  Superusers can still change anyone's role for orgs they belong to
- Don't deactivate users with 0 org memberships
- The "Delete User" modal is gone since users aren't deleted or deactivated

adrian-lara (approved):
Code changes 👍🏽 
Testing 👍🏽 
- I was able to change my role from member to viewer as a non-superuser.
- Upon being removed from my last org (by someone else), I was still able to log in. (I'm now prompted to go through the new org flow like any other org-less user.
- There's no longer a delete user modal when a user is being removed from their last org.

Issue #1643, opened 2018-05-22
Add ability for users to display labels and allow which labels to see in the list view


RDmitchell:
### Expected Behavior
Users would like the be able to select labels that could then be displayed in the List view (not just in Detail view). 

An example would be "compliance status" labels such as Complied, Rejected, Pending or New (or whatever labels the user creates). 

If they could select, for example, these 4 labels to appear in the list view (good UI design needed here !!), it would potentially make it easier for them to do a quick look through the list for records with certain labels, not just filtering by those labels.  The program wouldn't be filtering on those labels, just displaying them. 

The biggest issue seems to me to be a nice UI design for this, because labels can be long. But in the end it would be up to the users to select which labels to display.

I would recommend some thought by @alex about this feature, and then maybe we could have a design call with a few interested users to present them with some possibilities and get feedback. 

There was already an issue about this, #1567, which had much less thought put into it, so I am going to close that issue in favor of this one. 
### Actual Behavior 
Program does not have this functionality
### Steps to Reproduce
Program does not have this functionality
### Instance Information
instance: LBNL staging/ production
SHA: 36cc59d

RDmitchell:
I am going to add the labels for points, etc, from #1567 to this issue.

Ryoken:
Ok, here's the initial implementation.  On the labels admin page you can set whether the label is shown on the list:
<img width="1242" alt="Screen Shot 2021-04-21 at 8 11 26 PM" src="https://user-images.githubusercontent.com/564231/115738428-48401a00-a34a-11eb-852f-eb87b90e4cf6.png">

When marked as such and applied to a property/taxlot, they will now appear on the list under a new column.  This column starts collapsed and only shows a small color bar to indicate a label, which can be hovered over to see the full label text:
<img width="1049" alt="Screen Shot 2021-04-21 at 8 11 35 PM" src="https://user-images.githubusercontent.com/564231/115738432-49714700-a34a-11eb-9304-cb7fcca859de.png">

Clicking the expand arrow in the header resizes the column to display the full labels:
<img width="1123" alt="Screen Shot 2021-04-21 at 8 11 43 PM" src="https://user-images.githubusercontent.com/564231/115738433-49714700-a34a-11eb-9271-feeedbeb7712.png">

adrian-lara:
This looks good to me! @RDmitchell thoughts?

RDmitchell:
@Ryoken / @adrian-lara

[...]

PR #2679, opened 2021-04-27
1643 feat/add label column

Ryoken:
#### What's this PR do?
Adds a property to labels that denote whether the label shows up in a new column in the property/taxlot listing.  The labels column can be hovered over to see the full label text and clicked to expand.

Mark a label as showing uo in the list, assign label to a property and see it appear in the list.

https://github.com/SEED-platform/seed/issues/1643

Screen Shot 2021-04-21 at 8 11 26 PM
Screen Shot 2021-04-22 at 3 41 09 PM
Screen Shot 2021-04-21 at 8 11 35 PM
Screen Shot 2021-04-22 at 3 40 47 PM
Screen Shot 2021-04-21 at 8 11 43 PM


seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -92,10 +94,73 @@ angular.module('BE.seed.controller.inventory_list', [])
       } else {
         // No profiles exist
         $scope.columns = _.reject(all_columns, 'is_extra_data');
-      }
+      };
 
       $scope.restoring = false;
 
+      // Find labels that should be displayed and organize by applied id
+      $scope.show_labels = [];
```
macintoshpie, line 102: took me a sec to figure this out, how would you feel about changing this variable name to `$scope.labels_by_inventory_id` or something like that? A bit tough to infer what this is stored. Or maybe changing the comment above from `organize by applied id` to `organize by inventory id`?

seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -92,10 +94,73 @@ angular.module('BE.seed.controller.inventory_list', [])
...
+        setTimeout(() => {
+          $scope.gridApi.grid.getColumn('labels').width = $scope.get_label_column_width();
+          let icon = document.getElementById('label-header-icon');
+          icon.classList.add($scope.show_full_labels ? 'fa-chevron-circle-left' : 'fa-chevron-circle-right');
+          icon.classList.remove($scope.show_full_labels ? 'fa-chevron-circle-right' : 'fa-chevron-circle-left');
+          $scope.gridApi.grid.refresh();
+        }, 0);
+      };
+
+      $scope.get_label_column_width = function () {
```
macintoshpie, line 143: Nice! I'm not sure if we want to extend the column as wide as the labels go though. I tested adding a lot of labels (which is a semi-reasonable thing, IDK what the max number of labels typically would be though), and it prevents you from viewing the data (cant scroll to the right either):
Screen Shot 2021-04-30 at 3 13 41 PM

I reckon this could also happen with fewer labels with longer names.

I guess the alternative is to set a max-width and let the height of the row expand, or to set a max-width and make the td scrollable. I think either option is good

macintoshpie (commented):
This is very slick and looks great 😎 . I left a few code comments.

I thought it'd be nice if we automatically "refreshed" the labels column after a user adds or removes labels from the inventory list page. It appears the label filter search bar detects these changes, so maybe we just need to rerun the funcs that initialize the data.

Also, I wonder if we should add labels to a property's taxlot (or vice versa). E.g. this property is part of a taxlot (second row) but we don't display the taxlot's labels. This might get messy, so I wouldn't worry about it too much if it seems like a lot of trouble.
Screen Shot 2021-04-30 at 3 23 27 PM

seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -92,10 +94,73 @@ angular.module('BE.seed.controller.inventory_list', [])
...
 
+      // Find labels that should be displayed and organize by applied id
+      $scope.show_labels = [];
+      for (let n in labels) {
+        let label = labels[n];
+        if (label.show_in_list) {
+          for (let m in label.is_applied) {
+            let id = label.is_applied[m];
+            if (!$scope.show_labels[id]) {
+              $scope.show_labels[id] = [];
```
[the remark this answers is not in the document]
Ryoken: If this is still working the way I recall, it's creating a sparse array under the hood.  I'm find changing it to a dictionary however, that's definitely more common and readable.

seed/static/seed/js/controllers/inventory_list_controller.js
Ryoken: Renamed to "show_labels_by_inventory_id" for clarification.

seed/static/seed/js/controllers/inventory_list_controller.js
Ryoken: I set a max width of 1000px (which should rarely if ever be hit) and set the container to allow scrolling in the x direction.    If this ends up being not enough for a user I would recommend turning some labels off.

seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -98,17 +98,17 @@ angular.module('BE.seed.controller.inventory_list', [])
 
       $scope.restoring = false;
 
-      // Find labels that should be displayed and organize by applied id
-      $scope.show_labels = [];
+      // Find labels that should be displayed and organize by applied inventory id
+      $scope.show_labels_by_inventory_id = {};
```
macintoshpie, line 102: gracias 👍

seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -118,14 +118,14 @@ angular.module('BE.seed.controller.inventory_list', [])
...
-            let label = $scope.show_labels[id][i];
+        if ($scope.show_labels_by_inventory_id[id]) {
+          for (let i in $scope.show_labels_by_inventory_id[id]) {
+            let label = $scope.show_labels_by_inventory_id[id][i];
             labels.push('<span class="', $scope.show_full_labels ? 'label' : 'label-bar', ' label-', label.label, '">', $scope.show_full_labels ? label.text : '', '</span>');
             titles.push(label.text);
           }
         }
-        return ['<span title="', titles.join(', ') ,'" class="label-bars">', labels.join(''), '</span>'].join('');
+        return ['<span title="', titles.join(', ') ,'" class="label-bars" style="overflow-x:scroll">', labels.join(''), '</span>'].join('');
```
macintoshpie, line 132: nice

macintoshpie (approved):
Looks great!

Issue #1642, opened 2018-05-22
Data Analysis: Ability to define new calculated fields based on existing data


RDmitchell:
### Expected Behavior
New feature which would allow users to define new fields in SEED (after data has been imported), based on other fields in data that has already been imported, including the ability to define calculations based on existing data (fields). 

* This is a bit of a follow-on to #1640 -- it would be useful in defining DQ rules (as illustrated in the 2nd example in 1640) but would be useful in general for data analysis. 

* And it is also a follow-on to #1641 -- maybe that issue is really this issue, and it's not a formal report, but the user is able to define that new field and how the program would calculate it. 
 * For example, in 1641, the user could define those new % fuel type fields as desired, selecting the fields from the data that has been imported
   * % Electricity Mix = [Electricity Use - Grid Purchase (kWh)] / [Site Energy Use (kBtu)]
   * % Natural Gas Mix = [Natural Gas Use (therms)] / [Site Energy Use (kBtu)]
 * Then maybe in a user defined reporting function, the fields could be defined to be part of the report (#1641 )
### Actual Behavior 
Program currently doesn't have this functionality
### Steps to Reproduce
Program currently doesn't have this functionality
### Instance Information
instance: LBNL staging/ production
SHA: 36cc59d

macintoshpie:
@RDmitchell @nllong @adrian-lara 
Here's a wireframe of what I'm imagining this feature basically looks like to get things rolling. I'll get a time setup for all of us to align our ideas and discuss what's needed/possible.

Here are some questions we'll need to consider, no need to answer these right now, just something to start thinking about.
- what are the operations a user can use in the expression?
  - `-, +, /, *`
- should we only allow arithmetic, or also allow string operations?
  - maybe just arithmetic for now to keep things simple?
- who can create derived columns?
  - org admins?
- where do you create derived columns?
  - in org dashboard
- when are derived values updated? Is it applied to all properties/taxlots?
  - when derived column is created?
  - do you automatically apply all derived columns to new properties in imported data?
  - if updating existing properties on import, do you recalculate derived columns?
  - after manually editing/updating referenced field for a property?
- are derived columns cycle-specific? (if a user has cycle A and cycle B, imports to cycle B with new cols,
  do we add those cols to properties in cycle A?)
- can derived columns reference derived columns?
- can you edit derived columns after they're created?
  - implication is creating cyclical dependencies
- when importing, can a user use a derived column name as the column name for an import field?
[...]

PR #2682, opened 2021-05-12
Feat/derived columns

macintoshpie:
Currently, users can import data into SEED into the predefined core columns (e.g. Site EUI) or into extra data (e.g. My Custom Column) for an inventory.

#### What's this PR do?
Adds initial models, DerivedColumns, which will allow users to define expressions with parameters derived from inventory values.

For example, `$a / $b * 100`, where `$a` might reference  the column "Conditioned floor area" and `$b` might reference "Total floor area".

This implementation allows numerous operators/functions (see EXPRESSION_GRAMMAR in derived_columns.py). Currently expressions can reference any `Column`. If an inventory has missing or invalid data used in the expression, it should return `None`.

This PR only adds the basic models for storing these expressions as well as code for evaluating/validating them. I.e. there is no API for creating/reading/updating/deleting these models.

Nothing for now, see unit tests.

#1642


requirements/base.txt
```diff
@@ -59,6 +59,7 @@ requests==2.22.0
 lxml==4.6.3
 probablepeople==0.5.4
 xmlschema==1.1.1
+lark==0.11.3
```
macintoshpie, line 62: package for parsing/evaluating expression strings

seed/tests/test_derived_columns.py
```diff
@@ -0,0 +1,611 @@
...
+            atomic_no_params_st | st.sampled_from(template_params),
+            recursive_st_func
+        )
+    )
+
+    return expression, params
+
+
+class TestExpressionEvaluator(TestCase):
+    @no_deadline
```
macintoshpie, line 166: @adrian-lara 
```suggestion
class TestExpressionEvaluator(TestCase):
    """NOTE:
    This collection of tests uses the hypothesis testing library
    See: https://hypothesis.readthedocs.io/en/latest/index.html

    Instead of making assertions about specific data, we tell hypothesis what the data _basically_
    looks like, and it runs our tests many times over different types of data it generates.
    This allows us to test cases we would have never though of and find edge cases.

    We create data specifications (ie what the data should look like) with "strategies".
    We have written custom strategies for generating different types of expressions above.
    They all end with the `_st` suffix.
    We then tell hypothesis which strategy to use for a test (ie what the data input is) by using the
    `@given` decorator.
    """

    @no_deadline
```

seed/tests/test_derived_columns.py
```diff
@@ -0,0 +1,626 @@
...
+        derived_column = models['derived_column']
+        property_state = models['property_state']
+
+        # -- Act
+        result = derived_column.evaluate(property_state)
+
+        # -- Assert
+        self.assertIsNone(result)
+
+    def test_derived_column_evaluate_falls_back_to_passed_params_if_not_in_inventory(self):
```
adrian-lara, line 542: What's the use case for passing in "fallback" params? It seems related to not having an inventoryState object to base values off, but it isn't immediately clear to me when that might happen. 

I have a feeling I'm missing/forgetting something notable 😅

seed/migrations/0138_auto_20210512_0024.py
```diff
@@ -0,0 +1,51 @@
...
+                ('organization', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='orgs.Organization')),
+            ],
+        ),
+        migrations.CreateModel(
+            name='DerivedColumnParameter',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('parameter_name', models.CharField(max_length=50)),
+                ('derived_column', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='seed.DerivedColumn')),
+                ('source_column', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='seed.Column')),
```
adrian-lara, line 31: Not sure if this is the relevant line, but a question comes to mind here: What happens if a column is deleted that was associated to a DerivedColumn via a reference from this DerivedColumnParameter?

adrian-lara (approved):
A few open questions that may or may not lead to code changes. In general, the overall patterns make sense to me, and I don't see anything "off" there! 👍🏽  I'll go ahead and approve in case no code changes end up being needed. 

I'm less concerned with missing an edge case, etc given we'll later implement and use this in the frontend and test it more thoroughly during that point.

*Note/reminder that if we notice any performance hits during testing, we may consider limiting the number of tests being autogenerated for the different cases.

seed/models/derived_columns.py
```diff
@@ -0,0 +1,344 @@
...
+    # All parameters used in the expression must be linked to a Column through this
+    # model's `source_columns` field
+    expression = models.CharField(max_length=526, blank=False)
+    source_columns = models.ManyToManyField(Column, through='DerivedColumnParameter')
+
+    class Meta:
+        constraints = [
+            # name must be unique
+            models.UniqueConstraint(
+                fields=['name'], name='unique_name'
```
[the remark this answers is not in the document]
macintoshpie: Oops, good catch!

seed/tests/test_derived_columns.py
macintoshpie: Yeah I think I just added that without considering it too much. I think "falling back" onto a params dict is unlikely. But passing only a dict is somewhat likely. I just didn't want to make the arguments mutually exclusive, and had to decide a precedence.

seed/migrations/0138_auto_20210512_0024.py
macintoshpie: Great point, I'm going to use `PROTECT` on this for now until we decide how we want to handle this.
With `CASCADE` the `evaluate` method is just going to end up returning `None` every  time b/c it's missing that parameter for all inventory.

macintoshpie: See
- https://github.com/SEED-platform/seed/pull/2682/files#diff-fc0d01e7c8f83cc5d37ed1e01da02c0b9c6998a458e79e9f08eccf4f32626899R31
- https://github.com/SEED-platform/seed/pull/2682/files#diff-bfbe128f201eda3032ef868db0ae6d82527e6352043c9bcbf68e97ed3b81bc16R315

seed/models/derived_columns.py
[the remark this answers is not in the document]
macintoshpie: See
- https://github.com/SEED-platform/seed/pull/2682/files#diff-bfbe128f201eda3032ef868db0ae6d82527e6352043c9bcbf68e97ed3b81bc16R206

adrian-lara (approved):
Thanks for addressing the points I made on my last review! This looks good to go! 👍🏽

PR #2686, opened 2021-05-21
feat: add derived columns list view

macintoshpie:
Seed is adding ability for users to define columns derived from other columns

#### What's this PR do?
Adds a list API endpoint and an initial iteration of displaying them on the Frontend

Create some derived columns then browse the page. Note the edit/delete buttons do not work yet.
You can create some derived col data by running this command (note you should update the `organization_id` variable if your's isn't `1`)
```bash
python manage.py shell <<-'EOF'
"""
Creates derived columns (but does not create DerivedColumnParameters)
"""

from random import randint

from seed.models import *

BASE_NAME = 'Fancy Column'

# replace org id
organization_id = 1

expressions = [
    '$a + $b + $c',
    '$a / $b * 100',
    '$a * 3.14 + 123 % 100',
    '($a - $b) / ($c + $d)',
]


for i in range(10):
    dc = DerivedColumn.objects.create(
        name=f'{BASE_NAME} {randint(100, 1000)}',
        expression=expressions[randint(0, len(expressions) - 1)],
        organization_id=organization_id,
        inventory_type=DerivedColumn.PROPERTY_TYPE
    )
    print(f'Created derived column with id {dc.id}')
EOF
```

#1642 

Screen Shot 2021-05-20 at 7 52 53 PM

### DIFF
--- a/seed/serializers/derived_columns.py
+++ b/seed/serializers/derived_columns.py
@@ -1,0 +1,27 @@
+# !/usr/bin/env python
+# encoding: utf-8
+"""
+:copyright (c) 2014 - 2021, The Regents of the University of California, through Lawrence Berkeley National Laboratory (subject to receipt of any required approvals from the U.S. Department of Energy) and contributors. All rights reserved.  # NOQA
+:author
+"""
+from rest_framework import serializers
+
+from seed.models.derived_columns import DerivedColumn, DerivedColumnParameter
+
+
+class DerivedColumnParameterSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = DerivedColumnParameter
+        fields = '__all__'
+
+
+class DerivedColumnSerializer(serializers.ModelSerializer):
+    parameters = serializers.SerializerMethodField()
+
+    class Meta:
+        model = DerivedColumn
+        exclude = ['source_columns']
+
+    def get_parameters(self, obj):
+        derived_column_parameters = DerivedColumnParameter.objects.filter(derived_column=obj.id)
+        return DerivedColumnParameterSerializer(derived_column_parameters, many=True).data
--- a/seed/static/seed/partials/accounts_nav.html
+++ b/seed/static/seed/partials/accounts_nav.html
@@ -8,3 +8,4 @@
      --><a ui-sref="organization_sub_orgs(::{organization_id: org.id})" ng-if="::org.is_parent && auth.requires_owner" ui-sref-active="active" translate>Sub-Organizations</a><!--
      --><a ui-sref="organization_members(::{organization_id: org.id})" ui-sref-active="active" translate>Members</a><!--
      --><a ui-sref="analyses(::{organization_id: org.id})" ui-sref-active="active" translate>Analyses</a>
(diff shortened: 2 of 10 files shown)


seed/static/seed/partials/derived_columns_admin.html
```diff
@@ -0,0 +1,80 @@
...
+                            </th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                        <tr ng-if="derived_columns.length > 0" ng-repeat="derived_column in derived_columns">
+                            <td>
+                                {$:: derived_column.name $}
+                            </td>
+                            <td>
+                                {$:: derived_column.expression $}
```
macintoshpie, line 59: I'm going to come back and make the display of expressions prettier later (time letting). Just displays the expression with variable names in it for now

seed/views/v3/derived_columns.py
```diff
@@ -0,0 +1,45 @@
...
+
+class DerivedColumnViewSet(viewsets.ViewSet, OrgMixin):
+    serializer_class = DerivedColumnSerializer
+    model = DerivedColumn
+
+    @require_organization_id_class
+    @api_endpoint_class
+    @ajax_request_class
+    @has_perm_class('requires_viewer')
+    def list(self, request):
```
adrian-lara, line 25: Lower priority note for now, or this can be updated later once more endpoints are built out. We should add code to get the Swagger docs updated.

seed/serializers/derived_columns.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class DerivedColumnSerializer(serializers.ModelSerializer):
+    parameters = serializers.SerializerMethodField()
+
+    class Meta:
+        model = DerivedColumn
+        exclude = ['source_columns']
+
+    def get_parameters(self, obj):
```
adrian-lara, line 25: Mostly just curious... Where is this used, or is this for something DRF specific?

adrian-lara (commented):
This looks good! One really small change about adding a link. Also, there's one optional change and a question.

seed/serializers/derived_columns.py
macintoshpie: This is DRF specific, `get_<field name>` allows you to provide a function for getting the serialized value for a field. I ended up refactoring this bit in the most recent derived columns PR #2698

seed/static/seed/partials/accounts_nav.html
```diff
@@ -8,3 +8,4 @@
      --><a ui-sref="organization_sub_orgs(::{organization_id: org.id})" ng-if="::org.is_parent && auth.requires_owner" ui-sref-active="active" translate>Sub-Organizations</a><!--
      --><a ui-sref="organization_members(::{organization_id: org.id})" ui-sref-active="active" translate>Members</a><!--
      --><a ui-sref="analyses(::{organization_id: org.id})" ui-sref-active="active" translate>Analyses</a>
+        <a ui-sref="organization_derived_columns(::{organization_id: org.id, inventory_type: 'properties'})" ui-sref-active="active" translate>Derived Columns</a>
```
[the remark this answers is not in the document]
macintoshpie: Ah nice, I missed this, I'll add that

adrian-lara (approved):
Thanks for making the change! This is good to go 👍🏽

Issue #2689, opened 2021-05-24
BuildingSync: default auc:Recommended to True


macintoshpie:
## background
SEED and Audit Template are working on interoperability

## task
From AT, not all auc:Measure have auc:Recommended and SEED currently defaults to false. SEED should default to true

Source: https://docs.google.com/spreadsheets/d/1yrvHA0X2rj6ckulvNiED9WadhiztnRv5Xf9BJtP-HHs

nllong:
@macintoshpie you want me to tackle this one?

nllong:
nvm... just created the PR.

PR #2697, opened 2021-05-24
Default null measure recommendation to True

nllong:
### Any background context you want to provide?
Audit Template tool does not fill out this field. By default, it should be set to true because a person has spent time to fill out the measure meaning that it is probably recommended.

#### What's this PR do?
Defaults PropertyMeasure.recommended to True.

* `./manage.py test`
* Import the buildingsync_v2_0_bricr_workflow.xml and verify that there is one explicit False recommended measure. All the others should be true (there are two non defined entries in the updated XML)

#2689


seed/models/building_file.py
```diff
@@ -180,7 +180,8 @@ def process(self, organization_id, cycle, property_view=None):
             implementation_status = m['implementation_status'] if m.get('implementation_status') else 'Proposed'
             application_scale = m['application_scale_of_application'] if m.get('application_scale_of_application') else PropertyMeasure.SCALE_ENTIRE_FACILITY
             category_affected = m['system_category_affected'] if m.get('system_category_affected') else PropertyMeasure.CATEGORY_OTHER
-            recommended = str(m.get('recommended', 'false')).lower() == 'true'
+            # for some reason this is returning none if the field is empty. So none and true should both be true.
+            recommended = str(m.get('recommended', 'true')).lower() in ['true', 'none']
```
macintoshpie, line 184: just for context, we turn bsync into models in ~two steps:
1. parse bsync into some structure
2. turn structure into db models

for (1) we use `None` to signify we failed to parse the value from BSync (ie the XPath mapping returns no nodes / content), deferring how to handle that to the db/model level (ie here)

seed/models/property_measures.py
```diff
@@ -135,7 +135,7 @@ class PropertyMeasure(models.Model):
     description = models.TextField(null=True)
     implementation_status = models.IntegerField(choices=IMPLEMENTATION_TYPES, default=MEASURE_PROPOSED)
     application_scale = models.IntegerField(choices=APPLICATION_SCALE_TYPES, default=SCALE_ENTIRE_FACILITY)
-    recommended = models.BooleanField(default=False)
+    recommended = models.BooleanField(default=True)
```
macintoshpie, line 138: nice 👍

macintoshpie (approved):
Tested the import with the buildingsync_v2_0_bricr_workflow.xml file and it worked as expected. Nick's gainin that coder cred 💻 😎

seed/models/building_file.py
nllong: gotcha!

nllong:
@macintoshpie -- when you merge this into your derived columns, it might be a good idea to reorder the migrations...

Issue #2536, opened 2021-01-11
Portfolio Manager import fails with TypeError


Ryoken:
While running a Portfolio Manager import (via logging into ESPM) the app crashes while generating the mapping suggestions.  See sentry report here: https://sentry.io/organizations/nrel/issues/2096132816/?project=1193349&query=&statsPeriod=90d#

**Expected Behavior**
Site should not crash while importing properties.

**Actual Behavior**
Site did crash.

**Steps to Reproduce**
We were unable to reproduce this bug in testing.

**Note**
PM may be not serving the data consistently and we need to do some extra logging to make sure that this is (or isn't) the case.

Ryoken:
Since we could not reproduce, here is what we do know:

**data_importer/views.py:1317** is creating **import_file**, which in several lines and without being modified is accessed successfully with if **import_file.from_portfolio_manager:**.  However, the error is triggered when passing in **import_file.first_row_columns**, meaning the query was successful and the import file exists.  **first_row_columns** is derived directly from the field **cached_first_row**, which means the field was blank.  The **cached_first_row** function that appears to be called is from **data_importer/tasks.py**, which means either the parser is messing up or the import file has missing information.

adrian-lara:
@Ryoken Just a heads up that Nick and I were chatting about this and we agree that the appropriate next step is to add some additional logging somewhere when we get unexpected data from ESPM. See the Note that he added to the main ticket text

adrian-lara:
@Ryoken Any reason why this is no longer relevant? I'm assuming this hasn't happened recently? If that's the case, I'd agree that we can close unless this pops up again.

Ryoken:
@adrian-lara Yeah, the original ticket's problem hasn't cropped up again.  I can still keep this ticket open to add the logging mentioned in the comments, however, or move it into a new ticket.  Preference?

adrian-lara:
Ah yeah, adding the logging would be great. Let's just keep this ticket, and I'll throw a p-2 label on it

Ryoken:
I spent a little more time trying to fix this bug and again could not reproduce it, so I boiled down my debug to a few statements I think will be able to identify the problem, should it crop up again.

RDmitchell:
@macintoshpie -- not sure what to do with this one. No "monkey testing" for me to do.

macintoshpie:
@RDmitchell no testing needed on your end -- we added additional logging to help debug this if it happens again later. Thanks!

PR #2699, opened 2021-05-25
fix: added some debug to track the first row columns

Ryoken:
See linked ticket for information about the import bug.

#### What's this PR do?
Adds in some debug statements to track the import and creation of "first_row_columns" in case the bug ever show up again.

none

https://github.com/SEED-platform/seed/issues/2536

none

### DIFF
--- a/seed/data_importer/models.py
+++ b/seed/data_importer/models.py
@@ -796,7 +796,9 @@ class ImportFile(NotDeletableModel, TimeStampedModel):
         if not hasattr(self, '_first_row_columns'):
             if self.cached_first_row:
                 self._first_row_columns = self.cached_first_row.split(ROW_DELIMITER)
+                _log.debug("Using cached first row columns.")
             else:
+                _log.debug("No first row columns property or cache was found!")
                 return None
         return self._first_row_columns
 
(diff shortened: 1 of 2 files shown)


macintoshpie:
As we discussed in Slack, Ryo is going to check if we also want to persist/log the PM response somewhere from Nick or Adrian since they saw the original Sentry error (it's now missing from Sentry). Once we get feedback I'll come back to this

Ryoken:
After we decided to try and find a way to get the PM response logged on a failure, I dug into portfolio_manager.py and started messing with the response to try and re-create the same error.  The section handling the response is already pretty well covered with try/except blocks, so any obvious malformed or incorrect response should already be caught.  I tried setting the properties list to an empty list and to None, but both times the site handled it with error messages:

Screen Shot 2021-05-28 at 12 32 12 PM
Screen Shot 2021-05-28 at 12 37 55 PM

macintoshpie (approved):
Nice, this looks good to me from what I know about the original issue. Tagging @adrian-lara to take a glance at it as well since I think he knows a bit more about the issue

adrian-lara (approved):
This looks good to me. We'll know for sure if there are any other areas to add logging if the issue ever comes up again.

Issue #2657, opened 2021-04-06
BETTER integration


adrian-lara:
The BETTER team is separately working on the backend changes needed in the BETTER application. 

In the meantime, SEED can likely take steps to prepare for that. These steps need to be scoped out before implementing.

adrian-lara:
Capturing/summarizing separate discussions:
SEED will provide BETTER with meters that have start and end date-times. BETTER is able to take those and standardized them into calendarized monthly readings on its own. SEED won't have to do any extra work from that aspect.

Also, the eventual list of supported meter types for the BETTER integration are as follows. We'll support a subset of these to start and build incrementally going forward:
```
    ELECTRIC_GRID = "Electric - Grid"
    ELECTRIC_SOLAR = "Electric - Solar"
    ELECTRIC_WIND = "Electric - Wind"
    NATURAL_GAS = "Natural Gas"
    DIESEL = "Diesel"
    PROPANE = "Propane"
    COAL_ANTHRACITE = "Coal (anthracite)"
    COAL_BITUMINOUS = "Coal (bituminous)"
    COKE = "Coke"
    FUEL_OIL_1 = "Fuel Oil (No. 1)"
    FUEL_OIL_2 = "Fuel Oil (No. 2)"
    FUEL_OIL_4 = "Fuel Oil (No. 4)"
    FUEL_OIL_5_AND_6 = "Fuel Oil (No. 5 and No. 6)"
    DISTRICT_STEAM = "District Steam"
    DISTRICT_HOT_WATER = "District Hot Water"
    DISTRICT_CHILLED_WATER_ELECTRIC = "District Chilled Water - Electric"
    DISTRICT_CHILLED_WATER_ABSORPTION = "District Chilled Water - Absorption"
    DISTRICT_CHILLED_WATER_ENGINE = "District Chilled Water - Engine"
    DISTRICT_CHILLED_WATER_OTHER = "District Chilled Water - Other"
    KEROSENE = "Kerosene"
    WOOD = "Wood"
```

FYI @RDmitchell  @yroussev

yroussev:
@adrian-lara  thank you for the update.  Once the link is operational we can test @RDmitchell with some actual data for buildings that have  ELECTRIC_GRID = "Electric - Grid" and  NATURAL_GAS = "Natural Gas" as the energy sources.

RDmitchell:
[...]

PR #2700, opened 2021-05-27
Feat/better integration

aviveiros11:
Initial effort to connect SEED and BETTER via Bsync and Analysis framework.  This PR only tackles the first half of the deliverable:
 - Front end for user inputs for BETTER 
 - property with > 12 months Electricity and Natural Gas meter data converted into Bsync File
 - Analysis creation with the proper payload to BETTER

What this PR does not yet include:
- Actual requests to the BETTER API for building creation and subsequent analysis creation

Because this is likely a large PR I wanted to make sure I was going in the right direction preemptively and resolve any issues as I go.  

#### What's this PR do?
Lays the groundwork for eventual integration with the BETTER API

1. Upload demo properties from the attached file similar to the `bsyncr` demo
2. Match the properties on `PM Property ID' 
3. Upload the PM meter data file also attached
Make sure the 12 months of electricity and natural gas meters have been linked to `123 Location and meters lane`
4. Edit the property to include the `City`, `State`, `Groos Floor Area`, and `Postal Code` 
5. Run the BETTER analysis with arbitrary values (should fail with a message saying it is not connected to BETTER)
Additionally - if you would like to view the BuildingSync document produced from the pipeline you can do so in `../seed/analysis_pipelines/better.py/`. 
#2657 

#### Supporting Files:
[4 Properties - lat:long and address.xlsx](https://github.com/SEED-platform/seed/files/6555715/4.Properties.-.lat.long.and.address.xlsx)
[PM.Meter.Data.BETTER.input (1).xlsx](https://github.com/SEED-platform/seed/files/6650528/PM.Meter.Data.BETTER.input.1.xlsx)

## UPDATE (06/08)
1. BETTER API calls have been added and analysis results are now served as an HTML file.  The endpoint still has a `@login required` constraint however I have requested they switch this to token auth similar to the other endpoints.    

## UPDATE (06/14)
1. Should now work as expected with the test data


seed/analysis_pipelines/better.py
```diff
@@ -0,0 +1,448 @@
...
+        errors.append('Linked PropertyState is missing a gross floor area')
+    for meter in meters:
+        for meter_reading in meter.meter_readings.all():
+            if meter_reading.reading is None:
+                errors.append(f'{meter}: MeterReading starting at {meter_reading.start_time} has no reading value')
+    if errors:
+        return None, errors
+
+    # clean inputs
+    # BETTER will default if eGRIDRegion is not explicitly set
```
macintoshpie, line 228: Does this mean better will set a default eGRIDRegion value if not set? Or BETTER will fail?

seed/analysis_pipelines/better.py
```diff
@@ -0,0 +1,448 @@
...
+                                    E.ScenarioType(
+                                        E.CurrentBuilding(
+                                            E.CalculationMethod(
+                                                E.Measured()
+                                            )
+                                        )
+                                    ),
+                                    E.ResourceUses(
+                                        E.ResourceUse(
+                                            {'ID': 'Resource-' + str(11)},
```
macintoshpie, line 304: I'm guessing these IDs will be dehardcoded later?

seed/analysis_pipelines/better.py
```diff
@@ -0,0 +1,448 @@
...
+                                    ),
+                                    E.PremisesIdentifiers(
+                                        E.PremisesIdentifier(
+                                            E.IdentifierLabel('Custom'),
+                                            E.IdentifierCustomName(PREMISES_ID_NAME),
+                                            E.IdentifierValue(str(analysis_property_view.id)),
+                                        )
+                                    ),
+                                    E.Longitude(str(analysis_property_view.property_state.longitude)),
+                                    E.Latitude(str(analysis_property_view.property_state.latitude)),
```
macintoshpie, line 284: should this be replaced with the city, state, zip?

macintoshpie (commented):
Some initial comments, but this is looking good! I'll do a more full review early next week

seed/analysis_pipelines/better.py
aviveiros11: Yeah BETTER has a default region it assigns if none are provided so it will not fail.

seed/analysis_pipelines/better.py
aviveiros11: That is the intent.  Ideally, this ID would be populated by the meter type dynamically but I had trouble finding a solution that would filter back up to `ResourceUses` outside the context of the nested `for`.  Definitely open to suggestions or a quick pair to talk about strategy.

seed/analysis_pipelines/better.py
aviveiros11: geo lat/long are optional BETTER inputs so maybe I could throw in a check to use the values if the property has them and exclude them otherwise.

macintoshpie: ah gotcha, this seems good then 👍

macintoshpie (changes requested):
Looks good! I uploaded the files and tested the happy and sad paths and everything worked as expected. I think there's just one thing that I'd like to be tweaked (tasks.py)

## happy path
Screen Shot 2021-06-01 at 9 32 04 AM
Screen Shot 2021-06-01 at 9 32 15 AM

## property missing info
Screen Shot 2021-06-01 at 9 33 09 AM

I had to import the tasks to `analysis_pipeline/tasks.py` for this to work, which probably should get added to this PR. I'm not sure how it could work otherwise (unless you're not using seed/celery.py for settings?).

Last comment is that the BuildingSync file generated has a few small issues, which we can address now or later.
1. eGRIDRegion seems to be in the wrong place and we are dropping whatever state the user provides into the file (BSync wants short vesion)
Screen Shot 2021-06-01 at 9 40 45 AM
2. There can be duplicate TimeSeries ID's based on resource type (e.g. here an electricity and natural gas reading have the same ID)
Screen Shot 2021-06-01 at 9 40 28 AM
3. Natural gas in ResourceUse should have lowercase "g"
Screen Shot 2021-06-01 at 9 40 55 AM

yroussev:
@aviveiros11 BETTER userland comment, something to think about... currently the tool prefers to work with Natural-Gas and Grid Electricity, but wanted to mention that the tool should be able to work with other metered types of thermal energy that ESPM hosts (for example district heating or cooling). ESPM is currently limited to monthly normalized custom reporting to only natural gas and electricity, but that will likely expand in the future to other types of energy or it can be done in SEED.  If the data is coming from the raw meter data, that SEED can help with this to push monthly data to BETTER, and categorize it accordingly if not electricity or natural gas.

aviveiros11:
> @aviveiros11 BETTER userland comment, something to think about... currently the tool prefers to work with Natural-Gas and Grid Electricity, but wanted to mention that the tool should be able to work with other metered types of thermal energy that ESPM hosts (for example district heating or cooling). ESPM is currently limited to monthly normalized custom reporting to only natural gas and electricity, but that will likely expand in the future to other types of energy or it can be done in SEED. Since BETTER currently works with data coming from the raw ESPM meter data "Meter Entries", maybe SEED can help with normalizing to calendar month data prior to pushing to BETTER, and categorize it accordingly if not electricity or natural gas to other types of energy. The following 7 fossil fuel sources that BETTER works with for carbon reporting, so it is expected that data would be analyzed by the tool:
> 
> Coal (anthracite)
> Coal (bituminous)
> Coke
> Diesel
> Propane
> Wood
> District Steam
> District Hot Water
> District Chilled Water - Electric
> District Chilled Water - Absorption
> District Chilled Water - Engine
> District Chilled Water - Other
> Fuel Oil (No.1)
> Fuel Oil (No. 2)
> Fuel Oil (No. 4)
> Fuel Oil (No. 5 and No. 6)
> Kerosene
> Natural gas

Hey, @yroussev definitely!   Our intention here is to prove the basic functionality of the SEED <-> BETTER connection with a simple subset of fuel types and build the list incrementally to accommodate all of them you listed.

macintoshpie (commented):
Looks like the order of the building elements is a bit off, here's the fixed ordering:
Screen Shot 2021-06-02 at 8 38 13 AM

Also, could you update seed/analysis_pipelines/tasks.py to include you tasks?
```python
from seed.analysis_pipelines.better import (  # noqa: F401
    _prepare_all_properties,
    _finish_preparation,
    _start_analysis,
    _finish_analysis,
)
```

macintoshpie (approved):
Schweet! This all looks good to me 🐩

yroussev:
> > @aviveiros11 BETTER userland comment, something to think about... currently the tool prefers to work with Natural-Gas and Grid Electricity, but wanted to mention that the tool should be able to work with other metered types of thermal energy that ESPM hosts (for example district heating or cooling). ESPM is currently limited to monthly normalized custom reporting to only natural gas and electricity, but that will likely expand in the future to other types of energy or it can be done in SEED. Since BETTER currently works with data coming from the raw ESPM meter data "Meter Entries", maybe SEED can help with normalizing to calendar month data prior to pushing to BETTER, and categorize it accordingly if not electricity or natural gas to other types of energy. The following 7 fossil fuel sources that BETTER works with for carbon reporting, so it is expected that data would be analyzed by the tool:
> > Coal (anthracite)
> > Coal (bituminous)
> > Coke
> > Diesel
> > Propane
> > Wood
> > District Steam
> > District Hot Water
> > District Chilled Water - Electric
> > District Chilled Water - Absorption
> > District Chilled Water - Engine
> > District Chilled Water - Other
> > Fuel Oil (No.1)
> > Fuel Oil (No. 2)
> > Fuel Oil (No. 4)
> > Fuel Oil (No. 5 and No. 6)
> > Kerosene
> > Natural gas
> 
> Hey, @yroussev definitely! Our intention here is to prove the basic functionality of the SEED <-> BETTER connection with a simple subset of fuel types and build the list incrementally to accommodate all of them you listed.

Thanks for confirming @aviveiros11 , it is all very exciting work.  Thank you!

macintoshpie (commented):
Looking good, tried to run this locally, and added the BETTER_TOKEN to local untracked, but got this error in the console when I tried to start the analysis:

```
seed_celery    | [2021-06-10 19:46:32,933: ERROR/ForkPoolWorker-1] {"analysis_id": 19, "analysis_property_view": null, "user_message": "Unexpected error occurred.", "debug_message": "Caught unexpected exception occurred during analysis task and the analysis still exists. Failing the analysis and re-raising the exception.", "exception": "AnalysisPipelineException('BETTER service could not create building',)"}
```

I might have this misconfigured though, not too sure. Might need some more logging around here

seed/analysis_pipelines/better.py
```diff
@@ -435,7 +552,7 @@ def _run_better_analysis(building_id, config):
     except Exception as e:
         return None, [f'Failed to make request to better server: {e}']
 
-    if response.status_code != 200:
+    if response.status_code != 201:
         try:
             response_body = response.json()
             flattened_errors = [error['detail'] for error in response_body['errors']]
```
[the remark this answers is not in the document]
aviveiros11: oop - yep - with extra marinara 🥟

seed/models/analysis_output_files.py
```diff
@@ -14,10 +14,12 @@ class AnalysisOutputFile(models.Model):
     The AnalysisOutputFile is a file returned as output from an analysis.
     """
     BUILDINGSYNC = 1
+    HTML = 2
```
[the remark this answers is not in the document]
aviveiros11: we do 👍

macintoshpie:
@aviveiros11 going to hold off on another review until the next set of updates if that sounds good to you

seed/analysis_pipelines/better.py
```diff
@@ -221,6 +223,8 @@ def _build_better_input(analysis_property_view, meters):
         errors.append('Linked PropertyState is missing a State')
     if property_state.gross_floor_area is None:
         errors.append('Linked PropertyState is missing a gross floor area')
+    if property_state.property_type is None:
+        errors.append('Linked PropertyState is missing a property type')
```
macintoshpie, line 227: Documenting our discussion, we should validate the user has provided one of the required property_types, otherwise BETTER 500s

seed/analysis_pipelines/better.py
```diff
@@ -445,38 +448,21 @@ def _better_building_service_request(bsync_xml):
...
 
     headers = {
-        'accept': 'application/json',
         'Authorization': settings.BETTER_TOKEN,
-        'Content-Type': 'application/json',
+        'Content-Type': 'buildingsync/xml',
     }
     try:
-        response = requests.request("POST", url, headers=headers, data=json.dumps(bsync_xml))
+        response = requests.request("POST", url, headers=headers, data=bsync_content)
```
macintoshpie, line 461: Documenting our discussion, it might be nice to check `response.status_code` is 2xx before trying to parse the response

seed/analysis_pipelines/better.py
```diff
@@ -543,24 +526,33 @@ def _run_better_analysis(building_id, config):
...
+
+    # Gotta make sure the analysis is done
+    url = "{host}/api/v1/buildings/{id}/analytics/".format(host=HOST, id=building_id)
+
+    headers = {
+        'accept': 'application/json',
+        'Authorization': settings.BETTER_TOKEN,
+    }
+    try:
+        response = polling.poll(
```
macintoshpie, line 548: noice

macintoshpie (commented):
Tested locally and everything worked properly after:
- uploaded the provided spreadsheets
- updated the "correct" property to have City, State, Postal Code, Gross floor area, and Property type

Output here: https://gist.github.com/macintoshpie/378f0952d4b35df5964cf47144793c3f

Caveats:
- change point model for fossil fuels didn't work (from my understanding this is a BETTER issue however)
- there's additional validation of SEED inputs we should do at a later time (or force BETTER to validate - see next bullet)
- update the [BETTER schematron](https://github.com/BuildingSync/use-cases/tree/main/BETTER) and use it in the BETTER app

I think once we add validation for property type (see code comment) then this is ready for merging

seed/analysis_pipelines/better.py
```diff
@@ -459,8 +479,11 @@ def _better_building_service_request(bsync_xml):
     }
     try:
         response = requests.request("POST", url, headers=headers, data=bsync_content)
-        data = response.json()
-        building_id = data['id']
+        if response.status_code == 201:
+            data = response.json()
+            building_id = data['id']
+        else:
+            raise
```
macintoshpie, line 486: Will this work? Since we aren't in an `except` I'm not sure this will work as is
```suggestion
            raise Exception(f'Received non 2xx status from BETTER: {response.status_code}: {response.content}')
```

seed/analysis_pipelines/better.py
```diff
@@ -225,6 +243,8 @@ def _build_better_input(analysis_property_view, meters):
         errors.append('Linked PropertyState is missing a gross floor area')
     if property_state.property_type is None:
         errors.append('Linked PropertyState is missing a property type')
+    if property_state.property_type not in BETTER_PROPERTY_TYPES:
```
macintoshpie, line 246: 😎

macintoshpie (approved):
One last thing, but I think that should do it if you think the suggested change looks good

macintoshpie (approved):
\* chef's kiss *

PR #2706, opened 2021-06-04
Feat/Email post office

ebeers-png:
This is a feature by ClearlyEnergy, that we're moving to SEED from our tool BEAM.

#### What's this PR do?
Adds a post office to SEED. Users can create email templates in the organization settings, and then send emails en-masse to building owners. The email addresses are pulled from their owner_email field.

Email templates can include inventory fields using {{ field_name }}. For example, in a template you can write "Dear {{ first_name }}," and when sent en-masse the emails will be autofilled in from each building's information: "Dear John," "Dear Jane," et cetera.

The post office is fairly bare-bones, and there are more features I'd like to add in the future, such as a more detailed modal window in the inventory list, and a "preview email" button in the template page.

Templates can be created, edited, and deleted in the "Emails" organization tab. On the inventory list, we added an item for Emails to the "Actions" dropdown. You can select buildings in the list, navigate to Emails, and then select the template to send in the modal window. Emails will be created and added to a queue.

The post office may have to be configured further, to (for example) send emails on a schedule. See the [django-post_office page](https://github.com/ui/django-post_office) for more information.

The commits in this PR were cherry-picked from a long string of commits, so there _is_ a chance there are some code edits missing.

I wrote some quick Python tests in test_postoffice and test_views, but more unit testing should be done.

Templates page:
image


ebeers-png:
Made some edits addressing flake8 failures and migrations, all of which should be good to go now.

macintoshpie:
Hi @ebeers-png , thanks for the PR! Could you please update the functional tests so that they pass? It looks like there might be a new angular module, nWig, and it's not included in the tests: https://github.com/SEED-platform/seed/pull/2706/checks?check_run_id=2795309235#step:11:1851

Thanks!

ebeers-png:
Sure thing @macintoshpie, I'll take a look!

macintoshpie:
@ebeers-png Hi! Just wanted to check in to see if you've had the chance to take a look at the tests?
I can probably help out if needed, thanks!

ebeers-png:
@macintoshpie Hi, sorry about that, I was able to take a look today! It looks like we added ngWig to the list of scripts in base.html but not to AngularJSTests.html. I can add them in today, and see if there are any other tests we need to address after that.

seed/views/v3/postoffice.py
```diff
@@ -0,0 +1,70 @@
...
+        return PostOfficeEmail.objects.all()
+
+    def perform_create(self, serializer):
+        template_id = self.request.data.get('template_id')
+        inventory_id = self.request.data.get('inventory_id', [])
+        if self.request.data.get('inventory_type') == "properties":
+            state = PropertyState
+        else:
+            state = TaxLotState
+        properties = state.objects.filter(id__in=inventory_id)
```
macintoshpie, line 51: I think we might need add org id as a filter param here. As is, I think a user could pass IDs of states that aren't part of their org.

This wouldn't necessarily leak leak data (the email is still going to the property owner), but should probably be avoided.

seed/tests/test_views.py
```diff
@@ -1463,6 +1463,49 @@ def test_get_cycles(self):
         self.assertEqual(cycle['id'], self.cycle.pk)
         self.assertEqual(cycle['name'], self.cycle.name)
 
+    def test_postoffice(self):
```
macintoshpie, line 1466: Nice! Thanks for adding these!

macintoshpie (changes requested):
Look great! Thanks Elizabeth!

Just one comment on the code that I think we should address before merging (using org id in inventory query when sending emails). Once this is updated I think this PR is good to go

## misc notes
I was able to create / edit / rename / delete templates and select them for different properties from the inventory list view. I was able to use property state fields as values in the email templates, though I wasn't sure how to use extra data keys.

Down the road we might want to consider limiting which fields are accessible (maybe just pass the *State as a dictionary?). It makes me a bit nervous passing the model instance itself as context, I'm not sure if/what the implications could be to allow users to define what's in a template.

In the future it would also be nice to add documentation about what the column names are (maybe show a collapsable table above the WSIWYG editor?) -- I'm guessing most folks won't know what they are off the top of their heads.

A last concern is performance -- we've been recently been seeing long query times on the PropertyState table in production, but we should wait and see if this turns out to actually be an issue.

macintoshpie:
I'm going to go ahead and merge this into the temporary branch off of `develop` and tweak some things myself. Thanks again @ebeers-png !

Issue #2710, opened 2021-06-09
Derived Columns: users can view derived column results in inventory list view


macintoshpie:
Users should be able to create column profiles (list and detail) which contain the derived column. They should then be able to view (but not edit) derived column data in these respective views.

Also see parent issue #1642

RDmitchell:
@macintoshpie -- I can add a derived column to a Column List Profile, at the end of the list, but when I try to move it to somewhere else one the list, and save the profile, the Inventory list doesn't show the new position of the field, and when I go back to edit the profile, the field is again at the end of the list. 

Steps to reproduce:

* Create a derived column
* Add it to the Property Inventory List Profile
* It is at the end of the list
![image](https://user-images.githubusercontent.com/6314950/125539245-b44e173b-e79e-4924-ab98-5e8baf8f8b0c.png)
* Move it to the top of the list
![image](https://user-images.githubusercontent.com/6314950/125539333-ab6699bb-5520-4452-bf12-055debf83df1.png)
* Save the profile
![image](https://user-images.githubusercontent.com/6314950/125539379-c5e6cda5-eaae-40b5-b4df-d71bdfe99dcc.png)
* View the Inventory list -- the field is not in the new location, it is at the end of the fields
![image](https://user-images.githubusercontent.com/6314950/125539443-fc951618-d462-4e70-9f3b-0f3e9cae6f38.png)
![image](https://user-images.githubusercontent.com/6314950/125539526-d010d493-54f4-4721-8740-7eaa8cdc6f97.png)
* Go back to the profile -- the field is again at the end of the list
*

macintoshpie:
@RDmitchell you cant order derived columns currently, they are always placed at the end. Ideally we could prevent users from moving them around in the column list profile editor, but it was too much effort. So instead we have a note at the top of the page:

![Screen Shot 2021-07-14 at 2 55 56 PM](https://user-images.githubusercontent.com/18518728/125691738-79dee7b8-895a-41a4-857a-9473eebf4606.png)

RDmitchell:
@macintoshpie -- right I just figured that out today. I will close this issue !!

PR #2714, opened 2021-06-11
Feat/derived column inventory list

macintoshpie:
SEED is adding the ability for users to define columns which are evaluated based on other columns.

#### What's this PR do?
1. shows derived column values in the inventory list and inventory detail page
2. allows derived columns to be added to column list and detail profiles

1. create some derived columns from the organization section.
2. create/update column profile to include the derived columns
3. update/confirm one or more properties has all data required for these derived columns (so the value will not be null after evaluating)
4. view the inventory list and inventory detail page to see the results. You should not be able to edit derived column values.

#2710 

Screen Shot 2021-06-11 at 5 25 19 PM
Screen Shot 2021-06-11 at 5 24 55 PM
Screen Shot 2021-06-11 at 5 24 20 PM
Screen Shot 2021-06-11 at 5 24 01 PM


seed/static/seed/js/controllers/inventory_detail_controller.js
```diff
@@ -164,61 +229,6 @@ angular.module('BE.seed.controller.inventory_detail', [])
...
-      if ($scope.currentProfile) {
-        $scope.columns = [];
-        _.forEach($scope.currentProfile.columns, function (col) {
-          var foundCol = _.find(columns, {id: col.id});
-          if (foundCol) $scope.columns.push(historical_changes_check(foundCol));
-        });
-      } else {
-        // No profiles exist
-        $scope.columns = _.reject(columns, 'is_extra_data');
-      }
```
macintoshpie, line 220: this section of code was duplicated originally, which is why I cut/paste it above.

seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -721,6 +737,50 @@ angular.module('BE.seed.controller.inventory_list', [])
         });
       };
 
+      // evaluate all derived columns and add the results to the table
+      var evaluateDerivedColumns = function () {
```
macintoshpie, line 741: this could get out of hand for the inventory list... here we "batch" inventory IDs, then for every derived column evaluate that batch at once. Problem is that we aren't limiting concurrent requests here, we just fire them all off, not sure if that will cause problems down the line.

adrian-lara: I am following this approach and it does seem to work as expected. Maybe we just see how this reacts when stress tested? I suspect that it could definitely take some time to see the derived column values on inventory page load. 

I'm wondering the idea of sending each batch request independently and giving each request a resolve that updates the rows. I could see that possibly going badly but wanted to throw the idea out there.

Either way, this works and it's probably worth testing before changing. Just wanted to capture some thoughts.

seed/views/v3/derived_columns.py
```diff
@@ -172,3 +176,75 @@ def destroy(self, request, pk):
...
+            openapi.Parameter(
+                'inventory_ids',
+                openapi.IN_QUERY,
+                description='List of inventory IDs (i.e. Property or TaxLot)',
+                required=True,
+                type=openapi.TYPE_ARRAY,
+                items=openapi.Items(
+                    type=openapi.TYPE_INTEGER
+                )
+            )
```
adrian-lara, line 197: I don't want to expand scope and context switch too much, but as a "quickfix", what do you think about leaving a code comment here to revisit this and standardizing this via AutoSchemaHelper?

adrian-lara (commented):
Largely works as expected! A few comments but mostly capturing thoughts and considerations. We should get this up to dev1 sooner than later for testing.

However, there is one small bug that I suspect came from this PR: on the inventory list and detail pages, I can't seem to create a column list profile based on populated columns when it's my first profile created.

Screen Shot 2021-06-14 at 1 33 00 PM

seed/models/derived_columns.py
```diff
@@ -30,6 +32,9 @@ def _cast_params_to_floats(params):
         if isinstance(value, bool):
             continue
 
+        if isinstance(value, ureg.Quantity):
+            value = value.magnitude
```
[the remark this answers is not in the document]
macintoshpie: yep, I will make a note to add this to the create/edit page for derived columns

seed/static/seed/js/controllers/inventory_list_controller.js
macintoshpie: 👍  yeah let's stress test it first
I would like to use async/await for batches but couldn't figure out how to make it work with some of the angular stuff

seed/views/v3/derived_columns.py
macintoshpie: I'm not too worried about this yet, this is the only place so far we've had to create this custom swagger parameter, maybe once it pops up some more we can push it into the AutoSchemaHelper

adrian-lara (approved):
Thanks for fixing the bug as well as the broken tests!

Things work as expected 👍🏽

seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -949,7 +1015,7 @@ angular.module('BE.seed.controller.inventory_list', [])
       function currentColumns () {
         // Save all columns except first 3
         var gridCols = _.filter($scope.gridApi.grid.columns, function (col) {
-          return !_.includes(['treeBaseRowHeaderCol', 'selectionRowHeaderCol', 'notes_count', 'merged_indicator', 'id'], col.name) && col.visible;
+          return !_.includes(['treeBaseRowHeaderCol', 'selectionRowHeaderCol', 'notes_count', 'merged_indicator', 'id'], col.name) && col.visible && col.id;
```
macintoshpie, line 1018: this turned out to be an existing bug, fixed here (screenshot from develop branch)
Screen Shot 2021-06-15 at 9 58 06 AM

Issue #2711, opened 2021-06-09
Derived Columns: users can create data quality checks based on derived columns


macintoshpie:
Users should be able to create data quality checks based on derived column values. Note that the current implementation for derived columns assumes they're always numeric.

Also see parent issue #1642

RDmitchell:
Instance: dev1
SHA: 
Org: LBNL 200
Cycle: Derived Column Settings

This appears to be working (although not exhaustively tested)

* Defined a Data Quality rule to add a label if the value of the derived field "Percent Natural Gas Use" was > 50
![image](https://user-images.githubusercontent.com/6314950/126411443-1b580734-ba34-4748-9257-cc2186ecd576.png)
* I then ran the DQ rules on the Property Inventory List and 4 records were returned
![image](https://user-images.githubusercontent.com/6314950/126411503-87ea8539-170f-4c73-9343-f8e00067af16.png)
* When doing a filter for Natural Gas Use > 50, the same 4 records are displayed
![image](https://user-images.githubusercontent.com/6314950/126411529-58ed6b68-82c1-4acd-bfcc-f20685306560.png)
* There is a bug in displaying the labels in the list view, but going to the detail view for the appropriate records, the label is there
![image](https://user-images.githubusercontent.com/6314950/126411585-7f185889-6faa-4aad-8413-42f1cb06f394.png)

PR #2716, opened 2021-06-15
feat: allow usage of derived columns in data quality checks

macintoshpie:
SEED is adding a feature for users to define columns calculated based on other column values (derived columns)

#### What's this PR do?
Adds the ability for users to use derived columns in data quality rules.
Technically:
- adds `for_derived_column` field to the `Rule` model. I considered just using `table_name` to indicate the rule was for a derived column, but it turned out to overcomplicate things.

1. create a derived column, make sure an inventory has data for the necessary columns so the calculation can be made
2. create a dq rule which references that column
3. run a data quality check from the inventory list page

Also verify creating/editing/deleting dq rules still works for non-derived columns

#2711 

Screen Shot 2021-06-15 at 5 47 48 PM
Screen Shot 2021-06-15 at 5 48 30 PM


seed/models/data_quality.py
```diff
@@ -697,24 +716,28 @@ def _check(self, rules, row):
...
-                    continue
-
-                try:
-                    value = rule.str_to_data_type(value)
-                except DataQualityTypeCastError:
-                    self.add_result_type_error(row.id, rule, display_name, value)
-                    continue
+            if rule.for_derived_column:
+                derived_column = derived_columns_by_name[rule.field]
+                value = derived_column.evaluate(inventory_state=row)
```
adrian-lara, line 721: Good to see it was just a matter of adding the values of derived columns to each "record", and it doesn't seem necessary to update the code/logic that inspects the values.

adrian-lara (approved):
Code changes look good! 

Derived columns seem to be well integrated into DQ checks. 

I like how the Derived column names are grouped in the DQ page dropdown
Screen Shot 2021-06-21 at 12 21 01 PM

As expected the DQ checks work within both the import process (after column mapping) as well as within the inventory page.

Issue #2717, opened 2021-06-16
Derived Columns: users can export derived column results


macintoshpie:
From the inventory list page, users should be able to select properties, and if their current column profile includes derived columns then that column and the derived values should be included.

RDmitchell:
@macintoshpie -- In testing this, the export of the derived fields works, but it ONLY exports the derived fields, not any of the other fields. 

Instance: dev1
SHA: 9d9368cbf
Org: LBNL 200
Cycle: Derived Column Testing

Steps to reproduce: 
* There are 4 derived fields defined
![image](https://user-images.githubusercontent.com/6314950/126407919-e7b1ad8f-de16-4d3b-b446-c7535b3fa448.png)
* View the Inventory List View -- the Column List Profiles shows that the derived fields are included in the list
![image](https://user-images.githubusercontent.com/6314950/126407975-622b397b-c502-4862-b197-e1e9d168c304.png)
* In the Inventory List view the derived columns are displayed, as well as all the other non-empty fields; the checkboxes indicate that all the records are selected for export
![image](https://user-images.githubusercontent.com/6314950/126408267-e6c777d6-b33a-446b-baa8-ccea21e49f50.png)
* Using the Actions menu, select Export Selected, and this dialog box appears; Click the Export CSV button
![image](https://user-images.githubusercontent.com/6314950/126408481-5effe0be-af09-44a1-aa9d-46f7e64b3e55.png)
* the program exports a file with all the records, but with ONLY the derived columns, none of the other fields.
![image](https://user-images.githubusercontent.com/6314950/126408421-d61c0bc3-78ff-47d7-adc8-fb1b2a541e25.png)

macintoshpie:
@RDmitchell I wasn't able to replicate this. Could you please test this again? I wonder if maybe your column profile had been messed up before you clicked export (see #2749 )

SHA: d5beacab829c4580b93315aed4801dd1a6f912b7

RDmitchell:
@macintoshpie -- you are correct. I can't replicate this.

SHA: 24b503d2d

PR #2718, opened 2021-06-16
feat: add derived columns to exported results

macintoshpie:
SEED is adding the feature for users to define columns calculated from other columns (derived columns)

#### What's this PR do?
Adds derived columns and their values to exported sheets (CSV, excel) for derived columns part of the column profile they're using.

1. create derived column, have property that includes required columns
2. create a column list profile which includes the derived column
3. select properties in list view, then choose export
4. you should see the derived column and its data in the resulting file

#2717 

Screen Shot 2021-06-15 at 6 58 04 PM


seed/views/v3/tax_lot_properties.py
```diff
@@ -112,25 +111,23 @@ def export(self, request):
...
             column_name_mappings['property_notes'] = 'Property Notes'
             column_name_mappings['property_labels'] = 'Property Labels'
 
         elif hasattr(view_klass, 'taxlot'):
             select_related.append('taxlot')
-            prefetch_related = ['labels']
-            filter_str = {'taxlot__organization_id': org_id}
-            if ids:
-                filter_str['id__in'] = ids
+            filter_str['taxlot__organization_id'] = org_id
```
macintoshpie, line 130: a little code cleanup. Note that originally we were dropping the `cycle` filter param every time (we were always resetting `filter_str`), this fixes that as well.

seed/views/v3/tax_lot_properties.py
```diff
@@ -88,18 +87,12 @@ def export(self, request):
         """
         Download a collection of the TaxLot and Properties in multiple formats.
         """
-        cycle_pk = request.query_params.get('cycle_id', None)
-        if not cycle_pk:
-            return JsonResponse(
-                {'status': 'error', 'message': 'Must pass in cycle_id as query parameter'})
```
macintoshpie, line 94: removing cycle_pk because the code does not actually use it for filtering.

adrian-lara (approved):
Code changes look good and derived col values look to be added on all export types. 👍🏽

Issue #2691, opened 2021-05-24
BuildingSync: parse units for meter data


macintoshpie:
## background
SEED and Audit Template are improving interoperability. SEED does not currently check units when reading in meter data.

## task
SEED should read units for meter data and store the data accordingly.

Source: https://docs.google.com/spreadsheets/d/1yrvHA0X2rj6ckulvNiED9WadhiztnRv5Xf9BJtP-HHs

PR #2719, opened 2021-06-16
2691 feat/parse meter units

Ryoken:
See ticket.

#### What's this PR do?
Adds conversion factor to imported building sync meter readings.

Upload a building sync file with meter readings, then examine the database and see conversion_factor has been added.

https://github.com/SEED-platform/seed/issues/2691

A building sync file with a meter reading:
Screen Shot 2021-06-16 at 4 42 31 PM
The meter reading in the database with a conversion factor:
Screen Shot 2021-06-16 at 4 41 55 PM
The conversion factor list showing the correct value has been loaded:
Screen Shot 2021-06-16 at 4 49 21 PM


seed/data_importer/utils.py
```diff
@@ -78,6 +78,7 @@ def kbtu_thermal_conversion_factors(country):
 
     if country == "US":
         factors['Coal (anthracite)']['GJ'] = 947.82
+        factors['Coal (anthracite)']['Btu'] = .001
```
nllong, line 81: preceding zero throughout

nllong (changes requested):
Good job detangling this effort and adding the mappings! I made comments on several of the mappings and am really wondering if we need to add more `METER_TYPES`. I wonder how that will impact SEED in general. Thoughts?

Few requested changes...

💥🤘

seed/building_sync/mappings.py
```diff
@@ -207,18 +207,106 @@ def to_energy_type(energy_type):
...
+    return {
+        'Cubic Meters': 'cm (cubic meters)',
+        'kcf': 'kcf (thousand cubic feet)',
+        'MCF': 'Mcf (million cubic feet)',
+        'Gallons': 'Gallons (US)',              # assuming "US" for conversion_factor but could be "UK"
+        'Wh': 'Wh (Watt-hours)',
+        'kWh': 'kWh (thousand Watt-hours)',
+        'MWh': 'MWh (million Watt-hours)',
+        'Btu': 'Btu',
+        'kBtu': 'kBtu (thousand Btu)',
```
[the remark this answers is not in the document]
macintoshpie: I think we have to do singular here -- we're actually trying to map to these labels here (which actually come from PM):
https://github.com/SEED-platform/seed/blob/ceff4089a4b6cc1ebca3bef16a2cae36838d2e5c/seed/data_importer/utils.py#L81

seed/building_sync/mappings.py
```diff
@@ -207,18 +207,106 @@ def to_energy_type(energy_type):
...
+        'MMBtu': 'MBtu/MMBtu (million Btu)',
```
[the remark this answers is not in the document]
macintoshpie: Maybe not plural, same comment as here: https://github.com/SEED-platform/seed/pull/2719#discussion_r653646888

seed/building_sync/mappings.py
```diff
@@ -207,18 +207,106 @@ def to_energy_type(energy_type):
...
+        'therms': 'therms',
+        'lbs': 'Lbs. (pounds)',
+        'Klbs': 'kLbs. (thousand pounds)',
+        'Mlbs': 'MLbs. (million pounds)',
+        'Mass ton': 'Tons',                     # assuming "Tons" over "Tonnes (metric)"
```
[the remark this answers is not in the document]
macintoshpie: Hmm, if the data is always coming from BuildingSync though, is it ok to assume it's always going to imperial/US units?

seed/building_sync/mappings.py
```diff
@@ -207,18 +207,106 @@ def to_energy_type(energy_type):
     """converts an energy type from BuildingSync into one allowed by SEED
 
     :param energy_type: string, building sync energy type
-    :return: string
+    :return: int
```
macintoshpie, line 210: 🙏

seed/models/building_file.py
```diff
@@ -278,18 +289,23 @@ def process(self, organization_id, cycle, property_view=None):
...
                 meter.is_virtual = m.get('is_virtual')
+                if meter.is_virtual is None:
+                    meter.is_virtual = False
                 meter.save()
 
                 # meterreadings
-                # TODO: need to check that these are in kBtu already?
+                # ENERGY_TYPES index starts at 1, tuple indexes start at 0
+                meter_type = Meter.ENERGY_TYPES[meter.type - 1][1]
+                meter_conversions = self._kbtu_thermal_conversion_factors().get(meter_type, {})
```
macintoshpie, line 299: Hmm, maybe we shouldn't assume enums match indices here. Maybe we could do something like this?
```
dict(Meter.ENERGY_TYPES)[meter.type]
```
Note that dict(<list of tuples>) will create a dict where the keys are the first values, and the values are the second values

seed/models/building_file.py
```diff
@@ -278,18 +289,23 @@ def process(self, organization_id, cycle, property_view=None):
...
                 readings = {
                     MeterReading(
                         start_time=mr.get('start_time'),
                         end_time=mr.get('end_time'),
                         reading=mr.get('reading'),
                         source_unit=mr.get('source_unit'),
                         meter_id=meter.id,
-                        conversion_factor=1.00,  # assuming kBtu
+                        # default to a conversion_factor of 0 if the units specified are not found
+                        conversion_factor=meter_conversions.get(mr.get('source_unit'), 0.00)
```
macintoshpie, line 308: I think we might want to default to `1.0` -- if we do 0 then we will always get zero as a result... Not sure what the downstream effects are, but 1 seems "safe"

seed/tests/test_greenbutton_import.py
```diff
@@ -100,14 +100,14 @@ def test_green_button_import_base_case(self):
         self.assertEqual(meter_reading_10.start_time, make_aware(datetime(2011, 3, 5, 21, 0, 0), timezone=self.tz_obj))
         self.assertEqual(meter_reading_10.end_time, make_aware(datetime(2011, 3, 5, 21, 15, 0), timezone=self.tz_obj))
         self.assertAlmostEqual(meter_reading_10.reading, 1790 * 3.41 / 1000)
-        self.assertEqual(meter_reading_10.source_unit, 'kWh (thousand Watt-hours)')
-        self.assertEqual(meter_reading_10.conversion_factor, 3.41)
+        self.assertEqual(meter_reading_10.source_unit, 'Wh (Watt-hours)')
```
macintoshpie, line 103: just curious, do you know why these tests for greenbutton changed? I would assume they wouldn't be affected by BSync stuff

seed/building_sync/mappings.py
```diff
@@ -207,18 +207,106 @@ def to_energy_type(energy_type):
...
     """
     # avoid circular dependency
     from seed.models import Meter
 
-    energy_name = "Electric - Grid" if energy_type == 'Electricity' else energy_type
-    energy_name = energy_name.lower()
+    # valid energy type values from the schema (<xs:simpleType name="FuelTypes">) and their maps
+    # non-trivial or non-obvious mappings currently map to "Other:" and are flagged with a comment
+    # this mapping is important for unit generation... see "kbtu_thermal_conversion_factors"
+    energy_name = {
```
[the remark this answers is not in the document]
macintoshpie: @nllong would you be ok with merging as is (after your recommended changes) and creating a ticket to review these more in-depth at a later time?

macintoshpie (changes requested):
Nice! Just a few comments

seed/building_sync/mappings.py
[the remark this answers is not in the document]
Ryoken: Yeah, I'm game for pushing this through as the initial implementation and then making a task to come back and flesh it out with more thorough mappings.

seed/building_sync/mappings.py
```diff
@@ -207,18 +207,106 @@ def to_energy_type(energy_type):
...
+        'Electricity': 'Electric - Grid',
+        'Electricity-Exported': 'Other:',                           # other?
```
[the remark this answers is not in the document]
Ryoken: No issues that I am aware of, it's just using these to find the correct conversion factor.

seed/building_sync/mappings.py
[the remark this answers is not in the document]
Ryoken: We should add a new energy type to the schema then that covers this, say "Mass Tonnes (metric)".

seed/models/building_file.py
Ryoken: I like it, thanks!

seed/tests/test_greenbutton_import.py
Ryoken: If you load up the file it's importing you will notice the meters are reporting a uom code of '72' which is just watt hours (see seed/lib/mcm/reader.py).  However, 'Wh' weren't supported previously, so they were being converted to 'kWh' and all the tests had to mirror that.  Now that the parser can support Wh directly, the tests had to be updated.

seed/models/building_file.py
```diff
@@ -278,18 +290,25 @@ def process(self, organization_id, cycle, property_view=None):
                 meter.source = m.get('source')
                 meter.type = m.get('type')
                 meter.is_virtual = m.get('is_virtual')
+                if meter.is_virtual is None:
+                    meter.is_virtual = False
```
nllong, line 294: Nice, so meters are assume to be "real", which is what AT is asserting as well.

seed/building_sync/mappings.py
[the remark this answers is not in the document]
nllong: For carbon it will be kgCO2_e or tonnes CO2. 

image

Since we specify the units in the schema, I think we assume that the user put the data with the right units.

nllong (approved):
I think this is good for our current needs. We can revisit as needed when we start tracking more carbon and demand flexibility metrics.

seed/models/building_file.py
```diff
@@ -305,7 +305,7 @@ def process(self, organization_id, cycle, property_view=None):
                     MeterReading(
                         start_time=mr.get('start_time'),
                         end_time=mr.get('end_time'),
-                        reading=mr.get('reading'),
+                        reading=float(mr.get('reading', 0)) * meter_conversions.get(mr.get('source_unit'), 1.00),
```
macintoshpie, line 308: We store meter data as kbtu in the database and convert on the way out

macintoshpie (approved):
If the tests pass this should be good to go. I had to tweak the import of meter data so that the conversion was done properly. You can now see the converted values on the meter page. This example shows conversion of 500 gallons from bsync:
Screen Shot 2021-06-18 at 8 58 12 AM

seed/models/building_file.py
Ryoken: ah, good to know!  makes sense too, thanks for the catch

PR #2715, opened 2021-06-14
refactor: allow meter data import from PM data requests

macintoshpie:
#### Any background context you want to provide?
Users of SEED have requested the ability to import meter data from Data Request spreadsheets

#### What's this PR do?
Adds the ability for the existing meter imports flow to import Data Request meter data.

This has two technical implications:
- adding a new `Electricity - Unknown` source type for Meters. Data requests don't specify the type of data
- adding a new meter display type for `Electricity - Unknown`

Note: data requests do not include source meter IDs, so `Unknown` is use for the IDs of meters which come from Data Requests

#### How should this be manually tested?
1. download an Excel workbook from a data request which includes natural gas and/or electricity data. Either use the example file provided by OpenTech, or use SEED's test account and go to reports and for SEED City Test Report click "Download current report in Excel".
2. Make sure one or more of your properties have the appropriate PM ID which will link to the meter data being imported
3. import PM meter data like normally, select the data request file

#### What are the relevant tickets?
#2673 

#### Screenshots (if appropriate)
Screen Shot 2021-06-14 at 5 06 48 PM

Issue #869, opened 2016-03-23
Make import of Portfolio Manager Monthly meter data one step and include mapping


RDmitchell:
In the timeseries module, the program can read a PM multi-tab spreadsheet that has Properites, as well as Meter and Meter Consumption tabs. 

![image](https://cloud.githubusercontent.com/assets/6314950/14002107/eb9fbe36-f106-11e5-81e9-f1f880aa8568.png)

The program currently requires two import steps
1 - Import the spreadsheet as a regular PM file
2 - Import the spreadsheet as a PM file with "Energy Data"
![image](https://cloud.githubusercontent.com/assets/6314950/14002182/7d432864-f107-11e5-95c2-35a568dd77dd.png)

Step 1 goes through the standard mapping steps, but Step 2 does not. 

We should combine these two operations into one step, so that the mapping step is performed for all the tabs in the database, so that the correct fields will get mapped and matched properly. 

Right now I have a file that has two records in the Properties tab with the same address. Each has data in the Meter and Meter Consumption tabs, but the program only matches to one of the records. If the PM Property ID could also be mapped and set as a matching field, I think that the two records would both end up having the correct monthly data. 

This will involve some UI as well as probably some BE design.

nllong:
The design is negotiable, but I just wanted to assign @adrian-lara  this ticket to make sure it is on thy radar.

nllong:
@adrian-lara -- are you working on this? Right now the two step process is the default and unless you already started working on this, then I think we should postpone this specific task for now.

adrian-lara:
@nllong I'm not working on this, so 👍 to postponing it.

RDmitchell:
@nllong -- even though this is from the old CMU meter implementation, the issue is still relevant. So maybe we can add this to the list of "Meter" improvements for FY 2020?

adrian-lara:
Capturing a separate conversation about this: 

The solution we'd like to move forward with here is to add a prompt after importing properties from a file that has both inventory records as well as meter information. This prompt could say something like, 
> It looks like the file you've uploaded has accompanying meter reading data. Would you like to import those meters?

My initial thought is that we'd need to research the related effort behind the following:
- Creating an API endpoint to check whether a previously imported file has the expected tab containing meter information (with the exact tab name)
- Reusing an import file record for two different types of imports - the first being for property records and the second for meters. **Alternatively**, we could create a new import file record based on the same underlying file. Some work will likely be needed in either case.

[...]

PR #2721, opened 2021-06-17
869 feat/pm file single import

adrian-lara:
Previously, if a user had a Portfolio Manager file that contained both property and meter information (on separate tabs), they needed to upload the file twice in order to import both properties then meters.

#### What's this PR do?
This PR introduces a new button that shows up after importing an xlsx file that has one of two common ESPM meters tabs. This new button allows for the reuse of the underlying file to import the meters.

Import properties using either:
[example_espm_meter_files.zip](https://github.com/SEED-platform/seed/files/6673469/example_espm_meter_files.zip)

- ESPM data request file with meters in a separate tab "Monthly Usage"
- ESPM file with meters in a separate "Meter Entries" tab 

(Note that the different meter tabs have a different expected format. See #2715 for info on the newly accepted "Monthly Usage" format.)

#869 

Screen Shot 2021-06-17 at 3 53 19 PM


macintoshpie (approved):
Great! Tested locally with provided files and it works.

PR #2725, opened 2021-06-22
Chore/email template develop

macintoshpie:
ClearlyEnergy has provided code for using email templates (see #2706 ). I needed to tweak some things before merging into develop, so this branch is the result

#### What's this PR do?
- fixes a race condition in the email sending endpoint
- fixes the tox docs test

See testing notes in #2706


seed/views/v3/postoffice.py
```diff
@@ -64,7 +64,28 @@ def perform_create(self, serializer):
             )
 
             user = self.request.user
-            ptr.organization_id = org_id
-            ptr.user = user
-            ptr.template_id = template_id
-            ptr.save()
```
macintoshpie: Though in my original testing the emails seemed to work fine, after testing it again today the emails were missing context data. I realized that we were possibly causing async/race condition issues (emails are sent in celery), and found the commit before this addition and "reverted" it back to that. It is working again

PR #2733, opened 2021-07-08
Fix/better bsync file

aviveiros11:
After reviewing the example BSync files for BETTER - I noticed from SEED I was not including `ReadingType` or `LinkedPremises`.  This PR now produces BSync BETTER use-case compliant files.


macintoshpie (approved):
Sweet! 🏛️

Issue #2660, opened 2021-04-06
Data Analysis (Statistics)


adrian-lara:
@laurenliz22 , can you add details here around what SEED will be able to do as well as a rough idea of what the remaining work looks like here.

Initial Notes:
- analysis statistics were originally SEED to comstock
- add analysis focused on a current portfolio
    - e.g., how many types of buildings do we have, different PDF on square footage, etc

macintoshpie:
Closing as first iteration is completed. We can create new issues/tasks as needed.

PR #2742, opened 2021-07-13
feat(models): add stats to analysis.pys

laurenliz22:
Added code to develop analysis statistics around the inventory properties

#### What's this PR do?
API to create JSON output of analysis statistics around the inventory properties

run the API in Firefox filling in the organizaton_id and cycle_id: example = http://0.0.0.0:8000/api/v3/analyses/stats/?organization_id=1&cycle_id=2

https://github.com/SEED-platform/seed/issues/2660


macintoshpie (commented):
This is looking good! A few thoughts about what we could tweak / improve:
1. Having user-friendly chart and axis titles would be nice
2. having user friendly text in the table would be nice
3. I think we could improve the appearance of the table a bit. Maybe we could make it look more like the inventory list table with a little CSS?
Screen Shot 2021-07-26 at 12 24 57 PM

macintoshpie:
One more thought, if a chart will be empty then we should hide it from display

laurenliz22:
@macintoshpie This has been pushed up with the changes above.  Please reach out if you have any additional changes/questions.  Thanks!

macintoshpie (approved):
Awesome! This looks great ✨ 

I tweaked the charts so that if they're missing data it'll say so on the chart with some text
Screen Shot 2021-08-09 at 2 16 08 PM

I also tweaked the stats endpoint to standardize unit-aware fields just in case a user has imported data with different units.

I think this is ready to be merged! There is some followup work we will do to "finish" this page:
- summarize how many properties have non-null values for columns of interest (we have 
- ??? I think there might have been another but I can't remember currently

PR #2745, opened 2021-07-15
Update validation_client.py

macintoshpie:
Selection tool is no more, long live buildingsync.net

#### What's this PR do?
Get rid of selectiontool subdomain

Should be able to successfully import a bsync file


Ryoken (approved):
Seems to work fine!

PR #2762, opened 2021-07-21
refactor(better): map from ESPM property types to BSync types

### LINKED ISSUE #2752: Allow ESPM property types to be used
Currently we only allow BEDES compliant property types (to fit BSync schema requirements). These are different from the types provided by ESPM and used by BETTER.

We need to map these property types internally during the analysis.
See our types in `BETTER_PROPERTY_TYPES`, see BETTER's property types here:
Screen Shot 2021-07-20 at 2 22 39 PM

macintoshpie:
SEED is improving integration with BETTER

#### What's this PR do?
Requires property type in SEED to match those used by BETTER/PM rather than those used by BEDES/BSync

Upload the example file (see below). Map all required fields to SEED fields (City, State, Postal Code, Property Type, PM ID, ...)
Import meters from same file.

You should be able to create a BETTER analysis, and see that the "Building Type" in the BETTER html file is K-12

Closes #2752 

[PM_BETTER_INPUT.xlsx](https://github.com/SEED-platform/seed/files/6857341/PM_BETTER_INPUT.xlsx)

### DIFF
--- a/seed/analysis_pipelines/better.py
+++ b/seed/analysis_pipelines/better.py
@@ -42,24 +42,25 @@ import requests
 logger = logging.getLogger(__name__)
 
 HOST = "https://better-lbnl-development.herokuapp.com"
-# Acceptable BEDES defined BETTER inputs
-BETTER_PROPERTY_TYPES = ['Office',
-                         'Lodging',
-                         'Education',
-                         'Health care-Inpatient hospital',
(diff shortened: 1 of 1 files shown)


Ryoken (approved):
Looks good!

macintoshpie:
@aviveiros11 FYI

Issue #2764, opened 2021-07-22
Modify BuildingSync file creation to conform to BETTER geocoding requirements


aviveiros11:
BETTER parses the location of a building as `<Address>, <City>, <State>` in order to geocode properly.  SEED is currently sending just the `<City>`.

PR #2765, opened 2021-07-22
fix: add address line 1 to bsync file to send to BETTER

aviveiros11:
BETTER integration improvements
#### What's this PR do?
Adds address to BuildingSync file. 
View the bsync input file and optional upload to BETTER building create endpoint via bsync content
#2764

### DIFF
--- a/seed/analysis_pipelines/better.py
+++ b/seed/analysis_pipelines/better.py
@@ -377,7 +377,6 @@ def _build_better_input(analysis_property_view, meters):
             )
         )
     )
-    print(etree.tostring(doc))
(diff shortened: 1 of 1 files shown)


macintoshpie (approved):
Perfecto, tested and verified the address is being added to BSync:
Screen Shot 2021-07-22 at 11 47 53 AM

Also tested different location to see if BETTER was indeed finding the right weather station and it appears so (though we are tweaking/fixing this functionality on BETTER server as we speak)
Screen Shot 2021-07-22 at 11 52 10 AM

Issue #2770, opened 2021-07-22
[stretch] Add analysis results to original inventory PropertyState


macintoshpie:
After a user successfully runs a BETTER analysis for a property, the inventory list should display some key things parsed from that analysis

PR #2775, opened 2021-07-23
feat(better): add results to original property state

macintoshpie:
SEED is improving integration with BETTER

#### What's this PR do?
- makes request to get building analysis as JSON and stores it in the AnalsisPropertyView's parsed_results property
- adds a step to the BETTER pipeline to use the JSON results to update to _original_ PropertyState. Currently it stores the potential cost and energy savings.

Upload the file below. Import the required BETTER fields (City, State, Postal Code, Property Type) as well as the meters from that same file. Then Run the analysis for all or one of the properties. After it's completed, a new Column should be created for both of the columns mentioned above. Go to the column list profile settings, add those new BETTER columns (they both start with "BETTER") then view the inventory list and confirm the results were put into these columns.

[PM_BETTER_PORTFOLIO_INPUT.xlsx](https://github.com/SEED-platform/seed/files/6871487/PM_BETTER_PORTFOLIO_INPUT.xlsx)

#2770 

Screen Shot 2021-07-23 at 4 14 49 PM


aviveiros11 (approved):
Works as expected! 
Screen Shot 2021-07-27 at 8 04 35 AM

Issue #2753, opened 2021-07-20
Allow all BETTER supported meter types to be used in SEED


macintoshpie:
SEED's BETTER analysis should try to use all meters supported by BETTER.

This will also require an update to BETTER's BuildingSync parser.

PR #2776, opened 2021-07-26
feat: update resource use for all espm meter types

aviveiros11:
on-going improvement of BETTER integration
#### What's this PR do?
allows for dynamically populated resource uses in the bsync file for fuel type and unit assignment.  
view the bsync input file for a BETTER analysis
#2753 

NOTE: will work pending update of BETTER bsync parser.


macintoshpie:
@aviveiros11 can you take a look at my changes?

seed/analysis_pipelines/better.py
```diff
@@ -337,38 +359,36 @@ def _build_better_input(analysis_property_view, meters):
...
-                                            E.EnergyResource('Natural gas'),
-                                            E.ResourceUnits('MMBtu'),
-                                            E.EndUse('Heating')
-                                        ),
+                                        *[
+                                            E.ResourceUse(
+                                                {'ID': f'ResourceUse-{meter_idx:03}'},
+                                                E.EnergyResource(SEED_TO_BSYNC_RESOURCE_TYPE[meter.type]),
+                                                # SEED stores all meter readings as kBtu
+                                                E.ResourceUnits('kBtu'),
```
macintoshpie, line 367: Turns out seed stores all readings as kbtu, so no need to convert to anything else

aviveiros11: I think this should be fine just so long as ESPM does not have a meter type that is incompatible with `kBtu`.

PR #2780, opened 2021-07-27
patch boto, new oep

nllong:
* Upgrade boto3 and django-ses (this has probably already been done).
* Use the most recent version of OEP by default when building local containers. (Utilized SEED API v3.)


macintoshpie (approved):
Noice 🥇

Issue #2777, opened 2021-07-26
Investigate MeterReading model


macintoshpie:
Research needs to be done for this task.

I am concerned that we are storing meter readings incorrectly. [This comment](https://github.com/SEED-platform/seed/blob/dc41e7a22f4a9f447f1281a4b8e7314109834386/seed/models/meters.py#L158) caused me concern b/c I thought we _were_ using these fields to convert between units.

Please determine if this comment seems outdated. Please also add comments to the MeterReading model -- it seems that we are assuming _all_ readings are stored as kbtu in the database (converted using conversion_factor). If this is the case we should explicitly mention this and how it's used on when importing data as well as when it's displayed to the user (according to their org's unit preferences)

macintoshpie:
I think we store everything as kBtu

Here, when parsing PM meters I believe, we store the 'reading' as the original value multiplied by the conversion factor: https://github.com/SEED-platform/seed/blob/dc41e7a22f4a9f447f1281a4b8e7314109834386/seed/data_importer/meters_parser.py#L361

Here when gathering meter readings for a web request, SEED divides the reading value by the conversion factor:
https://github.com/SEED-platform/seed/blob/dc41e7a22f4a9f447f1281a4b8e7314109834386/seed/utils/meters.py#L108

Also we have this helper which is used in both places to get conversion factor to kBtu: https://github.com/SEED-platform/seed/blob/dc41e7a22f4a9f447f1281a4b8e7314109834386/seed/data_importer/utils.py#L62

PR #2782, opened 2021-07-28
docs: build docs around meter readings

aviveiros11:
Investigation of how we are storing meter readings.
#### What's this PR do?
SEED does in fact store all meter readings as 'kBtu' regardless of incoming units on import.  This is accomplished by a `conversion_factor` lookup via the meter type and raw units provided from `espm`.  However, when viewing the Meters tab on the Property Detail page, the original imported units and readings are shown.  

Documentation has been added to the model to hopefully provide information on the actual behavior of the model fields.  

#2777


seed/models/meters.py
```diff
@@ -142,6 +142,16 @@ def copy_readings(self, source_meter, overlaps_possible=True):
...
+    """
+    A meter reading represents the actual usage entry for a given meter.
+
+    NOTE: SEED stores all meter readings in kBtu.  The raw usage reading is converted
+    on import to kBtu using the conversion_factor which is determined by the meter type and raw units on import,
+    therefore, the reading field of this model will always be in kBtu.
+
+    The original units however when being displayed to the user ie. (Property Detail Meters tab)
+    will contain the original units and meter readings.
+    """
```
macintoshpie, line 154: EXCELLENT

macintoshpie (approved):
💣

Issue #2693, opened 2021-05-24
BuildingSync: allow bsync to import all current and future v2 docs


macintoshpie:
## background
SEED and Audit Template are improving interoperability. SEED currently only allows v2.0 BuildingSync files to be imported.

## task
Update SEED to allow importing of v2.0, 2.1, 2.2, 2.3, and other future 2.x versions. Note that there/s probably no need to update XPaths for parsing as all v2 files should not remove/reorganize elements.
You _might_ however have to modify the export feature to handle these different versions.
Update the default BSync ColumnMappingProfile to indicate it's valid for any v2 file rather than just 2.0
Update call to selection tool (validation client) to use the version parsed from the bsync file

Ryoken:
Be sure to update https://github.com/SEED-platform/seed/blob/6bef732840ebc4bbdb2ac2d9ccf1e6b2a1d2b3db/seed/building_sync/building_sync.py#L359 to v2.3 as well

Ryoken:
Stashing this for now: https://github.com/SEED-platform/seed/tree/2693-feature/bsync_support_v2

Ryoken:
https://github.com/SEED-platform/seed/pull/2784

macintoshpie:
No user testing needed at this time 😃

PR #2784, opened 2021-07-29
2693 feature/bsync support v2

Ryoken:
#### What's this PR do?
Adds support for all V2.x buildingsync files

Upload a version 2.3.0 buildingsync file (such as <a download href=https://buildingenergyscore.energy.gov/documents/example_SF_audit_report_BS_v2.3_033121.xml>this one</a>)
https://github.com/SEED-platform/seed/issues/2693


seed/building_sync/building_sync.py
```diff
@@ -365,15 +369,19 @@ def _parse_version(self):
...
             raise ParsingError('Expected BuildingSync element as root element in xml')
 
+        # first check for a version attribute in the buldingsync tag
+        if "version" in bsync_element.attrib:
+            return bsync_element.attrib["version"]
+
+        # second check if it's a file form Audit Template Tool and infer the version
+        # Currently ATT doesn't include a schemaLocation so this is necessary
+        if self._is_from_audit_template_tool():
+            return self.BUILDINGSYNC_V_LATEST
```
macintoshpie, line 383: Ah I didn't think about this, but if a file is missing the version attribute and it appears to be from ATT, then we should assume the bsync version is `2.0`, not the latest (v2.0 was the last version which didn't require a version attribute). Could you update the comment to say something like
```
# second check if it's a file form Audit Template Tool
# if so, then it must be a 2.0 file b/c that was the last version which didn't require @version
```

seed/building_sync/validation_client.py
```diff
@@ -13,16 +13,11 @@ class ValidationClientException(Exception):
...
-    files = [
-        ('file', file_)
-    ]
-
     return requests.request(
         "POST",
         VALIDATION_API_URL,
-        data=payload,
-        files=files,
+        data={'schema_version': schema_version},
```
macintoshpie, line 19: Ugh, I just found out that if someone uploads a v2.0 file, we will pass a version of 2.0 (which is correct according to our tags) but selection tool requires 2.0.0. Can we add a check in this function to change version to 2.0.0 if someone passes BUILDINGSYNC_V2_0?

seed/data_importer/tasks.py
```diff
@@ -1620,12 +1621,15 @@ def pair_new_states(merged_property_views, merged_taxlot_views):
 def _validate_use_cases(file_pk, progress_key):
     import_file = ImportFile.objects.get(pk=file_pk)
     progress_data = ProgressData.from_key(progress_key)
+    bs = BuildingSync()
+    bs.import_file(import_file.file)
```
macintoshpie, line 1625: Yet another gotcha, we allow users to upload zip files of BuildingSync files, so this causes the BuildingSync class to try and parse a zip. We might have to open the zip, get the version of the first file and validate them all as that version... What do you think about that? Alternatively we could iterate through all the files and validate them individually

seed/building_sync/validation_client.py
```diff
@@ -13,16 +13,11 @@ class ValidationClientException(Exception):
...
+        files={'file': (file_.name, open(file_.name, 'r').read(), 'application/xml')},
```
macintoshpie, line 20: We also allow folks to upload zips (and validate zips too). I didn't get the chance to test this yet, but I'm wondering if specifying this content-type will work.

Here's an example zip we use: seed/building_sync/tests/data/ex_1_and_buildingsync_ex01_measures.zip

seed/building_sync/building_sync.py
```diff
@@ -43,9 +43,13 @@ class ParsingError(Exception):
 class BuildingSync(object):
     BUILDINGSYNC_V2_0 = '2.0'
     BUILDINGSYNC_V2_2_0 = '2.2.0'
+    BUILDINGSYNC_V2_3_0 = '2.3.0'
```
macintoshpie, line 46: Not sure why I didn't add this earlier, but could you add 2.1.0 as well?

macintoshpie (changes requested):
Great job so far, glad this is looking like it's not requiring many big changes! Just a few comments about tweaks we'll have to make (sorry for more gotchas).

Could you also add these new schemas to the repo? We store them here:
seed/building_sync/schemas/

We should probably have some test which asserts there's a schema for every version we claim to support, but don't feel like you have to add that now.

These schema files are used when exporting files to make sure we structure the file correctly.

seed/building_sync/building_sync.py
Ryoken: gotcha!

seed/building_sync/validation_client.py
Ryoken: I instead just updated BUILDINGSYNC_V2_0 to be "2.0.0" as I don't believe it's ever used to verify "2.0".  Can you see any problems with this?

seed/data_importer/tasks.py
Ryoken: I honestly feel we should validate all the files if a ZIP is presented.  I'll refactor to accomodate!

seed/building_sync/validation_client.py
Ryoken: Ah, I had specified content type in an attempt to debug the file upload issue.  I don't think it's needed in either case however so I'll just remove it.

seed/building_sync/validation_client.py
macintoshpie: I'm worried that might cause some issues, our official tag is [v2.0](https://github.com/BuildingSync/schema/releases/download/v2.0/BuildingSync.xsd) (not 2.0.0) which people would use in schemaLocation to reference the schema, which we were parsing out to determine the BSync version originally (before v2.1 added the `@version` attribute).

seed/data_importer/tasks.py
```diff
@@ -1620,12 +1622,33 @@ def pair_new_states(merged_property_views, merged_taxlot_views):
...
+                with tempfile.TemporaryDirectory() as temp_dir:
+                    zip_file.extractall(path=temp_dir)
+                    for file_name in zip_file.namelist():
+                        bs = BuildingSync()
+                        bs.import_file(f'{temp_dir}/{file_name}')
+                        if found_version == 0:
+                            found_version = bs.version
+                        elif found_version != bs.version:
+                            raise Exception(f'Zip contains multiple BuildingSync versions (found {found_version} and {bs.version})')
+            import_file.refresh_from_db()
```
macintoshpie, line 1641: Nice! worked for me!

macintoshpie (approved):
Thanks! This looks great, I tweaked two things:
- added a migration to rename the default BuildingSync column mapping to just v2 instead of v2.0 now that we are importing more bsync versions
- fixed exports for other bsync file versions by tweaking get_schema method

If you could take a look at my changes and test it yourself by running the migration and importing / exporting a v2.3 bsync file, then I think this will be good to go!

Imported 2.3 file
Screen Shot 2021-08-05 at 8 02 49 AM
Screen Shot 2021-08-05 at 8 02 55 AM

Error when trying to import a zip with multiple versions of BSync (nice!)
Screen Shot 2021-08-05 at 8 06 36 AM

Ryoken:
Works for me!  Thanks for the additions.

Issue #2779, opened 2021-07-27
Add org-level setting for BETTER API token


macintoshpie:
Should work like the MAPQUEST API token. Should include helpful message about how to get the token

PR #2788, opened 2021-07-30
refactor(organization)!: add org-level BETTER token

macintoshpie:
SEED is improving BETTER integration

#### What's this PR do?
- adds field to organizations for storing BETTER API token. This means organizations are now required to provide their own token in order to run the analysis. ⚠️  As a result this PR requires a migration to be run
- refactor BETTER client and helpers to work a bit more cleanly now that the API token isn't a global variable from settings.

Note to reviewers: the changes to BETTER client are almost entirely whitespace, view the meaningful diffs by hiding the whitespace diffs:
Screen Shot 2021-07-30 at 1 32 24 PM

Run the migration.
Upload the linked file below adding required properties (see recent BETTER prs for more info).
On the org settings page, you should see a field for the BETTER API token. You should be able to update this field.
From the inventory list page, try to run an analysis (1) without a BETTER api token in the org settings, (2) with an invalid BETTER API token, and (3) with a valid BETTER API token. (1) should fail saying you need to add a token. (2) will fail with a message from better indicating your token is bad. (3) should succeed like normal.

[PM_BETTER_PORTFOLIO_INPUT.xlsx](https://github.com/SEED-platform/seed/files/6909670/PM_BETTER_PORTFOLIO_INPUT.xlsx)

#2779 

Org settings page with new better token field
Screen Shot 2021-07-30 at 1 27 59 PM

Notification when trying to create a BETTER analysis when the org token is invalid
Screen Shot 2021-07-30 at 1 29 05 PM

Notification when trying to create a BETTER analysis when the org token is missing
Screen Shot 2021-07-30 at 1 29 43 PM


aviveiros11 (commented):
Maybe we could add additional information to the user when adding their BETTER token to not include the `Token` prefix.  I foresee users reading the use of the prefix in the better documentation and trying to add the entire thing to SEED? 

Otherwise, worked as expected:

Without token: 
Screen Shot 2021-08-02 at 12 11 58 PM

With Invalid token: 
Screen Shot 2021-08-02 at 12 14 02 PM

aviveiros11 (approved):
Perfect 💪

PR #2798, opened 2021-08-04
chore: always use latest chrome in CI

macintoshpie:
We pinned the chrome version for testing previously. Now apt is complaining b/c when we try to install the package it would downgrade chrome.

#### What's this PR do?
Install latest stable chrome, then parse that version out to install the webdriver

NA


Ryoken:
Looks good!

Issue #2810, opened 2021-08-10
Bug: check meters tab on spreadsheet is broken


macintoshpie:
## context
We added a function to check if there's a meters tab after a file is imported to prompt a user to import the meters as well.

## bug
It doesn't work. E.g. I upload a BSync file and it asks if I want to import meters

RDmitchell:
Instance: dev1
SHA: 
Org: LBNL 405

I have seen inconsistent results for this for ESPM XLSX files with meter tabs; probably to do with how the spreadsheet it set up. 

Sometimes SEED figures out there is meter data and asks if you want to import it, but sometimes it doesn't. 

I think I will move this to the 12/31/2021 project, and assign it to me, so that I can try to figure out when it works and when it doesn't.

macintoshpie:
@RDmitchell thanks, sounds like we do have some tweaking to do. I agree that moving it to the new project board is best. I'll also tag myself on this to make sure it doesn't slip through the cracks.

RDmitchell:
@macintoshpie -- I observed this today, ie, I waited a few seconds and then the "Import Meters" option appeared when importing an ESPM file that had meters in it. 

I have recorded a video that shows this -- you can see the behavior at 3:47 minutes in.
Issue 2913.wmv: https://drive.google.com/drive/folders/1N2e7I2hm487hqnA6kTOgcPMbCPmwue8b?usp=sharing

macintoshpie:
I think I'm going to close this ticket if that's ok with you @RDmitchell , the original bug has been fixed.
I created this ticket to track the bad UX for a button showing up late: #2978

RDmitchell:
@macintoshpie -- sure that's fine !!

PR #2811, opened 2021-08-10
fix: correctly check if import file has meters tab

macintoshpie:
#### What's this PR do?
Fixes the bug in issue

- upload a BSync file, it should not prompt you to import meters
- import a spreadsheet with meters. it should prompt you to import meters

#2810


seed/static/seed/js/controllers/data_upload_modal_controller.js
```diff
@@ -653,7 +653,7 @@ angular.module('BE.seed.controller.data_upload_modal', [])
 
                 // Toggle a meter import button if the imported file also has a meters tab
                 dataset_service.check_meters_tab_exists($scope.dataset.import_file_id).then(function(result) {
-                  $scope.import_file_reusable_for_meters = result;
```
macintoshpie, line 656: whoops. I think `result` was always an object here. We care about result.data which indicates if it has a meter tab or not

seed/views/v3/import_files.py
```diff
@@ -185,12 +184,12 @@ def check_meters_tab_exists(self, request, pk=None):
...
             return JsonResponse({
                 'status': 'success',
-                'data': dumps(bool(meter_tab_check))
+                'data': bool(meter_tab_check)
             })
         except XLRDError:
             return JsonResponse({
                 'status': 'success',
-                'data': 'false'
+                'data': False
```
macintoshpie, line 192: whoops, don't stringify the boolean. strings are truthy

Ryoken (approved):

PR #2817, opened 2021-08-13
fix: handle additional edge cases for BETTER analysis

macintoshpie:
SEED is improving BETTER integration

#### What's this PR do?
- only pass meters and readings with 12 readings of values >= 1.0
- dont fail pipeline if one building fails to be created in BETTER

Import the CSV of PM data and meters (shared by Nick, ask me for access if you need it). Select all buildings and run the analysis. The analysis should skip some buildings b/c of two issues with these properties:
- they have meters with readings of 0.0 (you should see an explicit message about this next to the property)
- they have a location which BETTER fails to geocode (SEED doesn't know when this happens though, we just treat it as an unexpected error and continue)

The analysis should complete, and you should be able to view both the portfolio and building reports

### DIFF
--- a/seed/analysis_pipelines/better/buildingsync.py
+++ b/seed/analysis_pipelines/better/buildingsync.py
@@ -79,10 +79,20 @@ def _build_better_input(analysis_property_view, meters):
     if property_state.property_type not in BETTER_TO_BSYNC_PROPERTY_TYPE:
         errors.append(
             f"BETTER analysis requires the property's type must be one of the following: {', '.join(BETTER_TO_BSYNC_PROPERTY_TYPE.keys())}")
+
+    valid_meters_and_readings = []
     for meter in meters:
-        for meter_reading in meter.meter_readings.all():
-            if meter_reading.reading is None:
-                errors.append(f'{meter}: MeterReading starting at {meter_reading.start_time} has no reading value.')
+        readings = meter.meter_readings.filter(reading__gte=1.0)
+        if readings.count() >= 12:
+            valid_meters_and_readings.append({
+                'meter': meter,
+                'readings': readings,
+            })
+    if len(valid_meters_and_readings) == 0:
+        errors.append(
+            f'BETTER analysis requires at least one meter with 12 consecutive readings with values >= 1.0'
+        )
+
     if errors:
         return None, errors
 
@@ -176,12 +186,12 @@ def _build_better_input(analysis_property_view, meters):
                                         *[
                                             E.ResourceUse(
                                                 {'ID': f'ResourceUse-{meter_idx:03}'},
-                                                E.EnergyResource(SEED_TO_BSYNC_RESOURCE_TYPE[meter.type]),
+                                                E.EnergyResource(SEED_TO_BSYNC_RESOURCE_TYPE[meter_and_readings['meter'].type]),
                                                 # SEED stores all meter readings as kBtu
                                                 E.ResourceUnits('kBtu'),
                                                 E.EndUse('All end uses')
                                             )
-                                            for meter_idx, meter in enumerate(meters)
+                                            for meter_idx, meter_and_readings in enumerate(valid_meters_and_readings)
                                         ]
                                     ),
                                     E.TimeSeriesData(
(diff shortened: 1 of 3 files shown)


seed/analysis_pipelines/better/client.py
```diff
@@ -146,19 +146,30 @@ def is_ready(res):
             response, errors = res[0], res[1]
             if errors:
                 raise Exception('; '.join(errors))
-            return response['generation_result'] == 'COMPLETE'
 
+            if response['generation_result'] == 'COMPLETE':
+                return True
+            elif response['generation_result'] == 'FAILED':
+                raise Exception(f'BETTER failed to generate the portfolio analysis: {response}')
```
macintoshpie, line 153: new edge case, was getting a FAILED result b/c of bad meter data

seed/analysis_pipelines/better/client.py
```diff
@@ -197,7 +208,7 @@ def create_building(self, bsync_xml, better_portfolio_id=None):
         :param bsync_xml: str, path to BSync xml file for property
         :param better_portfolio_id: int | str, optional, if provided it will add the
             building to the portfolio
-        :returns: requests.Response building_id
+        :returns: tuple(int, list[str]), BETTER Building ID followed by list of errors
```
macintoshpie, line 211: before we would fail the entire pipeline if we failed to create a building, we should be more flexible and allow buildings to fail

seed/analysis_pipelines/better/buildingsync.py
```diff
@@ -79,10 +79,20 @@ def _build_better_input(analysis_property_view, meters):
     if property_state.property_type not in BETTER_TO_BSYNC_PROPERTY_TYPE:
         errors.append(
             f"BETTER analysis requires the property's type must be one of the following: {', '.join(BETTER_TO_BSYNC_PROPERTY_TYPE.keys())}")
+
+    valid_meters_and_readings = []
     for meter in meters:
-        for meter_reading in meter.meter_readings.all():
-            if meter_reading.reading is None:
-                errors.append(f'{meter}: MeterReading starting at {meter_reading.start_time} has no reading value.')
+        readings = meter.meter_readings.filter(reading__gte=1.0)
```
macintoshpie, line 85: make sure all readings we pass along are greater than or equal to 1 (BETTER requires this)

seed/analysis_pipelines/better/client.py
```diff
@@ -146,19 +146,30 @@ def is_ready(res):
...
+
+        POLLING_TIMEOUT_SECS = 300
         try:
             polling.poll(
                 lambda: self.get_portfolio_analysis(better_portfolio_id, better_analysis_id),
                 check_success=is_ready,
-                timeout=60,
-                step=1
+                timeout=POLLING_TIMEOUT_SECS,
+                step=10,  # wait 10 seconds between polls
```
macintoshpie, line 163: polling every second was a bit aggressive IMO

seed/analysis_pipelines/better/client.py
aviveiros11: 🔥

seed/analysis_pipelines/better/buildingsync.py
aviveiros11: Good look on that meter filter 🔥 .

seed/analysis_pipelines/better/client.py
aviveiros11: I agree, looking at it now I think BETTER recommends this in their API docs.  `Continue to poll the endpoint (perhaps every 10 seconds) until the generation_result field changes from IN_PROGRESS to another state.`

seed/analysis_pipelines/better/client.py
aviveiros11: This makes sense to me.

aviveiros11 (approved):
Looks good! 🤠

seed/analysis_pipelines/better/client.py
macintoshpie: Ooh good point, should probably be checking if the status is COMPLETE or not IN_PROGRESS. Will consider adding that later

Issue #2790, opened 2021-08-03
Inventory List columns disappear in Column List Profile after hiding a field in the list view


RDmitchell:
**Describe the bug**
"Hiding" a field in the Inventory list view causes all the fields in the Column List Profile for that list to be unchecked when you go to the profile. 
See also  #2749

**Expected Behavior**
It should be possible to Hide (and move) fields from the list view and have the program automatically update the Column List Profile being used.

**Actual Behavior**
See Steps to Reproduce

**Steps to Reproduce**
Org == LBNL 444
Cycle == 2020 Calendar Year

Here is a link to a video that shows the behavior
https://drive.google.com/file/d/1EXBDvC9gFhJb3VTrIofIVqkk8VpDYKJ0/view?usp=sharing

* I created a new user and org and clicked on the "Auto-Populate Sample Data" button to create the default sample data. But this bug happens in any Inventory List view, the auto-populated nature of the data is not the issue.
* Go to Inventory List view. The default fields for the "Auto-Populate" Column List Profile are displayed
* I decide to change the fields that are being displayed based on the Auto-Populate Column List Profile.
![image](https://user-images.githubusercontent.com/6314950/128083963-5c7c1ee2-f50c-4c04-a0e8-6af43c4092fb.png)
* I click on the pulldown list for the column called "Campus", and select the "Hide Column" option
![image](https://user-images.githubusercontent.com/6314950/128084097-0870d6cd-5e46-45e0-92b0-e2912dc2f45b.png)
* The Campus field is no longer displayed in the List view
![image](https://user-images.githubusercontent.com/6314950/128084227-6f151bc3-4751-4a0c-8960-302bbefa0537.png)
* Then click on the "Column List Profiles" link to view the rest of the fields in the "Auto-Populate" Column List Profile
![image](https://user-images.githubusercontent.com/6314950/128084362-d5a4995d-bb15-4411-be04-2e78222a8ebf.png)
* All of the fields that were selected for the "Auto-Populate" Column List Profile are now unselected
![image](https://user-images.githubusercontent.com/6314950/128084498-db7af0e0-2471-4718-82da-e812549cdde4.png)
* Go back to the Inventory List and no fields are displayed, just the labels column, expanded.
![image](https://user-images.githubusercontent.com/6314950/128084664-3ed4038d-d55f-4840-b0d1-38e1ff89476d.png)

**Instance Information**

Instance: dev1
SHA: 4d08f5057

RDmitchell:
[...]

PR #2821, opened 2021-08-16
fix: correctly filter columns when updating column profile

macintoshpie:
#### What's this PR do?
- updating the column list profile was broken on the inventory page when dragging/dropping column headers. This fixes that by properly filtering columns before POSTing

- go to inventory list page
- drag and drop one of the columns (should successfully update the column profile)
- go to column list profile page and verify it was updated
- go back to list profile and it should be correct according to your updated ordering

#2790

### DIFF
--- a/seed/static/seed/js/controllers/inventory_list_controller.js
+++ b/seed/static/seed/js/controllers/inventory_list_controller.js
@@ -1087,7 +1087,9 @@ angular.module('BE.seed.controller.inventory_list', [])
       function currentColumns () {
         // Save all columns except first 3
         var gridCols = _.filter($scope.gridApi.grid.columns, function (col) {
-          return !_.includes(['treeBaseRowHeaderCol', 'selectionRowHeaderCol', 'notes_count', 'merged_indicator', 'id'], col.name) && col.visible && col.id;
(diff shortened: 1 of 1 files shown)


seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -1087,7 +1087,9 @@ angular.module('BE.seed.controller.inventory_list', [])
       function currentColumns () {
         // Save all columns except first 3
         var gridCols = _.filter($scope.gridApi.grid.columns, function (col) {
-          return !_.includes(['treeBaseRowHeaderCol', 'selectionRowHeaderCol', 'notes_count', 'merged_indicator', 'id'], col.name) && col.visible && col.id;
+          return !_.includes(['treeBaseRowHeaderCol', 'selectionRowHeaderCol', 'notes_count', 'merged_indicator', 'id', 'labels'], col.name)
+            && col.visible
+            && !col.colDef.is_derived_column;
```
macintoshpie, line 1092: 2 new things:
1. don't include the labels column
2. ignore derived columns

Ryoken: simple enough.  Thanks!

Ryoken (approved):

PR #2827, opened 2021-08-23
Fix/placeholder

aviveiros11:
#### What's this PR do?
Uses ng-value instead of placeholder input attribute. 
run analysis without modifying the min_r_squared value


macintoshpie (approved):
:sunglasses: :cactus:

Issue #2828, opened 2021-08-23
Parse meter readings from Audit Template file


macintoshpie:
## background
Audit template stores meter readings in a couple different places, including `auc:TimeSeries` as well as `auc:AllResourceTotal`. SEED only pulls meter readings from `auc:TimeSeries` currently.

From what I can tell, peak readings are stored in the `auc:TimeSeries` elements, and point readings (e.g. monthly readings) are stored in `auc:AllResourceTotal`

## task
Parse the readings from `auc:AllResourceTotal`

PR #2829, opened 2021-08-23
refactor(bsync): parse meter data from AllResourceTotals

macintoshpie:
Audit Template tool puts some meter data inside of AllResourceTotal instead of TimeSeries (as we would expect)
Screen Shot 2021-08-23 at 10 44 29 AM

#### What's this PR do?
Parses meter readings from AllResourceTotals in addition to the TimeSeries readings.

NOTE: because SEED only allows one reading for a meter for a specific time span, we must create a _new_ meter to store these readings. This is fine for now, but we should consider removing that constraint and adding a "type" for a reading like BuildingSync's `auc:ReadingType` (e.g. point, median, mean, peak, etc). I add a "Site Energy Use" prefix to the original "meter" (ie ResourceUse) for the meter ID.

Upload a recent Audit template file which contains meter readings (peak, as well as regular readings). Verify meters were imported.

[example_SF_audit_report_BS_v2.3_081321.xml.zip](https://github.com/SEED-platform/seed/files/7034068/example_SF_audit_report_BS_v2.3_081321.xml.zip)

#2828 

Screen Shot 2021-08-23 at 12 25 11 PM

### DIFF
--- a/seed/building_sync/building_sync.py
+++ b/seed/building_sync/building_sync.py
@@ -237,16 +237,51 @@ class BuildingSync(object):
 
             # process the scenario meter readings
             for series_data in scenario['time_series']:
-                reading = {
+                meter_reading = {
                     'start_time': series_data['start_time'],
                     'end_time': series_data['end_time'],
                     'reading': series_data['reading'],
                     'source_id': series_data['source_id'],
                 }
-                reading['source_unit'] = meters[reading['source_id']].get('units')
+                meter_reading['source_unit'] = meters[meter_reading['source_id']].get('units')
 
                 # add reading to the meter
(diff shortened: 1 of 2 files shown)


aviveiros11:
@macintoshpie  when uploading the bsync file everything worked as expected.  One comment would be the inclusion of the meter with all 0 readings - is this supposed to be there? 
Screen Shot 2021-08-23 at 12 48 26 PM

macintoshpie:
@aviveiros11 yeah those are supposed to be there, they're actually included in the BSync file as 0.0's

aviveiros11:
Sweet, then this looks good - I will wait and let @nllong formally approve, but it should be all set.

Issue #2816, opened 2021-08-13
Users have to refresh analyses pages to get updated status


macintoshpie:
## context
SEED's org analyses page lists analyses and their status (creating, ready, running, failed, etc).

## problem
Users have to reload the page every time they want to see whether an analysis is ready to be run or if it finished (assuming they came from another page, e.g. they created an analysis on inventory list page then went to the analysis page.)

macintoshpie:
Possible fixes:
- add a button to refresh the analysis data so they don't have to reload the page
- automatically begin polling SEED when you find an analysis in an "active" (creating, running) state.

PR #2853, opened 2021-09-01
feat(analyses): auto-poll analysis progress on frontend

macintoshpie:
#### What's this PR do?
TLDR: for any page showing an analysis in a list or detailed view, SEED will now begin polling its status for updates so that it knows when to refresh it. This involved updating how progress data are created/tracked in the analysis pipelines.

SPECIFICALLY:
- add a method the the analyses_service which will continuously poll the status of an analysis until it determines theres no more progress data to track (`analyses_service.check_progress_loop(...)`)
- use method above in all views that render analyses so that they're refreshed when the status of the analysis changes
- part of accomplishing the above was adding a new endpoint `/analyses/<analysis id>/progress_key/` which returns the current `ProgressData` key (or null if there's no current `ProgressData` for the analysis).
  - This involved standardizing the progress data for all analysis pipelines. Progress data is now dependent on `Analysis.status`, and pipeline authors just need to use `pipeline.get_progress_data()` to get the current progress data.
  - As a result, pipeline authors must call the `AnalysisPipeline` base class methods for updating the status of the analysis instead of doing it directly (e.g. don't do `analysis.status = Analysis.COMPLETED; analysis.save()`). These methods are:
    - `set_analysis_status_to_ready`
    - `set_analysis_status_to_running`
    - `set_analysis_status_to_completed`
    - the existing `prepare_analysis`, `start_analysis`, `stop_analysis`, and `fail_analysis` methods already handle this for the remaining statuses.
   - the goal of these changes is to standardize the analysis pipeline web interface and hopefully simplify how pipelines are written - now pipeline authors just need to call these methods and not worry about finishing progress data, keeping track of progress data keys, updating the start/end time of the analysis, etc.

I'd recommend importing the anonymized JCC spreadsheet and meter data to start with. Then run try running different analyses and verify that the statuses are being updated correctly. It should NOT continue polling when the page is changed. It should NOT continue polling when the analysis finishes (fails, stopped, or completed).

Here are the pages affected (ie they should show live reloading as the analysis is being created/running/finishing):
- Analyses list page: `/app/#/accounts/<org id>/analyses`
- Analysis page: `/app/#/accounts/<org id>/analyses/<analysis id>`
- Analysis property view page: `/app/#/accounts/<org id>/analyses/<analysis id>/runs/<analysis property view id>` -- this one should actually refresh the whole page to show the results when the analysis is complete
- List of analyses for individual property: `/app/#/properties/<property_id>/analyses` -- this one should also refresh the whole page to view the results of the completed analysis
- Property details page with an analysis card: `/app/#/properties/<property_id>` (NOTE: might not see it here b/c the card is the most recently _completed_ analysis)

#2816 

Screen Shot 2021-09-03 at 1 56 02 PM
Screen Shot 2021-09-03 at 1 56 16 PM
Screen Shot 2021-09-03 at 1 56 54 PM
Screen Shot 2021-09-03 at 1 56 34 PM

### DIFF
--- a/seed/analysis_pipelines/better/pipeline.py
+++ b/seed/analysis_pipelines/better/pipeline.py
@@ -8,7 +8,6 @@ import logging
 import copy
 
 from django.db.models import Q, Count
-from django.utils import timezone as tz
 from django.core.files.base import ContentFile
 from celery import chain, shared_task
 
@@ -38,7 +37,6 @@ from seed.analysis_pipelines.better.helpers import (
     _update_original_property_state,
 )
 
-from seed.lib.progress_data.progress_data import ProgressData
 from seed.models import (
     Analysis,
     AnalysisInputFile,
--- a/seed/analysis_pipelines/better/pipeline.py
+++ b/seed/analysis_pipelines/better/pipeline.py
@@ -108,7 +108,7 @@ class BETTERPipeline(AnalysisPipeline):
             raise AnalysisPipelineException(
                 f'Analysis configuration is invalid: {"; ".join(validation_errors)}')
 
-        progress_data = ProgressData('prepare-analysis-better', self._analysis_id)
+        progress_data = self.get_progress_data(analysis)
 
         # Steps:
         # 1) ...starting
@@ -118,17 +118,15 @@ class BETTERPipeline(AnalysisPipeline):
         progress_data.save()
 
         chain(
-            task_create_analysis_property_views.si(self._analysis_id, property_view_ids, progress_data.key),
-            _prepare_all_properties.s(self._analysis_id, progress_data.key),
-            _finish_preparation.si(self._analysis_id, progress_data.key, start_analysis)
+            task_create_analysis_property_views.si(self._analysis_id, property_view_ids),
+            _prepare_all_properties.s(self._analysis_id),
+            _finish_preparation.si(self._analysis_id, start_analysis)
         ).apply_async()
-
-        return progress_data.result()
 
     def _start_analysis(self):
         """Internal implementation for starting the BETTER analysis"""
 
-        progress_data = ProgressData('start-analysis-better', self._analysis_id)
+        progress_data = self.get_progress_data()
 
         # Steps:
         # 1) ...starting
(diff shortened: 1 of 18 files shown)


seed/analysis_pipelines/better/pipeline.py
```diff
@@ -108,7 +108,7 @@ def _prepare_analysis(self, property_view_ids, start_analysis=False):
             raise AnalysisPipelineException(
                 f'Analysis configuration is invalid: {"; ".join(validation_errors)}')
 
-        progress_data = ProgressData('prepare-analysis-better', self._analysis_id)
+        progress_data = self.get_progress_data(analysis)
```
macintoshpie, line 111: note to self: update other pipelines to use these new methods on the Pipeline base class

macintoshpie: donezo

seed/analysis_pipelines/bsyncr.py
```diff
@@ -429,19 +416,13 @@ def _process_results(self, analysis_output_xml_file_ids, analysis_id, progress_d
...
+def _finish_analysis(self, analysis_id):
     """A Celery task which finishes the analysis run
 
     :param analysis_id: int
-    :param progress_data_key: str
     """
-    analysis = Analysis.objects.get(id=analysis_id)
-    analysis.status = Analysis.COMPLETED
-    analysis.end_time = tz.now()
-    analysis.save()
```
macintoshpie, line 441: e.g. new pipeline api

seed/static/seed/partials/inventory_detail_analyses.html
```diff
@@ -40,7 +40,9 @@ <h2>
                 <b>{$:: type $}</b> <small>({$:: analyses.length $})</small>
             </uib-tab-heading>
             <div class="tab-content-cards">
-                <div ng-repeat="analysis in analyses" ng-include="::urls.static_url + 'seed/partials/analysis_card.html'" class="tab-content-card">
+                <!-- NOTE: This is bad, using a controller created for someone else but it will work
+                     we should refactor this into a AnalysisCard component. -->
+                <div ng-repeat="analysis in analyses" ng-controller="analysis_details_controller" ng-include="::urls.static_url + 'seed/partials/analysis_card.html'" class="tab-content-card">
```
macintoshpie, line 45: This is bad! I'm using a controller that's really written for the analysis_details.html partial. But it works well enough here without writing another controller that does basically the same thing. We should probably refactor these partials (analysis_details.html and analysis_card.html) into components so that they can do their own thing more cleanly

Ryoken (approved):
Seems to be working, thanks!

Issue #2846, opened 2021-08-25
BETTER - Pin most recent BETTER analysis to top


RDmitchell:
**Describe the solution you'd like**
Show the most recent BETTER report in the list of analyses for a property. Then place the other analyses at the bottom of the screen.

Ryoken:
Analyses are now listed at the top and grouped by analysis type, with key information highlighted.

<img width="1142" alt="Screen Shot 2021-09-03 at 11 00 38 AM" src="https://user-images.githubusercontent.com/564231/132041943-88f524e9-7c64-42da-8041-d20073ee67d1.png">

Issue #2841, opened 2021-08-25
BETTER - Easy access to analysis report


RDmitchell:
**Is your feature request related to a problem? Please describe.**
Right now it is a bit convoluted to get from an Inventory Property to a BETTER analysis report of that property. 

You need to:
* From the Property List, click on the Detail view icon
* From the Property Detail view, click on the Analyses link at the top
* This takes you to the Org/Analyses screen, where you see a list with the Run ID
* Click on the Run ID link and you will then see first the BETTER Portfolio Report, and then the BETTER Building report

**Describe the solution you'd like**
Many less clicks, like 1 or 2, would be ideal

Here are a few ideas:
* **Add the Run ID field as a link to the Property List View**
  * In the Property List, have one of the BETTER fields be the Run ID, not just the Analysis ID. 
![image](https://user-images.githubusercontent.com/6314950/130723009-c8ae0c13-7ec7-4205-b084-c19a10e0b4d9.png)
  * That ID could be a link directly to the Building Report HTML file -- the first report to display would be the building report, not the Portfolio report
![image](https://user-images.githubusercontent.com/6314950/130723200-53770960-16b0-498d-abff-01c795203137.png)

* **Add the Run ID field as a link in the Property Detail View**
  * In the Property Detail view, have a link to the BETTER analysis
![image](https://user-images.githubusercontent.com/6314950/130723746-8236689b-0653-4349-bda8-72785528e4bf.png)
  * That would go directly to the Building Report HTML file, as shown above

* **Embed the BETTER Building Report directly into the Property Detail view, possibly after the Measures and Scenarios;**
![image](https://user-images.githubusercontent.com/6314950/130724282-65f7bd2d-45cb-44d3-a64b-a2a109438106.png)

RDmitchell:
These were just a few ideas. I'm sure all of you will come up with even better ones !

Ryoken:
The most recently run analysis will now show up on the top of the inventory detail page:

<img width="1138" alt="Screen Shot 2021-09-03 at 10 55 33 AM" src="https://user-images.githubusercontent.com/564231/132041526-13a0fd3b-855d-45f4-8b84-b49132c2635c.png">

In addition, the relevant calculated values also appear in the inventory list:

<img width="1141" alt="Screen Shot 2021-09-03 at 10 57 28 AM" src="https://user-images.githubusercontent.com/564231/132041653-8dbe8d35-1ba0-48b1-a9ac-9ab63dc668c6.png">

Issue #2740, opened 2021-07-13
Add ability for users to view analysis results in inventory list view


macintoshpie:
After a user runs an analysis, we might pull out particular data of interest. E.g. if a user runs an EUI analysis, they care about the EUI value. A user should be able to see this value in the inventory list view (e.g. as a column).

Implementation Notes:
- We've discussed adding the parsed result(s) to the cycle's property (ie PropertyState). This would probably be a part of the post-processing (e.g. finish_analysis for the Bsyncr pipeline)

macintoshpie:
This has been done for the BETTER analysis already by adding a postprocessing step (check the better pipeline file). This ticket should focus on ESTCP specific analyses

macintoshpie:
We are still not sure what the best solution is, here are some ideas we've come up with so far:
- just add it to the property state and save it
- use audit log to update the property state so that users can see what the old value was
- create an additional field for tracking analysis results in the property state

E.g. consider an analysis that calculates EUI. Should it just update the EUI? what if there was already an EUI imported from ESPM? Does SEED overwrite it? Or do we create a new, analysis-specific, column for storing this? Or do we create some datastructure (jsonfield) which keeps track of analysis results on its own?

E.g. consider an analysis which creates a totally unique value (ie there's no "blessed" SEED column for it, and it would probably never be imported/created by a user themselves).

RDmitchell:
@macintoshpie 

A few comments:
* Don't overwrite an EUI from ESPM with an EUI from BETTER -- use field names to keep them separate. 
* If there are BETTER values that are not part of the structured SEED db, just make then Extra Data fields which the user can decide whether or not to display
* Don't use the structured SEED fields as a way to standardize -- if a user wants to standardize, they should probably do a BEDES mapping, which will probably put all of the fields into Extra Data

It might make sense to have a separate tab (?) in the Inventory List view for BETTER (or other analysis results) with BETTER values having their own field names, and the same rules for being able to select fields to view from all three tabs (just like now you can select fields from both tax and property to display in the list view of either).

This seems like a big design topic that probably isn't really suited to discussion just in a Github issue. 

I recommend that we have a design call or two about this, possibly talking to various users who want to use BETTER -- San Francisco, Berkeley (?), Judicial Council of California, etc, to see what their preferences are.

PR #2854, opened 2021-09-01
2740 feat/property analyes results

Ryoken:
#### What's this PR do?
- Saves EUI analysis results to the property_state's extra_data
- Displays analysis results in a compact card format
- Show the most recently completed analysis on the property detailed view

Do all the analyses

https://github.com/SEED-platform/seed/issues/2740
https://github.com/SEED-platform/seed/issues/2846
https://github.com/SEED-platform/seed/issues/2841

Screen Shot 2021-09-01 at 12 46 33 PM
Screen Shot 2021-09-01 at 12 46 43 PM


RDmitchell:
@Ryoken -- will it be possible to hide these cards if the user doesn't want to view them?

Ryoken:
> @Ryoken -- will it be possible to hide these cards if the user doesn't want to view them?

The cards replaced the table for the property>analyses page, but we could put in a way to hide the card on the property detail page

seed/analysis_pipelines/eui.py
```diff
@@ -219,15 +220,26 @@ def _run_analysis(self, meter_readings_by_analysis_property_view, analysis_id, p
...
+        eui = _calculate_eui(meter_readings_by_analysis_property_view[analysis_property_view_id], property_state.gross_floor_area.magnitude)
         analysis_property_view.parsed_results = {
-            'EUI': _calculate_eui(meter_readings_by_analysis_property_view[analysis_property_view_id], property_state.gross_floor_area.magnitude),
+            'EUI': eui,
             'Total Yearly Meter Reading': sum(meter_readings_by_analysis_property_view[analysis_property_view_id]),
             'Gross Floor Area': property_state.gross_floor_area.magnitude
         }
         analysis_property_view.save()
+        property_state.extra_data.update({'analysis_eui': eui})
+        property_state.save()
```
macintoshpie, line 242: I think this might not be the property state whose extra data we want to update. This changes the AnalysisPropertyView's PropertyState. The one we want to update is the PropertyView's PropertyState.

For reference, this is how things are modeled:
Screen Shot 2021-09-01 at 2 25 55 PM

The PropertyView is what's viewed in the inventory list, and is used for data quality checks. The AnalysisPropertyView is a snapshot (copy) of a PropertyView for usage in the analysis. So we need to update the PropertyView's PropertyState we took a snapshot of with the analysis result

seed/models/properties.py
```diff
@@ -231,6 +231,17 @@ class PropertyState(models.Model):
...
+    # analyses_results = {
+    #     analysis_type: [ # ordered descending by timestamp
+    #         timestamp:
+    #         analysis_property_view_id:
+    #         data: { # depends on each analysis }
+    #         summary: # one-line summary for display
+    #     ]
+    # }
+    analyses_results = JSONField(default=dict, blank=True)
+
```
macintoshpie, line 244: Could we remove this and the migration for now? I do like this approach, but would like to hold off until we have the user discussion and time to chat about the pros and cons of different methods

seed/analysis_pipelines/eui.py
```diff
@@ -219,15 +220,26 @@ def _run_analysis(self, meter_readings_by_analysis_property_view, analysis_id, p
...
+        is_extra_data=True,
+        column_name="analysis_eui",
+        display_name="Analysis EUI",
+        organization=analysis.organization,
+        table_name='PropertyState',
+    )
+
     for analysis_property_view_id in meter_readings_by_analysis_property_view:
         analysis_property_view = AnalysisPropertyView.objects.get(id=analysis_property_view_id)
         property_state = PropertyState.objects.get(id=analysis_property_view.property_state.id)
```
macintoshpie, line 233: FYI I don't think you'd need to make this query, since `analysis_property_view.property_state` is the same thing as what you're getting

seed/models/analyses.py
```diff
@@ -70,6 +73,39 @@ def get_property_view_info(self, property_id=None):
             'cycles': list(analysis_property_views.values_list('cycle', flat=True).distinct())
         }
 
+    def get_highlights(self, property_id):
```
macintoshpie, line 76: nice! I like this

seed/views/v3/analyses.py
```diff
@@ -140,8 +140,7 @@ def retrieve(self, request, pk):
                 'message': "Requested analysis doesn't exist in this organization."
             }, status=HTTP_409_CONFLICT)
         serialized_analysis = AnalysisSerializer(analysis).data
-        property_view_info = analysis.get_property_view_info()
-        serialized_analysis.update(property_view_info)
+        serialized_analysis.update(analysis.get_property_view_info())
```
macintoshpie, line 143: I think we might want to add highlights here as well for consistency

RDmitchell:
> > @Ryoken -- will it be possible to hide these cards if the user doesn't want to view them?
> 
> The cards replaced the table for the property>analyses page, but we could put in a way to hide the card on the property detail page

Maybe we should wait and get feedback from users on our call tomorrow (if that functionality will be on dev1 by then) and get their reaction before making any changes to how it is working now.

macintoshpie (changes requested):
This looks _really_ good! In addition to the comments, one thing I found when viewing the card form the detail page was that clicking "Full Analysis" resulted in this error:
Screen Shot 2021-09-01 at 2 37 37 PM

Also when hovering over the full analysis link on the Details page it's not showing the "hand" icon for the cursor.

I ran into this issue when trying to view the analyses which had a better analysis as well as the EUI analysis:
```python
web_1          |   File "/seed/seed/models/analyses.py", line 103, in get_highlights
web_1          |     return [{'name': 'EUI', 'value': f'{extra_data["analysis_eui"]:,.2f}'}]
web_1          | KeyError: 'analysis_eui'
```
I think it relates to the comment I left about updating the PropertyView's PropertyState instead of the AnalysisPropertyView's propertystate.

macintoshpie (approved):
Awesome! This looks so nice.

We might later consider linking result highlights to AnalysisPropertyViews instead of Analyses -- it's a bit more of a one-to-one result currently, we probably need to chew on this.

Issue #2835, opened 2021-08-24
BETTER Reports - Units are defaulting to SI - Can they be set to either SI or IP?


RDmitchell:
**Is your feature request related to a problem? Please describe.**
Right now the BETTER reports that are obtained from the BETTER analysis in SEED are in SI units.

For example, the Gross Floor Area in the BETTER report defaults to meters squared
![image](https://user-images.githubusercontent.com/6314950/130698107-51306d94-9eff-4fa8-aaab-acc1a3528fd3.png)

While the SEED Property record has the units in foot squared, which came from Portfolio Manager.
![image](https://user-images.githubusercontent.com/6314950/130698255-f1719f68-4b17-46e8-b6ba-f893ae65afbd.png)

**Describe the solution you'd like**
See if it is possible to have BETTER generate the report in either SI or IP units, and when we send the request to BETTER to do the analysis, we specify the units we want, which I think in general, at least for now, we can default to IP. 

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

PR #2855, opened 2021-09-01
refactor(better): use IP units in standalone HTML report

macintoshpie:
BETTER HTML reports have unitted values. Previously those defaulted to SI, but an update to BETTER allows the API user to specify what they want

#### What's this PR do?
- request IP units from BETTER for standalone html reports

NOTE: future work would be to look at the seed org settings and request the units according to their settings.

Run a better analysis, the report should use IP units.

#2835

### DIFF
--- a/seed/analysis_pipelines/better/client.py
+++ b/seed/analysis_pipelines/better/client.py
@@ -185,9 +185,10 @@ class BETTERClient:
             'accept': 'text/html',
             'Authorization': self._token,
         }
+        params = {'unit': 'IP'}
 
         try:
-            response = requests.request("GET", url, headers=headers)
+            response = requests.request("GET", url, headers=headers, params=params)
             if response.status_code != 200:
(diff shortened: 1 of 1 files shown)


RDmitchell:
@macintoshpie -- and the units will be controlled by the option selected for the "Thermal Conversion Assumption", right? So if a user did want SI for the BETTER results, they would select "Canada". 

image

RDmitchell:
@macintoshpie -- ok -- should I an issue for future work so we don't forget ? I'm happy to do that if it's useful.

macintoshpie:
@RDmitchell that'd be great, thanks!

RDmitchell:
@macintoshpie -- I created a new issue but didn't assign it to a project.
https://github.com/SEED-platform/seed/issues/2856

nllong (approved):
looks bomber!

PR #2860, opened 2021-09-07
add explicit media path for dev users

nllong:
Media is often served by NGINX, however in dev mode the path was not found.

#### What's this PR do?
Add in the media path to `dev.py`. Path points to the users local `media` folder.

Run a BETTER analysis locally (not in docker) with the dev settings

 N/A


macintoshpie (approved):
🐮

PR #2862, opened 2021-09-07
fix: only use keys that aren't none

Ryoken:


macintoshpie:
The issue this branch was resolving (#2867 ) has been resolved by #2870

Issue #2867, opened 2021-09-13
Fix failing analysis tests


macintoshpie:
## task
- fix failing tests found here: https://github.com/SEED-platform/seed/runs/3537559397

macintoshpie:
No user testing needed 🕺

PR #2870, opened 2021-09-13
Fix/analysis views

macintoshpie:
A few tests broke recently

#### What's this PR do?
- fixes broken test / analyses page by handling case where a property is deleted after running an analysis
- deduplicate fetching PropertyView for analyses by creating a method on AnalysisPropertyView (`get_property_views`).
- fixes issue where links to property views are broken on the /analyses/<id> page by adding the `original_views` data to scope.
- removes link to property detail page if the property view no longer exists (user deleted it after running analysis)

- tests should pass
- should be able to click an address of a property on the /analyses/<id> page and it should go to the property detail page without errors.
- run an analysis, then delete the property from the inventory list. You should not be able to "click" the address to view the detail page.

#2867 

#### Screenshots
Example property where it was deleted after running analysis
Screen Shot 2021-09-13 at 4 45 11 PM

### DIFF
--- a/seed/models/analysis_property_views.py
+++ b/seed/models/analysis_property_views.py
@@ -90,3 +90,32 @@ class AnalysisPropertyView(models.Model):
                     ))
 
         return analysis_property_view_ids, failures
+
+    @classmethod
+    def get_property_views(cls, analysis_property_views):
+        """Get PropertyViews related to the AnalysisPropertyViews. If no PropertyView
+        is found for an AnalysisPropertyView, the value will be None for that key.
+
+        :param analysis_property_views: list[AnalysisPropertyView]
+        :return: dict{int: PropertyView}, PropertyViews keyed by the related AnalysisPropertyView id
+        """
+        # build a query to find PropertyViews linked to the canonical property and cycles we're interested in
+        property_view_query = models.Q()
+        for analysis_property_view in analysis_property_views:
+            property_view_query |= (
+                models.Q(property=analysis_property_view.property)
+                & models.Q(cycle=analysis_property_view.cycle)
+            )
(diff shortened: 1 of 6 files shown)


seed/models/analysis_property_views.py
```diff
@@ -90,3 +90,32 @@ def batch_create(cls, analysis_id, property_view_ids):
...
+
+        # get original property views keyed by canonical property id and cycle
+        property_views_by_property_cycle_id = {
+            (pv.property.id, pv.cycle.id): pv
+            for pv in property_views
+        }
+
+        return {
+            # we use .get() here because the PropertyView might not exist anymore!
+            apv.id: property_views_by_property_cycle_id.get((apv.property.id, apv.cycle.id), None)
```
macintoshpie, line 119: This is the line that technically fixes the broken test. Test was failing b/c we assumed every AnalysisPropertyView could successfully find a PropertyView with the same cycle and canonical property, yet we weren't creating the PropertyViews in the test setup

seed/static/seed/js/controllers/analysis_controller.js
```diff
@@ -33,6 +33,7 @@ angular.module('BE.seed.controller.analysis', [])
       $scope.users = users_payload.users;
       $scope.views = views_payload.views;
       $scope.view_id = $stateParams.view_id;
+      $scope.original_views = views_payload.original_views;
```
macintoshpie, line 36: fix linking to detail page

seed/models/analysis_property_views.py
macintoshpie: Note that this bug was also visible to site users when they go to one of the analyses pages and a property was deleted

Issue #2863, opened 2021-09-10
Make analyses page more easily accessible


nllong:
## context
Users view the list of all analyses they've run (or their org has run) by going to the Org page then clicking analyses.

## problem
This is inconvenient. Users should be able to access their analyses more easily.

## task
- decide where the link should go and implement it

macintoshpie:
@nllong @Ryoken which Analysis are they navigating to? Is this adding it to the collapsable menu?

macintoshpie:
One idea is to create a top level nav item in the side bar:

![Screen Shot 2021-09-13 at 11 48 26 AM](https://user-images.githubusercontent.com/18518728/133132167-aca58371-f1dd-47ae-8bd2-9ced882b0d4d.png)

Another idea is to make more of these org sub-pages accessible as well. In that case, we might consider refactoring the "+" icon in the upper right to link to these pages (including the analysis page). See #2014 
Another general solution to providing users access to subpages could be adding sub-items in the sidebar under the organization link.

macintoshpie:
From discussion:
- if no analyses, maybe link back to inventory page with note of how to create analysis

Ryoken:
Link on main nav:
<img width="1198" alt="Screen Shot 2021-09-14 at 4 40 20 PM" src="https://user-images.githubusercontent.com/564231/133344156-0c6295ba-873e-42d2-a5e9-7754db400c61.png">

Clicking on an brings you to it's page with a link back to the analyses list:
<img width="1196" alt="Screen Shot 2021-09-14 at 4 40 29 PM" src="https://user-images.githubusercontent.com/564231/133344159-fe7f4883-12f7-4b68-8f5b-a2dcac27d503.png">

Clicking on a run id brings you to the detail page with a link back to the analysis:
<img width="1197" alt="Screen Shot 2021-09-14 at 4 40 38 PM" src="https://user-images.githubusercontent.com/564231/133344161-405e5395-b71e-4dc1-bee7-e16e437228b8.png">

nllong:
This looks great. @RDmitchell -- any comments on the location of the Analyses button?

Ryoken:
> * if no analyses, maybe link back to inventory page with note of how to create analysis

Added this comment to https://github.com/SEED-platform/seed/issues/2866

PR #2874, opened 2021-09-14
refactor: moved analyses link

Ryoken:
#### What's this PR do?
- moves analyses link to sidebar
- adjusts navigation of analisi pages

Click through nav with analyses.

https://github.com/SEED-platform/seed/issues/2863

Link on main nav:
Screen Shot 2021-09-14 at 4 40 20 PM

Clicking on an analyses brings you to it's page with a link back to the analyses list:
Screen Shot 2021-09-14 at 4 40 29 PM

Clicking on a run id brings you to the detail page with a link back to the analysis:
Screen Shot 2021-09-14 at 4 40 38 PM


seed/templates/seed/_sidebar.html
```diff
@@ -29,6 +29,10 @@
             <div class="item_name" translate>Organizations</div>
             <div ng-show="logged_in" class="badge badge_menu">{$ organizations_count | number:0 $}</div>
         </a>
+        <a id="sidebar-accounts" ng-class="{item: true, active: is_active('/analyses/' + menu.user.organization.id), 'disabled-item': !logged_in}" title="{$:: 'Analyses' | translate $}" href="{% url "seed:home" %}#/analyses/{$:: menu.user.organization.id $}">
```
macintoshpie, line 32: I'm not sure how to fix this, but it looks like the org id isn't being updated when I change update it so the analyses link always goes to the analyses for the same org until I refresh the page.

Screen Shot 2021-09-15 at 12 14 31 PM


I think there's some kind of refresh event happening when the user updates the selected org, not sure how to integrate it with the sidebar though.

I think we also need to handle the case where the user is on the analyses page and they change the active org (it should refresh the analyses page for the new org). Again this is handled by some kind of refresh on other pages (e.g. try going to a property detail page then changing the org) but I don't know much else about it.

macintoshpie (commented):
Looks nice! One comment on refreshing the nav/page when the org changes. Also I think we need to update this link:
https://github.com/SEED-platform/seed/blob/2863-refactor/move-analyses-link/seed/static/seed/js/controllers/inventory_list_controller.js#L1059

seed/templates/seed/_sidebar.html
Ryoken: Good catch.  I'll take a look!

Ryoken: I've refactored the analyses pages to update when the org changes.  The org ID is no longer a URL param and the link behaves like the other nav links, getting org id from the user_service instead.

macintoshpie (approved):
💯 🔥

Issue #2873, opened 2021-09-14
Analysis cards show most recent highlights


macintoshpie:
## context
You can view all previous analyses for a property in the detail analyses tab

## problem
It looks like the highlights of the cards are only shown from the most recent analysis. E.g. here the EUI should be different on the cards. When you look at the analysis page you can see the EUIs are indeed different but the cards only show the most recent value.

![Screen Shot 2021-09-13 at 2 23 22 PM](https://user-images.githubusercontent.com/18518728/133338080-0de09b86-967b-4ebc-aad4-e70604e083d9.png)
![Screen Shot 2021-09-13 at 2 24 00 PM](https://user-images.githubusercontent.com/18518728/133338112-1e5d2ac7-a96a-427f-b212-a1595f1c669a.png)
![Screen Shot 2021-09-13 at 2 23 41 PM](https://user-images.githubusercontent.com/18518728/133338120-7fb13930-442e-406f-93b6-96a8e09ebe51.png)

PR #2875, opened 2021-09-14
fix: get analysis highlights from analysis results

macintoshpie:
SEED has analysis cards with highlights on some pages. The highlights of the cards were always the same and showed the most recent results.

#### What's this PR do?
- fixes issue of different analyses showing the same highlights.
  - this was fixed by pulling analysis results from the AnalysisPropertyView's parsed results instead of the PropertyView's PropertyState. PropertyState will always be updated with the most recent results, so pulling info from there will always get you the most recent data.
- added some formatting to the cards to show `N/A` for some highlights

Run an analysis twice for a property that should get different results (e.g. run one then edit gross floor area then run again). Verify the cards show different values.

#2873 

Screen Shot 2021-09-14 at 5 25 11 PM
Screen Shot 2021-09-14 at 5 25 45 PM

### DIFF
--- a/seed/models/analyses.py
+++ b/seed/models/analyses.py
@@ -88,7 +88,10 @@ class Analysis(models.Model):
 
         results = {}
         if property_id is not None:
-            results = self.analysispropertyview_set.filter(property=property_id).first().parsed_results
+            try:
+                results = self.analysispropertyview_set.get(property=property_id).parsed_results
+            except models.Model.DoesNotExist:
+                return []
         else:
             results = self.parsed_results
 
--- a/seed/models/analyses.py
+++ b/seed/models/analyses.py
@@ -6,10 +6,10 @@
 """
 from django.contrib.postgres.fields import JSONField
 from django.db import models
-from django.db.models import Q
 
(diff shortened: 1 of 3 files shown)


seed/models/analyses.py
```diff
@@ -75,46 +75,63 @@ def get_property_view_info(self, property_id=None):
...
+        :return: list[dict{}], a list of highlights as dictionaries, each including
+            a `name` and `value`
+        """
         if self.status < self.COMPLETED:
             return []
+
+        results = {}
         if property_id is not None:
-            analysis_property_view = self.analysispropertyview_set.filter(property=property_id).first()
+            results = self.analysispropertyview_set.filter(property=property_id).first().parsed_results
```
Ryoken, line 91: https://sentry.io/organizations/nrel/issues/2649369995/?project=1185731

Will this still be an issue after this refactor?  the analyses_property_view set could be None here.

macintoshpie: Hmm good catch, yeah we should handle that case

Ryoken (approved):

Issue #2865, opened 2021-09-10
Remove obj/dict/hash content from analysis card on inventory detail page


nllong:
See benchmark and savings data in this image:

![image](https://user-images.githubusercontent.com/1907354/132915514-c93af88d-a2b4-4df4-b22f-c9e4ed182955.png)

macintoshpie:
Personal opinion: don't store the config with nested dicts like we are currently for BETTER. This would fix this issue everywhere else. There's not really a reason that the config should look like:
```
{'min_r_squared': 0.6,
'benchmark_data': {'benchmark_data': 'DEFAULT'},
'savings_target': {'savings_target': 'CONSERVATIVE'},
'portfolio_analysis': True}
```

It should instead look like this:
```
{'min_r_squared': 0.6,
'benchmark_data': 'DEFAULT',
'savings_target': 'CONSERVATIVE',
'portfolio_analysis': True}
```

PR #2876, opened 2021-09-15
fix: removed nested config and changed broken min_model_r_squared key

Ryoken:
#### What's this PR do?
Removes nested config options and fixes a broken key that BETTER required.

Run a better analysis and ensure the config looks correct on the property detail page (see original ticket).

https://github.com/SEED-platform/seed/issues/2865

### DIFF
--- a/seed/static/seed/js/controllers/inventory_detail_analyses_modal_controller.js
+++ b/seed/static/seed/js/controllers/inventory_detail_analyses_modal_controller.js
@@ -42,14 +42,14 @@ angular.module('BE.seed.controller.inventory_detail_analyses_modal', [])
       ];
 
       $scope.better_savings_targets = [
-        {savings_target: 'CONSERVATIVE'},
-        {savings_target: 'NOMINAL'},
-        {savings_target: 'AGGRESSIVE'}
+        'CONSERVATIVE',
+        'NOMINAL',
+        'AGGRESSIVE'
       ];
 
       $scope.better_benchmark_options = [
-        {benchmark_data: 'DEFAULT'},
-        {benchmark_data: 'GENERATE'}
+        'DEFAULT',
+        'GENERATE'
       ];
 
       $scope.initializeAnalysisConfig = () => {
@@ -82,13 +82,6 @@ angular.module('BE.seed.controller.inventory_detail_analyses_modal', [])
           return;
         }
         $scope.waiting_for_server = true;
-        ['benchmark_data', 'savings_target'].forEach((key, index) => {
-          if (key in $scope.new_analysis.configuration) {
-            if (key in $scope.new_analysis.configuration[key]) {
-              $scope.new_analysis.configuration[key] = $scope.new_analysis.configuration[key][key];
-            }
-          }
-        });
         analyses_service.create_analysis(
           $scope.new_analysis.name,
           $scope.new_analysis.service,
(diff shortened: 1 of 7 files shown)


seed/static/seed/js/controllers/inventory_detail_analyses_modal_controller.js
```diff
@@ -81,7 +81,14 @@ angular.module('BE.seed.controller.inventory_detail_analyses_modal', [])
...
         }
-        $scope.waiting_for_server = true
+        $scope.waiting_for_server = true;
+        ['benchmark_data', 'savings_target'].forEach((key, index) => {
+          if (key in $scope.new_analysis.configuration) {
+            if (key in $scope.new_analysis.configuration[key]) {
+              $scope.new_analysis.configuration[key] = $scope.new_analysis.configuration[key][key];
+            }
+          }
+        });
```
macintoshpie, line 91: I'm wondering if there;s a way to avoid doing this. Could we tweak the analysis modal so that the ng-model doesn't create these nested values?

macintoshpie (changes requested):
Looking gewd. One comment on using ng-model. Also I noticed the analyses pages were doing some funky things with the configuration now:

Screen Shot 2021-09-15 at 1 44 51 PM

seed/static/seed/js/controllers/inventory_detail_analyses_modal_controller.js
Ryoken: AFAIK it's how the form is submitting the select fields but I'll look and see.

Ryoken:
> Looking gewd. One comment on using ng-model. Also I noticed the analyses pages were doing some funky things with the configuration now:
> 
> Screen Shot 2021-09-15 at 1 44 51 PM

That was fixed when this line was removed from the partial:
            <li ng-repeat="(key, value) in setting">{$:: key $} = {$:: value $}</li>

try force refreshing the page and see if it till does that?  It's only a display issue; that isn't actually saved on the model.

macintoshpie:
> That was fixed when this line was removed from the partial:
> <li ng-repeat="(key, value) in setting">{$:: key $} = {$:: value $}</li>
> try force refreshing the page and see if it till does that? It's only a display issue; that isn't actually saved on the model.

Looks like we do that in two other places I think:
https://github.com/SEED-platform/seed/blob/fix/simplify-config/seed/static/seed/partials/analyses.html#L47
https://github.com/SEED-platform/seed/blob/fix/simplify-config/seed/static/seed/partials/analysis_details.html#L22

seed/static/seed/js/controllers/inventory_detail_analyses_modal_controller.js
Ryoken: I had to change how the settings were stored on the controller by making them a list of keys instead of a list of objects to fix this.  A search said it wasn't being used anywhere else so I think this is safe.

seed/static/seed/js/controllers/inventory_detail_analyses_modal_controller.js
```diff
@@ -35,21 +35,21 @@ angular.module('BE.seed.controller.inventory_detail_analyses_modal', [])
...
+        'NOMINAL',
+        'AGGRESSIVE'
       ];
 
       $scope.better_benchmark_options = [
-        {benchmark_data: 'DEFAULT'},
-        {benchmark_data: 'GENERATE'}
+        'DEFAULT',
+        'GENERATE'
       ];
```
macintoshpie, line 53: Nice!

macintoshpie (approved):
Looks great! Thanks for cleaning up the analysis modal stuff too!

Issue #2886, opened 2021-09-17
Update values parsed from BETTER analysis


macintoshpie:
## context
After running a BETTER analysis, SEED pulls some results and links them to the property so they're viewable in the inventory list. SEED is pulling a few values including the measure recommendations, combined ghg reduction, combined cost savings

## task
- SEED should check if the analysis was successful for each model (electricity and nat gas), and pull out the individual cost savings and ghg reduction instead of combined
![image (5)](https://user-images.githubusercontent.com/18518728/133838996-c07e6def-c4e3-4f29-a2dd-91635a7e0844.png)

- if one of the models failed, the values associated with that energy type should be set to null on the property state
- also, for each model that failed, create an error message for that property

RDmitchell:
Can we also add under "Latest Message" something like like "Model Failed".

![image](https://user-images.githubusercontent.com/6314950/133846111-98fc7d1c-f0e5-418f-99e7-08631ba1a453.png)

macintoshpie:
👍  Good idea Robin. I'll add add that to the ticket

Issue #2884, opened 2021-09-16
Remove usage of address line 1 in BETTER analysis


macintoshpie:
## context
SEED currently uses (and requires) address line 1 from the property in the analysis.
BETTER doesn't actually require this, they just need city and state.

## task
- remove this requirement from the validation step for BETTER
- don't add address line 1 to the bsync file
- verify these changes are reflected in the BETTER use case: https://github.com/BuildingSync/use-cases/blob/main/BETTER/schematron/BETTER-1.0.0.sch

macintoshpie:
Update: BETTER currently tries to pull out this data and will cause a 500 if it doesn't exist. Waiting for this PR to get merged before we can actually remove it:
https://github.com/LBNL-ETA/better_lbnl/pull/72

nllong:
What is the status of this?

macintoshpie:
This is ready for review in a PR, @nllong I just tagged you to review if you have time now
https://github.com/SEED-platform/seed/pull/2887/files#diff-8143b1bdb8af06f8876b2526c1ff0608cbaf1e4c436ba87e2274d3ae6bd517d8L149

macintoshpie:
Donezo

Issue #2840, opened 2021-08-24
BETTER - show which records had 0 savings in the Analyses view


RDmitchell:
**Is your feature request related to a problem? Please describe.**
Right now it's a bit laborious to see which buildings have good (non-zero) results
![image](https://user-images.githubusercontent.com/6314950/130705066-22c4a5ff-02bf-4602-98ba-060853e32fdc.png)

**Describe the solution you'd like**
It would be nice to show in the Analyses list, what the basic savings values are, so it is easy to see the ones that are zero
![image](https://user-images.githubusercontent.com/6314950/130704917-7e8a2dc1-f78e-4990-8293-d7059021a15f.png)

**Describe alternatives you've considered**
could also make a default DQ rule to label all the records with 0 savings -- will make another issue for that. 

**Additional context**
n/a

macintoshpie:
New (slightly different) plan is to create a Message for users when we detect that BETTER failed to create a model (for elec or nat gas). We can discuss adding the "highlight values" (savings) to this table at a later time

RDmitchell:
instance: dev1
SHA: 4d262e985

This is what is shows, which is ok for now, but it seems like a nice addition would be to just show the derived fields as columns with their results, so the user could quickly see those results in this view. 

![image](https://user-images.githubusercontent.com/6314950/140236545-e7951480-959c-44c7-8fe9-e3925471cbd3.png)

PR #2887, opened 2021-09-20
Refactor/misc better improvements

macintoshpie:
#### What's this PR do?
- add energy, cost, and ghg savings specific to electricity and fuel (nat gas)
  - if BETTER reports that it failed to create a model for the energy resource (electricity or nat gas) then set those energy-specific savings to NULL. Also, create a warning message for the property (see screenshot)
- remove street address from the BuildingSync file sent to BETTER (i.e. no longer require that SEED properties have address line 1)

- Run a better analysis
- update your column profile to include the new BETTER columns mentioned above (e.g. BETTER Potential Electricity Energy Savings)
- verify that if the analysis failed to create a model for natural gas or electricity (look at the HTML page to see) then the energy-specific savings in the inventory list should be empty.

#2884 
#2886 
#2840 

Screen Shot 2021-09-20 at 2 15 41 PM
Screen Shot 2021-09-27 at 11 28 02 AM


seed/analysis_pipelines/better/pipeline.py
```diff
@@ -311,21 +311,48 @@ def _process_results(self, analysis_id):
     ]
 
     # gather all columns to store
+    BETTER_VALID_MODEL_E_COL = 'better_valid_model_electricity'
+    BETTER_VALID_MODEL_F_COL = 'better_valid_model_fuel'
     column_data_paths = [
         ExtraDataColumnPath(
-            'better_cost_savings_combined',
-            'BETTER Potential Cost Savings (USD)',
-            'assessment.assessment_energy_use.cost_savings_combined'
```
macintoshpie, line 318: @nllong these changes removes the combined cost, ghg, energy usage columns from the inventory list -- is that ok? we're now pulling out electricity/nat-gas specific values.

RDmitchell:
@macintoshpie -- what values are removed and what are added to the Inventory list?

macintoshpie:
@RDmitchell that's fair, I did remove some of the values which you see highlighted there in your screenshot. I will add the back!

RDmitchell:
@macintoshpie -- thanks ... just seems like it would be good to be able to match the values in the inventory list with the values on the report.

nllong (approved):
This looks great. Imported our test data and ran it through BETTER. 

image

I noticed the header of the BSync files still show BSync 2.2, is that correct?

```
<auc:BuildingSync xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:auc="http://buildingsync.net/schemas/bedes-auc/2019" xsi:schemaLocation="http://buildingsync.net/schemas/bedes-auc/2019 https://raw.github.com/BuildingSync/schema/v2.2.0/BuildingSync.xsd" version="2.2.0">
```

macintoshpie:
> I noticed the header of the BSync files still show BSync 2.2, is that correct?

Yeah that's what we generate, we even have a test which validates the bsync file we generate against the 2.2 schema. We should be fine bumping to 2.3 as well. Is that something we'd want to do?

nllong:
> > I noticed the header of the BSync files still show BSync 2.2, is that correct?
> 
> Yeah that's what we generate, we even have a test which validates the bsync file we generate against the 2.2 schema. We should be fine bumping to 2.3 as well. Is that something we'd want to do?

It would be great to bump to 2.3 if we have time today... whatcha think?

Issue #2904, opened 2021-09-28
Unable to run BSyncr analysis


macintoshpie:
**Describe the bug**
You cannot run BSyncr analyses

**Expected Behavior**
You should be able to run bsyncr analyses
<img width="569" alt="Screen Shot 2021-09-28 at 11 05 47 AM" src="https://user-images.githubusercontent.com/18518728/135124267-2c870d32-b058-4f71-b1df-5d5cde8e306c.png">

<img width="1207" alt="Screen Shot 2021-09-28 at 11 05 15 AM" src="https://user-images.githubusercontent.com/18518728/135124298-97ba4945-0177-4529-b375-08a92583a2b0.png">

**Actual Behavior**
SEED complains when you try to create the analysis:

**Steps to Reproduce**
Try to run a BSyncr analysis

**Instance Information**
SHA: 2.11.0.6247533c9

**Additional Context**

RDmitchell:
Instance: dev1
SHA: f3e9a4a4d
Org: LBNL 405

I ran a BSyncr analysis on a record from the JCC anonymized data that had good meter data, and it seems to generate the XML file and the graph (not exactly sure what the graph is but it exists !!), so I will close this issue
![image](https://user-images.githubusercontent.com/6314950/135353016-f56130f4-d14c-48f4-81c3-adb3dd75b504.png)

The XML file downloaded without issue
![image](https://user-images.githubusercontent.com/6314950/135353117-0d3b9fc9-37a9-4c95-82fe-bacffda7deb5.png)
As well as the PNG file
![image](https://user-images.githubusercontent.com/6314950/135353167-c9c977ed-a5e5-40c6-aa9e-7fdf91c2e37f.png)

Issue #2903, opened 2021-09-28
Analysis page fails to update analysis status after a step is "completed"


macintoshpie:
**Describe the bug**
If an analysis is in an "active" state (e.g. running) SEED should poll the backend to see when the analysis status changes.

**Expected Behavior**

Polling should continue until the analysis reaches a terminal state (failed, stopped). Also each time the analysis status changes it should be reflected on the page

**Actual Behavior**
It appears that SEED stops polling after the status changes once, and the page/item is not updated...

**Steps to Reproduce**
Start an analysis, go to analysis page. It will probably show "running" with a spinning icon. If you watch the Network tab you'll see it eventually gets 100% complete, but the analysis status on the page isn't updating...

**Instance Information**

SHA: 2.11.0.6247533c9

**Additional Context**
This was working previously, see #2853 
It's possible that when we moved the analysis page from under the organization page that something got tweaked.

RDmitchell:
instance: dev1
SHA: f3e9a4a4d
Org: LBNL 405

This seems to be working properly now; the Run Status shows "Creating", "Running" and "Completed" without the user having to refresh the browser. 
![image](https://user-images.githubusercontent.com/6314950/135343353-06eaa8e6-089c-4c30-b592-5fae21d3f39a.png)
![image](https://user-images.githubusercontent.com/6314950/135343389-909323d8-a3cc-4324-aeab-fb089221d86a.png)
![image](https://user-images.githubusercontent.com/6314950/135343418-80591d56-4d1e-473f-90b5-c3f098f0d4d5.png)

And for a set with more records
![image](https://user-images.githubusercontent.com/6314950/135343762-13e0bf8d-acd4-4375-b7a6-65be68f608a2.png)
![image](https://user-images.githubusercontent.com/6314950/135343782-1eca8c48-e639-497e-b49c-05afac29aa9c.png)
![image](https://user-images.githubusercontent.com/6314950/135343825-0028165e-56d6-452b-8ae8-42e759c5e4ed.png)

PR #2905, opened 2021-09-28
Fix/misc analyses bugs

macintoshpie:
See issues

#### What's this PR do?
- fix creation of BSyncr analyses
- fix polling of analyses from frontend
- add error handling for analysis polling (no linked issue, but found it when testing)

- run an analysis, on one of the analysis pages, the status of the analysis on the page should be updated when the status changes (e.g. when it goes from running to complete)

#2904 
#2903

### DIFF
--- a/seed/static/seed/js/controllers/analyses_controller.js
+++ b/seed/static/seed/js/controllers/analyses_controller.js
@@ -23,7 +23,7 @@ angular.module('BE.seed.controller.analyses', [])
       analyses_service,
       Notification,
     ) {
-      $scope.org = organization_payload.organization;
+      $scope.org = organization_payload;
       $scope.auth = auth_payload.auth;
       $scope.analyses = analyses_payload.analyses;
       $scope.users = users_payload.users;
--- a/seed/static/seed/js/services/analyses_service.js
+++ b/seed/static/seed/js/services/analyses_service.js
@@ -197,6 +197,12 @@ angular.module('BE.seed.service.analyses', [])
                           })
                       }
                     })
+                    .catch(data => {
+                      // yikes, something went wrong. Let the caller know the status
+                      // probably changed and let's bail
+                      status_update_callback(id)
+                        .then(() => no_current_task_callback(id))
+                    })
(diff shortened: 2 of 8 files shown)


seed/static/seed/partials/inventory_detail_analyses_modal.html
```diff
@@ -22,7 +22,7 @@ <h4 class="modal-title" id="manageLabelsModalLabel">Run Analysis</h4>
 
                 <div class="form-group" ng-if="new_analysis.service == 'BSyncr'">
                     <label class="control-label sectionLabel">BSyncr Model Selection</label>
-                    <select class="form-control" ng-model="new_analysis.configuration" ng-dropdown required>
+                    <select class="form-control" ng-model="new_analysis.configuration.model_type" ng-dropdown required>
```
macintoshpie, line 25: previously we were overwriting the entire configuration object which isn't what we want

seed/static/seed/js/services/analyses_service.js
```diff
@@ -197,6 +197,12 @@ angular.module('BE.seed.service.analyses', [])
                           })
                       }
                     })
+                    .catch(data => {
+                      // yikes, something went wrong. Let the caller know the status
+                      // probably changed and let's bail
+                      status_update_callback(id)
+                        .then(() => no_current_task_callback(id))
+                    })
```
macintoshpie, line 205: handle the case where we fail the analysis (and as a result the progress data) -- the called method will reject the promise which we weren't doing anything with before.

seed/static/seed/js/controllers/analyses_controller.js
```diff
@@ -23,7 +23,7 @@ angular.module('BE.seed.controller.analyses', [])
       analyses_service,
       Notification,
     ) {
-      $scope.org = organization_payload.organization;
+      $scope.org = organization_payload;
```
macintoshpie, line 26: The data passed in for organization_payload has changed -- it's no longer nested

Ryoken (approved):

Issue #2877, opened 2021-09-15
Tweak EUI Analysis


macintoshpie:
## context
SEED can run an EUI analysis on properties

## tasks
- we want the final EUI to be the average annual kBtu / sqft
  - verify we're doing the annual average
  - verify the meter data we're using is in kBtu (I think meters are stored in kBtu already, so make sure we're not modifying the readings: https://github.com/SEED-platform/seed/blob/00b7322388fcfe25683804d48477dd94f446d000/seed/models/meters.py#L148)
- update the highlights card to have `kBtu / sqft` or something similar for units
- update the extra_data column from `Analysis EUI` to something like `Fractional EUI (kBtu / sqft)` or something similar

macintoshpie:
For testing:
- check the EUI analysis card looks good and has units on the property details page
- Verify the column is now Fractional EUI

RDmitchell:
instance: dev1
SHA: 4d262e985

EUI card looks good, has units on property details page
![image](https://user-images.githubusercontent.com/6314950/140236774-873a5b4b-d89f-4575-8c61-18ade087037e.png)

Fields are 
* Fractional EUI 
* EUI coverage
![image](https://user-images.githubusercontent.com/6314950/140236881-25fc8db6-4a0e-4988-b003-b8c4056e36d7.png)

PR #2914, opened 2021-09-29
2877 refactor/adjust eui

Ryoken:
#### What's this PR do?
- EUI analysis uses 1 year of meter readings prior to the most recent reading, regardless of meter reading coverage
- use all meters
- meter readings that overlap reject that meter
- calculate annual coverage as a percentage of how much of that year was covered by meter readings
- display units and coverage on highlight card

Run an EUI analysis on a property with meter readings

https://github.com/SEED-platform/seed/issues/2877


seed/analysis_pipelines/eui.py
```diff
@@ -207,40 +219,49 @@ def _run_analysis(self, meter_readings_by_analysis_property_view, analysis_id):
...
         eui = _calculate_eui(meter_readings, area)
 
         analysis_property_view.parsed_results = {
-            'EUI': eui,
-            'Total Yearly Meter Reading': sum(meter_readings),
-            'Gross Floor Area': area
+            'Fractional EUI (kBtu/sqft)': eui['eui'],
+            'Annual Coverage %': eui['coverage'],
+            'Total Annual Meter Reading (kBtu)': eui['reading'],
+            'Gross Floor Area (sqft)': area
```
macintoshpie, line 258: noiiice 💯

seed/analysis_pipelines/eui.py
```diff
@@ -59,73 +62,83 @@ def _get_valid_meters(property_view_ids):
...
+    total_reading = 0
+    total_time = 0
+    for meter_id in meter_readings:
+        total_reading += meter_readings[meter_id]['reading']
+        total_time += meter_readings[meter_id]['time']
+    return {
+        'eui': round(total_reading / gross_floor_area, 4),
+        'reading': total_reading,
+        'coverage': 100 - round(100 * (TIME_PERIOD.total_seconds() - total_time) / TIME_PERIOD.total_seconds())
+    }
```
macintoshpie, line 141: I think I might have figured out why we saw > 100% for coverage on some properties, there could be two meters which report overlapping reading intervals, but wouldn't get filtered out by the "overlapping_meters" check above b/c they're separate meters
Screen Shot 2021-10-04 at 4 29 17 PM

I think including both of these meters and their readings is actually OK -- a building could have multiple separate meters. The tricky part now is handling the total time...

Maybe the simplest approach could be to keep a hash of datetime days (at midnight or something) from the meter readings. Then count the number of keys `len(meter_days_hash.keys())`... not perfectly accurate but pretty good and not too resource intensive/complex

Alternatively we could sort _all_ readings, and add up the diffs for non-overlapping days. Interested to hear what you think though

macintoshpie (commented):
It works! Nice job updating the highlights card as well. We found one bug when chatting with Sean which should get ironed out, and I had one comment about migrations but everything else looks great!

Also it worked on the ESTCP sample property which is 🏀 er

seed/analysis_pipelines/eui.py
```diff
@@ -59,73 +62,95 @@ def _get_valid_meters(property_view_ids):
...
+        meter_reading.end_time.year,
+        meter_reading.end_time.month,
+        meter_reading.end_time.day,
+    )
+
+    all_days = []
+    current_day = start
+    while current_day <= end:
+        all_days.append(current_day)
+        current_day += relativedelta.relativedelta(days=1)
```
macintoshpie, line 122: This appears to be a pretty inefficient algorithm, but should be fine for now. Technically constant time since all meter readings must fit within a year (365 days).

seed/tests/test_analysis_pipelines.py
```diff
@@ -1030,30 +1029,6 @@ def test_valid_meters(self):
         self.assertDictEqual(errors_by_property_view_id, {})
         self.assertNotEqual(meter_readings_by_property_view, {})
 
-    def test_overlapping_meters(self):
```
macintoshpie: I ended up removing this validation -- it is unlikely (if not impossible) for an individual meter in SEED to have overlapping reading intervals. I think overlapping readings for an individual meter would mean SEED has an internal data integrity issues

macintoshpie:
After some tweaks, it looks like we're getting a consistent report for coverage (i.e. values <= 100%)
Screen Shot 2021-10-06 at 12 25 36 PM
Screen Shot 2021-10-06 at 1 00 15 PM

Issue #2917, opened 2021-09-30
BETTER: fails to create Buildings


macintoshpie:
## context
I was trying to run a better analysis on a property which has a lot of 15 minute interval meter readings. SEED successfully creates the file for the analysis, but then we error out when trying to create the property in BETTER (this is the log from SEED in the BETTER pipeline):
```json
{
   "analysis_id":5,
   "analysis_property_view":2,
   "user_message":"Unexpected error from BETTER service. Please confirm your Organization's API token is correct and try again or contact the SEED administrators.",
   "debug_message":"Failed to create building for analysis property view 2: BETTER service could not create building with the following message: ('Connection aborted.', OSError(\"(32, 'EPIPE')\"))",
   "exception":"None"
}
```

SEED is successfully creating the BuildingSync file, but it is quite large, 25MB. Since BETTER seems to be working for other properties with less meter data, I think the issue is that BETTER can't handle these large files

## task
I _think_ this issue can be fixed for my current use case by turning the 15minute readings into monthly readings (resulting in way less readings in the BSync file). I'm not sure how we should solve this in the long term though. One consideration is seeing if minimizing our bsync files results in significant space savings

PR #2919, opened 2021-10-04
Refactor/better aggregate meters

macintoshpie:
NREL's ESTCP project includes sample data with spotty meter data which doesn't work well with BETTER by default. There were a few issues with the data that presented new challenges communicating with BETTER:
  1. There were _a lot_ of meter readings and the BSync file created ended up being much too large (see the referenced issue) because SEED is currently dumping meters as they are into the file. We needed to reduce the readings included in the BSync file
  2. The data was pretty incomplete -- there were some missing months and BETTER requires at least 12 _continuous_ months of readings (it assumes usage of 0 for missing months which wrecks the models)

#### What's this PR do?
Adds an option to the BETTER analysis to "preprocess" meters. This should probably be just a temporary implementation because not much thought/knowledge about meters was put into it.
This "preprocessing" basically:
- bins meter readings into monthly readings (i.e. aggregate readings or disperse them)
  - e.g. if you have multiple 15-minute-interval readings it will sum them up for the month
  - e.g. if a reading _spans_ or _straddles_ a month, the algorithm splits the energy usage according to the fraction of the time the reading covers a particular month. This is a big ⚠️ that this is not a great/perfect solution; it assumes even distribution of usage throughout the meter reading interval which is likely not true (especially for readings that span a seasonal change)
- drops outliers (was experiencing some weird months due to binning - especially on the starting/ending boundaries)
- interpolates missing months

Run better analysis with and without preprocessing. It should work.

#2917 

New option on BETTER analysis
Screen Shot 2021-10-01 at 3 07 25 PM

Description of new option
Screen Shot 2021-10-01 at 3 07 34 PM

### DIFF
--- a/seed/analysis_pipelines/better/client.py
+++ b/seed/analysis_pipelines/better/client.py
@@ -19,6 +19,24 @@ class BETTERClient:
     def __init__(self, token):
         self._token = f'Token {token}'
 
+    def test_token(self):
+        """Returns true if token is valid
+
+        :return: bool
+        """
+        # TODO: get BETTER to create a "token" endpoint for testing
+        url = f'{self.API_URL}/weathers/'
+        headers = {
+            'accept': 'application/json',
+            'Authorization': self._token,
+        }
+
+        try:
+            response = requests.request("GET", url, headers=headers)
+            return response.status_code != 401
+        except Exception:
+            return False
+
     def get_buildings(self):
         """Get list of all buildings
 
--- a/seed/analysis_pipelines/better/pipeline.py
+++ b/seed/analysis_pipelines/better/pipeline.py
@@ -35,7 +35,12 @@ from seed.analysis_pipelines.better.helpers import (
     _store_better_building_analysis_results,
     _store_better_portfolio_analysis_results,
 )
-from seed.analysis_pipelines.utils import get_json_path
+from seed.analysis_pipelines.utils import (
+    aggregate_meter_readings,
+    get_json_path,
+    interpolate_monthly_readings,
+    reject_outliers,
+)
 
 from seed.models import (
     Analysis,
@@ -44,6 +49,7 @@ from seed.models import (
     AnalysisPropertyView,
     Column,
     Meter,
+    MeterReading,
 )
 
 
@@ -66,6 +72,7 @@ def _validate_better_config(analysis):
         'savings_target',
         'benchmark_data',
         'portfolio_analysis',
+        'preprocess_meters',
     ]
 
     return [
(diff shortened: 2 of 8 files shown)


seed/analysis_pipelines/better/client.py
```diff
@@ -19,6 +19,24 @@ class BETTERClient:
     def __init__(self, token):
         self._token = f'Token {token}'
 
+    def test_token(self):
```
macintoshpie, line 22: This change is unrelated to the meter stuff -- we were using the /buildings/ api to check if tokens were valid when creating the analysis but as we created buildings this endpoint got much too slow. We should ask BETTER team to create a token endpoint for testing. For now we just use their weathers API and check if we got a 401

seed/analysis_pipelines/better/pipeline.py
```diff
@@ -93,9 +100,9 @@ def _prepare_analysis(self, property_view_ids, start_analysis=False):
 
         # ping BETTER to verify the token is valid
         client = BETTERClient(organization.better_analysis_api_key)
-        _, errors = client.get_buildings()
-        if errors:
-            message = '; '.join(errors)
+        success = client.test_token()
```
macintoshpie, line 103: 🚤

seed/analysis_pipelines/better/pipeline.py
```diff
@@ -139,6 +146,54 @@ def _start_analysis(self):
...
+            )
+        )
+        for meter in meters:
+            # filtering on readings >= 1.0 b/c BETTER flails when readings are less than 1 currently
+            readings = meter.meter_readings.filter(reading__gte=1.0).order_by('start_time')
+            if readings.count() >= 12:
+                selected_meters_and_readings.append({
+                    'meter_type': meter.type,
+                    'readings': readings,
+                })
```
macintoshpie, line 192: this matches the original implementation of gathering meters and readings from before

seed/analysis_pipelines/better/pipeline.py
```diff
@@ -139,6 +146,54 @@ def _start_analysis(self):
...
+            monthly_readings = aggregate_meter_readings(meter_readings)
+            monthly_readings = reject_outliers(monthly_readings)
+            # filtering on readings >= 1.0 b/c BETTER flails when readings are less than 1 currently
+            monthly_readings = [reading for reading in monthly_readings if reading.reading >= 1.0]
+            monthly_readings = interpolate_monthly_readings(monthly_readings)
+            if len(monthly_readings) >= 12:
+                selected_meters_and_readings.append({
+                    'meter_type': meter_type,
+                    'readings': monthly_readings
+                })
```
macintoshpie, line 174: Section for preprocessing meters. There are some interesting differences from our original method:
1. we combine readings from the same meter type (instead of keeping different meters separate)
2. we "bin" meters into calendar months
3. we reject outliers
4. we interpolate missing months

Ryoken (approved):
My only comment would be to include the translation for the HTML tooltip.  Looks good otherwise!

macintoshpie:
I will make a ticket to translate the tooltips!

PR #2923, opened 2021-10-05
refactor(ci): increase operations per run

macintoshpie:
SEED is using a stalebot for closing old issues. The stalebot is configured by default to do 30 "actions" per run (ie per day). This is slow to go through our issues, e.g. from a recent run:
```
Warning: No more operations left! Exiting...
Warning: If you think that not enough issues were processed you could try to increase the quantity related to the operations-per-run (​https://github.com/actions/stale#operations-per-run​) option which is currently set to 30
```

#### What's this PR do?
Increase our actions per day to `500`. If we hit our rate limit then GH will just pause the GH actions ability to make requests for 1 hour, and I can bump it down tomorrow.

> GitHub has a rate limit and if reached will block these API calls for one hour (or API calls from other actions using the same user (a.k.a.: the github-token from the repo-token option)).


nllong (approved):
noice--clear them all out 🧹

axelstudios (approved):

Issue #2941, opened 2021-10-13
Password reset error


nllong:
The password reset email errors with the following log message:

```
web
File "/usr/lib/python3.9/site-packages/django/urls/resolvers.py", line 694, in _reverse_with_prefix
web
raise NoReverseMatch(msg)
web
django.urls.exceptions.NoReverseMatch: Reverse for 'password_reset_confirm' with keyword arguments '{'uidb64': 'Mzg3', 'token': 'aufq82-d4ce0a0e96d256dccb82f039e3ae6926'}' not found. 1 pattern(s) tried: ['accounts/password/reset/confirm/(?P<uidb64>[0-9A-Za-z_\\-]+)/(?P<token>[0-9A-Za-z]{1,13}-[0-9A-Za-z]{1,20})/$']
```

To recreate:
* launch seed with SES enabled
* try to reset password from the `forgot password` link on the login screen
* enter email
* site will say "page not found"
* look at log

PR #2942, opened 2021-10-13
Password reset url

nllong:
The password reset confirm URL was erroring when calling reverse. Presumably this was a django 2.3 issue?

#### What's this PR do?
Fixes the URL to follow the pattern being sent so the reverse works.

Test on dev1 (new) if possible sense it has the SES enabled.

#2941


macintoshpie:
@nllong  I'm guessing we'll want to release this as a patch to v2.12?

nllong:
> @nllong I'm guessing we'll want to release this as a patch to v2.12?

Unfortunately. I was thinking the same.

Issue #2921, opened 2021-10-04
Analyses: Add a date to the Run Date even if the analysis fails


RDmitchell:
**Is your feature request related to a problem? Please describe.**
Right now it is not possible to know when an analysis fails because a date isn't entered into the "Run Date" column

**Describe the solution you'd like**
Add a date in the "Run Date" column even if the analysis fails
![image](https://user-images.githubusercontent.com/6314950/135916884-accd50e2-a230-4893-930d-78d5638f7ce0.png)

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

macintoshpie:
@RDmitchell Sorry for the confusion, we ended up choosing to add a "Created" column to the table (see in the screenshot you shared). We thought it was best to keep "Run Date" to mean "when did this analysis actually start running" and the creation date can cover the case where an analysis was never started because it failed to pass our initial validation.

I updated the title of this ticket to reflect that

PR #2954, opened 2021-10-14
refactor: moved start_time on analyses to when they are prepared

Ryoken:
see ticket

#### What's this PR do?
The initial timestamp is now created when the analysis is prepared, rather than started, so start times should show up for failed analyses.

Create an analysis on a property with invalid info, such as an EUI analysis on a property with no meters.  Note there is still a start time.

https://github.com/SEED-platform/seed/issues/2921

Screen Shot 2021-10-14 at 3 33 27 PM


Ryoken:
Screen Shot 2021-10-18 at 12 46 16 PM

Issue #2913, opened 2021-09-29
After importing Meter Data, program shows the previous Mapping screen


RDmitchell:
**Describe the bug**
After importing meter data to match to an existing set of properties that were previously imported and mapped, the program leaves the user on the Mapping Review screen from the previous import. 

It's confusing, because you think maybe you are supposed to click the Save Mappings button, but if you do that the program will say the data has already been mapped. 

**Expected Behavior**
It would be nice if the program asked if you want to view your properties, or just put you back in the Data Set list view

**Actual Behavior**
See the bug description

**Steps to Reproduce**
* File to import: JCC_ESPM_clean_anonymized-Allgoodreecords.xlsx
![image](https://user-images.githubusercontent.com/6314950/135345815-d325c997-068c-4c03-96b6-4dec866e1e6e.png)
* Import as a Spreadsheet
![image](https://user-images.githubusercontent.com/6314950/135345836-1bfc565b-f4e5-4a55-bdcc-997523cfa79d.png)
![image](https://user-images.githubusercontent.com/6314950/135345860-1bb171bb-6676-4b8b-a112-9f66249eb8a4.png)
* Successful upload, continue to data mapping
![image](https://user-images.githubusercontent.com/6314950/135345995-03bbd838-6a3b-4579-b435-bc8b5557948e.png)
* Data mapping from previously defined Mapping Profile
![image](https://user-images.githubusercontent.com/6314950/135346167-21e8b655-5124-4a2e-a13d-81347a14a61b.png)
* Click Map Your Data button and go to Mapping Review and click Saving Mappings
![image](https://user-images.githubusercontent.com/6314950/135346744-31d4f789-81a6-45b2-94b9-84e793a87894.png)
* Click Confirm mappings & start matching
![image](https://user-images.githubusercontent.com/6314950/135346882-b1f5fa30-7980-45f8-b9d2-9faa1bcb9d7d.png)
* Program matches data
![image](https://user-images.githubusercontent.com/6314950/135346920-cb9f9b92-f67d-483c-8b1a-608f05a00100.png)
* Program shows results from matching and I click "Import Meters from the same file" button
![image](https://user-images.githubusercontent.com/6314950/135347106-7b5814d5-779a-45b7-a3f8-46849b73766c.png)
* Program displays the meter information and I click Confirm
![image](https://user-images.githubusercontent.com/6314950/135347446-563a7812-8596-4108-bc1c-ac3f8664556b.png)
[...]

PR #2977, opened 2021-11-01
refactor: navigate to data set list view after saved mapping

perryr16:
The process for importing new Data Files from the Data Set List view kicks off a sequence of events to ensure the correct data is imported. The majority of information is presented via a modal overlayed above a relevant page. After the Data File has been imported and matched the background page remains on the Data Mapping screen. This is confusing as a user might be inclined to re-save their mappings. It would be preferable to have the Data Set List as the background page after the Data File has been mapped. 

#### What's this PR do?
Upon completion of the data mapping, the Data Set List view will be set as the background page allowing the user to add more files. A user has the option to view properties by clicking 'View my properties' or import meters from the same file.

Add a new Data File to a Data Set and follow the flow. After the data mapping has been saved, the background screen should transition to the Data Set List view. 

#2913

https://user-images.githubusercontent.com/58446472/139734835-a51e5c46-c271-4db9-ae77-4a6e02a80996.mov


macintoshpie (approved):
Looks good to me 👍 👍

Issue #2999, opened 2021-11-08
Bug: property details "notes" tab breaks links


macintoshpie:
## context
- go to the notes tab for a property, then try to click on any of the other tabs. I get an error (see below)
<img width="966" alt="Screen Shot 2021-11-08 at 9 00 21 AM" src="https://user-images.githubusercontent.com/18518728/140776230-fc203753-e075-4717-ba14-44237b886b47.png">

## tasks
- investigate and fix
- determine which release this bug was introduced. Consider creating a patch release to fix

RDmitchell:
Instance: dev1
SHA: 10ead69a9

Fixed

PR #3003, opened 2021-11-09
refactor: update scope to include inventory

perryr16:
Navigation breaks when on the inventory notes page

#### What's this PR do?
updates $scope to include property view_id

Navigate to Inventory > i > meters > notes > meters 
There should be no error messages dispalyed.

#2999


macintoshpie (approved):
Fixed it 🔨

Issue #2813, opened 2021-08-11
Update Portfolio Manager Defaults column mapping profile


RDmitchell:
**Describe the bug**
The current "Portfolio Manager Defaults" Column Mapping Profile does not do a good job of actually mapping Portfolio Manager fields.
![image](https://user-images.githubusercontent.com/6314950/129078204-b60bb167-729c-43e5-ae28-0dbd299f966e.png)

**Expected Behavior**
I expected that selecting that mapping profile would map the ESPM fields in the file I was importing

**Actual Behavior**
Only a few fields were actually mapped when I selected that mapping (using the JCC ESPM file from #2808)
![image](https://user-images.githubusercontent.com/6314950/129078295-30d65769-4f61-493c-8296-e27f09f2005f.png)

**Steps to Reproduce**
* import a Portfolio Manager file (such as the one referenced in #2808)
* In the first Mapping screen, select the "Portfolio Manager Defaults" mapping profile and you will see very few fields filled out (see screen shot above)

Here is a more reasonable mapping, but basically we need to go through the standard ESPM fields and map them to SEED fields.
![image](https://user-images.githubusercontent.com/6314950/129078864-2f9d9144-7fc1-4f26-8770-1dc90cbbafff.png)

I am happy to help with this task.

**Instance Information**

Instance: dev1
SHA: 24b503d2d

**Additional Context**
none

RDmitchell:
@nllong -- This is a tricky request because there are so many ESPM fields that are possible. The defaults currently there are OK, not great, but probably passable for now. 

Would it be possible to add these three also?
![image](https://user-images.githubusercontent.com/6314950/166852753-a609a8c2-08a0-4651-94c0-1db33a633f2d.png)

perryr16:
I can take a look at this

perryr16:
Hey @RDmitchell, when trying to reproduce the bug on production (https://seedv2.lbl.gov/) "Property Name" and "City" populated with the "Property Manager Defaults" profile. Could you pass along the file you were testing with?

https://user-images.githubusercontent.com/58446472/167028828-c342a28f-af1a-4911-966c-917c65ea7808.mov

RDmitchell:
instance: dev1
SHA: 400ca8895

I don't see this change implemented in this version on dev1
![image](https://user-images.githubusercontent.com/6314950/172684528-8d3e7dbd-c7be-4286-baf7-a548cc9dec9f.png)

Here is a link to a file to test with
[...]

PR #3006, opened 2021-11-10
2813 refactor/update pm default column mapping profile

perryr16:
When an organization is created a 'Portfolio Manager Defaults' profile is assigned based on the values in `/seed/lib/mappings/pm-mapping.json`. A user can select the 'Portfolio Manager Defaults' profile during the data mapping sequence. Applying the profile will iterate through the imported data headers and assign any ESPM default columns/headers to the SEED Headers. Headers are assigned by an exact match only, as a result variations of headers have been added (ex: `Street Address` and `Address 1` are assigned to `Address Line 1`)

A column will be assigned to the `/seed/lib/mappings/pm-mapping.json` file if it meets the following conditions:
- is a column in the Column database model
- is a column in the ESPM formatted excel file (ESPM headers can be viewed by "Downloading your entire portfolio to Excel" at https://portfoliomanager.energystar.gov/pm/home.html)
- is a column in the ESPM Portfolio Manager Metrics file (viewed by "Download the full list of Portfolio Manager metrics here" at https://www.energystar.gov/buildings/tools-and-resources/full-list-portfolio-manager-custom-reporting-metrics)

Headers that do not meet these conditions can be added to the 'Portfolio Manager Defaults' profile by manually entering the SEED headers and clicking the green checkmark to update the profile.

Mapped ESPM Headers 
- Street Address
- Street Address 2
- How Many Buildings? 
- City/Municipality
- City
- Custom Property ID 1 - ID
- eGRID Subregion
- ENERGY STAR Score
- Gross Floor Area
- Postal Code
- Property Name
- Portfolio Manager ID
- Portfolio Manager Property ID
- Parent Property ID
- Portfolio Manager Parent Property ID
- Property Type - Self-Selected
- Site EUI (kBtu/ft2)
- Weather Normalized Site EUI (kBtu/ft2)
- State/Province
- UBID
- Year Built

Unmapped ESPM Headers
- Building Count 
- Country 
- Construction Status 
- Federal Agency/Department 
- GFA Units 
- Is this Property Owned or Operated by the US or Canadian Federal Government?
- Irrigated Area 
- Irrigated Area Units 
- Is this an Institutional Property? (Applicable only for Canadian properties)
- Occupancy (%)
- Other State/Province 
- Parent Property Name

#### What's this PR do?

Adds mapping possibilities to the `pm-mapping.json` file.

- Create a new organization
- Import a JCC file 
- on the mapping screen, select 'Profile Manager Defaults' from the dropdown and apply
- Known headers will be mapped
- Headers unknown to the DB will not be mapped 

#2813 

#### Screenshots
Screen Shot 2021-11-17 at 10 22 02 AM


seed/lib/mappings/data/pm-mapping.json
```diff
@@ -152,6 +242,69 @@
...
+    "type": "float",
+    "schema": ""
+  },
+  {
+    "display_name": "Longitude",
+    "to_field": "longitude",
+    "to_table_name": "PropertyState",
+    "from_field": "Longitude",
+    "units": "",
+    "type": "float",
```
perryr16, line 305: Entries like this (Longitude vs longitude) could be matched logically in the mapping_controller with `.toLowerCase()`

perryr16:
This only applies to PropertyState headers, should it also apply to TaxLotState headers?

macintoshpie:
> This only applies to PropertyState headers, should it also apply to TaxLotState headers?

I don't think so, ESPM doesn't store/export tax lot data, right @nllong ? Also looking through the PM mapping json file there aren't any existing tax lot mappings, so I think you're good

nllong:
That is correct, ESPM does not include taxlot data. That is typically loaded separately in SEED and then matched using a jurisdiction tax lot id field.

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -152,6 +206,60 @@
     "type": "float",
     "schema": ""
   },
+  {
+    "display_name": "Home Energy Score ID",
```
macintoshpie, line 210: I don't think ESPM reports this

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -152,6 +206,60 @@
...
+    "to_field": "home_energy_score_id",
+    "to_table_name": "PropertyState",
+    "from_field": "Home Energy Score ID",
+    "units": "",
+    "type": "string",
+    "schema": ""
+  },
+  {
+    "display_name": "Jurisdiction Property Lot ID",
```
macintoshpie, line 219: I don't think ESPM reports this

seed/lib/mappings/data/pm-mapping.json
macintoshpie: That's a good point. I'm ok with this though to avoid some kind of unexpected match though it's pretty unlikely

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -152,6 +206,60 @@
...
+    "to_field": "jurisdiction_property_lot_id",
+    "to_table_name": "PropertyState",
+    "from_field": "Jurisdiction Property Lot ID",
+    "units": "",
+    "type": "string",
+    "schema": ""
+  },
+  {
+    "display_name": "Latitude",
```
macintoshpie, line 228: I just checked and espm doesn't repot lat/lng. I wonder if we should still keep this though. @nllong thoughts?

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -224,6 +332,15 @@
     "type": "float",
     "schema": ""
   },
+  {
+    "display_name": "Occupied Floor Area",
```
macintoshpie, line 336: I don't see this exact field reported from ESPM, though I wonder if there's another name for it? Maybe we could scan through the list to see if there's a potential match

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -233,6 +350,33 @@
...
+  },
+  {
+    "display_name": "Owner City/State",
+    "to_field": "owner_city_state",
+    "to_table_name": "PropertyState",
+    "from_field": "Owner City/State",
+    "units": "",
+    "type": "string",
+    "schema": "bedes"
+  },
```
macintoshpie, line 379: Not finding these as exported from ESPM

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -242,6 +386,15 @@
     "type": "string",
     "schema": "bedes"
   },
+  {
+    "display_name": "Owner Postal Code",
```
macintoshpie, line 390: Not found

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -269,6 +422,15 @@
     "type": "string",
     "schema": ""
   },
+  {
+    "display_name": "PM Property ID",
+    "to_field": "pm_property_id",
+    "to_table_name": "PropertyState",
+    "from_field": "Portfolio Manager ID",
```
macintoshpie, line 429: 💯  can't believe we didn't have this before -- just a note, looks like it's only included when importing from data exported as "Download your entire portfolio"

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -350,6 +512,15 @@
     "type": "float",
     "schema": ""
   },
+  {
+    "display_name": "Property Footprint",
```
macintoshpie, line 516: Not found

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -441,17 +666,26 @@
     "schema": ""
   },
   {
-    "display_name": "State Province",
-    "to_field": "state_province",
+    "display_name": "Space Alerts",
```
macintoshpie, line 669: Not found

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -395,6 +611,15 @@
     "type": "float",
     "schema": ""
   },
+  {
+    "display_name": "Site EUI Modeled",
```
macintoshpie, line 615: not found

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -441,17 +666,26 @@
     "schema": ""
   },
   {
-    "display_name": "State Province",
```
macintoshpie, line 444: Not sure we want to delete this. Though it is interesting that `state_province` doesn't exist in the PropertyState model... @nllong are we intentionally mapping to extra data fields for this?

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -485,6 +719,15 @@
     "type": "string",
     "schema": ""
   },
+  {
+    "display_name": "Use Description",
+    "to_field": "use_description",
```
macintoshpie, line 724: Let's map this to `property_type` instead

macintoshpie (changes requested):
Great first pass! This is definitely going to save users time/effort.

I found [this spreadsheet](https://www.energystar.gov/buildings/tools-and-resources/full-list-portfolio-manager-custom-reporting-metrics) which lists all column names exportable through custom exports. To review this, I searched that spreadsheet as well as an example ["export entire portfolio" spreadsheet](https://github.com/SEED-platform/seed/files/7541194/SEED_City_Test.xlsx) b/c column names can be different between these two. I tried to flag any inconsistencies I could find.

Sorry for the confusion on some of these, the example files I had originally shared had apparently been modified with additional columns which I was unaware of.

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -63,12 +81,21 @@
     "schema": ""
   },
   {
-    "display_name": "Created Date",
```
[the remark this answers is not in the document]
perryr16: Per @nllong's comments I am trimming columns that do not appear in ESPM exports

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -170,12 +188,21 @@
...
+  },
   {
     "display_name": "National Median Site EUI",
     "to_field": "national_median_site_eui",
     "to_table_name": "PropertyState",
     "from_field": "medianSiteIntensity",
-    "units": "kBtu/ft2",
+    "units": "kBtu/ft**2",
     "type": "float",
     "schema": ""
```
perryr16, line 207: Fields like this are present on the "Full list of Portfolio Manager custom reporting metrics" but not on the ESPM export. Since it is existing I chose to leave it in.

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -48,7 +57,7 @@
     "display_name": "Building Count",
     "to_field": "building_count",
     "to_table_name": "PropertyState",
-    "from_field": "Number of Buildings",
+    "from_field": "How Many Buildings?",
```
perryr16, line 60: `Number of Buildings` is represented as a string, `"One"`, while `How Many Buildings?" is represented as an integer, `1`

nllong: I've seen this in ESPM files. Is that what you are saying--that there are two fields, one is numeric and the other is string?

perryr16: Correct, both `Number of Buildings` and `How Many Buildings?` appear in the ESPM files. One is a string, the other is an integer.

nllong: Yeah, we need both of these, crazy right?!

seed/lib/mappings/data/pm-mapping.json
macintoshpie: Actually, I think we need to keep these as ft**2, which we check here:
https://github.com/SEED-platform/seed/blob/2898863714434ce928231a4f69563b2c4f877100/seed/utils/organizations.py#L30

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -386,6 +413,15 @@
     "type": "float",
     "schema": ""
   },
+  {
+    "display_name": "Site EUI",
+    "to_field": "site_eui",
+    "to_table_name": "PropertyState",
+    "from_field": "Site EUI (kBtu/ft²)",
+    "units": "kBtu/ft**2/year",
```
macintoshpie, line 421: 💯

seed/lib/mappings/data/pm-mapping.json
```diff
@@ -458,6 +521,15 @@
...
   },
+  {
+    "display_name": "State Province",
+    "to_field": "state_province",
+    "to_table_name": "PropertyState",
+    "from_field": "State",
+    "units": "",
+    "type": "string",
+    "schema": ""
+  },
```
macintoshpie, line 532: I'm wondering if this (as well as the two above) should have a `to_field` of just `state` -- `state_province` isn't a real SEED column from what I can see on the PropertyState model.
@nllong do you know why we originally had it mapping to state_province?

macintoshpie (changes requested):
Just a few more comments, but this is look great! So glad we're cleaning this up

seed/lib/mappings/data/pm-mapping.json
perryr16: I agree. I think this should be just `"to_field": "state"`

nllong: agreed

seed/lib/mappings/data/pm-mapping.json
macintoshpie: So are we trying to keep both of these or only import "how many buildings"?

macintoshpie (approved):
Great!

PR #3012, opened 2021-11-18
2515 refactor/progress bar sits at 33

perryr16:
When uploading new building files the user will eventually be shown a progress bar modal for "Finding inventory matches". This progress bar has 3 steps: "Geocoding", "Matching data", and "Pairing data". The second step can take a very long time to complete causing users to think that the program has crashed. 

#### What's this PR do?
This PR breaks the "Matching data" step into smaller more granular steps. Upon completion of each granular task, the progress bar will advance. This allows the progress bar to advance more frequently and in smaller steps, preventing the user from believing the program has crashed. 

The number of progress steps has been increased from 3 to 9 
The new steps order is:
- Geocoding
- Matching Data
- Merging Matched Properties (5x)
- Assigning Properties to Views
- Pairing Data

The "Matching Data" step gets hung up on a `.filter(...` function. 
The "Merging Matched Properties" steps are generated while iterating through `matched_id_groups`. Each time 1/5th of the properties have been matched a "Merging Matched Properties" message is generated and the progress bar increases.

- Import a file with 1000+ properties 
- Go through the upload process...
- Click "Confirm mappings & start matching" and watch the progress bar.

#2515 

#### Screenshots
Example of an import file with 2,500 properties

https://user-images.githubusercontent.com/58446472/142479210-82502043-95aa-49e2-b9e4-3447f2313d61.mov


seed/data_importer/match.py
```diff
@@ -256,6 +258,10 @@ def inclusive_match_and_merge(unmatched_state_ids, org, StateClass):
     # Collapse groups of matches found in the previous step into 1 -State per group
     merges_within_file = 0
     priorities = Column.retrieve_priorities(org)
+    total_ids = len(matched_id_groups)
+    batch = int(total_ids/5)
+    batches = [batch*i for i in range(5)]
```
perryr16, line 263: Open to suggestions on naming convention. If there were 200 ids`batch` is 1/5th (`20`), and `batches` are (`[0, 40, 80, 120, 160]`)

seed/data_importer/match.py
```diff
@@ -247,6 +248,7 @@ def inclusive_match_and_merge(unmatched_state_ids, org, StateClass):
     )
 
     # Group IDs by -States that match each other
+    # This filter function takes a significant amount of time to complete
```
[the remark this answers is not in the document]
macintoshpie: Check this out for how queries work (django uses this thing called QuerySets)
https://docs.djangoproject.com/en/3.2/topics/db/queries/#retrieving-objects

Not sure we want to try and optimize too much in this PR, but for future reference Django has some recommendations - you'll have to dig through for info about dbs and queries
https://docs.djangoproject.com/en/3.2/topics/performance/

seed/data_importer/match.py
```diff
@@ -256,6 +258,10 @@ def inclusive_match_and_merge(unmatched_state_ids, org, StateClass):
     # Collapse groups of matches found in the previous step into 1 -State per group
     merges_within_file = 0
     priorities = Column.retrieve_priorities(org)
+    total_ids = len(matched_id_groups)
```
macintoshpie, line 261: I like what you did here. It seems like we're stepping the progress data each we've gone through `batch` number of properties. A pretty traditional way of doing this is with the modulo operator. And we can use `enumerate` to also get the "index" we're working with. For example:
```python
batch_size = 100
for idx, ids in enumerate(matched_id_groups):
    if idx % batch_size == 0:
        print(f'I did another {batch_size} items!')
```

seed/data_importer/match.py
```diff
@@ -123,10 +123,11 @@ def match_and_link_incoming_properties_and_taxlots(file_pk, progress_key):
 
         # Within the ImportFile, merge -States together based on user defined matching_criteria
         log_debug('Start Properties inclusive_match_and_merge')
-        promoted_property_ids, property_merges_within_file_count = inclusive_match_and_merge(promoted_property_ids, org, PropertyState)
+        promoted_property_ids, property_merges_within_file_count = inclusive_match_and_merge(promoted_property_ids, org, PropertyState, progress_data)
 
         # Filter Cycle-wide duplicates then merge and/or assign -States to -Views
         log_debug('Start Properties states_to_views')
+        progress_data.step('Assigning Properties to Views')
         merged_property_views, property_duplicates_against_existing_count, property_new_count, property_merges_against_existing_count, property_merges_between_existing_count = states_to_views(
```
macintoshpie, line 131: I think we might need to add "sub progress" stuff for this one as well -- it's where we actually search the entire database for matching properties and merge them. So this will be particularly slow when there are many thousands of properties already in the database.

macintoshpie (changes requested):
I like the direction this is headed! We should probably add sub tasks for the taxlot sections as well (like we are for properties here).

Something that might be worth investigating is refactoring progressdata's
`increment_value` and `increment_by` so that we could make smaller increments for particular sections. But if that's too much effort to refactor then let's skip it

seed/data_importer/match.py
perryr16: That's so clean!

seed/data_importer/match.py
```diff
@@ -272,13 +276,16 @@ def inclusive_match_and_merge(unmatched_state_ids, org, StateClass):
 
             promoted_ids.append(merge_state.id)
 
+        if batch_size > 0 and idx % batch_size == 0 and progress_data:
```
perryr16, line 279: Too many conditions? If there are no matched_id groups, batch size will be zero, resulting in a zeroDivisionError and proress_data is an optional arg.

perryr16:
> Something that might be worth investigating is refactoring progressdata's `increment_value` and `increment_by` so that we could make smaller increments for particular sections. But if that's too much effort to refactor then let's skip it

Currently, this hasn't been refactored, but I'd be happy to go back and investigate

macintoshpie (changes requested):
Tested this a bit more with a larger file, 9,999 properties (see test_data_small.csv in this drive: https://drive.google.com/drive/folders/1e5-c5EpB4YkUlX442DD64IJpOCW-1No0)

It can still feel a bit stuck, e.g. It stayed at 93% for a really long time here

Screen Shot 2021-11-29 at 11 43 13 AM

Could you investigate what's happening there? Maybe there's another large expensive process that's not being tracked with the progress data.

perryr16:
Moving towards a parallel progress bar that shows the 'sub progress'

Issue #3018, opened 2021-11-22
Delete column timeout


macintoshpie:
## context
As an org owner, you can delete columns. This is no easy task for SEED because it has to update a lot of data in the database (i.e. remove the column and rehash *States).
We created the task `delete_organization_column` to handle this asynchronously for us. This works well, until the organization's number of *States is very large. In production, this ends up causing a timeout (harakiri) for the uWSGI worker because it takes too long to query these states and kick off the async task before responding to the user.

This action is actually blocked for these orgs with a large number of *States (i.e. they can't delete columns)

## task
- Don't evaluate the *State query in the request response cycle. This is where they are, you can see that we call `list` which immediately evaluates the queryset (which can take a really long time)

https://github.com/SEED-platform/seed/blob/2515-refactor/progress-bar-sits-at-33/seed/tasks.py#L335

https://github.com/SEED-platform/seed/blob/0b52f13c0f5bfe2b657173c6560a8349081fe388/seed/tasks.py#L340

RDmitchell:
@nllong -- can we close this, given that I believe DC was able to delete their fields successfully?

nllong:
yes, this has now been fixed!

PR #3021, opened 2021-12-03
3018 Delete Column Timeout Refactor

perryr16:
Authorized organization users have the ability to delete extra "PropertyState" or "TaxLotState" columns from the organization. If there are only a few "-States" then the process is quite quick. As more "-States" are added to the DB the column deleting process can become so time-consuming that it causes the webserver to timeout and 'harakiri' or kill itself. 

The general process is to query the database for all the "-States" that are both in the organization and have that specific column. Then as the columns are deleted the progress bar is updated showing `results deleted / total results` and the estimated time remaining. Both of these bits of data use the total number of results returned by the query. The original query is "lazy" and won't be evaluated until called. Asking for the length of the query will evaluate it and halt the process until the program has finished unpacking the results. If there are several million results this process can take upwards of 10 minutes.

#### What's this PR do?

This PR moves the evaluation process from the `seed_web` server into a series of nested tasks that will be handled in the `seed_celery` server. 

It also adds a step in the progress bar modal for `evaluating affected records`

- Login to any organization that has 1 million + "-States"
- Go to organization > column settings > scroll down and delete a column 
- While monitoring the network tab, you should see the process for `http://localhost/api/v3/columns/<column_id>/?organization_id=<organization_id>` complete in less than 1 second.

#3018 

#### Screenshots
Web process to initaite tasks 
Screen Shot 2021-12-03 at 1 32 55 PM

Celery task to find length of records 
Screen Shot 2021-12-03 at 1 33 02 PM

Celery task to delete records in batches 
Screen Shot 2021-12-03 at 1 33 30 PM

### DIFF
--- a/seed/static/seed/js/controllers/delete_column_modal_controller.js
+++ b/seed/static/seed/js/controllers/delete_column_modal_controller.js
@@ -24,10 +24,9 @@ angular.module('BE.seed.controller.delete_column_modal', [])
       $scope.delete = function () {
         $scope.state = 'pending';
         columns_service.delete_column_for_org(organization_id, column.id).then(function (result) {
-          $scope.state = 'running';
-          $scope.startTime = moment();
+          $scope.state = 'evaluating';
           $scope.interval = $interval(function () {
-            $scope.updateTime();
+            $scope.state == 'running' ? $scope.updateTime() : $scope.setRunningState();
           }, 1000);
           $scope.updateTime();
           uploader_service.check_progress_loop(result.data.progress_key, 0, 1, function (response) {
@@ -61,6 +60,12 @@ angular.module('BE.seed.controller.delete_column_modal', [])
             return $scope.formatTime(moment.duration(diff / progress - diff));
           }
         }
+      };
+
+      $scope.setRunningState = function () {
+        $scope.eta = $scope.etaFn();
+        $scope.eta ? 
+          ($scope.state = 'running', $scope.startTime = moment()) : null
       };
 
       $scope.updateTime = function () {
--- a/seed/tasks.py
+++ b/seed/tasks.py
@@ -361,6 +361,7 @@ def _delete_organization_column_evaluate(column_pk, org_pk, prog_key, chunk_size
         )
 
 
+@shared_task
 def _delete_organization_column_chunk(chunk_ids, column_name, table_name, prog_key, *args, **kwargs):
     """updates a list of ``chunk_ids`` and increments the cache"""
 
--- a/seed/tasks.py
+++ b/seed/tasks.py
@@ -326,12 +326,11 @@ def delete_organization_column(column_pk, org_pk, prog_key=None, chunk_size=100,
     progress_data = ProgressData.from_key(prog_key) if prog_key else ProgressData(
         func_name='delete_organization_column', unique_id=column_pk)
 
-    tasks = []
-    tasks.append(
-        _generate_tasks_to_delete_organization_column.subtask((column_pk, org_pk, progress_data.key, chunk_size))
-    )
+    chain(
+        _generate_tasks_to_delete_organization_column.subtask((column_pk, org_pk, progress_data.key, chunk_size)),
+        _finish_delete_column.subtask([column_pk, progress_data.key])
+    ).apply_async()
 
-    chord(tasks, interval=15)(_finish_delete_column.subtask([column_pk, progress_data.key]))
     return progress_data.result()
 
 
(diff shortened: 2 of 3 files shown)


seed/static/seed/js/controllers/delete_column_modal_controller.js
```diff
@@ -24,10 +24,9 @@ angular.module('BE.seed.controller.delete_column_modal', [])
       $scope.delete = function () {
         $scope.state = 'pending';
         columns_service.delete_column_for_org(organization_id, column.id).then(function (result) {
-          $scope.state = 'running';
-          $scope.startTime = moment();
+          $scope.state = 'evaluating';
           $scope.interval = $interval(function () {
-            $scope.updateTime();
+            $scope.state == 'running' ? $scope.updateTime() : $scope.setRunningState();
```
perryr16, line 29: New state while the celery task is evaluating the queryset for length. If we're still evaluating there is no need to update the time

seed/tasks.py
```diff
@@ -322,49 +323,50 @@ def _finish_delete_cycle(cycle_id, prog_key):
...
+    progress_data = ProgressData.from_key(prog_key)
     total = len(ids)
-
-    # total is the number of records divided by the chunk size
     progress_data.total = total / float(chunk_size)
     progress_data.data['completed_records'] = 0
     progress_data.data['total_records'] = total
     progress_data.save()
 
-    tasks = []
```
perryr16, line 352: since we are already inside of a task, simply run it. No need to add to a task list

perryr16:
Note that API requests to ESPM are returning 400 error codes causing django tests to fail.

seed/tasks.py
```diff
@@ -322,49 +323,50 @@ def _finish_delete_cycle(cycle_id, prog_key):
...
+                                           extra_data__has_key=column.column_name).values_list('id', flat=True)
     elif column.table_name == 'TaxLotState':
-        ids = list(
-            TaxLotState.objects.filter(organization_id=org_pk, data_state=DATA_STATE_MATCHING,
-                                       extra_data__has_key=column.column_name).values_list('id', flat=True)
-        )
+        ids = TaxLotState.objects.filter(organization_id=org_pk, data_state=DATA_STATE_MATCHING,
+                                         extra_data__has_key=column.column_name).values_list('id', flat=True)
 
+    """ Evaluate a queryset in a task allowing web processes to complete """
```
macintoshpie, line 352: I think you can delete this comment

seed/tasks.py
```diff
@@ -322,49 +323,50 @@ def _finish_delete_cycle(cycle_id, prog_key):
...
-
     progress_data = ProgressData.from_key(prog_key) if prog_key else ProgressData(
         func_name='delete_organization_column', unique_id=column_pk)
 
+    tasks = []
+    tasks.append(
+        _generate_tasks_to_delete_organization_column.subtask((column_pk, org_pk, progress_data.key, chunk_size))
+    )
+
+    chord(tasks, interval=15)(_finish_delete_column.subtask([column_pk, progress_data.key]))
```
macintoshpie, line 334: It looks like we just have a single task we kick off, which in that case we can just use a [celery chain](https://docs.celeryproject.org/en/stable/userguide/canvas.html#chains), e.g.
```python
chain(
    _generate_tasks_to_delete_organization_column.subtask(...),
    _finish_delete_column.subtask(...)
).apply_async()
```

seed/tasks.py
macintoshpie: I think the only benefit to creating a "task" list would be that we could parallelize the deletion, maybe making it `num_workers` times faster. Though I'm not that it matters too much here... I'm honestly fine keeping it how you have it. I would probably rename this function though since it's not actually generating any new tasks. Also, we should probably remote the `@shared_task` decorator on `_delete_organization_column_chunk` just so we know it's not actually used as a celery task.

macintoshpie (changes requested):
A few small requests, but this looks great!

macintoshpie (changes requested):
Fixes look great! Tested on a large number of properties and we immediately get an HTTP response on the delete request so no more timeouts
Screen Shot 2021-12-08 at 12 09 03 PM
 
The only issue I encountered was after it finished I ended up with this screen:
Screen Shot 2021-12-08 at 12 14 58 PM

Any idea if this is new/repeatable? Note that it was no longer polling progress data at this point, it just stopped at this page.

macintoshpie (approved):
Perf! ⚡

PR #3026, opened 2021-12-07
docs: remove conventional commit docs

macintoshpie:
Previously we suggested developers use conventional commits. This hasn't turned out to be very useful, so I think we should remove it.

#### What's this PR do?
- remove relevant docs
- remove the husky package which was used to enforce conventional commits


package.json
```diff
@@ -17,14 +17,12 @@
     "node": ">=8.10.0"
   },
   "devDependencies": {
-    "@commitlint/cli": "~11.0.0",
     "@commitlint/config-conventional": "~11.0.0",
```
axelstudios, line 20: @macintoshpie  Should this be removed, too?

macintoshpie: ope yeah, lemme do that

axelstudios (approved):

Issue #2890, opened 2021-09-21
Update deprecated JSONFields


macintoshpie:
## context
SEED uses JSONFields for some models. For version 4.0, Django is deprecating the fields we use in favor of a more generic field. See Django's release notes regarding this deprecation here: https://docs.djangoproject.com/en/3.2/releases/3.1/#jsonfield-for-all-supported-database-backends

## task
- refactor our current postgres JSONFields according to Django's guidelines

## warning logs
If you start the server (with django 3.1 currently) you will see these warning messages:
```
web_1          | System check identified some issues:
web_1          | 
web_1          | WARNINGS:
web_1          | data_importer.ImportFile.matching_results_data: (fields.W904) django.contrib.postgres.fields.JSONField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
web_1          |        HINT: Use django.db.models.JSONField instead.
web_1          | data_importer.ImportFile.raw_property_state_to_filename: (fields.W904) django.contrib.postgres.fields.JSONField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
web_1          |        HINT: Use django.db.models.JSONField instead.
web_1          | landing.SEEDUser.default_building_detail_custom_columns: (fields.W904) django.contrib.postgres.fields.JSONField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
web_1          |        HINT: Use django.db.models.JSONField instead.
web_1          | landing.SEEDUser.default_custom_columns: (fields.W904) django.contrib.postgres.fields.JSONField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
web_1          |        HINT: Use django.db.models.JSONField instead.
web_1          | orgs.Organization.display_meter_units: (fields.W904) django.contrib.postgres.fields.JSONField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
web_1          |        HINT: Use django.db.models.JSONField instead.
web_1          | seed.Analysis.configuration: (fields.W904) django.contrib.postgres.fields.JSONField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
web_1          |        HINT: Use django.db.models.JSONField instead.
web_1          | seed.Analysis.parsed_results: (fields.W904) django.contrib.postgres.fields.JSONField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
web_1          |        HINT: Use django.db.models.JSONField instead.
[...]

PR #3030, opened 2021-12-10
Update django.contrib JSONField to Django native JSONFields

nllong:
In Django 4, the contributed JSONFields will be deprecated. 

#### What's this PR do?
Use the `django.db.model.JSONField` for the previous versions. This also updates the migrations to use the new version so that the django.contrib library could be removed.

* Unit tests
* Monkey tests

#2890


seed/models/column_mapping_profiles.py
```diff
@@ -19,7 +18,7 @@ class ColumnMappingProfile(models.Model):
     )
 
     name = models.CharField(max_length=255, blank=False)
-    mappings = JSONField(default=list, blank=True)
+    mappings = models.JSONField(default=dict, blank=True)
```
nllong, line 21: the default should have been a `dict`, tested this change and didn't see any issues with it.

macintoshpie (approved):
Sweet! Did some monkey testing on existing DB as well as a fresh one and all seems well ⚡

PR #3037, opened 2021-12-14
2515 refactor/data upload progress bar

perryr16:
When uploading new building files the user will eventually be shown a progress bar modal for "Finding inventory matches". This progress bar has 3 steps: "Geocoding", "Matching data", and "Pairing data". The second step can take such a long time to complete the user may think that the program has crashed. 

#### What's this PR do?
This PR adds a parallel "sub progress" bar that updates from 0-100 {n} times for each of the three main steps. Geocoding and Pairing tasks only have 1 sub progress task while the Matching task has been broken up into 6 sub progress tasks. 
- Matching Data (1/6): Filtering Duplicate States
- Matching Data (2/6): Inclusive Matching and Merging
- Matching Data (3/6): Merging Unmatched States
- Matching Data (4/6): Merging State Pairs
- Matching Data (5/6): Promoting States
- Matching Data (6/6): Merge Views

Some tasks are dependent on the properties of the upload file and of the current database. As a result, some tasks may be very quick or non-existent.
- Import a file with about 100 properties
- Go through the upload process...
- Import a file with 1000+ properties 
- Go through the upload process...

`Note` that when tested with a file with 10,000+ properties attached to the DC database there was an internal server error (unrelated to this PR as it occurred in the develop branch as well)

#2515 

#### Screenshots
Data import from file with 250 properties:
Screen Shot 2021-12-17 at 11 31 32 AM

Screen Shot 2021-12-17 at 11 28 58 AM

Screen Shot 2021-12-17 at 11 29 13 AM

Screen Shot 2021-12-17 at 11 33 52 AM


seed/utils/match.py
```diff
@@ -429,3 +431,14 @@ def whole_org_match_merge_link(org_id, state_class_name, proposed_columns=[]):
             transaction.set_rollback(True)
 
     return summary
+
+
+def update_sub_progress_total(total, sub_progress_key=None, finish=False):
```
perryr16, line 436: Does this belong in /utils/match? It's being used in both /match and /tasks files. It could be in a new /util/progress_data file

seed/data_importer/match.py
```diff
@@ -186,7 +213,7 @@ def match_and_link_incoming_properties_and_taxlots(file_pk, progress_key):
     }
 
 
-def filter_duplicate_states(unmatched_states):
+def filter_duplicate_states(unmatched_states, sub_progress_key=None):
```
perryr16, line 216: Is this the best way to modify the existing functions? These fns are called in other locations that won't have `sub_progress_key` arguments. It works, but it leads to a lot of `if ...` statements

seed/data_importer/match.py
```diff
@@ -256,7 +295,8 @@ def inclusive_match_and_merge(unmatched_state_ids, org, StateClass):
     # Collapse groups of matches found in the previous step into 1 -State per group
     merges_within_file = 0
     priorities = Column.retrieve_priorities(org)
-    for ids in matched_id_groups:
+    batch_size = int(len(matched_id_groups) / 100) + (len(matched_id_groups) % 100 > 0)
```
perryr16, line 298: `(len(matched_id_groups) % 100 > 0)` is a simple way of rounding up without importing a library. This avoid a Zero division error

seed/data_importer/tasks.py
```diff
@@ -1163,16 +1166,21 @@ def geocode_and_match_buildings_task(file_pk):
...
-    celery_chain(_geocode_properties_or_tax_lots.s(file_pk, progress_data.key), post_geocode_tasks)()
+    sub_progress_data.total = 100
+    sub_progress_data.save()
 
-    return progress_data.result()
+    celery_chain(
+        _geocode_properties_or_tax_lots.s(file_pk, progress_data.key, sub_progress_data.key),
+        post_geocode_tasks)()
+
+    return {'progress_data': progress_data.result(), 'sub_progress_data': sub_progress_data.result()}
```
perryr16, line 1183: New return value has a few cascading effects in the tests

seed/data_importer/tests/integration/test_data_import.py
```diff
@@ -449,9 +449,10 @@ def test_map_all_models_xml(self):
         self.assertEqual(ps.count(), 1)
 
         # !! we should have warnings for our file because of the bad measure names !!
-        self.assertNotEqual({}, progress_info.get('file_info', {}))
-        self.assertIn(self.import_file.uploaded_filename, list(progress_info['file_info'].keys()))
-        self.assertNotEqual([], progress_info['file_info'][self.import_file.uploaded_filename].get('warnings', []))
+        self.assertNotEqual({}, progress_info.get('progress_data', {}))
```
perryr16, line 452: modify test for `geocode_and_match_buildings_task` new return value

macintoshpie: noice 👍

seed/static/seed/js/services/uploader_service.js
```diff
@@ -126,6 +126,56 @@ angular.module('BE.seed.service.uploader', []).factory('uploader_service', [
...
+              ) :
+              uploader_factory.check_progress_loop_main_sub(progress_argument, success_fn, failure_fn)
+          } else {
+            success_fn(data[0]);
+          }
+          
+        }, 750);
+      };
+
+      function update_progress_bar_obj(data, progress_arg) {
```
macintoshpie, line 154: Yoo this is nice, could you reuse it in the existing `check_progress_loop` above?

seed/static/seed/partials/data_upload_modal.html
```diff
@@ -200,17 +200,23 @@ <h4 class="modal-title" ng-switch-when="16" translate>PM_METER_IMPORT_RESULTS</h
...
-                    <div class="progress_bar_copy_bottom">{$ uploader.progress | number:0 $}% {$:: 'Complete' | translate $} {$ uploader.status_message ? ': ' + uploader.status_message : '' $}</div>
+                    <div class="progress_bar_copy_bottom"><strong>{$ uploader.progress | number:0 $}% {$:: 'Complete' | translate $} {$ uploader.status_message ? ': ' + uploader.status_message : '' $}</strong></div>
                     <!-- Show a hint that SEED is still working when the progress data hasn't been updated in 1 minute -->
                     <div ng-if="uploader.progress_last_updated != null && uploader.progress_last_checked != null && (uploader.progress_last_checked - uploader.progress_last_updated) / 1000 > 60"
                          class="progress_bar_copy_bottom">
                         <i class="fa fa-spinner fa-pulse fa-fw" style="padding-right: 0"></i>
                         SEED is still working. Please do not refresh or leave this page. (Last updated at {$ uploader.progress_last_checked | date:'hh:mm a' $})
                     </div>
                 </div>
+                <!-- Sub progress bar -->
```
macintoshpie, line 214: 💯

seed/utils/match.py
macintoshpie: Could it make sense to put it in `seed/lib/progress_data.py`?

macintoshpie (commented):
Some initial feedback. Everything is looking great though! I tested importing a 10k properties file into org with those properties already existing and it worked perfectly.

I'm going to do a bit more testing with other files and will come back for another review

macintoshpie:
Did some more testing and the files are being imported as expected. Once the previous comments are resolved I think this should be good to go

seed/utils/match.py
macintoshpie: I ended up not moving it, this is fine

PR #3053, opened 2021-12-21
Fix: electricity meter import from ESPM custom report

### LINKED ISSUE #3028: Fix: electricity meter import from ESPM custom report
## context
ESPM allows users to define custom reports for exporting data. We added the ability to import meters from the custom reports (see: #2673 and #2715 ). Recently ESPM renamed the column for electric meter data from "Electric Use - Monthly (kBtu)" to "Electric Use (Grid) - Monthly (kBtu)" (see: #3019 ) which broke the ability to import these types of files with electricity data.

## task
- Reference the new column for parsing meters
- Update the meter _type_ when parsing this type of file -- previously we used "Electricity - Unknown" b/c we weren't sure what type of electricity it was. Now we are, so we can use Electricity Grid for the meter type.
- Investigate, and fix, if we need to update the columns referenced when we check, after a successful properties import, if the file also includes meter data (which we then show the button that says "Import Meters from the same file")

haneslinger:
#### What's this PR do?
Fix `Electricity - GRID` in `METER_TYPE_MAPPING` during  ESPM request preprocessing and tests

import meters from ESPM to seed

Closes #3028


Ryoken (approved):
Looks good!

macintoshpie (approved):
Nice!

Issue #2891, opened 2021-09-21
Fix deprecated urls


macintoshpie:
## context
- Django is deprecating the urls method we are currently using, e.g.
```
RemovedInDjango40Warning: django.conf.urls.url() is deprecated in favor of django.urls.re_path()
```
See https://docs.djangoproject.com/en/3.2/releases/3.1/#id2

## task
- fix the urls that use the old format

PR #3057, opened 2021-12-28
Replace deprecated function django.urls.url

haneslinger:

done in tandem with [jwt-oauth2 #1](https://github.com/macintoshpie/jwt-oauth2/pull/1), which replaces deprecated functions in a dependencies

#### What's this PR do?

Replace deprecated function django.urls.url

```
python -Wd manage.py test
```
has no  warning for this function, in contrast to the same command on master.

#2891


macintoshpie (approved):
💯 nice!

Issue #2892, opened 2021-09-21
Fix usage of deprecated method request.is_ajax()


macintoshpie:
## context
Django 4.0 is deprecating `request.is_ajax()` which SEED uses. See https://docs.djangoproject.com/en/3.2/releases/3.1/#id2

## task
- get rid of this method call. We might need to implement an alternative method for determining if a request is an API request

PR #3058, opened 2021-12-28
Replace deprecated function request.is ajax

haneslinger:
# PUSH AFTER #3057 

#2892

### DIFF
--- a/seed/utils/api.py
+++ b/seed/utils/api.py
@@ -167,7 +167,7 @@ def get_api_request_user(request):
     """
     Determines if this is an API request and returns the corresponding user if so.
     """
-    if request.is_ajax():
+    if request.headers.get('x-requested-with') == 'XMLHttpRequest':
         return False
 
     return User.process_header_request(request)


seed/utils/api.py
```diff
@@ -167,7 +167,7 @@ def get_api_request_user(request):
     """
     Determines if this is an API request and returns the corresponding user if so.
     """
-    if request.is_ajax():
+    if request.headers.get('x-requested-with') == 'XMLHttpRequest':
```
macintoshpie, line 170: I don't think there's any meaningful difference here, but how would you feel about switching it to the way they had it implemented internally?
https://github.com/django/django/blob/03b733d8a8f5b5d8ca0a3cc32b7e6e6c046d95e2/django/http/request.py#L232

haneslinger: like this?
```
request.META.get('HTTP_X_REQUESTED_WITH') == 'XMLHttpRequest'
```
we could, but the [docs](https://docs.djangoproject.com/en/3.2/releases/3.1/#id2) recommend the current way. 
```
If you are writing your own AJAX detection method, request.is_ajax() can be reproduced exactly as request.headers.get('x-requested-with') == 'XMLHttpRequest'.
```

macintoshpie: oh shoot, I missed that, let's do that instead!

macintoshpie (approved):
Nice!

Issue #3048, opened 2021-12-17
Add frontend filtering for extra data columns


macintoshpie:
Blocked by #3042

Issue #3047, opened 2021-12-17
Add frontend functionality for filtering on canonical columns


macintoshpie:
Blocked by #3041 

Probably want to use UI Grid's `useExternalFiltering` functionality: http://ui-grid.info/docs/#!/tutorial/Tutorial:%20308%20External%20Filtering

PR #3060, opened 2021-12-29
Feat/inventory column filtering and sorting

macintoshpie:
Remaining Todos (won't be addressed in this PR):
- fix bug: if we allow users to filter properties based on taxlot columns (or vice versa) we will have column_name collisions. E.g. `created` for taxlots and `created` for properties (which we are unable to distinguish currently)
- fix bug: some extra_data columns are strings, and trying to use numeric filtering doesn't work
- improvement: find a way to improve how users "use" filtering and sorting. Two main issues:
  - currently we just debounce the filtering whenever a filter or sort is updated. This is confusing to user
  - find a way to summarize/display the sorts and filters applied currently so you can know what's going on even when you're not looking at a column

- SEED is creating a paginated inventory list, which requires sorting and filtering on the backend

#### What's this PR do?
- adds rudimentary filtering and sorting for the beta inventory list page
  - for the `/api/v3/properties/filter/` endpoint (and the same for taxlots), users can provide query parameters for filtering. The format is `<column_name>=<value>`, which is the Column.column_name. Users can optionally provide django operators like `<column_name>__lt` or `<column_name>__gte`, etc.
  - For the same endpoint, users can provide the `order_by` query parameter in the format `order_by=<column_name>` to sort the results according to that column's values. By default this is in ascending order, users can prefix the column name with a `-` for descending order: `order_by=-<column_name>`. Users can also provide multiple order_by params to sort by multiple columns: `order_by=site_eui&order_by=-created`

NOTE: these changes are all backwards compatible with the existing endpoint API usage

- monkey test the frontend filtering from the beta inventory list page. Try providing good and bad values. Bad values (e.g. a word like "hello" for a datetime column such as `created` should fail gracefully.

#3047 
#3048 
#3045 

https://user-images.githubusercontent.com/18518728/147616310-29957deb-4276-4143-b50e-2e32e06e9b24.mov

### DIFF
--- a/seed/static/seed/js/controllers/inventory_list_beta_controller.js
+++ b/seed/static/seed/js/controllers/inventory_list_beta_controller.js
@@ -829,14 +829,20 @@ angular.module('BE.seed.controller.inventory_list_beta', [])
       $scope.load_inventory = function (page) {
         const page_size = 100;
         spinner_utility.show()
-        return fetch(page, page_size).then(function (data) {
-          $scope.inventory_pagination = data.pagination;
-          processData(data.results);
-          $scope.gridApi.core.notifyDataChange(uiGridConstants.dataChange.EDIT);
-          modalInstance.close();
-          evaluateDerivedColumns();
-          spinner_utility.hide()
-        });
+        return fetch(page, page_size)
+          .then(function (data) {
+            if (data.status === 'error') {
+              Notification.error(data.message);
+              spinner_utility.hide();
+              return;
+            }
+            $scope.inventory_pagination = data.pagination;
+            processData(data.results);
+            $scope.gridApi.core.notifyDataChange(uiGridConstants.dataChange.EDIT);
+            modalInstance.close();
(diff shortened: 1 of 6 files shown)


seed/static/seed/js/controllers/inventory_list_beta_controller.js
```diff
@@ -599,20 +603,10 @@ angular.module('BE.seed.controller.inventory_list_beta', [])
         // Modify misc
         if (col.data_type === 'datetime') {
           options.cellFilter = 'date:\'yyyy-MM-dd h:mm a\'';
-          options.filter = inventory_service.dateFilter();
```
macintoshpie, line 602: don't need these filters and sorting anymore

haneslinger (approved):
LGTM 🥇

Ryoken:
Average load time for properties/filter endpoint:

before: 3.35 seconds
after: 3.62 seconds
improvement: -8%

Issue #3042, opened 2021-12-17
Add server-side filtering for extra data columns


macintoshpie:
future: comma-separated filters (basically an AND operator)
should support:
string: "contains"
numeric: <, >, <=, >=, ==, !=
dates: research how this works/should work
research other reqs by looking at frontend

macintoshpie:
#3048 implemented filtering for extra_data which works for some cases. Specifically, the cases where the data type inside of the extra data matches the data_type defined for that column. E.g. if the column is defined as numeric, and the data inside of extra data is actually numeric (instead of a string)

## more context
For context, when importing CSVs and Excel spreadsheets, extra data is handled differently. For Excel spreadsheets, we try to parse the value according to its actual type using cell metadata:
https://github.com/SEED-platform/seed/blob/8d838f1aa4fc10e656e97c29e7c2e5578a152998/seed/lib/mcm/reader.py#L346

For CSVs, we don't have this metadata so everything is parsed as strings (as far as I can tell). So all extra_data values are stored as strings in the DB. This obviously causes a problem if the user says their extra data column "foo" is a numeric type, but they imported the data from a CSV meaning filter comparisons (e.g. <, >, etc) won't work as expected.

## solutions
### casting
One approach is to use query annotations to cast json values into the type the user has specified according to column settings. This probably has performance implications, and definitely needs testing (e.g. how does it handle casting a series of values where not all can be cast to a valid value?). An example of how this query could work, assuming the column of interest is extra_data's "foo" field, the actual numeric field we care about ends up in the "foo_to_float" field:
```
PropertyState.objects.annotate(
  foo_to_text=Cast('extra_data__foo', output_field=models.TextField()), 
  foo_cleaned=Replace('foo_to_text', Value('"'), output_field=models.TextField()),
  foo_to_float=Cast('foo_cleaned', output_field=models.FloatField())
)
```
This is a three step process I found necessary:
  - convert the field to text (we're unable to cast jsonb directly to float)
  - remove the double quotes surrounding the text value (this is added by postgres to allow forwards/backwards translation to json)
  - finally cast the value to a float

Pros:
- requires no changes to the database or users' data
Cons:
- TBD: performance cost
- TBD: is this actually possible?

### clean the data
The second approach is to make sure our data in the DB in extra_data matches the expected type. This would require a few changes in SEED:
1. users must define data_type for all columns (not required currently)
2. on import, we try to cast the value to the specified column type, if it fails, it's set to null or something like that
[...]

PR #3070, opened 2022-01-06
Refactor/inventory filter

macintoshpie:
SEED is creating a paginated inventory list to improve performance. This requires moving sorting and filtering to the backend.

#### What's this PR do?
- updates backend to properly handle sorting and filtering for all extra_data columns (see linked issue for more context)
  - accomplished by casting any filtered/ordered extra_data column into the data_type defined by it's Column, and using that casted column for any filtering/ordering
- this comes with the caveat that if one or more extra_data values for the user's properties cannot be cast to the defined column type, the cast will fail and the user cannot filter/sort on that column. E.g., if the user has a numeric extra_data column `foo`, and one of the properties has the value `hello`, when trying to filter/sort on `foo` it will fail!

- upload a CSV (not an excel file!) that contains numeric data. map some of the numeric columns to extra data
- go to the beta inventory list page, and try filtering and sorting on that column, by default it won't work as expected b/c of your column settings (we will treat it as a string)
- go to your organization column settings page, and update these columns to be numeric or integers, then try filtering and sorting again and it should work as expected.
- also try setting the column type to something it cannot be cast to (e.g. on an extra data column containing string data, set it to numeric). When you try to filter/sort on that column it should fail gracefully with an error message about your "bad data"
- also assess the performance of this, if you have a cycle with a ton of properties, test the filter and sort on one of the extra_data columns and see if it feels sluggish

#3042 

Setup column settings:

Screen Shot 2022-01-06 at 11 21 08 AM
Screen Shot 2022-01-06 at 11 20 56 AM
Screen Shot 2022-01-06 at 11 20 15 AM

Demo showing how _uncasted_ (i.e. there's no data type specified) extra data columns fail to work when trying to filter

https://user-images.githubusercontent.com/18518728/148414062-fb2c696c-203f-43bb-a224-9f96bdc5bfd0.mov

Demo showing how _uncasted_ extra data columns fail to work when trying to sort

https://user-images.githubusercontent.com/18518728/148414160-4b6eb629-3f0a-4598-9e08-07e59a4b9614.mov

Demo showing how _casted_ (i.e. the column's data type is _number_ in the org setting) extra data columns work when trying to filter

https://user-images.githubusercontent.com/18518728/148414382-cf765a9b-ad33-4ed3-8108-e10b29a9bf94.mov

Demo showing how _casted_ extra data columns work when trying to sort

https://user-images.githubusercontent.com/18518728/148414485-51f2f97c-2e86-4fd5-8fb5-4b2a508f8cd0.mov

Screenshot example of error message when the user has data which does not match the Column.data_type (trying to cast a non-number to number)
Screen Shot 2022-01-06 at 11 22 39 AM

### DIFF
--- a/seed/search.py
+++ b/seed/search.py
@@ -19,7 +19,9 @@ from enum import Enum
 
 from functools import reduce
 
+from django.db import models
 from django.db.models import Q
+from django.db.models.functions import Cast, Replace
 from django.http.request import RawPostDataException, QueryDict
 from past.builtins import basestring
 
@@ -359,7 +361,42 @@ class QueryFilter:
             return Q(**q_dict)
 
 
-def _parse_view_filter(filter_expression: str, filter_value: str, columns_by_name: dict[str, dict]) -> Q:
+def _build_extra_data_filter_and_annotations(filter: QueryFilter, data_type: str) -> tuple[QueryFilter, dict]:
+    field_name = filter.field_name
+    full_field_name = f'state__extra_data__{field_name}'
+    text_field_name = f'_{field_name}_to_text'
+    cleaned_field_name = f'_{field_name}_cleaned'
+    final_field_name = f'_{field_name}_final'
+
+    annotations = {
+        text_field_name: Cast(full_field_name, output_field=models.TextField()),
+        cleaned_field_name: Replace(text_field_name, models.Value('"'), output_field=models.TextField()),
+    }
+    if data_type == 'integer':
+        annotations.update({
+            final_field_name: Cast(cleaned_field_name, output_field=models.IntegerField())
+        })
+    elif data_type in ['number', 'float', 'area', 'eui']:
+        annotations.update({
+            final_field_name: Cast(cleaned_field_name, output_field=models.FloatField())
+        })
+    elif data_type in ['date', 'datetime']:
+        annotations.update({
+            final_field_name: Cast(cleaned_field_name, output_field=models.DateTimeField())
+        })
+    elif data_type == 'boolean':
+        annotations.update({
+            final_field_name: Cast(cleaned_field_name, output_field=models.BooleanField())
+        })
+    else:
+        # treat it as a string
+        cleaned_value = annotations.pop(cleaned_field_name)
+        annotations[final_field_name] = cleaned_value
+
+    return QueryFilter(final_field_name, filter.operator, filter.is_negated), annotations
+
+
+def _parse_view_filter(filter_expression: str, filter_value: str, columns_by_name: dict[str, dict]) -> tuple[Q, dict]:
     """Parse a filter expression into a Q object
 
     :param filter_expression: should be a valid Column.column_name, with an optional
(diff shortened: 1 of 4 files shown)


seed/views/v3/properties.py
```diff
@@ -198,6 +199,14 @@ def _get_filtered_results(self, request, profile_id):
         except EmptyPage:
             property_views = paginator.page(paginator.num_pages)
             page = paginator.num_pages
+        except DataError as e:
+            return JsonResponse(
+                {
+                    'status': 'error',
+                    'message': f'Error filtering - your data might not match the column settings data type: {str(e)}'
```
macintoshpie, line 206: Looking for feedback on this -- not sure what a good error message would be. Is this semi-clear?

perryr16: Overall I think it's clear. If I had to critique:
- Include the column name. "Error filtering <column_name>...". 
- Could be useful to provide a link to the column settings page in the error message

perryr16 (approved):
This worked as expected when tested, thanks for the demo videos. I think the error message is ok (see my optional feedback).

RDmitchell:
@macintoshpie -- is this going to cause issues with existing data for the extra_data fields, if they were not properly defined originally?

macintoshpie:
@RDmitchell No this will not cause an issues with existing data -- we aren't modifying the data in any way

seed/views/v3/properties.py
Ryoken: When a correctly typed extra_data column has empty fields you will get errors.  Should (could?) we automatically handle empty and null fields by failing all filter criteria?
Screen Shot 2022-01-06 at 2 11 43 PM

Ryoken:
When a correctly typed extra_data column has empty fields you will get errors.  Should (could?) we automatically handle empty and null fields by failing all filter criteria?
Screen Shot 2022-01-06 at 2 11 43 PM

macintoshpie:
> When a correctly typed extra_data column has empty fields you will get errors. Should (could?) we automatically handle empty and null fields by failing all filter criteria?

Oooo yuck, we definitely do need to handle that... I need to do a bit of research on our options.

macintoshpie:
I'll mark this as a draft and try to fix the issue @Ryoken pointed out

seed/search.py
```diff
@@ -309,7 +315,111 @@ class FilterException(Exception):
...
+    stripped_field_name = f'_{column_name}_stripped'
+    cleaned_field_name = f'_{column_name}_cleaned'
+    final_field_name = f'_{column_name}_final'
+
+    annotations: AnnotationDict = {
+        text_field_name: Cast(full_field_name, output_field=models.TextField()),
+        # after casting a json field to text, the resulting value will be wrapped
+        # in double quotes which need to be removed
+        stripped_field_name: Replace(text_field_name, models.Value('"'), output_field=models.TextField()),
+        cleaned_field_name: NullIf(stripped_field_name, models.Value('null'), output_field=models.TextField())
```
macintoshpie, line 394: now properly handles null values in extra_data

macintoshpie:
@Ryoken should be fixed for null values now, could you give that another go?

seed/views/v3/properties.py
macintoshpie: > Include the column name.
Wish I could, but I don't think I have that information in the error object, I need to investigate this...

> Could be useful to provide a link to the column settings page in the error message
Not a bad idea, I'll see if I can add a link

Ryoken:
Filtering a very large set of data on an extra_data column clocked in on average at 1.31 seconds.  This is still over twice as fast as before pagination, so looks good!

Ryoken (approved):

Issue #3071, opened 2022-01-06
bug: saved filter/sort results in multiple initial queries


macintoshpie:
## context
SEED is creating a paginated version of the inventory list (currently beta). The frontend will save any filters/sorts the user applies (in localstorage) so that if the user goes to a different page then goes back to the inventory list page, it will load those filters/sorts.

It appears that the new implementation causes multiple calls to the API though, when we only want to do this once (3 times in this test):
<img width="699" alt="Screen Shot 2022-01-06 at 2 31 50 PM" src="https://user-images.githubusercontent.com/18518728/148440697-e8416d16-6bd2-455b-aa6c-148017c1ec25.png">

## task
- only hit the backend once (once the filters are fully determined/added to the table)

PR #3072, opened 2022-01-06
refactor: avoid multiple calls to load inventory

macintoshpie:
SEED is creating a paginated version of the inventory list

#### What's this PR do?
- fixes bug where the inventory was being fetched multiple times on page load. TLDR there's an issue where loading ui-grid settings it triggers our filter/sort event handlers when we don't want it to.

- test loading the beta inventory page with filters applied, then go to a different page, then go back. The network tab in dev tools should only show a single fetch to `/filter`

#3071


Ryoken (approved):
Ah, nice, I was hoping to address this multiple load thing.  Good simple state machine to handle it, too.

Issue #3073, opened 2022-01-06
Refactor Beta-Inventory List Filters View


Ryoken:
The filters and controls on the beta-inventory list need to be re-organized and formated to better communicate to users what is being filtered and how to update said filters.

RDmitchell:
Instance: dev1
SHA: 9a3a23722

This all looks good to me !!

PR #3075, opened 2022-01-07
Refactor/organize filter and labels

Ryoken:
see ticket

#### What's this PR do?
refactors the beta inventory list filters and settings

try filtering on the table and note the new filter in the upper right (see screenshots for examples)

https://github.com/SEED-platform/seed/issues/3073

If available, filtering by a label:
Screen Shot 2022-01-06 at 3 52 09 PM

Table filters show up in the filter list:
Screen Shot 2022-01-06 at 3 51 46 PM

Splits multiple column filters:
Screen Shot 2022-01-06 at 5 04 04 PM

Deleting only part of a filter updates the column filter:
Screen Shot 2022-01-06 at 5 04 15 PM

Actions have been organized and converted to a dropdown menu:
Screen Shot 2022-01-06 at 5 09 34 PM


seed/static/seed/js/controllers/inventory_list_beta_controller.js
```diff
@@ -122,6 +122,20 @@ angular.module('BE.seed.controller.inventory_list_beta', [])
       $scope.column_filters = []
       $scope.column_sorts = []
 
+      // remove editing on the filter list input (ngTagsInput doesn't support readonly yet)
+      let findFiltersListAttempts = 0;
+      const findFiltersList = setInterval(() => {
```
macintoshpie, line 127: so this makes the filters input read only? not crazy about the setInterval, but looks like its our best option

seed/static/seed/js/controllers/inventory_list_beta_controller.js
```diff
@@ -1136,6 +1169,33 @@ angular.module('BE.seed.controller.inventory_list_beta', [])
         }
       };
 
+      const operatorLookup = {
+        'ne': '!',
```
macintoshpie, line 1173: I think this might need to be `!=`, see video below -- `!=` end up turning into a `contains !<value>`

https://user-images.githubusercontent.com/18518728/148571455-5684d3df-cc1e-4889-896f-4a8b52e22f10.mov

seed/static/seed/js/controllers/inventory_list_beta_controller.js
```diff
@@ -1222,8 +1282,9 @@ angular.module('BE.seed.controller.inventory_list_beta', [])
               const subFilters = _.map(_.split(filter.term, ','), _.trim);
               for (const subFilter of subFilters) {
                 if (subFilter) {
-                  const {operator, value} = parseFilter(subFilter)
-                  $scope.column_filters.push({column_name, operator, value})
+                  const {string, operator, value} = parseFilter(subFilter)
+                  const display = [column_name, string, value].join(' ')
```
macintoshpie, line 1286: Do you think we could display the column's "display name" instead of the internal column_name here? Might be a better experience for users (e.g. showing `"Site EUI" < 10` instead of `site_eui < 10`)

macintoshpie (changes requested):
This is awesome! So nice to be able to see what's actually applied. A few comments:

1. Could we update the CSS s.t. it doesn't mess up the original inventory list view?
Screen Shot 2022-01-07 at 11 14 28 AM

2. I do think it would be useful to see the currently applied sort at the top, and providing a shortcut to "remove" it

3. For the new "Current filters" box, could we:
  - change the mouse cursor so that it's not the "text" cursor when hovered over (since we can't actually type stuff into it)
  - make the box + trashcan appear disabled when there aren't any filters applied (like we do for the labels box)
  - add a placeholder to the box which says something like "Add filters by updating the filter above the column" or something like that so users know that they can't use that box itself to add filters?

seed/static/seed/js/controllers/inventory_list_beta_controller.js
Ryoken: Ah yeah, good catch

seed/static/seed/js/controllers/inventory_list_beta_controller.js
Ryoken: I had to collect all the display names on page load to support this but it's working.  It isn't too terribly helfpul for extra_data though!

Ryoken:
Screen Shot 2022-01-11 at 5 06 58 PM

Added sorting, removed all editable styles and added a placeholder.  The way the tag plugin works the placeholder is always displayed so I made it a shorter string.  If this looks good I'll update translations and merge!

RDmitchell:
@Ryoken / @macintoshpie / @axelstudios / @nllong -- this screen seems very busy. 

I think there might need to be a bit of UX design applied. 

Any reason we can't keep Cycle and Column List Profile where they currently are?

macintoshpie:
@RDmitchell we can definitely iterate on this deign, but I think we'd like to get this first stab at it merged in. Then we can deploy it for you and you can test it and give us your feedback.

macintoshpie (approved):
Great!

RDmitchell:
Sure, that's fine

Issue #3046, opened 2021-12-17
enable actions for current page and across entire result set (on inventory list beta page)


macintoshpie:
## context
SEED is creating a paginated inventory list (the Beta inventory list). I.e. you can only see like 100 properties at a time on each page. This means running "actions" on selected properties needs to be modified (since you can currently only select properties on the page you're viewing)

## task
- Create new UI on frontend, where when the users clicks on the "select all" button (see image below) it will pop up a notification saying "All {num properties on page} on this page are selected. Select all {total properties from pagination data} properties". The "select all ... properties" text should be a link which when clicked just logs "hello world" or something (i.e. does nothing), then shows a message like "All {whatever} properties are selected. clear selection" where clicking clear selection will clear the selection

Seed's select all button:
<img width="547" alt="Screen Shot 2022-01-06 at 12 22 34 PM" src="https://user-images.githubusercontent.com/18518728/148424050-265d6bf6-fc65-45d2-852e-7a45f58524c0.png">

E.g. from gmail:
<img width="1291" alt="Screen Shot 2022-01-06 at 12 19 50 PM" src="https://user-images.githubusercontent.com/18518728/148423651-5dc17349-ebd2-445b-892d-db0f630b79e6.png">
<img width="1284" alt="Screen Shot 2022-01-06 at 12 19 56 PM" src="https://user-images.githubusercontent.com/18518728/148423669-c43c3056-b64f-44ce-9a80-1d4cc4455d65.png">

macintoshpie:
Some ideas:
- create/update an API so that we can fetch all inventory IDs for a given filter/sort query, then use those IDs when POSTing to the specific action's API
- update each action's API to allow filter/sort expressions, then they operate on that queryset

PR #3097, opened 2022-01-12
Feature/enable actions frontend

Ryoken:
See ticket.

#### What's this PR do?
Adds the ability to select the entire filtered set of properties and displays related feedback.

Load the inventory beta page and select properties (see screenshots).

https://github.com/SEED-platform/seed/issues/3046

At first nothing is selected.  An action for "Select All" appears and the action optgroup displays "0 selected":
Screen Shot 2022-01-12 at 2 49 59 PM

Selecting 1 property from the list updates the actions menu as well as displaying next to the pagination links:
Screen Shot 2022-01-12 at 2 49 32 PM

Clicking the "select all" checkmark in the table header will select the entire page:
Screen Shot 2022-01-12 at 2 50 10 PM

Clicking "select all" either in the actions menu or by the pagination links will update the selected display, remove/disable the "select all" link/action and check all visible rows on the table:
Screen Shot 2022-01-12 at 2 50 27 PM

When everything has been selected, any changes to the selection (including unchecking anything in the table) will update with only what's visibly selected:
Screen Shot 2022-01-12 at 2 50 36 PM


Ryoken:
Todo:  Update the selection when changing the page

seed/static/seed/js/controllers/inventory_list_beta_controller.js
```diff
@@ -1108,33 +1064,57 @@ angular.module('BE.seed.controller.inventory_list_beta', [])
...
+        // was the function called with a list of ids?
+        if (viewIds.length > 0) {
+          selectedViewIds = viewIds;
+
+        // if it appears everything selected, only get the full set of ids...
+        } else if ($scope.selectedCount == $scope.inventory_pagination.total) {
+          selectedViewIds = [];
+
+          $http.post('/api/v3/properties/filter/', {}, {
+            params: {
```
Ryoken, line 1079: Should we worry about the current cycle here?

perryr16 (approved):
Nice, this works as expected! I was able to select a single, then all 2500 properties and delete the selected properties. 

This doesn't prevent me from approving this PR, but my only suggestion would be to add a `Clear Selection` option in the dropdown and in the selection recap text. 
Screen Shot 2022-01-27 at 2 04 08 PM

perryr16 (changes requested):
One change request. After I deleted a partial selection of properties the inventory screen did not update (including "100 properties selected...") text. Can you refresh the page following an action?

https://user-images.githubusercontent.com/58446472/151444215-fd969c3e-58f5-4b3f-b7c1-e36211fd1cf1.mov

seed/static/seed/js/controllers/inventory_list_beta_controller.js
macintoshpie: That's probably not a bad idea, Let's add cycle here

seed/static/seed/js/controllers/inventory_list_beta_controller.js
```diff
@@ -1108,33 +1064,57 @@ angular.module('BE.seed.controller.inventory_list_beta', [])
...
+              organization_id: $scope.organization.org_id,
+              ids_only: true
+            }
```
macintoshpie, line 1082: As written, it looks like this will not appliy the user's filters and sorting, so I think we should add that here.
It might be better to use the inventory_service for the call as well instead of using $http

macintoshpie (changes requested):
Lookin good for the most part!

Ryoken:
Updated to use inventory_service, added select_none and ensured page updated after actions.

Ryoken:
Added translations.

seed/static/seed/js/services/inventory_service.js
```diff
@@ -54,14 +54,19 @@ angular.module('BE.seed.service.inventory', []).factory('inventory_service', [
...
         organization_id: organization_id,
-        page: page,
-        per_page: per_page || 999999999,
         include_related: include_related,
         ids_only: ids_only,
         ...format_column_sorts(column_sorts),
         ...format_column_filters(column_filters),
       };
 
+      if (ids_only) {
```
Ryoken, line 63: The endpoint wont allow ids_only and page/per_page at the same time.

macintoshpie: Nice!

macintoshpie (approved):
💯  Great job! This works so well - I'm very pleased with this

perryr16 (approved):
Works like a charm, thanks for making those changes.

Issue #3081, opened 2022-01-10
refactor: move filtering of querysets into generalized function


macintoshpie:
## context
SEED is creating a paginated inventory list page. There is duplicated code in both the taxlot and property `/filter` endpoint. Specifically, we want to create a helper function which can create a PropertyView or TaxlotView QuerySet based on the provided info (filter query params, cycle, org, etc)

## tasks
- move the parts which construct the *View queryset into a function that's usable for both taxlots and properties
- use the function in the properties and taxlots `_get_filtered_results` methods

nllong:
Not really an easy way to test this on the front end. Closing as this has been complete for a few months.

PR #3100, opened 2022-01-13
3081 refactor/util property taxlot fiter

perryr16:
When viewing the property or taxlot inventory a user may filter by column contents. The backend code to filter properties or taxlots is identical and duplicated across two functions. 

#### What's this PR do?
This PR moves the functionality to a single helper function that can be imported to both property and taxlot files. There are slight differences between the two and a new argument `state_type` has been introduced to differentiate between 'property' and 'taxlot' filtering. 

Filter the inventory for both properties and taxlots and ensure correct functionality.

#3081 

n/a

### DIFF
--- a/seed/utils/filter_state.py
+++ b/seed/utils/filter_state.py
@@ -1,0 +1,169 @@
+"""
+:copyright (c) 2014 - 2022, The Regents of the University of California, through Lawrence Berkeley National Laboratory (subject to receipt of any required approvals from the U.S. Department of Energy) and contributors. All rights reserved.  # NOQA
+:author
+"""
+from typing import Optional
+
+from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
+from django.db.utils import DataError
+from django.http import JsonResponse
+from rest_framework import status, viewsets, generics
+from rest_framework.request import Request
+from seed.lib.superperms.orgs.models import Organization
+from seed.models import (VIEW_LIST, VIEW_LIST_PROPERTY,
+                        Column, ColumnListProfile,
+                         ColumnListProfileColumn,  Cycle,
+                         PropertyView)
+from seed.models import TaxLotProperty, TaxLotView
+from seed.search import build_view_filters_and_sorts, FilterException
+from seed.serializers.pint import (apply_display_unit_preferences)
+
+
+def _get_filtered_results(request: Request, profile_id: int, state_type: str='property'):
+        page = request.query_params.get('page', 1)
+        per_page = request.query_params.get('per_page', 1)
+        org_id = request.query_params.get('organization_id')
+        cycle_id = request.query_params.get('cycle')
+        # check if there is a query paramater for the profile_id. If so, then use that one
+        profile_id = request.query_params.get('profile_id', profile_id)
+
+        if not org_id:
+            return JsonResponse(
+                {'status': 'error', 'message': 'Need to pass organization_id as query parameter'},
+                status=status.HTTP_400_BAD_REQUEST)
+        org = Organization.objects.get(id=org_id)
+
+        if cycle_id:
+            cycle = Cycle.objects.get(organization_id=org_id, pk=cycle_id)
+        else:
+            cycle = Cycle.objects.filter(organization_id=org_id).order_by('name')
+            if cycle:
+                cycle = cycle.first()
+            else:
+                return JsonResponse({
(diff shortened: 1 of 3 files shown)


seed/utils/filter_state.py
```diff
@@ -0,0 +1,175 @@
...
+from seed.models import (VIEW_LIST, VIEW_LIST_PROPERTY, VIEW_LIST_TAXLOT,
+                         Column, ColumnListProfile, ColumnListProfileColumn,
+                         Cycle, PropertyView)
+from seed.models import TaxLotProperty, TaxLotView
+from seed.search import build_view_filters_and_sorts, FilterException
+from seed.serializers.pint import (apply_display_unit_preferences)
+from typing import Union, Optional, Literal
+
+
+def _get_filtered_results(request: Request, state_type: Union[Literal['property'], Literal['taxlot'], None], profile_id: int):
```
perryr16, line 21: The typing should be `state_type: str` however mypy threw errors. I'll look into this as it's not clear at the moment.

seed/utils/filter_state.py
```diff
@@ -0,0 +1,175 @@
...
+    page = request.query_params.get('page', 1)
+    per_page = request.query_params.get('per_page', 1)
+    org_id = request.query_params.get('organization_id')
```
perryr16, line 24: changed from `org_id = self.get_organization(request)`, however this non-class function does not have access to `self` methods.

seed/utils/filter_state.py
macintoshpie, line 21: Will have to import HttpResponse from django.http -- I think this makes it a bit clearer that it's fully handling the HTTP response as well as filtering.

I think we should be a bit more general about the `state_type` name since it's really filtering on *Views, so maybe use `inventory_type` instead

I removed the `Union` with `None` -- I'm not sure if we want to allow users to pass None here right? What would that mean? I also simplified the `Literal`s, you can list multiple literals in the annotation

Also, I tweaked the function name. I removed the leading `_` b/c it's not really a private function -- we're using it in other files. Also I think a more general name (inventory) instead of state might be clearer b/c this is technically filtering for *Views

```suggestion
def get_filtered_inventory(request: Request, inventory_type: Literal['property', 'taxlot'], profile_id: int) -> HttpResponse:
```

seed/utils/filter_state.py
```diff
@@ -0,0 +1,175 @@
...
+            },
+            status=status.HTTP_400_BAD_REQUEST
+        )
+
+    # This uses an old method of returning the show_columns. There is a new method that
+    # is prefered in v2.1 API with the ProfileIdMixin.
+    if state_type == 'property':
+        view_list_state = VIEW_LIST_PROPERTY
+    elif state_type == 'taxlot':
+        view_list_state = VIEW_LIST_TAXLOT
```
macintoshpie, line 117:
```suggestion
    if state_type == 'property':
        profile_inventory_type = VIEW_LIST_PROPERTY
    elif state_type == 'taxlot':
        profile_inventory_type = VIEW_LIST_TAXLOT
```

I think this might be a bit more clear about what it is (an enum for ColumProfile.inventory_type)

macintoshpie (changes requested):
ya love to see this much code deletion! This is great, just a few comments about some naming/typing.

Screen Shot 2022-01-14 at 9 33 45 AM

seed/utils/filter_state.py
```diff
@@ -0,0 +1,175 @@
...
+            status=status.HTTP_400_BAD_REQUEST)
+    org = Organization.objects.get(id=org_id)
+
+    if cycle_id:
+        cycle = Cycle.objects.get(organization_id=org_id, pk=cycle_id)
+    else:
+        cycle = Cycle.objects.filter(organization_id=org_id).order_by('name')
+        if cycle:
+            cycle = cycle.first()
+        else:
```
Ryoken, line 41: This might be small but would we benefit from changing 'else' to 'if not' and bumping it back an indent?  Then we could check both if finding a cycle and if looking up a cycle by id failed.

Ryoken (approved):
Looks good!  Made one optional suggestion.

seed/utils/filter_state.py
perryr16: This makes sense to me

macintoshpie (approved):
💯  v nice

PR #3101, opened 2022-01-14
Update docker's pgdata path

nllong:
We fixed this in helm, but missed doing it in these files. For unknown reasons, the `pgdata` field needs to be passed to the containers now. The postgres base image may have changed.

#### What's this PR do?
* update the docker-compose files by adding in the `pgdata` env var.

* create required docker volumes (seed_pgdata, seed_media)
* docker-compose up 
* verify that the database authorization works and that you can log into the site

n/a

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -10,8 +10,6 @@ version: '3.4'
 services:
   db-postgres:
     container_name: seed_postgres
-    environment:
-      - PGDATA=/var/lib/postgresql/data/pgdata
     volumes:
       - ref_seed_pgdata:/var/lib/postgresql/data
   web:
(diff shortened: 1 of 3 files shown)


docker-compose.dev.yml
```diff
@@ -10,6 +10,8 @@ version: '3.4'
 services:
   db-postgres:
     container_name: seed_postgres
+    environment:
+      - PGDATA=/var/lib/postgresql/data/pgdata
```
macintoshpie, line 14: docker-compose.dev is intended to extend docker-compose.yml, so I don't think we have to redeclare this
```suggestion
```

macintoshpie (approved):
Interesting, curious what tipped you off on this problem?

Issue #3093, opened 2022-01-11
refactor: enable inventory `/filter` endpoints to return only IDs


macintoshpie:
## context
SEED is creating a paginated inventory list. All of the "Actions" require the IDs of the selected properties, but since we are not fetching all properties (b/c we are paginating) if the user wants to "select all" before an action, we need to fetch all of the IDs that match for usage in the "Action".

## task
- add a query parameter to the properties and taxlots `/filter` endpoint which will result in only the IDs of the *Views being returned (rather than all of the column data). By default, this option should be False if not explicitly provided to support backwards compatibility. Not sure what the best param name would be, maybe something like `only_ids=true`, `only_ids=false`

nllong:
Not really an easy way to test this on the front end. Closing as this has been complete for a few months.

PR #3102, opened 2022-01-14
3093 refactor/inventory filter ids only

perryr16:
With paginated inventory lists, it's not critical to return the entire column data when selecting properties or taxlots that are not on page 1.

#### What's this PR do?
Based on a query parameter `?ids_only=true`, the backend will return a json list of property or taxlot ids. If this query param is not passed the function will default to the existing functionality allowing backward compatibility. This should improve performance when applying filters to the entire inventory.

- Run new test /seed/tests/test_views.py:InventoryViewTests.test_get_properties_select_all. 
- Can be tested when the front end "select all" features are implemented.

#3093

### DIFF
--- a/seed/tests/test_views.py
+++ b/seed/tests/test_views.py
@@ -697,7 +697,7 @@ class InventoryViewTests(AssertDictSubsetMixin, DeleteModelsTestCase):
             'ids_only', 'true'
         ), data={}, content_type='application/json')
         self.assertEqual(response.status_code, 400)
-        result = response.json() 
+        result = response.json()
         self.assertEqual(result['success'], False)
         self.assertEqual(result['message'], 'Cannot pass query parameter "ids_only" with "per_page" or "page"')
 
@@ -715,7 +715,6 @@ class InventoryViewTests(AssertDictSubsetMixin, DeleteModelsTestCase):
             'ids_only', 'true'
         ), data={}, content_type='application/json')
         self.assertEqual(response.status_code, 400)
-
 
     def test_get_properties_pint_fields(self):
         state = self.property_state_factory.get_property_state(
--- a/seed/tests/test_views.py
+++ b/seed/tests/test_views.py
@@ -662,6 +662,15 @@ class InventoryViewTests(AssertDictSubsetMixin, DeleteModelsTestCase):
         self.assertEqual(len(results), 4)
         self.assertEqual(results, [1001, 1002, 1003, 1004])
 
+        response = self.client.post('/api/v3/properties/filter/?{}={}&{}={}'.format(
+            'organization_id', self.org.pk,
+            'ids_only', 'TrUE',
+        ), data={}, content_type='application/json')
+        result = response.json()
+        results = result['results']
+        self.assertEqual(len(results), 4)
+        self.assertEqual(results, [1001, 1002, 1003, 1004])
+
         response = self.client.post('/api/v3/properties/filter/?{}={}&{}={}&{}={}&{}={}'.format(
             'organization_id', self.org.pk,
             'page', 1,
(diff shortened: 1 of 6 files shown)


Ryoken:
I think it's looking good.  It is still possible to get a filtered list of ids only with this method but that still works for what we need (just pass in no filters).

Ryoken (approved):

seed/utils/inventory_filter.py
```diff
@@ -88,6 +88,12 @@ def get_filtered_results(request: Request, inventory_type: Literal['property', '
     if f'{inventory_type}_view_ids' in request.data and request.data[f'{inventory_type}_view_ids']:
         views_list = views_list.filter(id__in=request.data[f'{inventory_type}_view_ids'])
 
+    if request.query_params.get('ids_only') == 'true':
+        id_list = [view.id for view in views_list]
+        return JsonResponse({
+            'results': id_list
+        })
```
macintoshpie, line 95: I think it _might_ be more efficient to do something like this instead to get the ids -- I think this would make it happen at the DB level instead of evaluating the entire view in Python:
```
id_list = views_list.values_list('id', flat=True)
```

Also, I'm a bit conflicted about how this query param results in the pagination parameters being ignored... For our use case it definitely makes sense, but is a bit unexpected from an API perspective I think... Thoughts?

macintoshpie (changes requested):
Nice! One comment

seed/utils/inventory_filter.py
macintoshpie: Ah yeah, here's a note from the docs:
```
values() and values_list() are both intended as optimizations for a specific use case: retrieving a subset of data without the overhead of creating a model instance. This metaphor falls apart when dealing with many-to-many and other multivalued relations (such as the one-to-many relation of a reverse foreign key) because the “one row, one object” assumption doesn’t hold.
```
https://docs.djangoproject.com/en/4.0/ref/models/querysets/#django.db.models.query.QuerySet.values_list

Ryoken: hmm... what if we renamed the query parameter to "all_ids_only" or something similar?

macintoshpie: Hmm, kind of lead me to believe it's not applying the filtering (which it still is). I think the least surprising option is to just paginate on this as well. On frontend we can pass some large page size number for this request, and/or loop until all pages are fetched

Another option could be to assert that page/page_size aren't passed if `ids_only=true`, return a 400 (bad request) and indicate in the message why they can't do that?

macintoshpie:
@perryr16 another thing I missed in the review, we need to update the docs for the endpoints, e.g. here for properties:
https://github.com/SEED-platform/seed/blob/327b140ff9f08b139b98b5b7b71f0efc05f51e3d/seed/views/v3/properties.py#L333

seed/utils/inventory_filter.py
```diff
@@ -49,6 +50,15 @@ def get_filtered_results(request: Request, inventory_type: Literal['property', '
...
         })
 
+    if ids_only and (per_page or page):
+        return JsonResponse({
+            'success': False,
+            'message': 'Cannot pass query parameter "ids_only" with "per_page" or "page"'
+        }, status=status.HTTP_400_BAD_REQUEST)
+
+    page = page or 1
+    per_page = per_page or 1
```
perryr16, line 60: assign default values to page and per_page after evaluating if the correct query params were passed

seed/tests/test_views.py
```diff
@@ -653,17 +653,24 @@ def test_get_properties_select_all(self):
...
             'ids_only', 'true',
         ), data={}, content_type='application/json')
         result = response.json()
         results = result['results']
         self.assertEqual(len(results), 4)
         self.assertEqual(results, [1001, 1002, 1003, 1004])
 
+        response = self.client.post('/api/v3/properties/filter/?{}={}&{}={}'.format(
+            'organization_id', self.org.pk,
+            'ids_only', 'TrUE',
```
macintoshpie, line 667: gEnErAl bOoLeAns

macintoshpie (approved):
💯

Issue #3020, opened 2021-11-30
Backup prod postgres docker volume


macintoshpie:
## context
Historically we've created a postgres dump for the production database.

We would like to create a backup for the postgres volume _itself_ -- not just run pg dump and store that result.

## task
- update the backup script to tar the docker volume and store it

macintoshpie:
@axelstudios @nllong Is this the script which does backups for production?
https://github.com/SEED-platform/seed/blob/develop/bin/backup_database.sh

axelstudios:
@macintoshpie That's the script that's currently used in production (via cronjob), there's a [different script for k8s](https://github.com/SEED-platform/seed/blob/develop/docker/backup_k8s/backup_database.sh)

nllong:
> @macintoshpie That's the script that's currently used in production (via cronjob), there's a [different script for k8s](https://github.com/SEED-platform/seed/blob/develop/docker/backup_k8s/backup_database.sh)

Just echoing Alex's comment. Note that the k8s script is built into [this docker](https://hub.docker.com/repository/docker/seedplatform/seed-backup-k8s) container and that is what is used in the k8s CronJob task.

haneslinger:
[Docker recommends](https://docs.docker.com/storage/volumes/#backup-restore-or-migrate-data-volumes) backing up a volume by simply launching a container, giving it the volume, and taring it. idk if this would be much faster than I pg_dump -  what's more how safe it is.

[Postgres say](https://www.postgresql.org/docs/11/backup-file.html)s backing up via a tar /usr/local/pgsql/data  is an inferior method it pg_dump as it would require stoping the server, but it goes on to recommend making a "consistent snapshot" of the whole data directory instead, which wouldn't  require stopping the db.

I'm not sure to what "consistent snapshot" refers, but a quick google found me [Kubes Volume Snaphots](https://kubernetes.io/docs/concepts/storage/volume-snapshots/). idk if it stores the whole thing or just diagnostics

nllong:
Perhaps we store both the pg_dump version (which is what we have now) and a tar'd up volume. I would think that a shared tarball would be must faster to restore than restoring a whole pg dump. Maybe k8s volume snapshots are the way to go... give it a whirl. Let us know if you want us to test restoring a tarball volume!

PR #3103, opened 2022-01-14
Backup prod postgres docker volume

haneslinger:

This is a script to restore the dumped back up db and tar it. It's much quick to open that way. Called [here](https://github.com/SEED-platform/seed-helm-charts/pull/3)
https://github.com/SEED-platform/seed/issues/3020


nllong (approved):
looks great! Did we get past the disk pressure issue now?

haneslinger:
yes! The disk pressure was not coming from the db itself, but the process of restoring it. We now do that process in the volume scratch, which has plenty of storage.

PR #3104, opened 2022-01-18
Merge building sync files

haneslinger:

this is drafted. I'm just working out the necessary changes before rebaseing.


seed/models/building_file.py
```diff
@@ -378,11 +396,10 @@ def process(self, organization_id, cycle, property_view=None):
 
             # set the property_state to the new one
             property_state = merged_state
-        elif not property_view:
+        elif not property_view and promote_property_state:
             property_view = property_state.promote(cycle)
         else:
-            # invalid arguments, must pass both or neither
-            return False, None, None, "Invalid arguments passed to BuildingFile.process()"
```
haneslinger, line 385: I don't understand how this else would ever trigger. Do `PropertyViews` evaluate weird?

seed/data_importer/match.py
```diff
@@ -115,11 +115,21 @@ def match_and_link_incoming_properties_and_taxlots(file_pk, progress_key):
     tax_lot_initial_incoming_count = incoming_tax_lots.count()
 
     if incoming_properties.exists():
+        # If importing BuildingSync, we will not just skip duplicates like we normally
+        # do. Since we don't skip them, they will eventually get merged into their "duplicate".
+        # We do this b/c though the property data might be the same, the Scenarios, Measures,
+        # or Meters might have been updated. The merging flow is able to "transfer"
+        # this data, while skipping duplicates cannot.
```
haneslinger, line 122: maybe this goes in the docstring?

macintoshpie: Hmm, I could go either way. I think since this is more implementation specific a user using the function shouldn't need to be aware of this, and I kind of like that it's close to the code it's relevant to

seed/data_importer/tasks.py
```diff
@@ -1137,40 +1137,30 @@ def geocode_and_match_buildings_task(file_pk):
...
-        source_type = source_type_dict.get(import_file.source_type, ASSESSED_RAW)
-
-        # get the properties and chunk them into tasks
-        qs = PropertyState.objects.filter(
-            import_file=import_file,
-            source_type=source_type,
-            data_state=DATA_STATE_MAPPING,
-        ).only('id').iterator()
+    # get the properties and chunk them into tasks
+    qs = (
```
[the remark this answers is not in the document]
macintoshpie: sure, here `qs` stands for "QuerySet" (a django query result). But technically this is misnamed b/c we're calling `.iterator()` returns an iterator... So maybe we could rename it if you wanted to be more descriptive/accurate

seed/models/building_file.py
```diff
@@ -329,18 +331,34 @@ def process(self, organization_id, cycle, property_view=None):
...
+                    assert num_updated == 0 or num_updated == 1
+                    if num_updated == 0:
+                        # create the reading
+                        new_values = {
+                            'meter_id': meter.id,
+                            'start_time': mr.get('start_time'),
+                            'end_time': mr.get('end_time'),
+                        }
+                        new_values.update(defaults)
+                        MeterReading.objects.create(**new_values)
```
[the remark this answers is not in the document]
macintoshpie: I started typing one thing but then decided it might not be right (see below). TLDR I think it's worth reverting this to the original as you suggested, and writing a bunch of tests to verify meters are handled how we expect, then fix as needed.


ORIGINAL THOUGHTS:
Oof great question... Meters are a actually a little different from Scenarios and Measures -- instead of getting linked to the State or View, they are directly linked to the Canonical Property... Which means we might not want to just drop the old meters and readings...
@nllong what do you think? Here are the cases I can imagine:
1. User has existing meter with readings at times A, B, C. Then imports bsync which matches that meter id, and only has a new, non-overlapping reading D. I'd think we would add this reading to the existing meter
2. User has existing meter with readings at times A, B, C. Then imports bsync which matches that meter id, and has a new reading which starts before B and ends after B (i.e. consumes B's span). I'm not sure what we do here, do we enforce/check overlapping readings anywhere/should we be? If we don't care, we just create a new reading.
3. Same as 2, but instead of full overlap, it is a partial overlap with an existing reading.

seed/models/building_file.py
macintoshpie: probably why I deleted it

seed/models/tax_lots.py
```diff
@@ -372,11 +372,6 @@ def coparent(cls, state_id):
 
         return coparents, len(coparents)
 
-    @classmethod
-    def merge_relationships(cls, merged_state, state1, state2):
-        """Stub to implement if merging TaxLotState relationships is needed"""
-        return None
-
```
macintoshpie, line 379: We need to revert this as well

macintoshpie (changes requested):
I made some responses to your original comments.

I think our next steps should be to delete old irrelevant tests, and add new ones. We can chat about the different cases we might want to test

macintoshpie (approved):
💯 Awesome!

Issue #3029, opened 2021-12-10
Clean up Prod MEDIA directory


macintoshpie:

nllong:
The SEED production media folder has the following:

```
➜  media du -h -d 1
299M	./vendors
 24K	./landing
 21M	./seed
632K	./rest_framework_swagger
1.1M	./admin
6.2G	./uploads
548K	./rest_framework
1.3M	./buildingsync_files
 52K	./django_extensions
6.5G	.
```

Everything except `uploads` and `buildingsync_files` are remnants of how we used to collect the static resources in the media directory, that is not longer the case. The files in the other directories can be removed.

nllong:
Note that on the production database there are also these folders that needed to persist:

```
analysis_output_files
analysis_input_files
```

nllong:
This will be deployed shortly to production when it can be run.

PR #3107, opened 2022-01-20
management task to cleanup deleted uploaded

nllong:
SEED does not delete data files after the users deletes them from the Data Files page.

#### What's this PR do?
* Adds a management task that can be run to update the paths of all the ImportFiles records to `/seed/media/uploads/`
* Deletes the datafiles from disk but keeps the record
* Saves 1.6GB on the production data volume

* Within a docker instance (otherwise you will need to change the /seed/media/uploads path in the rename
* Create several imports (no need to go through mapping)
* Delete a record
* Run the script `./manage.py delete_unused_uploaded_files`. You can pass `--org_id <orgid>` if desired

Addresses #3029


seed/data_importer/migrations/0016_auto_20220119_1347.py
```diff
@@ -0,0 +1,29 @@
...
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='importfile',
+            name='export_file',
+        ),
+        migrations.RemoveField(
+            model_name='importrecord',
+            name='import_completed_at',
```
macintoshpie, line 19: Probably need to remove this:
https://github.com/SEED-platform/seed/blob/3029-media-volume-data/seed/data_importer/models.py#L501

seed/management/commands/delete_unused_uploaded_files.py
```diff
@@ -0,0 +1,109 @@
...
+        parser.add_argument('--org_id',
+                            default=None,
+                            help='Name of a specific organization to operate',
+                            action='store',
+                            dest='org_id')
+
+    def handle(self, *args, **options):
+        # In local dev and production the files are in the media/uploads folder.
+        # Folder has: 4491 files
+        # Database has: 7306 files
```
macintoshpie, line 28:
```suggestion
```

seed/management/commands/delete_unused_uploaded_files.py
```diff
@@ -0,0 +1,109 @@
...
+        for f in files:
+            filename = f.file.name
+            if os.path.exists(filename):
+                # don't do anything if the path exists
+                continue
+
+            # the filename/path is not correct and needs to be updated
+            # put it in a tuple of db object, oldname, new name
+            new_base_path = "/seed/media/uploads"
+            if filename.startswith(new_base_path):
```
macintoshpie, line 58: Won't this case for fixing the upload path get skipped if we `continue` here?

seed/data_importer/managers.py
```diff
@@ -12,3 +12,7 @@ class NotDeletedManager(models.Manager):
 
     def get_queryset(self, *args, **kwargs):
         return super().get_queryset(*args, **kwargs).exclude(deleted=True)
+
+    def get_all(self, *args, **kwargs):
+        """method to actually return all fields so we can do database/filesystem cleanup"""
```
macintoshpie, line 17:
```suggestion
        """Method to return ALL ImportFiles, including the ones where `deleted == True` which are normally excluded.
        This is used for database/filesystem cleanup."""
```

seed/data_importer/models.py
```diff
@@ -534,112 +526,6 @@ def SUMMARY_ANALYSIS_ACTIVE_KEY(cls, pk):
     def SUMMARY_ANALYSIS_QUEUED_KEY(cls, pk):
         return 'SUMMARY_ANALYSIS_QUEUED%s' % pk
 
-    def prefixed_pk(self, pk, max_len_before_prefix=(SOURCE_FACILITY_ID_MAX_LEN - len('IMP1234-'))):
```
macintoshpie, line 537: 🧹

seed/management/commands/delete_unused_uploaded_files.py
macintoshpie: Also I just noticed the comment above about updating the filepath/path isn't totally true - we don't seem to be doing that here. Am I misunderstanding?

seed/management/commands/delete_unused_uploaded_files.py
```diff
@@ -0,0 +1,109 @@
...
+            self.stdout.write('Done renaming', ending='\n')
+        else:
+            self.stdout.write('Not renaming, will not continue, exiting')
+            exit()
+
+        # now go through and find the deleted=True and remove the records. Since
+        # ImportFile is a NonDeletableModel, we've added a new method in the
+        # model manager to return the default queryset with the deleted bool
+        # selected. A typical filter doesn't work because the deleted field is on
+        # an inherited model.
```
macintoshpie, line 92:
```suggestion
        # now go through and find the deleted=True and remove the records
```

macintoshpie (changes requested):
Worked when testing locally, just a few suggestions/comments

seed/data_importer/migrations/0016_auto_20220119_1347.py
nllong: done, thanks for catching this one!

seed/management/commands/delete_unused_uploaded_files.py
nllong: if the file path is formed correctly or empty then it will skip that file, but continue to the next file. Maybe this is being confused with `break`?

nllong:
@macintoshpie  -- if this looks good (and the test pass), can you merge down? Grazie 🙏

macintoshpie (approved):
💯

PR #3113, opened 2022-01-27
fix(docker): use default PGDATA for postgres

macintoshpie:
We added PGDATA to docker compose b/c that's what we're doing in our k8s deployment. But that's not necessary for our local docker-compose, and prevents us from creating new (i.e. empty) volumes and using them with the postgres container.

The changes originally occurred in #3101 

#### What's this PR do?
- removes PGDATA from docker-compose files

- spin it up locally with an existing DB, also try creating a new empty volume and using that for the postgres volume


nllong (approved):
looks great!

Issue #2976, opened 2021-10-28
Export Properties and Tax Lots Progress Bar


perryr16:
**Is your feature request related to a problem? Please describe.**
While exporting your properties and tax lots from the inventory page, a spinner is presented as the program works to export the file. The spinner does not give any information to the user. It could be 3 seconds, it could be 30 minutes.

**Describe the solution you'd like**
A progress bar would improve the user's experience, giving the user an idea of when the export will finish.

**Additional context**
![export_properties_spinner](https://user-images.githubusercontent.com/58446472/139336471-a561dc1a-c28c-4fbc-a991-628fe74a8712.png)

RDmitchell:
@nllong -- should this be closed?

nllong:
No, I think you still need to test!

PR #3114, opened 2022-01-31
2976 export properties progress bar

perryr16:
The process for exporting properties or tax lots can leave the user wondering how long it will take. The current process has the user watch a spinner utility until the export has finished. This spinner utility gives no indication of the expected time to complete the export. It could be 3 milliseconds, or it could be 3 hours.
Screen Shot 2022-02-01 at 11 18 30 AM

#### What's this PR do?
This PR adds a progress bar feature giving the user a sense of completion. The progress bar shows a percentage complete.
Screen Shot 2022-02-01 at 11 22 13 AM

Inventory > Select 1000 + properties > Actions > Export selected > Export in any format. 
A progress bar will display as the properties are exported.

#2976


seed/static/seed/js/controllers/inventory_list_controller.js
```diff
@@ -1013,6 +1013,7 @@ angular.module('BE.seed.controller.inventory_list', [])
         $uibModal.open({
           templateUrl: urls.static_url + 'seed/partials/export_inventory_modal.html',
           controller: 'export_inventory_modal_controller',
+          backdrop: 'static',
```
perryr16, line 1016: This prevents the modal from closing if the user clicks outside of the modal. I found accidentally closing the modal and losing the visual progress to be annoying. Open to reverting to the original functionality.

seed/static/seed/js/controllers/export_inventory_modal_controller.js
```diff
@@ -30,27 +32,40 @@ angular.module('BE.seed.controller.export_inventory_modal', []).controller('expo
...
+          export_type: export_type,
+          include_notes: $scope.include_notes,
+          progress_key: data.data.progress_key
+        }, {
+          params: {
+            organization_id: user_service.get_organization().id,
+            inventory_type: inventory_type
+          },
+          responseType: export_type === 'xlsx' ? 'arraybuffer' : undefined
+        })
```
perryr16, line 67: hit backend for a new `ProgressData` object, use that returned `progress_key` to `check_progress_loop` and `tax_lot_properties/export`

perryr16:
The only place the `/tax_lot_properties/export` function is tested is within the api tests. I'd be happy to update the api tests to incorporate this new simple endpoint if anyone thinks that's a valuable step.

seed/views/v3/tax_lot_properties.py
```diff
@@ -168,18 +186,32 @@ def export(self, request):
...
+    @ajax_request_class
+    @has_perm_class('requires_member')
+    @action(detail=False, methods=['GET'])
+    def start_export(self, request):
+        """
+        Generate a ProgressData object that will be used to monitor property and tax lot exports
+        """
+        org_id = self.get_organization(request)
+
+        progress_data = ProgressData(func_name='export_inventory', unique_id=org_id)
```
macintoshpie, line 210: IDK if this is totally necessary, but do we need to "clear" the progress data in case a user recently exported another file? Another option could be to tweak the unique_id so that we don't end up with overlap

macintoshpie (commented):
I didn't get to test this locally, @Ryoken could you take a look and test locally as well?

seed/views/v3/tax_lot_properties.py
```diff
@@ -88,6 +91,15 @@ def export(self, request):
         Download a collection of the TaxLot and Properties in multiple formats.
         """
         org_id = self.get_organization(request)
+
+        if request.data.get('progress_key'):
+            progress_key = request.data['progress_key']
+            progress_data = ProgressData.from_key(progress_key)
+        else:
+            progress_data = ProgressData(func_name='export_inventory', unique_id=org_id)
+            progress_key = progress_data.key
```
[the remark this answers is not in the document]
perryr16: Yeah, this is to allow backward compatibility. I was a bit hesitant to update the API tests, this code allowed me to avoid that

seed/views/v3/tax_lot_properties.py
perryr16: Something like `progress_data.delete()`?

Ryoken:
When I attempt to export I get a 403 with the message "Organization does not exist".  I have tried it on the regular and the beta inventory, and after bringing docker back up.

Screen Shot 2022-02-07 at 11 42 22 AM

seed/views/v3/tax_lot_properties.py
Ryoken: Clearing the progress would prevent concurrent exports.  I've made the unique_id more unique to prevent said overlap!

Ryoken:
Tested on a significant number of properties and found it to work as epxected!

Ryoken (approved):

Issue #3078, opened 2022-01-07
add ability to import data loggers and sensors for a property


macintoshpie:
There is a request to support importing Meter data for IAQ related data (e.g., zone temperatures, zone RH, CO2 levels, etc). Ideally a meter is connected to many sensors, but we can flatten this and have:

* IAQ - Zone 1 - Drybulb Temperature
* IAQ - Zone 2 - Drybulb Temperature
* IAQ - Zone 1 - Relative Humidity
* IAQ - Zone 2 - Relative Humidity

Data would need to be imported from a new button on a Property Detail page, or from within the ESPM-style format. The backend would need to support flexible naming of the meters because right now (I think) we only support specifically named meters.

An additional feature/requirement is the ability to roll up results to daily, weekly, monthly with different aggregators such as average, sum, min, max, first, last.

nllong:
@kflemin @haneslinger  -- feel free to break this into other tickets as needed. 

Once we get data in the system, then let's pull in @laurenliz22 to do data sciency aggregations.

RDmitchell:
I may be able to get some of the IAQ interval data. If I do I will put a link to it here.

nllong:
@RDmitchell  -- that would be very helpful! Especially if it is in CSV format.

RDmitchell:
@nllong -- I received two IAQ interval data files. 

* awair-omni_example.csv
  * 15 minute interval data
  * I have requested unit definitions for the fields
* AVP_example.csv
  * hourly interval data
  * all fields that need units have them
  * I have requested the definition for the fields "AQI US" and "AQI CN"

Here is a link to the folder containing those files
https://drive.google.com/drive/folders/1h0eezReaMaND45gH6Kl062QC6RjmWykU?usp=sharing

nllong:
Great! @kflemin and @haneslinger FYI. Let us know if you need access to that folder!

nllong:
AQI is air quality index, i think. Not sure the CN vs US descriptor.

RDmitchell:
Here are definitions for the fields for **awair-omni_example.csv**

score = indoor air quality score, 0-100, 0 is poor, 100 is best
temp = air temperature (degree C)
humid = relative humidity (%)
co2 = carbon dioxide (ppm, parts per million)
voc = volatile organic compounds (ppb, parts per billion) 
pm2.5 = particulate matter <2.5 um (ug/m3)
light = ambient light (lux)
noise = ambient noise (dBA)

[...]

PR #3115, opened 2022-02-03
Add sensors

haneslinger:

This Looks like alot, but these addtions **deeply** mirrors the `meters` Model, view, controller, and UI  

These files all have meter counter parts. I truly copied, pasted, and slightly modified. I would recommend reviewing them by comparing them with there twin, which has the same file path, just with "meter" instead of "sensor":
- `seed/static/seed/js/controllers/inventory_detail_sensors_controller.js` 
- `seed/static/seed/js/services/sensor_service.js`
-  `seed/static/seed/partials/inventory_detail_sensors.html`
- `seed/utils/sensors.py`

The additions to `seed/static/seed/js/seed.js` mirriors the code block right above it
The added functions to `seed/views/v3/properties.py` mirror the meter functions in the same file. 

I checked the meter functions I copied, they don't appear to be tested, so I didn't add testing either.

#### What's this PR do?

This PR add Models, API, and UI for Sensors and SensorReadings. It only add gets, no creates sets, or deletes

You will have to go into a django shell and manually create sensors to test it 😬

https://github.com/SEED-platform/seed/issues/3078

image


haneslinger:
closing in favor of https://github.com/SEED-platform/seed/pull/3134

Issue #3049, opened 2021-12-17
Add frontend filtering of inventory by labels


macintoshpie:
blocked by #3044

RDmitchell:
@Ryoken -- how should I test this?

Ryoken:
This is part of the new inventory list beta page.  To test this part, just make sure the label filtering is working as expected:
 
 - you can filter on any label that is currently applied to any property
 - AND requires all listed labels, OR requires any of the listed labels
 - EXCLUDE will exclude properties with all listed labels (works like AND)
 - removing/adding a label to a property should update the page

Issue #3044, opened 2021-12-17
Add server-side filtering by labels


macintoshpie:
See front end functionality

nllong:
This is a backend change and has been running without issue for many months, closing.

PR #3125, opened 2022-02-14
Feat 3044/filter labels serverside

Ryoken:
See tickets.

#### What's this PR do?
Adds the ability to filter on labels with the new paginated beta inventory.

Add and remove labels on various properties, noting the label logic.

https://github.com/SEED-platform/seed/issues/3044
https://github.com/SEED-platform/seed/issues/3049

### DIFF
--- a/seed/static/seed/js/controllers/inventory_list_beta_controller.js
+++ b/seed/static/seed/js/controllers/inventory_list_beta_controller.js
@@ -369,6 +369,7 @@ angular.module('BE.seed.controller.inventory_list_beta', [])
         modalInstance.result.then(function () {
           //dialog was closed with 'Done' button.
           get_labels();
+          $scope.load_inventory(1);
         });
       };
 
--- a/seed/static/seed/js/controllers/inventory_list_beta_controller.js
+++ b/seed/static/seed/js/controllers/inventory_list_beta_controller.js
@@ -736,12 +736,15 @@ angular.module('BE.seed.controller.inventory_list_beta', [])
         }
 
         // add label filtering
-        let ids = undefined;
+        let include_ids = undefined;
+        let exclude_ids = undefined;
         if ($scope.selected_labels.length) {
           if ($scope.labelLogic === 'and') {
-            ids = _.intersection.apply(null, _.map($scope.selected_labels, 'is_applied'));
-          } else if (_.includes(['or', 'exclude'], $scope.labelLogic)) {
-            ids = _.union.apply(null, _.map($scope.selected_labels, 'is_applied'));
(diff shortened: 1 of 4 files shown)


perryr16 (commented):
The `exclude` filter didn't work as I expected on 'partial' matches. If this is by design then please ignore this comment. 

Example: 
No filtering:
Screen Shot 2022-02-14 at 3 21 05 PM



After filtering to exclude "red", "green" properties, those that have only one label are still returned. Note that the property with "red/green/orange" was excluded as expected
Screen Shot 2022-02-14 at 3 21 18 PM

perryr16 (changes requested):
The `and` filter works great if there are exact or partial matches. However, if the `and` filter does not meet any property condition then it returns all values. I would expect it to return nothing instead. 
Screen Shot 2022-02-14 at 3 27 50 PM

perryr16 (changes requested):
This one might be tricky because refreshing the page resets the filters. The filters don't seem to recognize new labels unless the page is refreshed. 


https://user-images.githubusercontent.com/58446472/153958456-b24cf612-6f16-40e2-8bfd-ad5df5f26ae9.mov

Ryoken:
> This one might be tricky because refreshing the page resets the filters. The filters don't seem to recognize new labels unless the page is refreshed.
> 
>  Screen.Recording.2022-02-14.at.3.36.01.PM.mov

Ah, thanks, that's perfect.  The labels don't appear to be saving and also are not triggering the filter correctly...

macintoshpie:
@Ryoken looks like this isn't finished, but I was imagining there'd be some new backend functionality where you could include labels when hitting the `filter` endpoint, and it would use those labels to filter properties. Were you thinking of a different approach?

Ryoken:
> @Ryoken looks like this isn't finished, but I was imagining there'd be some new backend functionality where you could include labels when hitting the `filter` endpoint, and it would use those labels to filter properties. Were you thinking of a different approach?

I am utilizing the existing "is_applied" list on the label and "include_ids" list on the backend to filter on labels.  This allows us to do all the filtering we want with just one call to the label API, which we were already making.

Ryoken:
- AND filtering was updated to list 0 results if there weren't any matches.
- EXCLUDE filtering is working if you expect it to work like "OR", excluding any property with any listed label rather than all of them.  I want to add another EXCLUDE to make this more clear (EXCLUDE ALL and EXCLUDE ANY for example).
- The saved label filters were being reset on each load, causing issues with re-applying them or navigating.  This should be fixed!

perryr16 (approved):
This works great, Ryo. All my previous issues have been addressed. "exclude" works as expected, "and" returns none if no properties meet the condition, and updated labels are reflected in the filter immediately.

PR #3110, opened 2022-01-24
upgrade: celery 4.4.7 to celery 5.2.2

perryr16:
#### Any background context you want to provide?
Upgrade Celery from version 4.4.7 to version 5.2.2

Celery version 4 introduced a new naming convention that switched from an all uppercase variable to a lowercase variable: `CELERY_QUEUES` becomes `task_queues`. Version 4 and 5 are backward compatible and changing variables names is not required. However, Celery version 6 will no longer be backward compatible, making this a good opportunity to implement the new naming convention and iron out any bugs.

[This document outlines the process for upgrading an app from 4.x to 5.x](https://docs.celeryproject.org/en/stable/history/whatsnew-5.0.html#upgrading-from-celery-4-x),
[This document shows the variable name conversion from lowercase to uppercase](https://docs.celeryproject.org/en/stable/userguide/configuration.html#std-setting-task_always_eager)

Many of the uppercase celery variables did not match the values in the variable conversion document, but looking back at [a PR from 2017](https://github.com/SEED-platform/seed/pull/1374/files#) shows that most of the uppercase celery variables originally corresponded to variables found in the conversion document. 

Celery has a command to auto-upgrade variable names file by file: `celery upgrade settings path/to/file.py`
There are a number of variables that look like celery variables but are not picked up by the auto-upgrade command or on the conversion document. These have been unmodified.
- `CELERY_PIDFILE`
- `CELERY_ENABLED` 
- `CELERY_LOG_LEVEL` 
- `CELERY_BROKER_BACKEND`
- `CELERY_RESULTS_EXPIRED`

The only variable that threw errors when changed was `CELERY_TASK_ALWAYS_EAGER`. See comment in the changelog.

#### How should this be manually tested?

Perform basic functionality that uses celery. e.g. add new data files. Monitor the logs for errors and warnings.

#### What are the relevant tickets?
#3096

PR #3131, opened 2022-02-16
upgrade: celery 5.2.2 and tasks per child

perryr16:

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ task manager (Celery) with:
 
 ```
 ./manage.py runserver
-celery -A seed worker -l info -c 4 --maxtasksperchild 1000 --events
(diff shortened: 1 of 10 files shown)


macintoshpie (approved):
Working locally for me, and tests are passing so I think we're good.

Just do document our discussion, the original PR #3110 changed the settings file to use the lowercase config vars, and removed the `CELERY_` prefix. We now think we actually need to keep the config the way it already is, see https://docs.celeryproject.org/en/v5.2.2/django/first-steps-with-django.html

This kinda conflicts with some release note we saw for 5.0, which said "hey don't use all caps, they'll be deprecated in 6.0" but we're just going to keep them for now since that's what 5.0 says to do...

PR #3152, opened 2022-03-02
update map popups with display default field settings

### LINKED ISSUE #2998: Geocoding -- allow user to specify field to appear in text associated with map pin
**Describe the bug**
I seems that the text in the map pin is based only on Address Line 1. If that field is not mapped, then the pin text will say "Null".
image
In this case, the address field was mapped to Street Address, and that field was set in the Geocoding field settings, but the map doesn't display that field data. (It occurs to me that maybe I shouldn't be able to specify an extra data field as a geocoding field, since the program doesn't know that "Street Address" represents a physical address).
image

**I guess this is somewhere between a bug and an enhancement, because it seems like the solution to this issue would be to use the Settings / Default Display Fields / Property Display Field might be what should display as the text for the map pin.** 

**Expected Behavior**
I was thinking that the Street Address field would appear in the text for the pin. 

**Actual Behavior**
Null appears instead

**Steps to Reproduce**
see description

**Instance Information**
Instance: ad707077bc25e403eafed1eb5cc6874d-8554735.us-east-1.elb.amazonaws.com
SHA: 6a31f3d

kflemin:
Map popups sometimes show up with null text within it (hard coded to address_line_1).  This PR fixes this.

#### What's this PR do?
Updates the popups on the inventory map to display property display field or taxlot display field set in the organization's default display fields.

Modify the default display fields and refresh the map. Click on a popup and verify that the text has changed and matches the selected display field.

Resolves #2998 

Screen Shot 2022-03-02 at 3 34 09 PM
Screen Shot 2022-03-02 at 3 33 42 PM

### DIFF
--- a/seed/static/seed/js/controllers/inventory_map_controller.js
+++ b/seed/static/seed/js/controllers/inventory_map_controller.js
@@ -11,6 +11,8 @@ angular.module('BE.seed.controller.inventory_map', [])
     '$uibModal',
     'cycles',
     'inventory_service',
+    'user_service',
+    'organization_service',
     'labels',
     'urls',
     function (
@@ -21,6 +23,8 @@ angular.module('BE.seed.controller.inventory_map', [])
       $uibModal,
       cycles,
       inventory_service,
+      user_service,
+      organization_service,
       labels,
       urls
     ) {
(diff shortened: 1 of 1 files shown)


nllong (approved):
noicely done! 
image

It seems hard to find exactly where to click on the popups, are you seeing that too? It seems like the clicking area is 4 pixels, haha.

PR #3159, opened 2022-03-08
refactor flilteredReadings to exclude cols and rows

perryr16:
When viewing an individual property's meters, one can choose to filter by meter type (or scenario). There is a dropdown menu where a user can select or unselect meter types. Below the dropdown is a table of all meters. Their column headers tend to have a timestamp and a meter type entry (ex: 'Start Time', 'End Time', 'Electricity...', 'Natural Gas...'). Unselecting a meter type will only remove the meter type column, not the row. This is unhelpful as rows with no meter data are still visible.

#### What's this PR do?
Updates the filterByMeterLabels function to remove rows that do not contain the selected meters. The function would return a row if it contained a "timeColumn" header: `timeColumns = ['start_time', 'end_time', 'month', 'year'];`. Since all meter entries contian a `start_time` or `end_time`, all rows would be returned, but without the meter readings. 

The new function will only return a row if it contains the selected meter label. Columns are filtered in a different area and have remained unchanged.

Inventory > Property with multiple meter types > meters > unselect a meter type > the table rows should only reflect meters that contain a selected meter type

#3159

**OLD:**
One Meter Selected: Showing rows that have empty values for selected meter type.
Screen Shot 2022-03-09 at 10 00 36 AM

No Meters Selected: Showing all meters, but no meter data
Screen Shot 2022-03-09 at 10 00 48 AM

**NEW:**
All Meters Selected:
Screen Shot 2022-03-09 at 8 34 37 AM

One Meter Selected: Rows without selection removed
Screen Shot 2022-03-09 at 8 34 25 AM

No Meters Selected: All rows removed
Screen Shot 2022-03-09 at 8 33 39 AM


seed/static/seed/js/controllers/inventory_detail_meters_controller.js
```diff
@@ -162,18 +162,9 @@ angular.module('BE.seed.controller.inventory_detail_meters', [])
...
         var timeColumns = ['start_time', 'end_time', 'month', 'year'];
         var selectedColumns = meterLabels.concat(timeColumns);
-        var filteredReadings = readings.map(function (reading) {
-          return Object.entries(reading).reduce(function (newReading, _ref) {
-            var key = _ref[0],
-              value = _ref[1];
-
-            if (selectedColumns.includes(key)) {
-              newReading[key] = value;
-            }
```
perryr16, line 172: Because the `timeColumns` are included in `selectedColumns`, this function will return every row, while removing unselected columns.

Ryoken (approved):
Looks good!

PR #3166, opened 2022-03-15
Add data logger creation

haneslinger:
#### What's this PR do?
Adds the DataLogger concept.


nllong:
Hey @haneslinger  -- looks like there are some conflicts with develop on this now. Can you resolve when you get a chance?

seed/api/v3/urls.py
```diff
@@ -69,6 +70,7 @@
 api_v3_router.register(r'postoffice', PostOfficeViewSet, basename='postoffice')
 api_v3_router.register(r'postoffice_email', PostOfficeEmailViewSet, basename='postoffice_email')
 api_v3_router.register(r'progress', ProgressViewSet, basename="progress")
+api_v3_router.register(r'data_loggers', DataLoggerViewSet, basename="data_logger")
```
nllong, line 73: silly request, can you move this to be alphabetical? Looks like `data_quality_checks` is also out of order.

seed/models/sensors.py
```diff
@@ -6,21 +6,39 @@
 from seed.models import Property
 
 
-class Sensor(models.Model):
-    sensor_property = models.ForeignKey(
+class DataLogger(models.Model):
+    property = models.ForeignKey(
```
nllong, line 10: nice. I like that we are attaching this onto the Property (similar to how sensor was on the property).

seed/models/sensors.py
```diff
@@ -6,21 +6,39 @@
...
     description = models.CharField(max_length=2047, default="")
 
     sensor_type = models.CharField(max_length=63)
     units = models.CharField(max_length=63)
 
-    column_name = models.CharField(unique=True, max_length=255)
+    column_name = models.CharField(max_length=255)
+
+    class Meta:
+        unique_together = ('data_logger', 'display_name')
```
nllong, line 40: 👍

seed/static/seed/partials/inventory_detail_sensors.html
```diff
@@ -33,10 +33,26 @@ <h2>
 </div>
 <div class="section_content_container" ng-cloak style="padding-left:15px; padding-right:15px;">
 
-    <div id="sensor-info-container">
-        <div class="btn-group" style="padding-top: 15px;">
-            <button type="button" ng-click="open_sensor_upload_modal()" class="btn btn-primary" translate>UPLOAD_SENSOR_BUTTON</button>
+    <div>
+        <div class="btn-group" style="padding-top: 15px; display: flex; gap: 15px;">
+            <button type="button" style="border-radius: 4px;" ng-click="open_data_logger_upload_modal()" class="btn btn-primary" translate>UPLOAD_DATA_LOGGER_BUTTON</button>
```
nllong, line 38: I'm thinking we should have "Add Data Logger" which is just a modal form to create the data logger. Then use your already created "Upload Sensor Readings" to the move on to that step.

Also when giving a random CSV file, the `SENSOR_CONTENTS_TITLE` doesn't resolve (see image)

image

nllong (approved):
Looks great. A few comments. 

image

kflemin (approved):
looks good!  Just a couple of things:
1. In the Sensor table (the 2nd table), can you move the Datalogger column to be column #2 next to display name? I think people will want to filter by datalogger and it's all the way at the end of the table :)
2. when I tried this on my local, changing the sensorreading interval reset the filtering on the table. Just an FYI.
3. I would add "filter by type" (with a dropdown of types) and "filter by Datalogger" with a dropdown of datalogger next to the "filter sensors" button for the sensor readings table.  maybe even a "reset filtering" button.  It will be too tedious to check / uncheck those checkboxes on a regular basis.

haneslinger:
> looks good! Just a couple of things:
> 
>     1. In the Sensor table (the 2nd table), can you move the Datalogger column to be column #2 next to display name? I think people will want to filter by datalogger and it's all the way at the end of the table :)
> 
>     2. when I tried this on my local, changing the sensorreading interval reset the filtering on the table. Just an FYI.
> 
>     3. I would add "filter by type" (with a dropdown of types) and "filter by Datalogger" with a dropdown of datalogger next to the "filter sensors" button for the sensor readings table.  maybe even a "reset filtering" button.  It will be too tedious to check / uncheck those checkboxes on a regular basis.

I just adressed the first point. The next two points are addressed in my follow up PR, #3182

Issue #2508, opened 2020-12-19
Inventory Detail View: Text bubbles stay on the screen even if you scroll down the page


RDmitchell:
**Describe the bug**
In the Inventory Detail View, if you click on a field a text bubble appears and will stay on the screen until you click it again, even if you scroll down the page.

**Expected Behavior**
Don't have the text bubbles "stick" even if you click on a field. If you hover or click on a field, have the text bubble appear and if you move off the field, the text bubble disappears.

Emulate the behavior on the Inventory List view
![image](https://user-images.githubusercontent.com/6314950/102678921-6ad5e280-4160-11eb-894d-0358a8d6fb83.png)

**Actual Behavior**
* In the Inventory Detail view, if you click on a field, a text bubble appears.
![image](https://user-images.githubusercontent.com/6314950/102678814-a4f2b480-415f-11eb-845a-1f9f616897b9.png)
* If you click on another field, another text bubble appears
![image](https://user-images.githubusercontent.com/6314950/102678831-c5227380-415f-11eb-8d82-ff523525a2f8.png)
* If you then scroll down the page, the text bubbles stay in their original locations, and are no longer associated with the original field
![image](https://user-images.githubusercontent.com/6314950/102678862-03b82e00-4160-11eb-8506-c3664a4da06a.png)
* In order to get them to go away, you have to go find their original field and click on it again.
* Very annoying

**Steps to Reproduce**
See actual behavior above

**Instance Information**

Instance: seedv2 production 2.7.6
SHA: e8d5cde48

**Additional Context**
San Jose reported this

RDmitchell:
Happens on both chrome and firefox browsers

Ryoken:
I believe the clicking on and off part is intentional, as it allows you to copy the actual cell value and not what's displayed.  If this functionality is no longer required I can make the tooltips standard html title attributes that appear when you hover.  Otherwise, I can make it where only one tooltip can be open at a time, and/or scrolling will clear open tooltips.  Any thoughts or preferences?

RDmitchell:
@Ryoken -- I think we want to be able to copy the cell value, so I guess making it so that only one tooltip can be open at a time, and scrolling will clear open tooltips. I think the latter is the most important, and is what I expected to happen.

RDmitchell:
Instance; Staging
SHA: 9de35f190

@nllong -- this doesn't appear to be included on staging

RDmitchell:
Instance: staging
SHA: 6ffea0837

[...]

PR #3169, opened 2022-03-15
fix: popovers now open one at a time and disappear with scrolling

Ryoken:
See ticket

#### What's this PR do?
Makes popovers on the inventory detail page only show one at a time, and scrolling will clear them.

Load any detail view, click on a field, click on another field, scroll the page.

https://github.com/SEED-platform/seed/issues/2508

see ticket

### DIFF
--- a/seed/static/seed/js/controllers/inventory_detail_controller.js
+++ b/seed/static/seed/js/controllers/inventory_detail_controller.js
@@ -118,7 +118,7 @@ angular.module('BE.seed.controller.inventory_detail', [])
       $scope.users = users_payload.users;
 
       // handle popovers cleared on scrolling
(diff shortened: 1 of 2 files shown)


perryr16 (approved):
Works as expected. Approved!

It's not part of the ticket, but I would have expected the popover to appear if a user hovers over a detail. Seems it's only triggered if it is clicked.

Ryoken:
> Works as expected. Approved!
> 
> It's not part of the ticket, but I would have expected the popover to appear if a user hovers over a detail. Seems it's only triggered if it is clicked.

That's actually by design and the way it was before this ticket.  I believe the intent is to allow copying the actual value (which is what's displayed in the popup).

Issue #2831, opened 2021-08-23
Deleting Property Inventory doesn't delete the BETTER analyses


RDmitchell:
**Describe the bug**
I deleted all the records from a Property Inventory list, and then looked to see if the related Analyses had also been deleted, but they hadn't.

**Expected Behavior**
I expected the analyses to also be deleted

Not sure if the program should delete the related analyses, but it seems like it should. Or maybe ask the user if they want those to be deleted when the properties are deleted. If the user says No then it is up to them to delete them "by hand". 

**Actual Behavior**
See the bug description

**Steps to Reproduce**
* Import properties into SEED
* Run BETTER analyses on them
* Delete the properties
* Look to see if the Analyses for those properties still display in Organizations/Analyses

**Instance Information**

Instance: dev1
SHA: 696f653a4
Org: LBNL 444
Cycle: BETTER JCC Test

**Additional Context**
None

RDmitchell:
@nllong / @macintoshpie -- I didn't put this into the BETTER project because there isn't a Triage column there for you to assess the priority.

RDmitchell:
I would like to keep this open.

nllong:
I think this needs to happen too!

Ryoken:
What should happen if the property is connected to a batch analysis?

nllong:
Good question. If a property is attached to a batch analyses, then perhaps we throw a message saying...

"The following properties will be removed; however, the results will still be part of the batch analysis. You may prefer to rerun the analysis.

<property a>
<property b>

Would you like to continue?
"

RDmitchell:
Do we really want to keep analysis results for buildings that are no longer there? 

Seems like we don't but not sure ... it would be nice to ask some users but I'm not sure I know of anyone using BETTER via SEED.

RDmitchell:
hmmm .... I guess warning the user is probably the best strategy so that at least they have control over how to handle it.

RDmitchell:
Instance; dev1
SHA: 505 error
Org: LBNL 405

Worked as expected

PR #3170, opened 2022-03-16
2831 fix/delete analysis with property

Ryoken:
see ticket

#### What's this PR do?
Gives a use the option to also delete associated analyses when deleting a property.  Deleting 1:1 analyses is checked by default, whereas you must opt in to delete batch analyses.

Add some properties, run some analyses (including batches), delete used properties, ensure analyses are gone.

https://github.com/SEED-platform/seed/issues/2831

There are two properties and two analyses, one single and one batch.  Attempting to delete the property in both results in:
Screen Shot 2022-03-15 at 8 25 55 PM
Attempting to delete the property just in the batch results in:
Screen Shot 2022-03-15 at 8 34 41 PM


perryr16 (approved):
Approved! Works as expected, the checkboxes when deleting properties are very intuitive. 

it might be nice to see the number of analyses deleted on the 'success page'.

Screen Shot 2022-03-18 at 8 21 59 AM

nllong (approved):
This work as expected. If a property is part of a batch, then deleting the building leaves the building in the batch. But if it is a single analysis then it deletes as expected.

Note that the column filtering wasn't working for me on the new Inventory List page, could be a fluke.

PR #3173, opened 2022-03-18
Update display of BETTER host configuration

nllong:
On development, or any custom configured `BETTER_HOST` the URL link would always go to production.

#### What's this PR do?
* Adds BETTER_HOST to the org API
* Point to the documentation based on the host URL and display it in the text.

Load settings page for an org and look at the BETTER key section.

n/a


perryr16 (approved):
Works as expected. approved.

Issue #3163, opened 2022-03-10
Recognize Empty not working as intended


RDmitchell:
**Describe the bug**
******
From Andrew Held, DC Department of Energy & Environment
Wanted to flag a bug we uncovered in SEED. The Recognize Empty column field is triggering even when the field in question is not in the upload. For example, if I upload a file with the field ENERGY STAR Score present it will upload accordingly, however if I were to then upload a file without the field ENERGY STAR Score SEED assumes that ENERGY STAR Score is empty for all records and overwrites it with null. I believe the intent of this option was to allow valid null values from Portfolio Manager to overwrite old data, but it seems it’s too aggressive it it’s current iteration.

Is this something we could get addressed in a new version of SEED soon? We’re about to enter the reporting season and it’s pretty important that we have this feature.

**Expected Behavior**
See above

**Actual Behavior**
See above

**Steps to Reproduce**
See above

**Instance Information**

Instance: LBNL production 
SHA: 9de35f190

nllong:
DOEE has confirmed that this fix is working.

PR #3176, opened 2022-03-18
3163 fix/differentiate null and missing

Ryoken:
see ticket

#### What's this PR do?
Updates "recognize_null" to not affect imports where a column is missing.

- Pick a field (ie: energy_star).
- turn "recognize_null" OFF for this field
- import a file with the field present and non-empty, note it's presence on the property
- import another file with the field missing completely, note the value on the property doesn't change
- import another file with the field present and empty, note the property value still doesn't change
- turn "recognize_null" ON for this field
- import a file with the field present and non-empty, note it's presence on the property
- import another file with the field missing completely, note the value on the property doesn't change
- import another file with the field present and empty, note the property value finally changes

https://github.com/SEED-platform/seed/issues/3163

"recognize_null" OFF... note the value remains
Screen Shot 2022-03-18 at 2 41 14 PM

"recognize_null" ON... note value only changes once the column is BOTH present and empty
Screen Shot 2022-03-18 at 2 31 04 PM
Screen Shot 2022-03-18 at 2 35 58 PM
Screen Shot 2022-03-18 at 2 36 50 PM

### DIFF
--- a/seed/lib/merging/merging.py
+++ b/seed/lib/merging/merging.py
@@ -186,7 +186,7 @@ def merge_state(merged_state, state1, state2, priorities, ignore_merge_protectio
     ).values_list('column_name', flat=True)
 
     default = state2
-    null = None # noqa: F481
+    null = None  # noqa: F481
     state2_present_columns = [column["to_field"] for column in eval(state2.import_file.cached_mapped_columns)]
     for attr in can_attrs:
         recognize_empty = attr in recognize_empty_columns
--- a/seed/lib/merging/merging.py
+++ b/seed/lib/merging/merging.py
@@ -153,7 +153,7 @@ def _merge_extra_data(ed1, ed2, priorities, recognize_empty_columns, ignore_merg
         if (val1 and val2) or recognize_empty:
             # decide based on the priority which one to use
             col_prior = priorities.get(key, 'Favor New')
-            if (ignore_merge_protection or col_prior == 'Favor New') and (state2_present_columns and key in state2_present_columns):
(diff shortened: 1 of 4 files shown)


nllong:
I have a feeling that all of the Django test errors are happening because the state2 model is None. I'm wondering if the state1 model is being used by itself at times???

Ryoken:
> I have a feeling that all of the Django test errors are happening because the state2 model is None. I'm wondering if the state1 model is being used by itself at times???

It appears to be that state2 is built by hand and the import files are only populated as needed for that test.  I put in some checks to ensure the sate has what we need before using it.

nllong (changes requested):
Hey @Ryoken  -- i tested this and glad that I tried it with an extra data field. It worked fine with a db-based field, but with an extra data field it just overwrote everything.

I think we need to pass in the import_file columns here https://github.com/SEED-platform/seed/blob/3163-fix/differentiate-null-and-missing/seed/lib/merging/merging.py#L138-L163.

Ryoken:
Ah, good catch!  I will revisit this asap

nllong:
> Ah, good catch! I will revisit this asap

cool thanks. If you get something in before noon on Monday, I'll take a look!

nllong:
The recognize null is now working as expected for both canonical fields and extradata fields; however, it looks like a few tests are failing. I did not investigate those errors but I think they are worth a deep dive to figure out what the right logic should be.

Also, I doubt it is related but the column filters on beta properties is not working on my computer anymore. I'm on safari and it was bypassing the cache.


image

PR #3186, opened 2022-03-28
fix: removed unused function call

Ryoken:
The beta inventory filters stopped working.

#### What's this PR do?
Fixes the beta inventory filters:  there was a function call that was no longer needed as loading the labels happens on init instead of on loading the inventory.

Use the beta inventory page and filter the list down.

---

---


perryr16 (approved):
Looks good

Issue #2844, opened 2021-08-25
BETTER -- show a list of all buildings in Org/Analyses, not just grouped by Portfolio


RDmitchell:
**Is your feature request related to a problem? Please describe.**
It is not possible to see all the buildings that have a BETTER analysis in a complete list -- right now they are only viewable grouped by Portfolio. 

**Describe the solution you'd like**
I would like also like to be able to see a complete list, not grouped by Portfolio.
![image](https://user-images.githubusercontent.com/6314950/130728123-9b5d9a71-58aa-428c-a7ae-acff7fea07b8.png)
![image](https://user-images.githubusercontent.com/6314950/130728313-55a8d600-13ca-40dd-824d-128caaf2cecb.png)

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

RDmitchell:
I would like to keep this open.

Ryoken:
The referenced PR adds support for listing the properties on the analyses page, but it does not add BETTER results columns.  This should be address in another ticket where we look at how to customize these lists for different analyis types.  Does this work for you @RDmitchell?

RDmitchell:
@Ryoken -- do you want me to make a separate ticket for showing the results (as values, not the HTML fancy reports) on the analysis page?

Ryoken:
> @Ryoken -- do you want me to make a separate ticket for showing the results (as values, not the HTML fancy reports) on the analysis page?

I think that would be best.  We will need to think about and design how different types of analyses will be listed here in this table.

PR #3187, opened 2022-03-28
2844 refactor/list analysis

Ryoken:
see ticket

#### What's this PR do?
Adds a tab to the analyses page that lists all analyses by property, rather than by analysis.

Run analyses on various properties and check the analyses page.

https://github.com/SEED-platform/seed/issues/2844

Screen Shot 2022-03-28 at 5 10 43 PM


RDmitchell:
The tabs seem like a nice solution ! Thx !

perryr16 (commented):
It seems the "By Analysis" tab has a lot more information. I think it might be useful to have similar info on the "By Property" tab like "analysis name", "created", "run status". Just a thought, happy to discuss. Otherwise it works like a charm

perryr16 (changes requested):
The "By Analysis" tab is sorted from newest to oldest, while "By Property" tab is sorted from oldest to newest. Can you change the sorting of "By Property" to `Run ID` descending?

Ryoken:
> The "By Analysis" tab has a lot more information. I think it might be useful to have similar info on the "By Property" tab like "analysis name", "created", "run status". Just a thought, happy to discuss. Otherwise, it works like a charm

Good point.  We are making a ticket soon to discuss what we want showing up in this table and how to display it.

haneslinger (approved):
LGTM

Issue #3146, opened 2022-02-24
Extra Data Fields: Need to be able to specify number of decimal points 


RDmitchell:
**Describe the bug**
For extra data fields that are imported, the numbers are shown to many decimal points.

**Expected Behavior**
The fields are set to "Number" in the spreadsheet
![image](https://user-images.githubusercontent.com/6314950/155621633-d8225595-928a-4765-86ea-49e4b2ae0325.png)

The fields are set to "Number" in the field definitions under Column Settings
![image](https://user-images.githubusercontent.com/6314950/155621667-b89f7a0e-6d7d-42d5-983c-db3591f9ff9c.png)
![image](https://user-images.githubusercontent.com/6314950/155621675-d78870f6-fbf9-4255-9f97-fbcd80939d4f.png)

Under Org/Settings/Units, the Number of decimal point to display is set to 2
![image](https://user-images.githubusercontent.com/6314950/155613191-b335f534-f2be-4c6c-8284-af9e62763237.png)

But when I import the data, the fields are not governed by the number of decimal points -- I suspect this is because they are in the Extra Data field.
![image](https://user-images.githubusercontent.com/6314950/155621911-76153b6d-1c9e-4a04-a831-923c10829df3.png)

I think it is important for users to be able to control the number of decimal points in all their data. This issue came up when importing some test data for the Better Buildings Campaign.

**Actual Behavior**
See above

**Steps to Reproduce**
See above

**Instance Information**

Instance: dev1
SHA: adf8e4dd5
Org: LBNL 21
Cycle: 2021 Calendar Year

Ryoken:
Turns out the issue wasn't really "related_data" related.  The system was only rounding values marked as "float", but now should round values marked as "number" as well.

RDmitchell:
Instance: dev1
SHA: 500 error
Org: LBNL 21

Worked as expected

PR #3191, opened 2022-03-31
Fix Extra Data Merge

Ryoken:
See ticket

#### What's this PR do?
Adjusts the extra_data merging logic to handle edge cases.

- Pick an extra_data field
- turn "recognize_null" OFF for this field
- import a file with the field present and non-empty, note it's presence on the property
- import another file with the field missing completely, note the value on the property doesn't change
- import another file with the field present and empty, note the property value still doesn't change
- turn "recognize_null" ON for this field
- import a file with the field present and non-empty, note it's presence on the property
- import another file with the field missing completely, note the value on the property doesn't change
- import another file with the field present and empty, note the property value finally changes

https://github.com/SEED-platform/seed/issues/3146


perryr16 (approved):
Worked as expected. Approved!

Issue #3188, opened 2022-03-30
Letter spacing for some derived columns is odd in Inventory List


RDmitchell:
**Describe the bug**
I defined some Derived Columns that have all uppercase words in the name -- the program is putting spaces between each capitalized letter 
Here is how they look in the Defined Columns view
![image](https://user-images.githubusercontent.com/6314950/160936137-8eb6f7fa-e27b-45d3-97b6-3ae30d3c111a.png)
Here is how they are displayed in the list view
![image](https://user-images.githubusercontent.com/6314950/160935942-d2828bb4-0a6a-485c-8032-6f607821bc5f.png)
Here is how they look in the column profiles settings
![image](https://user-images.githubusercontent.com/6314950/160936048-1d356c02-ebbc-4c8d-96b5-8c8ba2292926.png)

**Expected Behavior**
No spaces 

**Actual Behavior**
See description

**Steps to Reproduce**
see description

**Instance Information**

Instance: Dev1
SHA: a6ffde154
Org: LBNL 300

**Additional Context**
n/a

RDmitchell:
Instance: dev1
SHA: get a 500 error when trying to get SHA
Org: LBNL 300

Fixed !
![image](https://user-images.githubusercontent.com/6314950/165413690-6932aca1-c07a-4c5c-a128-22efcf6f1996.png)

PR #3197, opened 2022-04-01
3188 bug/derived cols letter spacing

perryr16:
When viewing the inventory list, derived columns with all caps titles appeared with spaces inserted between each consecutive capital letter: `TEST column` is displayed as `T E S T column`. 

#### What's this PR do?
This is an angular ui-grid "feature". If a column does not have a `.displayName` then one will be generated and spaces will be inserted between consecutive capital letters. When the inventory lists are loaded in the controller, the derived columns will be assigned a `.displayName` based on `.name`

- Generate a derived column with at least one all capital word in the title.
- Go to Column List Profile and enable `√` the new derived column
- Go to inventory or Inventory (beta) and ensure the title of the derived column does not have spaces inserted between each capital letter

#3188

NEW
Screen Shot 2022-04-01 at 1 17 30 PM

OLD
Screen Shot 2022-04-01 at 1 16 54 PM


Ryoken (approved):
Looks good, nice and easy

PR #3202, opened 2022-04-06
fix: fields marked 'number' also get rounded

Ryoken:
See ticket.

#### What's this PR do?
Checks for field type 'number' as well when rounding a display value.

Check the list view to ensure all extra_data numbers are being rounded as per the organization setting.

https://github.com/SEED-platform/seed/issues/3146


perryr16 (approved):
Works as advertised. Approved.

PR #3237, opened 2022-05-04
Mark /seed as safe

axelstudios:
Recent versions of git no longer allow users to interact with local git repos when the file owner permissions differ from the current user.

#### What's this PR do?
Updates the system-wide git config for all users (`/etc/gitconfig`) to mark `/seed` as a safe directory, so that the `uwsgi` user can query the current git revision despite the files being owned by `root`.

1. Run `docker-compose up -d`
2. Check that the About page loads the git sha correctly, instead of returning a 500 error.


nllong (approved):
worked 
image

PR #3241, opened 2022-05-06
2813 refactor/update pm default column mapping revisit

perryr16:
`description pending...`


RDmitchell:
instance: dev1
SHA: 400ca8895

I don't see this change implemented in this version on dev1
image

Here is a link to a file to test with
[2 - example-data-ESPM-Alerts-UBID.xlsx](https://github.com/SEED-platform/seed/files/8863956/2.-.example-data-ESPM-Alerts-UBID.xlsx)

PR #3251, opened 2022-05-12
3216 column description field

### LINKED ISSUE #3216: Column Description Field
**Is your feature request related to a problem? Please describe.**
On the Column Detail page can we add a field for column description. They would default to the column name until edited. When you are looking at a detailed column view of a property (or someplace where column names are referenced) a pop up window would show the definition. 

**Describe alternatives you've considered**
The other idea instead of having a pop up window is have column names link to a "definition" that when it's clicked on the user takes them to the summary tab and shows them the column name/scheme/definition/number of populated rows. It would also be nice to have that be exportable to an excel doc (or xml) for easy metadata sharing.

**Additional context**
It's hard when we have over 900 columns for users to always know what they're looking at.

anchapin:
Resolves #3216 

#### What's this PR do?
Adds a column description field to columns that defaults to column name and can be edited as necessary. It also will appear when hovering over column headings in inventory list and inventory list beta. Didn't add to other pages yet.

Add data and then go to organization/column settings to see column description field. Edit one of the column descriptions and save changes. Then go to inventory list and hover over that column heading and ensure that the popup matches the revised column description.
 
#3216


seed/models/columns.py
```diff
@@ -543,6 +611,7 @@ class Column(models.Model):
     table_name = models.CharField(max_length=512, blank=True, db_index=True)
 
     display_name = models.CharField(max_length=512, blank=True)
+    column_description = models.TextField(max_length=1000, blank=True, default=None)
```
axelstudios, line 614: FYI, max_length on text fields isn't enforced

anchapin: Thanks for the heads up. I did find this excerpt from [Django Fields documentation](https://docs.djangoproject.com/en/4.0/ref/models/fields/#textfield).

> If you specify a max_length attribute, it will be reflected in the [Textarea](https://docs.djangoproject.com/en/4.0/ref/forms/widgets/#django.forms.Textarea) widget of the auto-generated form field. However it is not enforced at the model or database level. Use a [CharField](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.CharField) for that.

I don't think we are using the Textarea widget but wasn't sure if I should leave the max_length in there in case.

seed/migrations/0164_column_column_description.py
```diff
@@ -0,0 +1,33 @@
...
+    # Go through all the organizations
+    for org in Organization.objects.all():
+        columns = Column.objects.filter(organization_id=org.id)
+
+        for col in columns:
+            if col.display_name is None or col.display_name == "":
+                col.column_description = col.column_name
+            else:
+                col.column_description = col.display_name
+            col.save()
```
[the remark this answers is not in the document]
anchapin: I just made this change and created a new commit.

Issue #3253, opened 2022-05-12
Verify if Inventory List is pulling all the property states


nllong:

haneslinger:
It's not. 🤙

Issue #3000, opened 2021-11-08
Purge History - Slow loading of inventory detail page (D.C.)


macintoshpie:
## context
We heard reports from D.C. that loading the property detail page can be very slow. It is worth following up with them to verify this and understand their usage of SEED so we can better diagnose their problems.

From what I can recall, they update properties quite often, resulting in a large "history" of properties (i.e. AuditLog or the columns you see of changes on the details page).

One proposed solution is to delete older property states, keeping a shorter history.

~~Semi-related, I believe Andrew Held has mentioned that sometimes a "history entry" is created when it really shouldn't be (e.g. nothing really changed in the property, it's an exact match of what previously existed, so no new item is required). Again it's probably best to follow up with them to verify this. If this is the case, it might be worth verifying our matching (deduplication) is working as it should for their use case. This might need to be put into a separate ticket.~~

UPDATE:
We confirmed the issue is that there is too "long" of a history of properties. The impact is not only loading the details page, but also pairing new data to existing records is extremely slow, and loading data for paring properties to tax lots is extremely slow (this might be a separate issue due to SEED loading all details for this screen when it only needs to load like 3 fields).

They confirmed that they think deduplication is working as intended so need to worry about that for now.

## tasks
- shorten the history

macintoshpie:
This needs more discussion and research (see unknowns below). But here are some thoughts

There are two methods for pruning the history:
1) based on time (e.g. purge older than 6 months)
2) based on history length (i.e. log chain length)

There are two general strategies for implementing pruning:
a) maintaining the tree as we edit it (prune when we update it in the Django code); note this is probably only a valid option for (2) above
b) creating a scheduled task (i.e. cron) to prune the tree

Pros of (a):
- no need for a job
Cons of (a):
- requires updating our Django code
Unknowns of (a):
- how hard is it to "centralize" this process so that anywhere history is updated we don't have to account for this? E.g. could we use a hook, use a new method, or something else so that this change doesn't require significant changes throughout our codebase?
- Could this cause unexpected performance issues by doing this cleanup?

Pros of (b):
- Django code doesn't need to worry about this, no code changes there
Cons of (b):
- requires a cron job, which SEED has none of currently
Unknowns of (b):
- How feasible is it to search for these *States that need to be pruned through the entire DB?

haneslinger:
Notes from our chat with DC:

- They are cool with us deleting all but the 10 newest States per Property.
[...]

PR #3264, opened 2022-05-20
Speed up inventory beta page

haneslinger:

DCs inventory beta page is slow ([issue](https://github.com/SEED-platform/seed/issues/3000#issuecomment-1104192349)). There's a lot that can be speed up about this page. This PR tackles the lower hanging fruit, and drop the load time from 15s to 6s on my local version with the DC db.

#### What's this PR do?

1. (45b7ffd) create and use a "brief" version of all the column_list_profiles and only get the full version of the one we need.
2. (1253a4a  to f3b3637) Only get the properties once instead of twice.
3. (6b556cc) Remove unused `GET  org`
4. (5638852) create and use a "brief" version of get org

To  get the speed up, your going to need to get DC db on your local machine and compare this branches load time to master.
Else, open the network tab of devtools and notice the lack of redundant calls. 

[Slow times first doumented](https://github.com/SEED-platform/seed/issues/3000#issuecomment-1104192349)
[Investigation began](https://github.com/SEED-platform/seed/issues/3253)


nllong (approved):
This is great. It worked as expected with the DC data. Down to about 10 seconds, at least on my machine.

Ryoken (approved):
Looks good and works as expected.  Sped up my large data set!

Issue #3175, opened 2022-03-18
Store results of carbon analysis into SEED db-based fields


nllong:
Update Analysis to store carbon results back into SEED db-based fields. This is only for the Carbon Metrics Analysis (not the BETTER analysis).

Ryoken:
We are currently calculating and saving three values with the CO2 analysis:
 - Average Annual CO2 (kgCO2e)
 - Total Annual Meter Reading (MWh)
 - Annual Coverage %

How would these translate to the fields?
 - Total GHG Emissions (MtCO2e/year)
 - Total Marginal GHG Emissions (MtCO2e/year)
 - Total GHG Emissions Intensity (kgCO2e/ft2/year and kgCO2e/m2/year)
 - Total Marginal GHG Emissions Intensity (kgCO2e/ft2/year and kgCO2e/m2/year)

nllong:
I think we need to update the CO2 analysis to report as actual fields now. Here is what I propose:

1. Add in an CO2 analysis setting to ask if the results should override the default property columns
2. Average Annual CO2 (kgCO2e) -> MtCO2e (so just divide by 1000)
3. Total Marginal GHG -> Total Marginal GHG Emissions (MtCO2e/year) [added in #3174]
4. Total GHG Emissions Intensity (kgCO2e/ft2/year and kgCO2e/m2/year) -> Average Annual CO2 divided by gross floor area in kgCO2/ft2/year
5. Total Marginal GHG Emissions Intensity -> Total Marginal GHG Emissions Intensity (kgCO2e/ft2/year and kgCO2e/m2/year) [added in #3174]

RDmitchell:
@nllong -- see also #3174

nllong:
The does what it is supposed to do. Ticket #2605 will cover the ability to have multiple units which is a bit more complicated.

PR #3267, opened 2022-05-23
feat: added an option for saving data

Ryoken:
see ticket

#### What's this PR do?
adds a checkbox when making a CO2 analysis to enable saving results to the property

- run a CO2 analysis without the option checked and verify no value is saved to the property state
- run another with the option checked and see the value

https://github.com/SEED-platform/seed/issues/3175


perryr16 (approved):
approved!

Issue #3243, opened 2022-05-06
Add timezone to cataloguer


nllong:

nllong:
That's correct, it is a canonical field on the property record now.

RDmitchell:
Here is a reference for the correct format
https://gist.github.com/heyalexej/8bf688fd67d7199be4a1682b3eec7568

And specifically for the US
![image](https://user-images.githubusercontent.com/6314950/178828218-1db10f12-9a88-455a-a344-57df20b42cf1.png)

I'm going to add an issue to document this for the users.

RDmitchell:
I can add Property Time Zone to a file, import it, map the field to Property Time Zone, and it displays in the Inventory Property List. 
![image](https://user-images.githubusercontent.com/6314950/178829762-3487e53d-de32-4fe7-84fa-e796d128a697.png)

I also see it in the Column Settings view
![image](https://user-images.githubusercontent.com/6314950/178829891-206199fb-0843-4d5b-a592-8fea78bc7f00.png)

It is unclear to me how it manifests itself in the Sensor data, but I am going to close this issue, since the basic functionality is complete.

nllong:
It isn't connected throughout the sensors yet, that task still needs completed.

PR #3273, opened 2022-05-27
Add property timezone as a database field

nllong:
Meters need to know the timezone in order to have the correct time stamps in the meter data. It made sense to add the timezone string to the property which makes the assumption that the meters are all within the property and that the property does not span multiple time zones.

#### What's this PR do?
* Add property_timezone
* Update `example-data-properties.xlsx` with a column for timezones
* Format of the timezone needs to follow the `pytz.all_timezones` format. 

* import new data that has a column for timezone
* map to new field
*Note that the timezone string is not validated. Downstream code will need to ensure the string is in correct format*

#3243 

image


RDmitchell:
Sensors also need the timezone functionality.

haneslinger:
> Sensors also need the timezone functionality.

This is for the sensors. They will all share their property's timezone.

RDmitchell:
Perfect !

haneslinger (approved):
Thank you!

RDmitchell:
Thank you !!

seed/models/properties.py
```diff
@@ -131,7 +131,23 @@ def copy_meters(self, source_property_id, source_persists=True):
 
 
 class PropertyState(models.Model):
-    """Store a single property. This contains all the state information about the property"""
+    """Store a single property. This contains all the state information about the property
+
+    For property_timezone, use the pytz timezone strings. The US has the following and a full
+    list can be creates by running calling pytz.all_timezones in Python:
```
nllong, line 137:
```suggestion
    list can be created by calling pytz.all_timezones in Python:
```

Issue #3287, opened 2022-06-03
Add ability to API to filter property/labels by label names


nllong:
**Is your feature request related to a problem? Please describe.**

The API on v3/properties/labels currently returns *all* the labels with all of the property views that are applied. You are able to downselect with property view IDs are desired by passing in a `selected` argument, but the result is still a list of all the labels with/without that property view id attached.

**Describe the solution you'd like**

Add query or data param to select which labels are of interest, preferably by name (not label id)

nllong:
@RDmitchell  - this is a backend change that isn't testable

PR #3288, opened 2022-06-03
add label_names filter to properties/labels

nllong:
For SEED <-> Salesforce connection, having a list of which properties have a subselection of labels would make it much easier.

#### What's this PR do?
Add `data={label_names=['label 1 name', 'label 2 name']` to the api/v3/properties/labels end point

* make sure website behaves normal
* Use postman or something similar to test the updated endpoint. 
* There is a modified unit test for this functionality

#3287


seed/views/v3/properties.py
```diff
@@ -199,12 +201,22 @@ def labels(self, request):
         Returns a list of all labels where the is_applied field
         in the response pertains to the labels applied to property_view
         """
-        labels = Label.objects.filter(
-            super_organization=self.get_parent_org(self.request)
```
nllong, line 203: We should never be looking at a parent's org for labels because each org is responsible for their own labels.

anchapin (commented):
It works for me. No comments

anchapin (approved):
I was able to test it and confirm that it works as intended.

Issue #3289, opened 2022-06-03
Unable to create labels in suborg


nllong:
Create a sub org and try to create a label using the UI.

nllong:
Marking as p-1 to keep it from becoming stable.

RDmitchell:
I will test on the latest dev solution

RDmitchell:
Instance: dev1
SHA: f9558b8bb
Org LBNL 101 SubOrg2
RM+20

It is possible to add labels to data in a sub org, but the labels are not showing up in the List View, at least for the Tax Lot tab
Jurisdiction Tax lot ID: 10102/5B193 has a label, as seen in the detail view (added in List view)
![image](https://user-images.githubusercontent.com/6314950/197083517-ea97672f-4d14-4766-b23d-c425f9bb8a4a.png)
But in the list view, you can't display the labels
![image](https://user-images.githubusercontent.com/6314950/197083681-790d6bc2-1f5f-4283-ab85-7a072f8e148c.png)

The same is true for the Property Tab List View
This record has a label
![image](https://user-images.githubusercontent.com/6314950/197084032-634d7c18-800d-47e8-8693-2bf4fd75bfc4.png)

But it doesn't appear in the List View
![image](https://user-images.githubusercontent.com/6314950/197084172-f1984f89-208d-4dcd-8a63-4deb31629164.png)

RDmitchell:
The reason why the labels aren't showing up in the List View is because I didn't specify "Show in List", which is set in Org/Labels.
![image](https://user-images.githubusercontent.com/6314950/197255167-4f474339-408b-4d7d-bf25-f35a8f7fc6d8.png)

When I go to the Org/Labels view for the suborg (LBNL 101 SubOrg 2) I don't see any labels. 
![image](https://user-images.githubusercontent.com/6314950/197255114-c5e4f0ef-7c4e-44ad-a0f5-7f84d65ad0b4.png)

But in the Property List View for that Suborg, I am able to add labels "on the fly"
![image](https://user-images.githubusercontent.com/6314950/197255719-048ae46f-6459-4d0b-a72e-0d3490fcdeb8.png)
![image](https://user-images.githubusercontent.com/6314950/197255617-b09a6d56-4739-405b-b802-128aaa24e5fd.png)

RDmitchell:
When I try to create a label "on the fly" from the Property List, when I click the Done button, I get a 500 error
[...]

PR #3290, opened 2022-06-03
fix listing of labels for suborgs

nllong:
A suborg's labels are not a function of the parent orgs. If you have a suborg and create a label, then the label is never seen because the creation of the label is at the suborg.

#### What's this PR do?
* query the existing org, not the parent org

* Create a suborg
* Add a new label in the sub org
* Before this PR notice how the labels won't appear
* On this branch, notice how the label is now there
* Try to assign the label to a propety

#3289 

image


isalanglois:
@nllong please review

PR #3298, opened 2022-06-09
Add id

haneslinger:

I forgot id the first time around

image
image


seed/models/sensors.py
```diff
@@ -15,6 +15,7 @@ class DataLogger(models.Model):
     )
 
     display_name = models.CharField(max_length=255)
+    identifier = models.CharField(max_length=2047, default="")
```
nllong, line 18: For very long ids

nllong (approved):

RDmitchell:
Can we call this field "Datalogger ID" ??

seed/models/sensors.py
haneslinger: 😬

haneslinger:
> Can we call this field "Datalogger ID" ??

we can in the UI

RDmitchell:
Datalogger ID in the UI is fine ... thx !

Issue #3302, opened 2022-06-09
Error Running BETTER


Ryoken:
**Describe the bug**

> seed_celery    | Traceback (most recent call last):
seed_celery    |   File "/usr/lib/python3.9/site-packages/celery/app/trace.py", line 451, in trace_task
seed_celery    |     R = retval = fun(*args, **kwargs)
seed_celery    |   File "/usr/lib/python3.9/site-packages/celery/app/trace.py", line 734, in __protected_call__
seed_celery    |     return self.run(*args, **kwargs)
seed_celery    |   File "/seed/seed/analysis_pipelines/pipeline.py", line 233, in _run_task
seed_celery    |     raise e
seed_celery    |   File "/seed/seed/analysis_pipelines/pipeline.py", line 151, in _run_task
seed_celery    |     return func(*args, **kwargs)
seed_celery    |   File "/seed/seed/analysis_pipelines/better/pipeline.py", line 466, in _process_results
seed_celery    |     column_description=column_data_path.column_description,
seed_celery    | AttributeError: 'ExtraDataColumnPath' object has no attribute 'column_description'

**Expected Behavior**

Creating a BETTER analysis should work.

**Actual Behavior**

It does not.

nllong:
This ticket is no longer valid. The latest GHG column fix #3332 updated how columns are handled in Analyses.

PR #3303, opened 2022-06-09
fix: column path does not have a column description

Ryoken:
See ticket.

#### What's this PR do?
Allows for making BETTER analysis again by removing the column_path attreibute that doesn't exist.

Try making a BETTEER analysis.

https://github.com/SEED-platform/seed/issues/3302

### DIFF
--- a/seed/analysis_pipelines/better/pipeline.py
+++ b/seed/analysis_pipelines/better/pipeline.py
@@ -463,7 +463,7 @@ def _process_results(self, analysis_id):
             is_extra_data=True,
             column_name=column_data_path.column_name,
             display_name=column_data_path.column_display_name,
-            column_description=column_data_path.column_description,
+            column_description=column_data_path.column_display_name,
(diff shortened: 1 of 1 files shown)


perryr16 (approved):

anchapin (commented):
@perryr16 this change makes the issue go away with BETTER not running, but is there an underlying issue as to why there isn't a "column_description" in this case?

perryr16:
@anchapin that was my first thought as well.  `column_data_path` is the iterable of `column_data_paths` which is an array of `ExtraDataColumnPath`s. While `column_description` is an attribute of `Columns`, `column_description` is not defined on `ExtraDataColumnPath`, therefore it wont exist.

see /seed/analysis_pipelines/better/helpers.py line 51 for it's definition:
`ExtraDataColumnPath = namedtuple('ExtraDataColumnPath', ['column_name', 'column_display_name', 'unit_multiplier', 'json_path'])`

Issue #2760, opened 2021-07-21
Derived Columns: allow derived column definitions and parameters to be used in multiple derived column definitions


RDmitchell:
**Is your feature request related to a problem? Please describe.**
Right now, the derived column feature requires the user to define the same fields over and over for each individual definition, which gets tiresome and is error prone.

**Describe the solution you'd like**
* Allow the user to define a set of parameters separately from the derived column definitions and then those parameters can be used over and over again in different derived column definitions.
* Also allow the defined derived columns to be used as parameters in other derived column definitions. 

**Describe alternatives you've considered**
None

**Additional context**
See this issue (#1641) for an example:
* When defining the fields for the Percent Electricity Use and Percent Natural Gas Use, each definition needed the parameter "Total Site Energy Use kBtu".
* It was necessary to define that parameter separately for each of the definitions for Electricity and Natural Gas, and they had to have different names (at least I think they did).
* It would have been much better to be able to define it once and then use that variable in both definitions
  * definition for Natural Gas
![image](https://user-images.githubusercontent.com/6314950/126409839-e5146dfb-c4ee-41e5-951b-8668e5b55437.png)
  * definition for Electricity
![image](https://user-images.githubusercontent.com/6314950/126409931-64c405db-6cfd-4c46-a612-dc525be60f0b.png)

An example of being able to use a Derived Field as a parameter can be seen in this example
* I defined a derived field to calculate the floor area of each property record, as that data wasn't in the original file.
![image](https://user-images.githubusercontent.com/6314950/126409499-a9fc6edc-34a5-4615-915f-95a8fe319c88.png)
* It would now be very useful if I could use that derived column, "Floor Area Derived Field", as a parameter in other derived columns.

RDmitchell:
@macintoshpie -- here is the feature I was mentioning on Slack

RDmitchell:
@nllong / @perryr16 -- This was closed because it was marked stale, but it is still very relevant. Without this feature the user has to redefine fields over and over within new derived fields if they are using them in other derived fields.

RDmitchell:
@isalanglois / @nllong -- can this be closed since it is now possible to use derived columns to derive other columns?

nllong:
Yes this is now complete.

PR #3305, opened 2022-06-10
2760 feat/recursive derived column use

perryr16:
To integrate SEED with BPS there are two features that need to be added to the Derived Column Behavior
1. derived columns can be reordered in the inventory list views 
   - The current functionality places columns first and derived columns second. Derived columns can be reordered within themselves, but will always be listed after the columns. 
3. derived columns can use derived columns in their definitions 
   - The current functionality only allows Columns to be used to define a Derived Column

Both changes require Derived Columns to act like (regular) Columns.

#### What's this PR do?
The primary change is to generate a one to one relationship between the `Column` model and the `DerivedColumn` model. This allows a `DerivedColumn` to be replaced with its related `Column` when being displayed. When it needs to be evaluated the logic will trace the `Column` to its `DerivedColumn` where normal functionality still exists. 

When a `DerivedColumn` is created the logic will automatically generate a `Column` with a property `.derived_column` that will point towards the `DerivedColumn.id`. The existing `Column`s will have a `derived_column` property with a value of `None`.

When defining a derived column its source column may be a `Column` with a related `DerivedColumn`. The logic will traverse the source column tree to determine if the source column contains a reference to the current column. If so an error message will be displayed indicating a circular definition has been created, and the user will be unable to save the derived column.

Now that derived columns may contain a reference to other derived columns, only derived columns that do not contain a reference to a derived column may be deleted. If a user attempts to delete a derived column that contains a reference to another derived column the user will be warned and the derived column will remain.

Any test file will work assuming there is a column with integer values.
You will need to run the new migration

- Define a derived column with another derived column 
    - Create a derived column based on a column with integer values 
    - Create a second derived column that uses the first derived column in its definition (ex: dc2 = dc1 + 10)
    - Confirm values on inventory or beta list

- Test derived columns for circular definitions 
    - to do so, you'll need to create 3 or more derived columns to prove the logics nested capabilities 
    - create a derived column "dc1". It's definition isn't important as we'll be changing it soon.
    - create dc2 = dc1 
    - create dc3 = dc2
    - edit dc1 = dc3 
    - There should be a circular error warning.

- Attempt to delete derived column with definition containing a derived column
    - best to test after testing for circular definitions 
    - attempt to delete dc1 or dc2. Because dc3 is dependent it should prevent a successful deletion. The error message should be clear
    - delete dc3, then dc2, then dc1 successfully.

- Reorder derived columns in the inventory list 
    - generate a few derived columns
    - on the inventory list and beta list click and drag derived columns until they are positioned to the left of regular columns 
    - refresh and confirm they are in the same position.
    - go to column list profile and click and drag derived columns above regular columns 
    - save the profile
    - view the inventory list or beta list and confirm they are in the correct position.

#2760


anchapin:
Everything works other than I noticed that if you rename a derived column, if you try to select the revised derived column in the Source column drop down menu, you still see the original name in the list and not the latest version of the derived column name. I'm not sure if this should be fixed for this pull request or if a new issue should be made.

perryr16:
> Everything works other than I noticed that if you rename a derived column, if you try to select the revised derived column in the Source column drop down menu, you still see the original name in the list and not the latest version of the derived column name. I'm not sure if this should be fixed for this pull request or if a new issue should be made.

Thanks for catching this. I'll update to include that functionality.

seed/static/seed/js/controllers/derived_columns_editor_controller.js
```diff
@@ -130,8 +136,52 @@ angular.module('BE.seed.controller.derived_columns_editor', [])
...
+        if (!source_column.derived_column) {
+          return false
+        } else if (source_column.derived_column == $scope.derived_column.id) {
+          return true
+        } else {
+          let source_derived_column = $scope.derived_columns.find(dc => dc.id == source_column.derived_column)
+          let nested_derived_columns = check_parameters_for_nested_derived_columns(source_derived_column.parameters)
+
+          const current_column = property_columns_payload.find(col => col.derived_column == $scope.derived_column.id);
+          console.log('RECURSION?', nested_derived_columns.includes(current_column.id)) // delete this
```
Ryoken, line 157: Still need this deleted?

seed/static/seed/js/controllers/inventory_detail_settings_controller.js
```diff
@@ -171,33 +156,26 @@ angular.module('BE.seed.controller.inventory_detail_settings', [])
 
       var currentColumns = function () {
         const columns = [];
-        const derived_columns = [];
+        // const derived_columns = [];
```
Ryoken, line 159: Is this needed?

seed/static/seed/partials/inventory_detail_settings.html
```diff
@@ -48,7 +48,6 @@ <h2><i class="fa fa-cogs"></i><span translate>Column Order/Visibility</span></h2
 <div class="section_content_container" ng-cloak style="clear:both; overflow:hidden;">
     <div class="section_content with_padding" style="overflow:hidden;">
         <h3 translate="NUMBER_COLUMNS_AVAILABLE" translate-values="{ count: (data.length | number) }"></h3>
-        <p class="text-danger">NOTE: ordering of all derived column (<i class="fa fa-link"></i>) rows is currently ignored; they will all appear at the end of the list when viewed.</p>
```
Ryoken, line 51: hooray!

Ryoken (commented):
It worked when I added the first derived column.  After adding the second with a reference to the first, I got an error trying to display the inventory detail page.

I only have two derived columns:
Screen Shot 2022-06-15 at 1 54 00 PM
Screen Shot 2022-06-15 at 1 54 47 PM
Screen Shot 2022-06-15 at 1 55 08 PM

Here is my column profile:
Screen Shot 2022-06-15 at 1 57 04 PM

I get two errors displaying the inventory list:
Screen Shot 2022-06-15 at 1 58 30 PM

Screen Shot 2022-06-15 at 1 58 06 PM

anchapin (approved):
From my perspective everything works now.

Ryoken (approved):
Looking good, no more errors!

seed/migrations/0170_column_derived_column.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.2.13 on 2022-05-11 18:29
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('seed', '0169_auto_20220616_1028'),
```
nllong, line 10: Is there any work to do to make previous existing columns work with this new migration?

perryr16: Yes. We would need to run a one-time script to create a related `column` for each `derived_column`. I believe this could be done by adding some logic to the migration.

perryr16: The updated migration will generate a `derived_column = None` attribute for each existing column, and each derived column will generate a related column. 

Users will have to reselect their derived columns in their column list profiles

nllong:
This is looking good. I confirmed with the production database that:
* you can't map to a derived column
* column migrations appear to have worked without issues

Few other things that I noticed.

1) You can still map to the derived column name, even though it should error as a duplicate/reserved column name on the mapping screen. If the user does end up mapping, then it crashes the organization.

image

```
  File "/Users/nlong/working/seed/seed/seed/views/v3/columns.py", line 105, in list
    columns = Column.retrieve_all(organization_id, inventory_type, only_used)
  File "/Users/nlong/working/seed/seed/seed/models/columns.py", line 1450, in retrieve_all
    raise Exception("Duplicate name '{}' found in columns".format(c['name']))
Exception: Duplicate name 'new column name_85428' found in columns
```

2) If you have a derived column displayed in your inventory, then rerun "show only populated columns", then it will not keep any derived columns.

image

I think this is okay, because we won't really know if a derived column has a value, so maybe we just put a warning in the "Show only populated columns" modal.

RDmitchell:
I suspect in the long run we will want to include derived column when running "Show only populated columns", but for now I think this is fine. It will be good to get some user feedback before making that kind of change.

PR #3309, opened 2022-06-15
include filename in post

nllong:
When uploading a file from Python requests into SEED, the object should take a tuple with (filename, filecontent), this wasn't happening, which made the name "file" on the frontend. This wasn't breaking anything, but it helps users use the API correctly if they are using this file as an example of how to use the API.

#### What's this PR do?
Pass file as `{file': (pathlib.Path(upload_filepath).name, pathlib.Path(upload_filepath).read_bytes())}`

If the unit tests pass, then we should be good. This file is run as part of the API integration tests.

n/a


perryr16 (approved):
Syntax was close! Tests pass locally.

Issue #3313, opened 2022-06-17
Add property search on exact result match


nllong:
The properties/search endpoint performs a `icontains` on the `identifier` field which is nice unless we know exactly what result we want. For example if the custom IDs are 

B-1
B-2
...
B-10
B-11

Then the search with `api/v3/properties/search?identifier=B=1` results in B-1* (so B-1, B-10, B-11). In some cases we really want only "B-1". It is suggested to a new identifier filter for exact matches.

<img width="1539" alt="image" src="https://user-images.githubusercontent.com/1907354/174221306-77f0668f-8278-4de9-bdda-32c1bc585d34.png">

nllong:
This is an API update. No testing needed @RDmitchell

Issue #3174, opened 2022-03-18
Add canonical GHG fields


nllong:
For BPS we need to promote GHG fields to canonical fields in the SEED database.

Add the following:
```
Total GHG Emissions (MtCO2e/year)
Total Marginal GHG Emissions (MtCO2e/year)
Total GHG Emissions Intensity (kgCO2e/ft2/year and kgCO2e/m2/year)
Total Marginal GHG Emissions Intensity (kgCO2e/ft2/year and kgCO2e/m2/year)
```

We should make these Pint/Unit aware.

Second step will be to update the carbon analysis to save the results into the correct fields (see #3175).

RDmitchell:
@nllong -- here is what we now have -- 3 different CO2 emissions from 3 different sources (not sure what the source would be for the new canonical fields -- presumably the CO2 calculation?)

This seems very confusing to me, and potentially to the users. Seems like we need a way to distinguish between them with a slightly different naming convention. Should the new fields be called "SEED calculated Total GHG ..." or something like that, so that someone wouldn't populate those fields with ESPM fields for the same data, for example? 

![image](https://user-images.githubusercontent.com/6314950/171737010-4397b2fc-43da-41af-b527-446dec46cf6a.png)

RDmitchell:
@nllong -- for the new canonical fields, where are the units defined / displayed? We don't have a place for them in Column Settings, although that does seem like the right place to define them
![image](https://user-images.githubusercontent.com/6314950/171737945-fbafa699-5774-4c17-90e9-243532f2246e.png)

RDmitchell:
Units are displayed in the Analysis section
![image](https://user-images.githubusercontent.com/6314950/171738716-eec642e0-f481-4b35-9c55-1db5b5d2987c.png)

But not for the fields in Property List View
![image](https://user-images.githubusercontent.com/6314950/171739108-7fced353-d1a0-4431-8379-a4a2aa8d6e1d.png)

In the Property Detail view, there are no units displayed and the values for the same thing, ie, Total GHG Emissions, are in different units within the same page display
![image](https://user-images.githubusercontent.com/6314950/171739739-34072992-d223-4b81-bee5-168aec25fcd7.png)

RDmitchell:
@nllong -- I think this needs more work

Ryoken:
I'll do my best to address every question!

Column List Profile
- The red columns are added to the property state, can come from any source and have dynamic units.
[...]

PR #3314, opened 2022-06-17
add ghg serializers

nllong:
The serializers for the GHGs columns were failing. Having a search return `icontains` is not always the best option if we know exactly the field that we want to query.

#### What's this PR do?
* adds the pint serializers
* adds new `identifier_exact` filter to the properties/search endpoint

* unit tests, monkey tests

#3174 
#3313


Ryoken (approved):
Looks good to me!

Issue #3325, opened 2022-06-22
fix of duplicate BETTER GHG Fields


RDmitchell:
**Describe the bug**

For orgs that did BETTER analysis before a certain time, extra data fields were created for the GHG data. We then added those fields as canonical fields, so now it is possible to get duplicate fields, which causes the org to be inaccessible. 

Need to identify the orgs that this is an issue for, and delete the extra data fields (on the server, not through the UI)

Orgs that I know about: JCC, LBNL 405

If possible, it would be good to make a list of orgs to clean up, as not all orgs using BETTER need this to happen (if they just started doing BETTER analysis, for example)

here are the duplicate fields
![image](https://user-images.githubusercontent.com/6314950/175103358-dce99f0e-1ea3-41a7-b684-5863f810319c.png)

**Instance Information**

Instance: production
SHA: 6ffea0837

nllong:
Here are my notes to fix this on dev for the `SEED (Demo)` org.

![image](https://user-images.githubusercontent.com/1907354/175406101-cd558c04-3e59-4ceb-adc7-d9f749c88745.png)

With error:
```
web File "/seed/./seed/models/columns.py", line 1444, in retrieve_all
web raise Exception("Duplicate name '{}' found in columns".format(c['name']))
web Exception: Duplicate name 'better_ghg_reductions_combined_79925' found in columns
```

```
select organization_id, orgs_organization.name, column_name, seed_column.id, is_extra_data, seed_column.created from seed_column
inner join orgs_organization on seed_column.organization_id = orgs_organization.id
where seed_column.column_name LIKE '%better_ghg%'
order by organization_id, created
```

Look for duplicates. If there are duplicates in the column_name, then find the column_descriptions that have kgCO2e (these are the old units), then get the column_id (my computer uses the period as the 1000 separator), so in the case below, 79925, 79932, 79933.

![image](https://user-images.githubusercontent.com/1907354/175406152-fb699446-60df-46b9-93d3-bff4b2336034.png)

![image](https://user-images.githubusercontent.com/1907354/175406172-edecafdf-f4f1-4bbc-aad9-52c38d46677b.png)

If the delete errs due to column profiles, then grab and delete those records as well.

![image](https://user-images.githubusercontent.com/1907354/175406197-e133d156-f8a5-403d-a06f-254222a24eb0.png)

Also look for total_ghg_* items. They may be shown as extra_data fields and exist in the column_mapped table, and will need to be removed too.

```
[...]

PR #3332, opened 2022-06-25
fix duplicate column errors

nllong:
The SEED analysis functionality is able to create new columns as needed. In some cases, the `column_name` may already exist in the database and it will create a new instance of the column with the same `column_name` but different `display_name` or `column_description`. This was causing duplicate column errors in the database.

#### What's this PR do?
1. creates a migration to remove certain duplicate fields. It only checks for mapped fields (where `table_name` is not None). If there is more than one of the fields, it removes the most recent. This may cause data loss if the user ran an analysis that used this field after the GHG fields were updated. Presently, there are only a few orgs (and they are all test orgs) that are experiencing this issue. The migration should handle any mappings.
2. Update the Column.objects.get_or_create method on the analyses to first `get_or_create` on the minimum "non-user" configurable data. After finding or creating the column, then assign the `column_display_name` and `column_description`. It only assigns the fields if it is created since the user may have renamed the display name or column description. 

* Run the migrations on your local DB and make sure that everything works.
* For bonus points, download the latest DB dump and run on the production, staging, or dev1.

#3325


Ryoken (approved):
After reordering the migrations I Iwas able to get eeverythign running as expected!

nllong:
@axelstudios  - can you review this one too before we merge?

seed/migrations/0171_auto_20220624_1427.py
```diff
@@ -0,0 +1,41 @@
...
+                # for column in columns:
+                #     print('Found multiple total_ghg_emissions columns, what to do?')
+                #     print(serializers.serialize('python', [column], ensure_ascii=False))
+                columns[0].delete()
+        
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('seed', '0170_column_derived_column'),
```
nllong, line 36:
```suggestion
        ('seed', '0171_update_bae_assets'),
```

need to rebase migrations, again.

isalanglois:
@axelstudios  Can you review this?

Issue #3327, opened 2022-06-23
Cloudflare error when loading Sensor data


RDmitchell:
**Describe the bug**

When loading some sensor data into SEED, the program produces a CloudFlare error. It is not possible to recover from that -- the Sensor page can no longer be displayed and the inventory record must be deleted (and then potentially you have to start over trying to load sensor data).

**Expected Behavior**

Should be able to load sensor data without a CloudFlare happening, and most of the time it seems to work properly.

**Actual Behavior**
![image](https://user-images.githubusercontent.com/6314950/175423435-a44fe439-eb00-4484-83c8-d177964ec826.png)

**Steps to Reproduce**

I will try to come up with a reproducible set of steps to see this error. Not sure if I will succeed

**Instance Information**

Instance: Dev1
SHA: 5f181142b

RDmitchell:
I think this might have to do with how much data has been loaded into the Sensor tables, because it doesn't seem to be tied to a particular sensor data file.

RDmitchell:
Here is a link to a movie showing the problem. It happens on import of the 3rd set of data
https://drive.google.com/file/d/1JnPV9N9MhKuKvrccjO5xqyInIv3E25gG/view?usp=sharing

RDmitchell:
This folder has all the files needed to work on this issue
https://drive.google.com/drive/folders/1JkMP-dlXG_vy7Sx58e8BU7C9RjSl5y2O?usp=sharing

axelstudios:
@RDmitchell @haneslinger 
The issue is that once you have a certain number of sensor readings uploaded, the  `/api/v3/properties/3211746/sensor_usage/` endpoint can take longer than 60 seconds to return, and CloudFlare immediately terminates the request if it takes longer than that.  We need to decide if we actually want to show all data at once on that page, or if we want to take an approach similar to the Inventory List Beta page with pagination to reduce the amount of data being transferred:

### After uploading Grade 2 (125,789 readings):
The `sensor_usage` endpoint takes 35 seconds, and transfers 34MB of uncompressed data

### After subsequently uploading Grade 1 (68,026 readings):
The `sensor_usage` endpoint takes 57 seconds, and transfers 52MB of uncompressed data

### After subsequently uploading Grade 3 (67,425 readings):
_**The `sensor_usage` endpoint takes at least 60 seconds before CloudFlare terminates the request, making the page permanently broken until the property itself is deleted**_

## Additional bugs
[...]

PR #3333, opened 2022-06-29
Paginate sensor readings

haneslinger:
1. Go to any property
2. Add a datalogger
3. add sensors with `example-sensor-metadata.xlsx`
4. add sensor readings with `example-sensor-readings.csv`
5. Observe that the sensor readings are paginated.

#3327


perryr16 (approved):
Approved. This makes the sensor reading table much easier to navigate

axelstudios (approved):
Tested with Robin's big data, looks good

PR #3334, opened 2022-06-29
Dismiss button works, No missing cycle, fix grid height

haneslinger:


Ryoken (approved):
Looks good!

axelstudios (changes requested):
The `Upload Sensors` button is now more broken than before:
image

That translation template should probably be `Uploading FILE to DATASET`

PR #3358, opened 2022-07-08
updating bae version and removing separate lxml dependency

kflemin:
Update lxml to 4.9.1
BAE also has a lxml dependency which creates mismatches.
Removing the separate lxml dependency in favor of the one within BAE


nllong (approved):
sweeeeeet!

Issue #3343, opened 2022-07-06
Analysis results slow to load when connected to lots of properties


nllong:
**Describe the bug**

When running a "large" analysis with many properties and with potential lots of resulting data, the loading of the property detail page can be really slow. 

**Expected Behavior**

**Actual Behavior**

**Steps to Reproduce**

* Load example better data
* Make sure to let the meter data import
* Add BETTER API Key to org settings
* Select all of the properties
* Select `Run Analysis` and then `BETTER`
* Wait
* View a property detail page. it will take 20 seconds to load.
 
**Instance Information**

dev1
Org ID: 371

**Additional Context**

PR #3359, opened 2022-07-08
Speed up list analysis

haneslinger:
#### What's this PR do?

The first commit speeds up the db queries and serialization of the "views" returned by `/api/v3/analyses/` by about 50%
The second makes returning the views optional, as they aren't used. That speed things up _a lot_

1. Create analysis ([as described here](https://github.com/SEED-platform/seed/issues/3343#issue-1296292747)) 
2. Go to property detail page. 
3. Say "wow, so fast."

https://github.com/SEED-platform/seed/issues/3343

### DIFF
--- a/seed/models/analysis_property_views.py
+++ b/seed/models/analysis_property_views.py
@@ -95,19 +95,19 @@ class AnalysisPropertyView(models.Model):
         property_view_query = models.Q()
         for analysis_property_view in analysis_property_views:
             property_view_query |= (
-                models.Q(property=analysis_property_view.property)
-                & models.Q(cycle=analysis_property_view.cycle)
+                models.Q(property_id=analysis_property_view.property_id)
+                & models.Q(cycle_id=analysis_property_view.cycle_id)
             )
         property_views = PropertyView.objects.filter(property_view_query).prefetch_related('state')
 
         # get original property views keyed by canonical property id and cycle
         property_views_by_property_cycle_id = {
-            (pv.property.id, pv.cycle.id): pv
+            (pv.property_id, pv.cycle_id): pv
             for pv in property_views
         }
 
         return {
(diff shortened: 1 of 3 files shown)


seed/models/analysis_property_views.py
```diff
@@ -95,19 +95,19 @@ def get_property_views(cls, analysis_property_views):
         property_view_query = models.Q()
         for analysis_property_view in analysis_property_views:
             property_view_query |= (
-                models.Q(property=analysis_property_view.property)
-                & models.Q(cycle=analysis_property_view.cycle)
+                models.Q(property_id=analysis_property_view.property_id)
```
nllong, line 98: I'm pretty sure this is functionally the same, but being explicit that it is the `_id` is nicer.

seed/static/seed/js/services/analyses_service.js
```diff
@@ -32,7 +32,7 @@ angular.module('BE.seed.service.analyses', [])
 
       const get_analyses_for_canonical_property = function (property_id) {
         const org = user_service.get_organization().id;
-        return $http.get('/api/v3/analyses/?organization_id=' + org + '&property_id=' + property_id).then(function (response) {
+        return $http.get('/api/v3/analyses/?&include_views=false&organization_id=' + org + '&property_id=' + property_id).then(function (response) {
```
nllong, line 35: 🔥

nllong (approved):
looks great, the loading of a single property after running 100 properties through better was *much* faster!

seed/models/analysis_property_views.py
haneslinger: I could be wrong, but I think before it was retrieving the whole cycle and property from the db, which cost us an extra query per loop.  

I remember less queries being logged after the change, but there were a lot of logs, so I could be confused.

Issue #3393, opened 2022-07-18
Control the number of decimal points being displayed in derived columns


RDmitchell:
**Is your feature request related to a problem? Please describe.**

right now, it appears that the default number of decimal places to display for derived columns is about 16, which is way too many.  Although in a few cases, there are 1 or 2 decimal places

![image](https://user-images.githubusercontent.com/6314950/179431023-518f1dc2-58bc-47bd-a064-7bd2be39bc61.png)

**Describe the solution you'd like**

*Default to 2 decimal places
*Add an option for each derived field as to how many decimals to display, ie, could be zero, so that 24.3433 becomes 24.

**Describe alternatives you've considered**

none

**Additional context**

Instance: production seedv2
SHA: 9d54e72b7

nllong:
closing for release. Reopen if something goes wrong.

PR #3397, opened 2022-07-18
3393 derived col 2 decimal pt hardcode

perryr16:
Derived Column evaluations came back with any range of decimal points.

#### What's this PR do?
Adds a hardcoded default of 2 decimal points

`NOTE: This is a temporary but solid fix for the demo. Adding a decimal count setting to the derived column editor for each derived column is currently in development.`

- create a derived column that will evaluate to values with more than 2 decimal points. 
- confirm there are only 2 decimal points

#3393 

dc2 = PM Property ID / 7
dc3 = dc2 / 7
Screen Shot 2022-07-18 at 10 15 32 AM


perryr16:
Moving in a different direction - using organization settings

Issue #3394, opened 2022-07-18
Fix when selecting delete doesn't delete records in new view


RDmitchell:
**Describe the bug**

In the new, used to be beta, view of the Tax Lot tab, selecting the Delete option from the Actions menu doesn't do anything. 
![image](https://user-images.githubusercontent.com/6314950/179431809-e7f62e51-2d20-4b4c-b8ac-249b29979b9f.png)

If I go to the legacy tax lot tab list view, it displays this modal when I select Actions/Delete, which is the appropriate behavior
![image](https://user-images.githubusercontent.com/6314950/179431876-6ff221d5-6711-4faa-b1d1-5b6a340eaa31.png)

**Expected Behavior**

It should ask if I want to delete the records

**Actual Behavior**

Nothing happens

**Steps to Reproduce**

Go to the Tax Lot Tab in list view
Select record or records to delete
Select the Delete option from the Actions pulldown
Nothing happens

**Instance Information**

Instance: production seedv2
SHA: 9d54e72b7

nllong:
closing for release! Reopen if this issue is still happening.

PR #3398, opened 2022-07-18
fix: taxlots->selectall->delete now works

Ryoken:
The new inventory list had a bug where batch deleting did not work for taxlots.

#### What's this PR do?
Allows batch deletes for taxlots.

Load taxlots, select all, delete.

https://github.com/SEED-platform/seed/issues/3394


haneslinger (approved):
cool beans.

At some point we should probably refactor it file so get_taxlots and get_properties share their dulpicate code to avoid problems like this.

but that's not this ticket. 👍

PR #3410, opened 2022-07-20
Celery-queue API endpoint

axelstudios:
#### What's this PR do?
Creates an API endpoint for superusers to see the number of active, reserved, and scheduled tasks.  This is helpful for understanding if the celery queue is at max concurrency, or just in use.  I opted for an API endpoint rather than a managed task.

Note: you may see a `AssertionError at /api/v3/celery_queue/` if celery isn't running, and sometimes nondeterministically (no idea why)

##### Sample response
```json
{
  "active": {
    "total": 4,
    "tasks": [
      "seed.data_importer.tasks._save_raw_data_create_tasks",
      "seed.data_importer.tasks._save_raw_data_chunk"
    ]
  },
  "reserved": {
    "total": 0
  },
  "scheduled": {
    "total": 0
  },
  "maxConcurrency": 4
}
```

As a superuser, fetch the `/api/v3/celery-queue` endpoint while celery task are running (or aren't)
#3315


nllong (approved):
looks great! I had to run precommit to get the tests passing, but the functionality was there!

PR #3424, opened 2022-07-21
Fix modal usage

haneslinger:
Currently, using the "show only populated column" when you have no profile list returns a 400. That's because mal formatted data was given to the profile list modal. This fixes that.


Ryoken (approved):
Thanks for cleaning that up!  Looks good.

perryr16 (approved):
Works as expected, approved!

Issue #3392, opened 2022-07-17
Fix error filtering doubles as strings


nllong:
If there is an existing filter on a field that has a unit (e.g., site_eui) and the user is filtering in the Legacy property list, then when moving to the new property list, it breaks.

![Image](https://user-images.githubusercontent.com/1907354/179421316-2dda25b4-6bed-4075-b823-3bd3d225ca10.png)

The user is unable to delete the filter and reset the state. Current workaround seems to be clearing the local cache or going to the Property (Legacy) list and deleting the filters, then going back to the new Property list.

perryr16:
A few notes as this does not seem to be an easy fix

The source of the issue is the legacy page uses the HTML table to apply filters while the new inventory list uses hand-rolled functions in the backend to apply sorting and filtering

The backend filter functions had no way to handle a `is not null` query (`PropertyView.objects.filter(state__site_eui__isnull=False)`. I have since added a conditional to allow `is not null` queries for default columns. I have been unable to apply the same fix for extra data as they are a nested dictionary and the same style of query returns all property views regardless of the extra data value (`PropertyView.objects.filter(state__extra_data__data_1__isnull=False`)

Below is an example of the nested nature of the extra data in property states
```
PropertyState: {
...
site_eui: 123,
extra_data: {
   data_1: 123,
   data_2: 'test'
}
...
}
```
Filtering for derived columns is also not functional and will require further debugging.

PR #3428, opened 2022-07-25
3392 filter error between inventory lists

perryr16:
Filters of `is not None` break on the new inventory list.

#### What's this PR do?
Fixes filtering for `is not None` for default canonical columns

Go to inventory legacy and search filter a column that has empty values using `!=""` (double quotes, not single)
Navigate to inventory list and confirm there are no errors

#3392

Screen Shot 2022-07-25 at 2 58 32 PM


perryr16:
Please leave this ticket open as it does not address the error when filtering on extra data columns. Because extra data columns values are within a nested dictionary, checking for `is not None` will be more challenging

seed/search.py
```diff
@@ -548,6 +549,9 @@ def build_view_filters_and_sorts(filters: QueryDict, columns: list[dict]) -> tup
     new_filters = Q()
     annotations = {}
     for filter_expression, filter_value in filters.items():
+        if filter_expression[-4:] == '__ne' and filter_value == '':
+            filter_expression = filter_expression.replace('__ne', '__isnull')
+            filter_value = False
```
perryr16, line 554: I don't believe this is the most elegant solution, but it should work until that solution is found

Ryoken: Eew, yeah, that's pretty gross.  But based on what we talked about I can't see a better way to do this.

Ryoken (approved):

perryr16:
Just to document another option. Currently, we are always using the `filter` function, we could add an argument and a conditional to `exclude` on specific queries

nllong (approved):
nice, fix works. 🥳

PR #3432, opened 2022-07-29
allow users to pick metering dates for better analysis

### LINKED ISSUE #2953: BETTER: allow user to specify meter data to use for analysis
As a user, I want to select a range of meter data when running the BETTER analysis, so that I can <compare a building’s analysis results, avoid including “bad” meter readings, ???>

## previously discussed solution
- provide start/end date inputs (calendar selector like when creating cycles) when configuring the analysis
- add a checkbox next to the inputs which is effectively "Use all meter data" -- when selected, the date inputs are disabled (maybe effectively making the start/end dates `null` in the configuration).
- default the start/end date to the start/end of the current cycle
- allow users to edit these dates
- verify start < end
- use dates in analysis task to select meter readings

kflemin:
#### What's this PR do?
- provides start/end date inputs (calendar selector like when creating cycles) when configuring the analysis
- adds a toggle button which is effectively "Use all meter data" -- when selected, the date inputs are hidden. when toggled to "Select Meter Range", calendar inputs are displayed.

default the start/end date to the start/end of the current cycle

- Allow users to edit these dates
- Verifies start < end
- Uses dates in analysis task to select meter readings
- If dates selected don't match any meter data, no data will be sent, analysis will be created, but will most likely fail.  Unsure whether we want to return any details to the user when no meter data was retrieved....

Create a BETTER analysis using the modal.

Fixes #2953 

Create a new analysis: (showing the new toggle button...if set to "Use All Meter Data", dates don't show up
Screen Shot 2022-07-29 at 1 30 39 PM
Toggle to "Select meter Data Range" shows inputs and calendar icons. Fields are prepopulated with the cycle start/end dates:
Screen Shot 2022-07-29 at 1 30 45 PM
User can use calendar popup to pick different dates:
Screen Shot 2022-07-29 at 1 30 51 PM
Analyses page shows the new config fields: "select_meters" when set to "select" will use the dates entered. if set to "all" will not use dates and will instead use all meters stored.
Screen Shot 2022-07-29 at 1 30 11 PM
Analysis detail page shows the new config fields as well.
Screen Shot 2022-07-29 at 1 30 01 PM


seed/analysis_pipelines/better/pipeline.py
```diff
@@ -179,7 +204,17 @@ def get_meter_readings(property_id, preprocess_meters):
         )
         for meter in meters:
             # filtering on readings >= 1.0 b/c BETTER flails when readings are less than 1 currently
-            readings = meter.meter_readings.filter(reading__gte=1.0).order_by('start_time')
+            readings = []
+            if 'select_meters' in config and config['select_meters'] == 'select':
+                try:
+                    readings = meter.meter_readings.filter(start_time__range=[value1, value2], reading__gte=1.0).order_by('start_time')
```
nllong, line 210: 🔥

nllong (approved):
image

Looks great! I just removed one of the debug print statements.

PR #3435, opened 2022-08-02
Added inventory documents to media view & tests

ebeers-png:
#### What's this PR do?

- Adds directory handling for InventoryDocument media files, so that files that have been uploaded can be downloaded on a server running nginx
- Adds testing for retrieving an InventoryDocument file as an org member to test_media.py

When running SEED with nginx (like on a server), navigate to a property details page, upload a .pdf document, and then try downloading that document.

#3437


nllong (approved):
nice, thanks! 🏆

And thanks for writing some tests!

Issue #3426, opened 2022-07-22
Cross Inventory Filtering/Sorting


Ryoken:
The new inventory list breaks when you:

- view the property list and sort/filter on a taxlot column
- view the taxlot list and sort/filter on a property column

Ryoken:
We are going to disable sorting/filtering on related columns until a more elegant solution is found.

RDmitchell:
hmmm, seems like we need to make sure this stays in the queue ... kind of important.

Ryoken:
https://github.com/SEED-platform/seed/pull/3436

Temporary fix in place!

ebeers-png:
Some of the orgs I have in BEAM (for DC and Cambridge) need this one working -- does there happen to be a timeline on it?

anchapin:
Should this issue be closed due to #3499 being merged? Or did that one only address filtering and sorting still needs to be fixed for this issue to close?

isalanglois:
@anchapin let's bring this up in today's meeting

PR #3436, opened 2022-08-02
3426 fix/cross inventory filter

Ryoken:
See ticket.

#### What's this PR do?
disables sorting and filtereing on related columns

see screenshot

https://github.com/SEED-platform/seed/issues/3426

Screen Shot 2022-08-02 at 2 48 40 PM
Screen Shot 2022-08-02 at 2 48 22 PM


RDmitchell:
Seems like we need to have an issue to fix this, so that this isn't the final solution. Thx

perryr16 (approved):
This works. I agree we should generate a new ticket to track this, but I think this is an acceptable guardrail until we can give the user full functionality.

PR #3505, opened 2022-09-06
Add filter groups to insight page

haneslinger:


perryr16 (approved):
This worked great! My only comment (and it may not be related to this PR) is that selected filter_groups aren't being saved to the data_view. For example, I have 2 total filter_groups but only want 1 to be saved to the data_view. I select 1 and save, but when viewing it always shows all filter_groups

haneslinger:
Fixed!

anchapin (approved):
It seems to be working to me.

Issue #3499, opened 2022-09-01
Enable filtering of columns that appear in both inventory types


anchapin:
For example, the "Address Line 1" column for Properties can't be filtered in the Tax Lot version of Inventory List.

Ryoken:
https://github.com/SEED-platform/seed/pull/3515

RDmitchell:
Issue #3426 is also about cross Inventory Filtering / Sorting.

RDmitchell:
instance: staging
SHA: bb2fdd425
Org: LBNL Test 200

Was able to filter on an Tax Lot field in the Property List tab.

However, when I tried to add a label to those filtered records, I got several 500 errors
![image](https://user-images.githubusercontent.com/6314950/205788241-7b4b8e07-8740-4a50-89e6-1aadce533d12.png)

RDmitchell:
the filter worked using text, ie, filtering on Parcel Owner == Lucky University
![image](https://user-images.githubusercontent.com/6314950/205789067-cd9574d6-c480-4bd1-a0f0-f17e0caf8abe.png)

However, it didn't seem to work when I filtered on the Jurisdiction Tax Lot ID for the case where there were multiple tax lot IDs associated with a record
![image](https://user-images.githubusercontent.com/6314950/205789471-4333bd91-aa48-42b4-9d38-70d6223e02f3.png)
But it doesn't seem to work even if there is only one Tax Lot ID associated with a record
![image](https://user-images.githubusercontent.com/6314950/205789751-d5a0da3a-a632-422a-9808-879da4cbcc86.png)
But it did work filtering with a number on the address field
![image](https://user-images.githubusercontent.com/6314950/205789968-b1573fbc-bfe4-44a6-aac1-c849ddff5563.png)

RDmitchell:
Instance: Dev1
SHA: 294626cbd
Org: LBNL Test 200
Cycle: 2020 Compliance Cycle

**Filtering for blank values in a Tax Lot field -- Does not work**
In this example, there are 14 records in the cycle. 
2 of them do not have a value in the Jurisdiction Tax Lot ID column
![image](https://user-images.githubusercontent.com/6314950/211952520-db80cab4-09bf-489c-9d37-265a55dad853.png)

If I filter for the records with the Tax Lot field blank, ie, enter "" in the filter field, no records are returned. But the two records with that field blank should be returned
[...]

PR #3515, opened 2022-09-12
fix: only collect columns that aren't related

Ryoken:
When the backend attempted to filter on a column that had a non-unique name, it would pull the most recent column instead of the appropriatee column.  This came up when "related data" was causing the same column name to be present twice.

#### What's this PR do?
Excludes the columns with "related data" when building the list of column s by table name so the correct column should always be present.

Ensure filtering works on a field that isn't related but has a related column by the same name (such as "Address Line 1").

https://github.com/SEED-platform/seed/issues/3499


anchapin:
I'm still getting the same issue. See screenshot below for an example using "Address Line 1" column.
image

Ryoken:
> I'm still getting the same issue. See screenshot below for an example using "Address Line 1" column. image

uhhh, what on earth happened to my commit?  the changes I made were gone, and the commit is now showing that I changed half the file that I didn't even touch.

I'm going to re-push my change and see if that resolves it.

anchapin (approved):
It works for me now. Thank you for this fix.

haneslinger (approved):
Nice work!

Issue #3494, opened 2022-09-01
Refresh list of dataviews when on metrics overview page


aviveiros11:
When on the metrics overview page (haven't yet clicked a dataview), the delete "X" button needs to refresh the list of dataviews or refresh the page. Currently it deletes the dataview on the backend, but the dataview remains in the available list. - (P3)

PR #3522, opened 2022-09-16
Revised the Custom Reports page to remove custom report from list after deletion

anchapin:
#### What's this PR do?
Updates the $scope.data_views to remove the deleted item.

Delete a custom report and see if it gets removed from the list.

#3494


Ryoken:
This isn't working as described.  Before when deleting a custom report it would load the report list page (which wouldn't have the deleted report).  Now when I delete it, nothing happens and I am still looking at the deleted report's page (See url).

First delete worked but did nothing on the ui, second delete caused the error:
Screen Shot 2022-09-16 at 10 35 33 AM

Ryoken:
Ah, after talking with @anchapin it makes sense.  The redirect should happen if the currently loaded report is deleted, otherwise it should stay on the current page and just refresh the list.

Ryoken (approved):

Issue #3354, opened 2022-07-08
Enable ability to not map a column in SEED


nllong:
**Is your feature request related to a problem? Please describe.**

Currently, in SEED you are required to map all the fields in the imported file to a field in SEED. 

**Describe the solution you'd like**

It would be nice to have the ability to turn off specific fields that don't need to be mapped. For example, in some benchmarking data, there is monthly energy consumption that really shouldn't go into the SEED columns, but should be uploaded into meters. (Uploading into meters is a different problem that this ticket is not addressing.) 

**Describe alternatives you've considered**

Since each mapped column has to go to a unique field in SEED, there really is no way to make this happen without deleting the columns in the raw data being uploaded.

**Additional context**

Ideally, a checkbox can be placed next to each column row to enable/disable mapping of that column.

RDmitchell:
Yes this would be a great feature.

PR #3523, opened 2022-09-16
Enable ability to not map a column in SEED

haneslinger:
#3354

image


Ryoken (commented):
I don't know if it's related to this PR or not, but when I tried to import one of my usual quicky files I got an error:
Screen Shot 2022-09-16 at 1 41 19 PM


I tried another of my files and it worked.  While playing with it, I noticed that if I omitted all required fields I wouldn't get the typical error and was allowed to click on.

Typical error message when both Property ID and UBID are not mapped:
Screen Shot 2022-09-16 at 1 40 32 PM

Error not appearing when both are omitted:
Screen Shot 2022-09-16 at 1 40 42 PM

Ryoken (approved):

PR #3548, opened 2022-09-26
3537 Add tzdata dependency to docker images

axelstudios:
In order for docker to support the `TZ` environment variable (e.g. `TZ: America/Denver`) the `tzdata` dependency must be in the base image.  Otherwise logs will only ever appear with UTC timestamps

#### What's this PR do?
Adds `tzdata`

#3537


axelstudios:
> I know that we can't upgrade the python tz package because of some issues with how the new package handles stuff.
> 
> I think this should be okay since it is a system level package, just want to make sure that we recognize this comment: https://github.com/SEED-platform/seed/blob/develop/requirements/base.txt#L19-L20

Yeah, this works at the OS level so it shouldn't affect Python.  I've been using this in my local docker configs for a while because I hate having to remember which tz offset to apply to UTC to figure out the log timestamps...

nllong (approved):
cool, thanks for explaining.

PR #3550, opened 2022-09-26
3532 Fixed nav gaps

axelstudios:
#### What's this PR do?
Adds html comments so that whitespace doesn't create gaps between links

#3532

Before:
NavGaps


aviveiros11 (approved):
thank you for spotting this  - it would've driven me nuts if I spotted it after release.

anchapin (approved):
Nice!

Issue #3416, opened 2022-07-21
Add button to export charts to image format


anchapin:
Button should be overlayed on top of the chart in the top right corner and export to .png at the very least

PR #3561, opened 2022-09-28
Add ugly export buttons to charts

haneslinger:
I've added the download button to the charts. They aren't pretty, but they're there. Looks like placing them _on_ the canvas will require a little pixel math 🤮  (best docs I can find on it is this 45 minute youtube video https://www.youtube.com/watch?v=vT7jjVDA1IY).  Not impossible to do, just not fun, so I wanted to be sure we wanted to before I sink time into it.

https://github.com/SEED-platform/seed/issues/3416
image
image


aviveiros11:
Download works as expected and for a beta lgtm :shipit:

aviveiros11 (approved):

RDmitchell:
@haneslinger -- I don't think they are so ugly (!) ... they are pretty standard icons, so they should work just fine !!

Issue #3552, opened 2022-09-26
Custom reports: use the same color for each axis per filter group


anchapin:

haneslinger:
Can I get a little more clarification?

Right now, each line is defined by 1) It's filter group 2) it's aggregation method  and 3) the left or right axis.

left and right axis lines are differentiated by solid or dot lines, but if all lines of a filter group should be the same color, how do we tell aggregation methods apart? 


![Image](https://user-images.githubusercontent.com/30241543/196818762-97471efb-b498-495c-a470-bebc8042c6d4.png)

anchapin:
To clarify, the color should be the same for each filter group and aggregation method pairing for both axes. I forgot about the potential for there to be more than one aggregation method for each filter group.

RDmitchell:
Instance: dev1
SHA: 294626cbd
Org: BPS Test

In Custom Reports, it does look like the criteria specified in this issue have been implemented. 

There is a separate color for each Filter Group and for each Aggregation method for each filter group. 
![image](https://user-images.githubusercontent.com/6314950/211438251-21ed0236-6c3d-4bb9-8fd2-10171b0ea0aa.png)

The graph seems pretty hard to read, and probably needs some work, but it seems like this issue has been completed, so I am going to close it.

PR #3587, opened 2022-10-24
Use the same color for each axis per filter group

haneslinger:
https://github.com/SEED-platform/seed/issues/3552
image


anchapin (approved):
Works as intended! Thanks.

nllong (commented):
late to the party... but looks great!

Issue #3261, opened 2022-05-18
Remove Campus field (currently a Boolean)


RDmitchell:
**Describe the bug**

Campus (a Boolean field) is automatically added to records on import, but it is not useful or used by anyone.

**Expected Behavior**

Don't generate that field

**Actual Behavior**

That field exists

**Steps to Reproduce**

Import a record into SEED and you will see that Campus is automatically added
![image](https://user-images.githubusercontent.com/6314950/169105776-44d4331c-3ebd-422a-ae9b-a62b0d982ab9.png)

**Instance Information**
Instance: all / seedv2
SHA: Seedv2: 6ffea0837

RDmitchell:
I am presuming this field is still present. I will verify with testing on latest dev server 

If it is still there, we need to think about what to do about that.

RDmitchell:
instance: staging
SHA: bb2fdd425
Org: LBNL 529

I created a new org and imported a file into it. 

I then looked at the column list profile and the Campus field was not in the list. 
![image](https://user-images.githubusercontent.com/6314950/206034181-92b4fa8a-d0d1-47df-9150-4aef0043a5a5.png)

PR #3614, opened 2022-11-14
Remove use of campus on the property model

nllong:
The `campus` field has never been used and treated as a second class citizen. The original intent of this field was to denote that a property was a campus, meaning it had a PM Parent Property ID. The campus field was then to be denoted by the PM Parent Property ID equality the PM Property ID.

#### What's this PR do?

* removes the use of Property.campus
* Migration added for removing the field from the columns table. I confirmed that there was only Property.campus fields in the production database.

* @axelstudios  - can you confirm that `campus` was not used in the hashing of the records?
* Monkey test after running the migration.

#3261


seed/migrations/0180_remove_property_campus.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 3.2.15 on 2022-11-13 19:45
+
+from django.db import migrations
+
+def remove_old_campus_column(apps, schema_editor):
```
nllong, line 5: special part of the migration to remove the `campus` column across all orgs

seed/models/columns.py
```diff
@@ -139,7 +137,6 @@ class Column(models.Model):
 
     # These are columns that should not be offered as suggestions during mapping
     UNMAPPABLE_PROPERTY_FIELDS = [
-        'campus',
```
nllong, line 142: campus is now mappable!

nllong:
I'm pretty sure that this will require us to rehash the property states. If so, then let's add `state` and `zip` to the normalized address, then rehash then, since it will be required. @axelstudios  -- can you confirm?

seed/migrations/0174_fix_ghg_columns.py
```diff
@@ -3,14 +3,15 @@
 from django.db import migrations
 # from django.core import serializers
 
+
 def remove_old_ghg_columns(apps, schema_editor):
     """create a default BuildingSync column mapping preset for each organization"""
     Organization = apps.get_model("orgs", "Organization")
 
     better_columns = [
-        'better_ghg_reductions_combined',
```
nllong, line 11: simply formatting

seed/migrations/0165_column_column_description.py
```diff
@@ -18,6 +19,7 @@ def forwards(apps, schema_editor):
                     col.column_description = col.display_name
                 col.save()
 
+
```
nllong, line 22: simply formatting

seed/migrations/0169_auto_20220616_1028.py
```diff
@@ -19,7 +19,7 @@ def add_new_bsync_mappings(apps, schema_editor):
                 'from_field_value': 'text',
                 'from_units': None,
                 'to_table_name': 'PropertyState',
-                'to_field': 'Audit Template Building Id', 
+                'to_field': 'Audit Template Building Id',
```
nllong, line 22: simply formatting

seed/migrations/0170_column_derived_column.py
```diff
@@ -3,13 +3,14 @@
...
 
+
 def forwards(apps, schema_editor):
     DerivedColumn = apps.get_model("seed", "DerivedColumn")
     Column = apps.get_model("seed", "Column")
 
     with transaction.atomic():
         Column.objects.all().update(derived_column=None)
-        
+
```
nllong, line 13: simply formatting

seed/migrations/0172_update_at_building_id_target.py
```diff
@@ -13,6 +13,7 @@ def update_at_building_id(apps, schema_editor):
                     mapping['to_field'] = "audit_template_building_id"
             profile.save(update_fields=['mappings'])
 
+
```
nllong, line 16: simply formatting

nllong:
> I'm pretty sure that this will require us to rehash the property states. If so, then let's add `state` and `zip` to the normalized address, then rehash then, since it will be required. @axelstudios -- can you confirm?

Okay, after inspecting more, I think [this line](https://github.com/SEED-platform/seed/blob/develop/seed/models/properties.py#L404-L405) shows that only the state object is passed to the hash_object method and since the `campus` is on `property` (not `property_state`), then this update will *not* require a database rehashing.

axelstudios (approved):

axelstudios:
@nllong Looks good to me - Running migrations against the seedv2 database took a grand total of 12 seconds, and no `hash_object`s were harmed.

Issue #3600, opened 2022-11-10
Property List New: !"" and "" don't seem to be working


RDmitchell:
**Describe the bug**

in the Property List view, the filter boxes should accept the following
* !"" == Not blank
* "" == blank

using either of these in the new property list view returns the wrong number of records

**Expected Behavior**

For this inventory list, there are 10 blank records and 4 non-blank records for the Geocoding Confidence field
![image](https://user-images.githubusercontent.com/6314950/201002177-3b0fdcdb-c298-449d-bc69-45efb1f5989f.png)

**Actual Behavior**

Entering "" which should show blank records, no records are displayed
![image](https://user-images.githubusercontent.com/6314950/201002293-d2ff8ce4-f500-4806-aeb6-a9a50d985215.png)

Entering !"" which should show the records with data, the 10 blank records are displayed
![image](https://user-images.githubusercontent.com/6314950/201002361-abe37c03-18be-4367-8b1a-4d7b72ad91c5.png)

**Steps to Reproduce**

see above

**Instance Information**

Instance: dev1
SHA: fbd9c503f

RDmitchell:
Instance: dev1
SHA: 92acc007d
Org: LBNL Test 200
Cycle: 2018 Compliance Cycle
Column List Profile: 2018 Compliance Year Property Fields

See #3759 which expands on what the problems are with some of the filtering. 

But I am still seeing problems using "" and !"" on Tax Lot fields in the Property tab.

RDmitchell:
Instance: seeddemostaging
SHA: 992862f09
Release: 2.17.1

Org: LBNL Test 200

On the Property tab, this seems to be working for Property fields

But there seem to be some issue with the Tax Lot fields. 

I will make a new issue for that.

PR #3618, opened 2022-11-15
3499 allow filtering on all columns

haneslinger:
#### What's this PR do?
see ticket 

I'd use the `seed/tests/data/san-jose-test-properties.csv` and `seed/tests/data/san-jose-test-taxlots.csv` data

[#3499](https://github.com/SEED-platform/seed/issues/3499)

image
image
image


RDmitchell:
Is this deployed to dev1? I can't really review it unless it is there, which I suspect it isn't just yet. 

Once it is deployed, if it can be moved into the Test category, then I will see it to review/test it

haneslinger:
@RDmitchell Can you look at the screenshots and confirm this is the excepted behavior?

RDmitchell:
@haneslinger -- looking at the screenshots, it appears that there are tax lot fields displayed in the Property tab, and that they have data in them, but the only filtering screenshot I see is the 2nd one, where I guess "g" is used to filter all the fields with a "g" in them, which appears to be working correctly. 

However, the 3rd screenshot, which appears to be filtering by "" which means "no data" isn't filtering properly because it is showing records that have data in them, but I think this is related to another bug, which is #3600 

It might be worth running through all the types of filters that are possible for a field, which are shown here (I think this covers most if not all of them):

image

haneslinger:
> However, the 3rd screenshot, which appears to be filtering by "" which means "no data" isn't filtering properly because it is showing records that have data in them, but I think this is related to another bug, which is https://github.com/SEED-platform/seed/issues/3600

The functionally present on the legacy page and functionally desired here (as confirmed by @nllong) is that for each property, if _any_ of its associated taxlots pass the filter, _all_ taxlots remain shown.

RDmitchell:
I didn't realize that the tax lots records you were displaying were all associated with one property. 

I think it makes sense that you would show all the associated tax records, even if they are not blank for that field. 

On the production server, SHA 09736c1d6, this functionality seems pretty broken, but the proposed implementation seems good to me. 

And we can always reevaluate if we get feedback from users about it.

haneslinger:
DO NOT PUSH YET.

This PR breaks existing filter_groups, as the query dicts must now contain the column name followed by their id. 

I think there's gotta be some custom migration stuff.

nllong:
Considering that filter groups are not really used yet on production, could we skip the custom migration? Are there other issues that may creep in because of this change?

Thinking out loud, since columns are tied to orgs, then the id appended to the column​_name should always be the same across all cycles and filter groups within an org. If someone changes a column name, which is not common, then it would break their  existing filter group—as expected.

— nicholas
________________________________
Da: Hannah Eslinger ***@***.***>
Inviato: Wednesday, November 23, 2022 11:00:13 AM
A: SEED-platform/seed ***@***.***>
Cc: Long, Nicholas ***@***.***>; Mention ***@***.***>
Oggetto: Re: [SEED-platform/seed] 3499 allow filtering on all columns (PR #3618)


CAUTION: This email originated from outside of NREL. Do not click links or open attachments unless you recognize the sender and know the content is safe.

DO NOT PUSH YET.

This PR breaks existing filter_groups, as the query dicts must now contain the column name followed by their id.

I think there's gotta be some custom migration stuff.

RDmitchell:
Users do change field names, so we need a way to let them know that their Filter Groups will be broken when doing that.

Issue #3582, opened 2022-10-19
Give users with viewer member role access to see charts in Insights tab


anchapin:
If a user has the "viewer" member role, they don't see the data in the insights tab.

For example, the Project Overview page looks like this:

![Image](https://user-images.githubusercontent.com/6326294/196778032-a756917a-ec91-481a-88fe-0a6cd2297c3b.png)

Instead of this:

![Image](https://user-images.githubusercontent.com/6326294/196778181-2acbba41-7d78-4db6-9a48-2c88e97ca2ce.png)

nllong:
we need this for the next release @isalanglois

isalanglois:
@nllong I put the Assign label on this ticket to make sure this gets addressed. We can bring it up in our meeting tomorrow

nllong:
Doh, I think we forgot to assign this one. @anchapin  -- can you take care of this before the release?

nllong:
@haneslinger  -- do you think you can tackle this one before the release too? @anchapin - I think your plate is full, so trying to redistribute a bit.

RDmitchell:
Instance: dev1
SHA: c8b79f627
Org: BPS Test
User: RDM+622

My account is set to Viewer, and I can see the Insights Program Setup, but can't edit it, which seems correct

![Image](https://user-images.githubusercontent.com/6314950/219141985-040c83c4-24fb-42ed-8ffa-6ab281a88c05.png)

I can see the Program Overview screen

![Image](https://user-images.githubusercontent.com/6314950/219142132-972660ab-88c6-490c-885c-0774497c330e.png)

I can see the Property Insights graph

![Image](https://user-images.githubusercontent.com/6314950/219142247-15b29414-0b74-4661-ba11-ae4421b27270.png)

However, the Custom Reports graph doesn't appear to be displaying, and also the Actions option with Edit and Delete is available, which I don't think it should be. And there appears to be a 404 error.

![Image](https://user-images.githubusercontent.com/6314950/219143153-9c23ce8c-a612-400a-a30b-eb578af16f90.png)

This is what I see for Custom Reports when I am logged in as owner, ie, I see the graph, which I don't see when logged in as Viewer

![Image](https://user-images.githubusercontent.com/6314950/219143708-a32beb49-14cb-4cf0-a6a0-acb6f9359d24.png)

@isalanglois / @nllong -- I am moving this to the Q2 project, and I guess for starters, it will have "No Status"

RDmitchell:
[...]

PR #3619, opened 2022-11-15
Remove nosetests and fix precommit

nllong:
#### Any background context you want to provide?
Nosetests have been out of development for several years and no longer a preferred solution for django testing. I also needed to run a specific test with debug within VSCode, which wasn't supported by nosetest (at least no easily).

#### What's this PR do?
* removes the django-nose package and related code
* update the pytest.ini to support running with existing db
* fix the name of the normalize address test
* update flake8 pre-commit to point to github (was gitlab) and ran `pre-commit autoupdate` to get the latest versions

#### How should this be manually tested?
* verify that `./manage.py test` still works as it used to
* test using `pytest` method or just `pytest --collect-only` to save time

image

* test running a test or debug in VSCode (choose `pytest` as the testing framework with VSCode)
image

#### What are the relevant tickets?

#3617

PR #3620, opened 2022-11-16
Fix permissions for viewers to see program overview

haneslinger:
@kflemin, Just checking this line wasn't important.

[Give users with viewer member role access to see charts in Insights tab#3582](https://github.com/SEED-platform/seed/issues/3582)


nllong:
It needs to be changed to `requires_viewer`, otherwise it would be viewable to anyone (I think). 

https://github.com/SEED-platform/seed/blob/develop/seed/lib/superperms/orgs/decorators.py#L120

The decorator has some redundancy but should allow us to easily extend in the future for other roles... such as level-based access control.

nllong:
Also, the pre-commit fix is in #3619, if you can review that one quickly to fix all our failing tests, thanks!

nllong (approved):
sweet that fixed it. I logged in with a viewer only account and the page shows up.

image

It does appear that the viewer has permissions to change the Program Metrics Page, so we need to resolve that. Verify that `@has_perm_class` has been added to all of those API endpoints.

nllong:
Actually, take back the last comment about permission. The user doesn't have permissions to change it, but the toast says config was updated. So we need to catch that the API result (below) is translated to the appropriate toast notification.

```
{"status": "error", "message": "Permission denied"}
```



image

Issue #3497, opened 2022-09-01
Add ability to track multiple compliance programs


anchapin:

anchapin:
Yes, that's correct.

PR #3623, opened 2022-11-23
Added the ability to create multiple compliance programs

anchapin:
Previously there was only ever one compliance program available from the GUI. Additionally, you weren't able to associated a compliance program with a filter group.

#### What's this PR do?
This PR adds the ability to create, edit, and delete compliance programs from the Program Setup page. Additionally, you can associate a filter group with a compliance program. Only one filter group can be selected for each compliance program.

Go to Program Setup page in Organization Settings and create, edit, and delete a compliance program. Ensure to try associating a filter group with it also.

Note that I ended up merging a few compliance metric related migrations. In hindsight I realized that there was a non-compliance metric migration that depended on one of the migrations that I removed. Therefore, I revised that migration as well. However, you might need to start fresh and migrate all again.

#3497


seed/migrations/0177_compliance_metric.py
```diff
@@ -17,25 +17,22 @@ class Migration(migrations.Migration):
             fields=[
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('name', models.CharField(max_length=255)),
-                ('start', models.DateTimeField()),
-                ('end', models.DateTimeField()),
+                ('start', models.DateTimeField(blank=True, null=True)),
```
kflemin, line 20: should users be able to enter program compliance metrics that have no start and end dates? I feel like that will cause issues on the insights pages...

Issue #3593, opened 2022-11-08
Help Dan with his Energy Star Portfolio Manager Import error


haneslinger:
From Dan: 

Currently, when I try to use the Import Portfolio Manager data to add data to SEED, I’m getting an Internal Server Error after entering my ESPM credentials. I’ve double checked that the credentials are correct, and this feature did work in the past.

![Image](https://user-images.githubusercontent.com/30241543/200662164-79712540-082a-45c5-be44-e4739daec56a.png)

 
We are using SEED version: 2.15.1.faa47b69b

haneslinger:
(https://github.com/SEED-platform/seed/pull/3592)

haneslinger:
[groups.io](https://lists.buildingenergytools.org/g/SEEDusers/topic/seed_bug_espm_data_pull/94746918?p=,,,20,0,0,0::recentpostdate/sticky,,,20,2,0,94746918,previd%3D9223372036854775807,nextid%3D1607714259070094586&previd=9223372036854775807&nextid=1607714259070094586) thread

RDmitchell:
@haneslinger / @nllong -- it seems like we have resolved all of Dan's issues? Can we close this?

haneslinger:
I believe this was fixed [here](https://github.com/SEED-platform/seed/pull/3632)

RDmitchell:
@nllong -- I guess they deploy their own instance, so because I communicated to him that I fixed this issue, can we close it, until we hear from him that with the new release it still isn't working?

Issue #3322, opened 2022-06-21
Can't log in to Portfolio Manager from SEED


RDmitchell:
**Describe the bug**

When trying to log into Portfolio Manager to automatically import data, the City of Berkeley is getting a 

**Expected Behavior**

Should be able to log in to a user's Portfolio Manager account, select a template and have the data from that report template automatically uploaded into SEED 

**Actual Behavior**

The program is giving a 500 error

See the video in the folder below, from the **beginning to 1:55** in, which shows the exact problem
G:\My Drive\SEED Shared Folder\Software Development\Issues\Issue 3322
SEED_Issue-3322.mp4

**Steps to Reproduce**

See the video referenced above.

**Instance Information**

Instance: production
SHA: 6ffea0837

**Additional Context**

@areagan-cob at the City of Berkeley reported this problem.

nllong:
@aviveiros11  -- can you verify if this is still an issue?

aviveiros11:
Hey @nllong or @RDmitchell could I get access to the SEED shared drive to take a look at that video.  Please use my viveiros.austin@gmail.com email.

RDmitchell:
@aviveiros11 -- I just shared the folder with you that has the video. Let me know if you have any issues accessing it.

aviveiros11:
I was unable to reproduce the behavior with PM however I did suspect that we weren't handling the case of an account without report templates but realized we actually DO handle that case as well [here](https://github.com/SEED-platform/seed/blob/3a18e847299e90f7c1169a248ceb97a86f8b67dc/seed/views/v3/portfolio_manager.py#L287)

![Image](https://user-images.githubusercontent.com/31779424/187259803-d2b9adac-9c0f-4fb9-ac90-0782bb353794.png)

RDmitchell:
seems like we should close this and we can always open another bug if needed.

RDmitchell:
@nllong -- maybe someone should get in touch with Ammon at the city of Berkeley to see if they are still having this problem, rather than closing without checking with them.

aviveiros11:
I'd be happy to reach out if you give me an up-to-date email 😄

RDmitchell:
@aviveiros11  / @areagan-cob (Ammon) -- maybe the two of you can connect and see if Berkeley is still having this problem)

areagan-cob:
[...]

PR #3632, opened 2022-11-29
Fix get template with no children

haneslinger:
non-empty arrays where parsed with `replace('"[{', '[{').replace('}]"', '}]')`, but not the empty ones. Thus, if `childrenRows` was empty, we'd try to iter through an empty array like it was a string, "[]".

Note, replacing `replace('"[{', '[{').replace('}]"', '}]')` with `replace('"[', '[').replace(']"', ']')` doesn't work, as `[` and `]` are uses in the data and it all gets messed up. 

 #3593 [#3322](https://github.com/SEED-platform/seed/issues/3322)


nllong (approved):
nice fix! I'll merge and get this on staging for testing!

Issue #3634, opened 2022-11-30
Note's API return null for other model's view_id


nllong:
**Describe the bug**

When looking at a note via the API the view_id of the other model (property, taxlot) is null. 

**Expected Behavior**

The other "model" view id should not exist in the resulting API

**Actual Behavior**

**Steps to Reproduce**

**Instance Information**

Server Instance: `dev1, dev2, production, staging, etc`
Server SHA: `abcd123`

**Additional Context**

For example -- http://localhost:8000/api/v3/taxlots/1/notes/4/ below shows `property_view_id: null`
```json
{
  "id":4,
  "note_type":"Note",
  "organization_id":3,
  "property_view_id":null,
  "taxlot_view_id":1,
  "user_id":1,
  "name":"Manually Created",
  "text":"this is a new text",
  "log_data":{
  },
  "created":"2022-11-30T08:25:28.918083-08:00",
  "updated":"2022-11-30T08:25:28.918103-08:00",
  }
}
```

RDmitchell:
@nllong / @kflemin -- is this something I can test through the UI? or is it a backend fix that I guess should just be closed if it passes the automated tests?

PR #3635, opened 2022-11-30
Fix note's API result to not include related model's view ID

nllong:
The Note's API returns both the `property_view_id` and `taxlot_view_id` even though only one of the two will be populated. 

#### What's this PR do?
Removed the `null` view id from the return by overloading the `to_representation` method

* create notes on a property and a taxlot
* monkey test the CRUD operations in the UI
* Also updated the API documentation

#3634 

#### Screenshots (if appropriate
Before:
image

After:
image


nllong:
@haneslinger  -- you want to check this PR out real quick? Should be very straightforward...

haneslinger (approved):
Looks good.

Issue #3586, opened 2022-10-21
Apply labels to the currently selected properties only


anchapin:
When in the inventory list, if you have a label selected and click on the "exclude" button, then the list will update appropriately. However, if you select all visible properties and go to add or remove labels, the changes are applied to the properties with the label that is excluded also.

For example, I have 9 properties total and 4 have the "Compliant" label.

![Image](https://user-images.githubusercontent.com/6326294/197222960-870b379e-e4d2-427b-b7c6-024418d8249a.png)

Here's what it looks like when I select the "Compliant" label and click exclude.

![Image](https://user-images.githubusercontent.com/6326294/197223106-d441a850-b332-4346-a56e-0f98173573cd.png)

Then when I select all and try to apply the "Violation" label, it shows that the changes will be applied to 9 properties instead of 5.

![Image](https://user-images.githubusercontent.com/6326294/197223379-8ba732c7-dce9-4a0d-a183-3787feaa1242.png)

Here's the final version with the "Violation" label applied to all 9 properties.

![Image](https://user-images.githubusercontent.com/6326294/197223681-166427b4-120a-4f6e-9b60-b1fb8d9d8360.png)

Note that if a filter group is selected, then you select all properties and apply a label, it works as intended.

anchapin:
Note that it also appears to apply labels to all properties when you select a label and then use the "And" button. So it seems to ignore the filtering of properties by label when applying or removing labels.

RDmitchell:
Instance: staging
SHA: 
Org: LBNL 505

I don't think this is working quite right yet. 

Here are all my records with their labels

RDmitchell:
@haneslinger -- I think a little bit more work needs to be done on this, unless I am misinterpreting what the program should be doing.

haneslinger:
This in not what this ticket is about. "Exclude" excludes only properties with all the labels applied, not just one.

RDmitchell:
This is working properly. 

If I filter by a label, with EXCLUDE chosen, then add another label to those records, only those records get the label applied.

PR #3637, opened 2022-12-01
Apply label to only selected

haneslinger:
[#3586](https://github.com/SEED-platform/seed/issues/3586)


anchapin (approved):
It works as expected now!

Issue #3638, opened 2022-12-01
Add ability for create, update, and delete to meters


nllong:
Phase 1 -- move the `meters` endpoint as a nested route under properties.

RDmitchell:
instance: staging
SHA: bb2fdd425
Org: LBNL Test 200
Cycle: Meter Import Testing

Under meters, the Filter Meters pulldown doesn't display the meters and no data is displayed for the meters that exist
![image](https://user-images.githubusercontent.com/6314950/205785060-2fef2c42-0515-4fea-820c-32dcabc74144.png)

on Production, before this update to delete meters, the Filter Meter pulldown is displayed, and there is data associated with the meters (same org and cycle in both cases).
![image](https://user-images.githubusercontent.com/6314950/205785253-5c10ba6f-3c0f-4d9d-923c-13cdaaea407b.png)

RDmitchell:
I think the records are not displayed because I am clicking on the Filter Meters button and can't select anything, so maybe it defaults to no meters selected. 

If I come to the meter data for a record by clicking on the lightning bolt icon in the Inventory List view, then I see the data
![Image](https://user-images.githubusercontent.com/6314950/205785897-67ff9309-8b37-4b36-82a3-424c2e9d1c70.png)


If I then click twice on the Filter Meters button, the data for the meters disappears.
![Image](https://user-images.githubusercontent.com/6314950/205785958-0f299045-9a48-40d5-833c-fefb14342129.png)

RDmitchell:
@nllong -- I think this needs to go back to "In Progress".

PR #3639, opened 2022-12-01
Add create, update, retrieve, and delete to Meter API

nllong:

Meters are only importable via files (ESPM or Green Button). This prevented the ability to delete or update meters.

#### What's this PR do?
* update version of pre-commit autoflake8
* nest meter API under properties
* Add "manual entry" type of meter source, which is not used yet.
* Create meter serializer to handle CRUD
* Add "alias" to meter fields

#### How should this be manually 

Most of the functionality for this PR is handled through the API. Ideally, using postman (or similar) try:
* Creating new meter
* Deleting meter
* Updating

#3638 
#3640


seed/data_importer/utils.py
```diff
@@ -408,9 +408,12 @@ def usage_point_id(raw_source_id):
     """
     Extracts and returns the usage point ID of a GreenButton full uri ID.
     """
-    id_split = raw_source_id.split('/')
-    usage_point_index = next(i for i, substr in enumerate(id_split) if substr == "UsagePoint") + 1
-    return id_split[usage_point_index]
+    if raw_source_id and '/' in raw_source_id:
```
nllong, line 411: only split if it exists...

seed/tests/test_meter_views.py
```diff
@@ -24,15 +28,120 @@ def setUp(self):
...
+
+        # faker class for properties
+        self.property_view_factory = FakePropertyViewFactory(organization=self.org)
+
+    def test_create_meter(self):
+        property_view = self.property_view_factory.get_property_view()
+        url = reverse('api:v3:property-meters-list', kwargs={'property_pk': property_view.id})
+
+        payload = {
+            'property_id': property_view.property.id,
```
nllong, line 63: I'm pretty sure this is correct, but a second set of eyes will be nice! 

The thought is that we find the "property" using the property_view_id that is in the URL (which is the `reverse` command in line 60, then the meter is attached to the "property" ID which is found through the property_view.

seed/tests/test_property_views.py
```diff
@@ -1522,8 +1522,7 @@ def test_property_meters_endpoint_returns_a_list_of_meters_of_a_view(self):
         }
         gb_gas_meter = Meter.objects.create(**meter_details)
 
-        url = reverse('api:v3:properties-meters', kwargs={'pk': self.property_view_1.id})
-        url += f'?organization_id={self.org.pk}'
+        url = reverse('api:v3:property-meters-list', kwargs={'property_pk': self.property_view_1.id})
```
nllong, line 1525: existing API is now under `property-meters-list` and does not require the passing of the org_id.

seed/serializers/meters.py
```diff
@@ -0,0 +1,35 @@
...
+        model = Meter
+        exclude = ('property', 'scenario',)
+
+    def to_representation(self, obj):
+        result = super().to_representation(obj)
+
+        if obj.source == Meter.GREENBUTTON:
+            result['source_id'] = usage_point_id(obj.source_id)
+
+        result['scenario_name'] = obj.scenario.name if obj.scenario else None
```
haneslinger, line 34: I think you want this inverted, I think it would error rn if `obj.scenario is None`.
```suggestion
        result['scenario_name'] = None if not obj.scenario else obj.scenario.name
```

seed/tests/test_meter_views.py
```diff
@@ -2,17 +2,23 @@
...
+    kbtu_thermal_conversion_factors as conversion_factors
 from seed.landing.models import SEEDUser as User
-from seed.tests.util import DataMappingBaseTestCase
+from seed.test_helpers.fake import FakePropertyViewFactory
+from seed.tests.util import DeleteModelsTestCase
 from seed.utils.organizations import create_organization
 
 
-class TestMeterValidTypesUnits(DataMappingBaseTestCase):
+class TestMeterValidTypesUnits(DeleteModelsTestCase):
```
haneslinger, line 18: Why this change?

seed/tests/test_meter_views.py
haneslinger: I'm a little confused. The request is `POST /api/v3/property_meters_list/<property_view.id>`, then in the body we give it the `property_id`. But like, why would the body need to specify the property, as it is a given because of the `property_view.id` in the url? 


The request being made here looks right, the actual endpoint seems a bit repetitive.

haneslinger (changes requested):

haneslinger:
Just small things.

nllong:
These unit tests are killing me. Something non-deterministic is happening that I'm still trying to debug.

seed/tests/test_meter_views.py
nllong: I updated this so that the meter's viewset `perform_create` method grabs the `property_pk`. The body had the property because that was legacy, it has now been removed.

seed/serializers/meters.py
nllong: hmm, either works. We have lots of tests where `obj.scenario` is None and it works as expected.

seed/tests/test_meter_views.py
nllong: `DataMappingBaseTestCase` has a bunch of helper methods for running the mapping and merging. It inherits from `DeleteModelsTestCase` which simply makes sure that all the models are empty when the test suite runs.

haneslinger (approved):
🎊

PR #3640, opened 2022-12-01
Add delete meters UI

haneslinger:

### DIFF
--- a/seed/static/seed/js/controllers/inventory_detail_meters_controller.js
+++ b/seed/static/seed/js/controllers/inventory_detail_meters_controller.js
@@ -59,7 +59,7 @@ angular.module('BE.seed.controller.inventory_detail_meters', [])
       };
 
       var resetSelections = function () {
-        $scope.meter_selections = _.map(sorted_meters, function (meter) {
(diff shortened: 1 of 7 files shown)


nllong (approved):
Looks great! I say we get this into the release and then handle the "editing" and "creating" next.

image

nllong (changes requested):
Hmm, it looks like the "monthly" and "yearly" aggregations aren't working anymore. Can you take a look at what might be causing it?

haneslinger:
good catch! fixed. I don't know if I like `refresh_meters_and_readings` and `refresh_readings` as two different functions. I will rethink this

nllong:
> good catch! fixed. I don't know if I like `refresh_meters_and_readings` and `refresh_readings` as two different functions. I will rethink this

Great thanks! Yeah, this page could use a little love. Probably good to create patterns that we can use on sensors too.

RDmitchell:
Instance: staging
SHA: bb2fdd425
Org: LBNL Test 200
Cycle: Meter Import Testing

The UI for deleting the meters exists. 
image

And it deletes the meter data, in this case the solar data
image

PR #3646, opened 2022-12-02
BPS Data Final Updates

anchapin:
#### What's this PR do?
Removed the finished files (the spreadsheets with all of the relevant columns of data) since it is unnecessary and copied over the ESPM files into the folder. Made note in readme that meters need to be uploaded separately as xlsx.


nllong (approved):
sweet, thanks!

Issue #3648, opened 2022-12-02
Refactor how meters are selected.


haneslinger:

nllong:
@RDmitchell  -- once deployed to dev1, check the meters page for a property. Mess around with the table, make sure all is good.

RDmitchell:
This seems to be working well, but I think we need a spinner when deleting a meter -- it took some time to do, and in the interim it looked like the delete hadn't worked.

RDmitchell:
@haneslinger -- I'll close this issue and make a new one (lower priority) for adding a spinner when deleting a meter

PR #3649, opened 2022-12-02
Refactor meter selection

haneslinger:
[#3648](https://github.com/SEED-platform/seed/issues/3648)


nllong (changes requested):
This is great--thanks for getting this in so quick!

When you change "Filter By" to "scenarios", the old "Filter Scenarios" pops us. I think we just want to remove the entire "Filter By" dropdown because the table can be used to filter now.

Also, if you delete a meter using the action button it doesn't remove the row until you deselect/select another meter.


image

haneslinger:
> When you change "Filter By" to "scenarios", the old "Filter Scenarios" pops us. I think we just want to remove the entire "Filter By" dropdown because the table can be used to filter now.

I'm unfamiliar with senerios, so I want to make sure I understand: You want all of the Scenarios related stuff off of this page?

> Also, if you delete a meter using the action button it doesn't remove the row until you deselect/select another meter.

Hm, it wasn't doing that or my end, I'll investigate.

nllong (approved):
This looks good!

The "deleting not removing the meter row" is still happening, but it is random. I'm not sure if I can pinpoint when it doesn't work. I suggest we merge this and see what happens on dev1.

🎉

Issue #3657, opened 2022-12-06
Speed up the Program Overview page for large Orgs


anchapin:
With the LBNL Test 200 organization on seeddemostaging, the Program Overview page takes at least 2 minutes to load. The issue is with the evaluate endpoint of the compliance metrics api call. Refer to the screenshot below.

![Image](https://user-images.githubusercontent.com/6326294/205958662-3d8e77d5-184b-457b-af49-071d1f99bd0b.png)

haneslinger:
[here's](https://github.com/SEED-platform/seed/issues/3673) the next step.

RDmitchell:
Instance: dev1
SHA: 4b8753257
Org: 387

I am currently getting a bunch of 500 errors trying to view the Property Insights page. 

There are 150 records in this cycle (2021)

![Image](https://user-images.githubusercontent.com/6314950/207992609-a42b04e8-924c-481c-a744-64bbd03e31c6.png)

RDmitchell:
Instance: dev1
SHA: 294626cbd

The program pages are much more responsive now. 

I will close and we see other performance issues with more data, etc, we can open a new issue.

PR #3664, opened 2022-12-06
Return only need columns in compliance evaluate

haneslinger:
#### What's this PR do?
idk if this will speed up the request, but it definitely shortens the response. 

I got like 200 properties. before and after this commit, the eval request takes ~3 secs, but the response size goes from 750KB to 55KB. the diff like scale with the number of properties. 

It's a start. 

I'll keep working, only push if we can't wait for more optimization.

### DIFF
--- a/seed/models/compliance_metrics.py
+++ b/seed/models/compliance_metrics.py
@@ -72,7 +72,18 @@ class ComplianceMetric(models.Model):
         # get properties (no filter)
         # property_response = properties_across_cycles(self.organization_id, -1, cycle_ids)
         # get properties (applies filter group)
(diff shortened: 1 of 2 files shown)


kflemin (approved):
looks good! thanks for fixing this!

RDmitchell:
@haneslinger -- what issue does this apply to?

haneslinger:
https://github.com/SEED-platform/seed/issues/3657

haneslinger:
Okay, So I think the other thing slowing this down is `properties_across_cycles_with_filters`. In this function, we query for the property views one for each cycle. It would be quicker to query once for all the views with a cycle within a list. 

working on that fix.

Issue #3673, opened 2022-12-08
In properties_across_cycles_with_filters, only query for the views once.


haneslinger:

haneslinger:
https://github.com/SEED-platform/seed/pull/3664#issuecomment-1340188607

RDmitchell:
@isalanglois  -- Not something I can test

isalanglois:
@haneslinger  Can you identify the process for testing for this ticket

haneslinger:
I did this [here](https://github.com/SEED-platform/seed/pull/3674) we are good to close.

PR #3674, opened 2022-12-08
Refactor properties_across_cycles_with_filters so it only makes one P…

haneslinger:
…ropertyViews query

Refactor `properties_across_cycles_with_filters` so it only makes one PropertyViews query. On a db with 3 cycles and 15ish properties in each cycle, this about halves the response time of `/api/v3/compliance_metrics/1/evaluate/` (1200 ms to 600 ms, roughly, still high). I believe the improvement in performance will scale with the number of cycles.  

https://github.com/SEED-platform/seed/issues/3657
https://github.com/SEED-platform/seed/issues/3673


anchapin:
I tested this new change and it works as expected. Once the spinner is added this can be merged.

haneslinger:
@anchapin I fixed the spinner!

Issue #3625, opened 2022-11-28
Property Insights Graph Tooltips


kflemin:
clean up tooltips on compliance graphs - they should match the style shown in the data table below the graph.
 - don't need "ID: ..." in the tooltip
 - don't need the label "name:" in the tooltip
 - use the user-selected identifier specified in the program metric setup page
- Ensure the tooltip is not too "twitchy" when hovering/clicking on it. May need to implement a mouse click to toggle show / hide  instead of hover

RDmitchell:
Instance: dev1
SHA: 05efb37bf

This looks good.

PR #3679, opened 2022-12-09
Nav property on click in property insight page

haneslinger:
If the points are close to each other and you click two, idk how it picks which you go to 🤷 lmk if you know a fix.
https://github.com/SEED-platform/seed/issues/3625


anchapin (changes requested):
I'm not sure why (I tried tracking it down to no avail), but when I use this branch I don't see any data points on the graph (see first image below) compared to the develop branch (second image below).

No Data Showing Up
image

Develop Branch
image

haneslinger:
@kflemin would you please check if you experience the same error as chapin? We fixed it with a rebase, but there's no go reason that should have worked.

kflemin:
@haneslinger @anchapin, I just pulled this branch, created a metric (I had none), and now when I navigate to the program insights page (or property insights), the graph doesn't load. I get this error: 
File "/Users/kflemin/repos/seed/seed/models/compliance_metrics.py", line 75, in evaluate
    display_field_id = Column.objects.get(table_name="PropertyState", column_name=self.organization.property_display_field).id
  File "/Users/kflemin/.pyenv/versions/seed-3.9.1/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/kflemin/.pyenv/versions/seed-3.9.1/lib/python3.9/site-packages/django/db/models/query.py", line 439, in get
    raise self.model.MultipleObjectsReturned(
seed.models.columns.Column.MultipleObjectsReturned: get() returned more than one Column -- it returned 2!

nllong:
@kflemin  - the multiple objects returned is due to not passing the org id to the column.objects.get call. @anchapin and I found this and fixed it on his branch--not sure if that a PR has been created for it yet though. The TLDR is that the column call will return the total of orgs (so you must have had 2).

PR #3680, opened 2022-12-09
Support import of a csv file that has lots of spaces

nllong:

When importing `.csv` files that have many spaces, the Python Sniffer can interpret the delimiter as a space, which causes the headers to be space-delimited instead of comma-delimited. This happen with only a large amount of rows like the ones below

```csv
id, address, a, b, c, square footage, d, total area
1,15013 Denver West Parkway,,1,," 10,000.00 ",," 10,500.00 "
2,15014 Denver West Parkway,,1,," 15,000.00 ",," 20,500.00 "
3,15015 Denver West Parkway,,1,," 20,000.00 ",," 50,500.00 "
4,15016 Denver West Parkway,,1,," 30,000.00 ",," 100,500.00 "
...
```

#### What's this PR do?
* Since it MCM parser is expecting CSV files, then force the format to be CSV if the Sniffer detects something otherwise.

* Create a long file (1000 rows) with the example above and verify it loads

#3675

### DIFF
--- a/seed/lib/mcm/reader.py
+++ b/seed/lib/mcm/reader.py
@@ -11,6 +11,7 @@ elsewhere.
 
 """
 import json
+import logging
 import mmap
 import operator
 import re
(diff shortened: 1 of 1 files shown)


kflemin (approved):
nice catch!!

Issue #3645, opened 2022-12-02
Source EUI import conversion issue


RDmitchell:
Deployment: BBBC (Building Benchmark BC)

I've noticed that there are a couple of bugs in SEED related to the Source EUI:
1. The Source EUI isn't converting from GJ/m2 to kBtu/ft2
2. When we upload a blank or #VALUE!, the "generic" value displayed in SEED is 15.00

![image](https://user-images.githubusercontent.com/6314950/205360909-aa883623-88a5-407f-8767-5c85242f222e.png)
![image](https://user-images.githubusercontent.com/6314950/205360926-c81bebae-6895-4a5d-a432-393a355e8408.png)

If you need more details or additional information, please let me know.

Carol Gomez, QA Analyst/Product Support
OPEN Technologies
opentech.eco
carol@opentech.eco | 778-834-4078

nllong:
I am looking into this. I am unable to recreate the `Source EUI` conversion issue. Are you sure that you are mapping to the "canonical" source EUI column? In the column mapping page it is the one with the green background (see image below). The green background column is the one that is unit aware and able to be converted. Once you select it, then the units will pop up for you to convert.

<img width="503" alt="image" src="https://user-images.githubusercontent.com/1907354/206723580-48126220-96cb-407b-9d67-44e6f39a27c9.png">

I was able to recreate the `#VALUE` issue and looking in fixing before out patch release next week.

@RDmitchell  -- can you communicate this to Open?

nllong:
Nice find on the `#VALUE`! I created an excel file with various Excel "errors" and tested my patch. I will do some creanup and get this in the next release (next week!)

Here's the test file I used for future reference.

Thanks again!

<img width="1115" alt="image" src="https://user-images.githubusercontent.com/1907354/206728531-06b53521-b0a3-4055-9f18-7863b45871f0.png">

RDmitchell:
I posted info from this issue (as well as a link to the issue) in response to Open's groups.io / users post.

RDmitchell:
@nllong -- I assume you will put this in the "To Deploy" column and then when that is done, into the Test column. Thx

RDmitchell:
From Dan

RDmitchell:
Instance: dev1
SHA: 

Here is the test file with the Excel issues
![image](https://user-images.githubusercontent.com/6314950/208182782-5fa2035c-eeda-4854-964e-8133330a9555.png)

Here is the mapping screen with this fix, which looks good.
[...]

PR #3685, opened 2022-12-09
Fix #VALUE! errors when importing xlsx files

nllong:

If the excel file being imported has cell errors, then the program should make that cell None. However, the cells were being converted into the integer of the ERROR type from Excel.

#### What's this PR do?

* Catch excel errors; Map errors to None and Blanks/Empties to ''

* Test file can be found on the referenced ticket

#3645

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
   # As a result, the DRYest method for now is to use a testing matrix and
   # use conditional steps for actually running the tests.
   unittests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


nllong:
> Makes sense. Do we want a test for this?

I added tests to the `example-data-properties-postal.xlsx` file. Also rebased to develop... wanna check again?

haneslinger:
@nllong What happen here? why are there no commits? Were they in a different Pr?

nllong:
So somehow, this has already been merged to develop see (https://github.com/SEED-platform/seed/blob/develop/seed/lib/mcm/reader.py#L378-L379). I imagine that it was probably pulled into another branch merged with another fix. Everything is there in develop. so I'll close.

Issue #3683, opened 2022-12-09
Custom Reports Page: Make sure the left and right axis have the same range


isalanglois:

RDmitchell:
Instance: dev1
SHA: 
Org: BPS Test

This appears to be implemented properly

![Image](https://user-images.githubusercontent.com/6314950/211693839-b92b9340-e0a3-486c-b1d4-cd7899d2cf52.png)

PR #3695, opened 2022-12-13
Make sure the left and right axis have the same range

haneslinger:
Not perfect but pretty good.

[#3683](https://github.com/SEED-platform/seed/issues/3683)
image


anchapin (approved):
It looks much better, thanks!

Issue #3676, opened 2022-12-08
Admin: Can't delete orgs


RDmitchell:
Instance: staging
SHA: 

As an Admin, I wanted to delete an organization. 

The program looked like it was deleting it
![Image](https://user-images.githubusercontent.com/6314950/206576512-eff4faf7-4d06-4fcf-82c8-b3fce6d34d07.png)

But it wasn't actually deleted
![Image](https://user-images.githubusercontent.com/6314950/206576540-79537754-4247-4bbf-96cc-7d126b3f439c.png)

RDmitchell:
Instance: Dev1
SHA: 05efb37bf

Added derived column to data in org, then deleted it successfully.

PR #3700, opened 2022-12-14
Delete Derived Columns Before Org

axelstudios:
Organizations couldn't be deleted if they used derived columns

#### What's this PR do?
Deletes derived columns prior to the cascaded-delete of the organization

(Optionally test this workflow beforehand on the `develop` branch to demonstrate the conditions for failure)
1. Create a new organization
2. Create a derived column within that organization
3. Delete the organization from the admin page

#3676


haneslinger (commented):
This make sense, but wouldn't make more sense to add `on_delete=models.CASCADE` to the foreign key keeping the derived columns from deleting?

axelstudios:
> This make sense, but wouldn't make more sense to add `on_delete=models.CASCADE` to the foreign key keeping the derived columns from deleting?

@haneslinger I wanted to preserve the original intent, which was to prevent users from deleting a column that was being used in a derived column, unless they first manually remove the derived column.

Having said that, it looks like when derived columns were added the handling for this case was skipped, so it still "allows" you to delete a source column of a derived column (it removes the key and contents from all extra_data fields, but the column itself remains)

haneslinger:
> @haneslinger I wanted to preserve the original intent, which was to prevent users from deleting a column that was being used in a derived column, unless they first manually remove the derived column.

that makes sense. 

> Having said that, it looks like when derived columns were added the handling for this case was skipped, so it still "allows" you to delete a source column of a derived column (it removes the key and contents from all extra_data fields, but the column itself remains)

Do we want to make a ticket for that?

axelstudios:
Already created #3701  ;)

PR #3702, opened 2022-12-14
Fix switching filter group

haneslinger:
This fixes what filtering by other inventory type broke


anchapin (approved):
This appears to be working as intended now. Thanks

Issue #3708, opened 2022-12-15
Update import from portfolio manager to reflect new api


haneslinger:
https://portfoliomanager.energystar.gov/pm/reports/reportData has changed, breaking our tests. It now has a JSON response, which will clean up our code but it is now at https://portfoliomanager.energystar.gov/pm/reports/reportData/MY_REPORTS_AND_TEMPLATES

update our api to reflect these changes.

RDmitchell:
How do I test this issue? Do I try logging into my ESPM account from the SEED import modal?

RDmitchell:
@nllong -- is logging in to my ESPM account, selecting a Custom Report Template and getting data a good enough test? 

If it is, I have been able to do that, and I can close this issue.

isalanglois:
@nllong or @haneslinger please confirm the best process for testing this ticket and then close.

haneslinger:
I think we are good to close.

PR #3709, opened 2022-12-15
Fix SEED to work with updated ESPM API, remove SEED ESPM API v2.1

nllong:
In the last couple of days, ESPM tests have been failing. Looking at the network calls in ESPM, there appear to be some changes to the request URLs and return types.

#### What's this PR do?
* The list of reports is now returned as JSON instead of string-encoded JSON. This is a great improvement on the ESPM side, but it only pertains to this endpoint. 
* The download report no longer uses the name of the report, just the template ID. Again, another great ESPM improvement.

* Unit tests
* Connect to the SEED City Test account and verify that content downloads and loads

#3708 

image


seed/views/v3/portfolio_manager.py
```diff
@@ -278,21 +282,24 @@ def get_list_of_report_templates(self):
...
 
-        template_object = self.parse_template_response(response.text)
+        # This endpoint now is JSON and not encoded JSON string. So just return the JSON.
+        template_object = response.json()
 
         # We need to parse the list of report templates
-        if 'customReportsData' not in template_object:
-            raise PMExcept('Could not find customReportsData key in template response; aborting.')
-        templates = template_object['customReportsData']
+        if 'reportTabData' not in template_object:
```
nllong, line 300: the URL changed and the root key of the dict.

seed/views/v3/portfolio_manager.py
```diff
@@ -405,21 +412,23 @@ def generate_and_download_template_report(self, matched_template):
...
+                json_authenticated_headers['Accept'] = 'application/json'
+                json_authenticated_headers['Content-Type'] = 'application/json'
+                response = requests.get(self.REPORT_URL, headers=json_authenticated_headers)
             except requests.exceptions.SSLError:
                 raise PMExcept('SSL Error in Portfolio Manager Query; check VPN/Network/Proxy.')
             if not response.status_code == status.HTTP_200_OK:
                 raise PMExcept('Unsuccessful response from report template rows query; aborting.')
 
-            template_objects = self.parse_template_response(response.text)['customReportsData']
+            template_objects = response.json()['reportTabData']
```
nllong, line 431: we are being presumptuous that this key exists, but it has worked in the past, so leaving.

seed/views/v3/portfolio_manager.py
```diff
@@ -278,21 +282,24 @@ def get_list_of_report_templates(self):
...
+            json_authenticated_headers['Accept'] = 'application/json'
+            json_authenticated_headers['Content-Type'] = 'application/json'
+            response = requests.get(self.REPORT_URL, headers=json_authenticated_headers)
 
         except requests.exceptions.SSLError:
             raise PMExcept('SSL Error in Portfolio Manager Query; check VPN/Network/Proxy.')
         if not response.status_code == status.HTTP_200_OK:
             raise PMExcept('Unsuccessful response from report template rows query; aborting.')
 
-        template_object = self.parse_template_response(response.text)
```
haneslinger, line 290: I think we can remove `parse_template_response` now

haneslinger (approved):
Makes sense 👍

seed/views/v3/portfolio_manager.py
nllong: great catch. removed.

nllong:
@haneslinger  -- can you review once more? I removed the `parse_template_response` and also the portfolio manager end point for API v2.1 because it was broken now anyway...

seed/api/v2_1/urls.py
```diff
@@ -19,7 +18,6 @@
 router.register(r'taxlots', TaxLotViewSet, basename="taxlots")
 router.register(r'scenarios', ScenarioViewSet, basename="scenarios")
 router.register(r'tax_lot_properties', TaxLotPropertyViewSet, basename="tax_lot_properties")
-router.register(r'portfolio_manager', PortfolioManagerViewSet, basename="portfolio_manager")
```
haneslinger, line 22: is it being used by the ui anywhere? i dont thing so, im just checking

haneslinger (approved):

seed/api/v2_1/urls.py
nllong: no, it isn't. We have a settings var to disable v2 API endpoints... we should eventually just get rid of all v2* API endpoints. Everyone should be using v3.

PR #3710, opened 2022-12-15
Enter edit mode state on any change in custom report 

### LINKED ISSUE #3492: DataView update edit button value to save after toggling
After toggling any dataview attributes (filter group, cycle, column) the 'Edit' button should transition to 'Save' - (P3)

kflemin:
Fixes #3492

It's a little bit wonky when  either filter group or cycle is clicked (you kind of have to click twice to make the change since the first click just reveals the checkboxes)


haneslinger:
How should this be tested? I can't find the change in behavior.

haneslinger (approved):
It works useless you try to modify the dataview before the graph loads, but I don't think thats this PRs fault.

PR #3711, opened 2022-12-15
Update the BPS Sample Data AT Files

### LINKED ISSUE #3478: Pull in an AT building with measures

anchapin:
The property name in the AT files hadn't been updated yet. Additionally, there was meter data associated with one of the files that was still from the source AT property. 

#### What's this PR do?
I updated the property names and removed the meter data.

Additionally, I added a disclaimer to the readme indicating that only the identifiable data in the AT files were revised to match the sample BPS data.

Try importing an AT file and make sure that the property name matches the BPS sample data.

Fixes #3478


RDmitchell:
@anchapin -- where is the updated XML file?

RDmitchell:
@anchapin -- never mind, I found them.

RDmitchell:
@anchapin -- can I test this by importing one of your new files into dev1? I tried doing that and am getting this error, but not sure this should be working on dev1

image

Here is the file I was importing, that I copied from one of your examples -- not sure it is actually correct, as I did copy and paste from the repo to create it. 

https://drive.google.com/file/d/1yzNvN36D2CScuAU0cWBavXr-3jqy0AWM/view?usp=share_link

nllong (approved):
looks good. I fixed the resource use tag that was missing. I like the address in the ID.

PR #3714, opened 2022-12-16
Fix build and publish scripts by pinning version of Shapely

nllong:
`develop` publish builds are failing.

The shapely dependency comes from pnnl-buildingid library which does not have any version constraints on shapely, thus it grabbed the latest shapely which was released on 12/12/2022.
```
pnnl-buildingid==2.0.0
  - click [required: Any, installed: 8.1.3]
  - click-log [required: Any, installed: 0.4.0]
    - click [required: Any, installed: 8.1.3]
  - shapely [required: Any, installed: 1.8.5.post1]
```

#### What's this PR do?
* force the use of 20.04 for most of the builds. This is not likely the issue, but now we will be explicit on what version of Ubuntu we are using.
* Force Shapely 1.8.5.post1 as 2.x version does not build the geos library correctly.

n/a

n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
   # As a result, the DRYest method for now is to use a testing matrix and
   # use conditional steps for actually running the tests.
   unittests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


haneslinger (approved):
Nice

nllong:
added more notes in the PR text. Always constrain your dependencies ✌️

PR #3715, opened 2022-12-16
Fix issue with evaluating compliance metric

### LINKED ISSUE #3697: Fix issue when both energy and emission metrics are included with Program
[traceback.txt](https://github.com/SEED-platform/seed/files/10228172/traceback.txt)

The issue appears to be with the property_response line of models/compliance_metrics.py. I was able to quickly get the Program Overview page to work by adding an if statement to check whether the metric has an actual_emission_column and adding ".id" to the end of the actual and target emission columns (which would only be included if emission column exists), but I couldn't get the Property Insights page to work.

Therefore, I'm creating this ticket hoping someone else can help me with this.

I've attached the traceback file from Chrome dev tools. It seems to be around line 1778 that it gets to the affected code sections.

anchapin:
There was an issue with how the column_ids were created for the property_response in the compliance_metrics model.

#### What's this PR do?
It adds checks for whether actual energy, target energy, actual emission, and target emission values are present before adding them to column_ids.

Create a Program with only energy metric, only emission metric, and both metrics. Then make sure each of the three Programs will load on Program Overview and Property Insights pages.

Fixes #3697

### DIFF
--- a/seed/models/compliance_metrics.py
+++ b/seed/models/compliance_metrics.py
@@ -75,12 +75,19 @@ class ComplianceMetric(models.Model):
         display_field_id = Column.objects.get(table_name="PropertyState", column_name=self.organization.property_display_field, organization=self.organization).id
         # array of columns to return
         column_ids = [
-            display_field_id,
-            self.actual_energy_column.id,
-            self.target_energy_column.id,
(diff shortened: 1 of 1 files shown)


kflemin (approved):
👍

Issue #3627, opened 2022-11-28
Toggle show/hide in Property Insights Page Legend


kflemin:
Add ability to show/hide whiskers, compliant datapoints, and non-compliant datapoints in the legend on the Program Insights page to help with the "busy-ness" of the graph with lots of properties

RDmitchell:
Instance: dev1
Release: 2.17.1
SHA: 7a529191a

@kflemin -- how do I test this? I don't see a way to show/hide anything in Program Insights

anchapin:
You just click on the legend. For example, if you want to hide the distance to target then you just click on it in the legend. I created an issue (#3764) related to improving the UX of this new feature. Feel free to add to that as you see fit.

RDmitchell:
Looking at the PR for this, I see now that users can click the legend options "on and off" to show the points or not. 

It's pretty subtle, but it does work, so I am going to close this issue. But I might make another one per @anchapin in the PR for this that we need to let the user know that this feature exists. 

I think this is part of the larger INSIGHTS UI rework. The graph is very "tall" so as @anchapin says, the legend is off the bottom of the screen, the "Need to configure your program" takes up way to much vertical space, etc. 

I will add these comments to an INSIGHTS UI rework issue.

RDmitchell:
See #3804 for some redesign ideas

RDmitchell:
@anchapin -- right I figured that out. But it's really a bit subtle for users.

PR #3726, opened 2022-12-20
Allow users to hide property insight dataset

haneslinger:
[Toggle show/hide in Property Insights Page Legend#3627](https://github.com/SEED-platform/seed/issues/3627)


seed/static/seed/partials/insights_property.html
```diff
@@ -67,21 +67,25 @@ <h2 style="font-size:18px;"> {$:: 'Property Insights' | translate $}</h2>
     			</div>
     			<div class="chart-legend">
     				<div class="title">Chart Legend</div>
-    				<div class="legend-row row">
+    				<div class="legend-row row" ng-click="toggle_dataset_visibility(0)">
     					<div class="col-md-3 legend-icon-col"><img class="legend-icon" src="{$ static_url $}seed/images/teal_circle.svg" alt="blue circle marker"></img></div>
-    					<div class="col-md-9 text">Compliant</div>
+    					<div class="col-md-9 text" ng-hide="insightsChart.legend.legendItems[0].hidden">Compliant</div>
+    					<div class="col-md-9 text strikethrough" ng-hide="!insightsChart.legend.legendItems[0].hidden">Compliant</div>
```
haneslinger, line 73: There's gotta be a better way to do this.

anchapin (commented):
This works as intended. However, I'm wondering if we need to tweak the UI to let the user know that this is an option. Let me know if you think these changes should be a separate ticket and I will approve this one and make a new ticket.

Here are some ideas:

- The legend is half off the screen on my computer. Try moving the legend under the chart and orient it horizontally.
- Add text header to legend section indicating that the user can toggle the datasets or maybe add checkboxes for each dataset that default to being checked and then the user would know that they can uncheck the box to hide that dataset.

RDmitchell:
@haneslinger / @anchapin -- wow this is subtle. I think we do need to have some way for users to know how to do this without reading documentation about it.

Issue #3696, opened 2022-12-13
Revise Programs so that filter groups with filters on extra data columns can be used.


anchapin:
When I use a filter group in a Program that filters on an extra data column, I get a 500 error when trying to load Program Overview and Property Insights pages.

isalanglois:
@anchapin Is this still an issue?

haneslinger:
@anchapin , I can't recreate this. Let's do it together?

RDmitchell:
@anchapin / @haneslinger -- what is the status of this issue? Were you able to reproduce it? 

Do I need to test it?

haneslinger:
It is fixed. 

If you go to the insights/program page, and use a filter that filters on extra columns, it used to break. It doesn't anymore.

RDmitchell:
instance: dev1
SHA: 26aac34d8

This appears to be working. I can use a filter group that is based on filtering extra data fields when defining a filter group in custom reports. And the program generates the graph properly.

PR #3727, opened 2022-12-28
Fix filter group with extra column work on revised programs

haneslinger:
[#3696](https://github.com/SEED-platform/seed/issues/3696)


anchapin (approved):
This works! Thanks

Issue #3439, opened 2022-08-05
Migration 0170 can create duplicate columns


ebeers-png:
**Describe the bug**

The migration 0170_column_derived_column creates new Column objects from existing derived columns, but doesn't check whether there are already columns with the same column_name. This can lead to duplicate columns being created, causing the Column model's retrieve_all to raise an exception and preventing the inventory list and column settings pages from loading.

**Steps to Reproduce**

1. Migrate database to before 0170_column_derived_column (preferably without derived columns existing already)
2. Create a derived column that shares a name with an existing column
3. Run the migration
4. Try accessing the inventory list for that column's organization

**Additional Context**

The fix might be to name the new column differently, or to allow duplicate column_names if one is linked to a derived column?

It'd also be good to explicitly set is_extra_data while creating the new columns (for clarity; they default to is_extra_data=False, not sure if that's intentional), and perhaps to skip creating a new column in the if the derived col is already linked to one.

perryr16:
Thanks for documenting this issue. Below are some notes I had regarding this.

- During validation, prevent Derived Columns from being named after an existing column. 
- Create a migration that finds any derived columns that share a name with an existing column and update the derived column to have a unique name by adding a `_derived_column` suffix. If that Derived Column already has an associated Column, its name will also need to be updated.

nllong:
Robin -- there isn't a need to test this on our servers.

PR #3728, opened 2022-12-28
Update derived column migration to prevent conflicting column names and prevent duplicate column names

nllong:
* See the original issue from @ebeers-png #3439 
* This update makes us that *maybe* forcing column names to be unique via a database constraint would be a good idea. Uniqueness constraint would be a combination of `org`, `table_name`, and `column_name`

#### What's this PR do?
* updates the migration to check if the creation of the column from the derived column will result in a name conflict
* adds in the `is_extra_data` to the column creations per the original issue's suggestion

Testing is a bit complicated since most databases have probably already run this migration. I've documented my testing in case you don't want to test this workflow.

* Checkout SEED to a version before the `0170` migration existed (I used sha `d43fdfc2e`)
* In a clean database since you went back to a pretty old commit, import some data
* Create a derived column with the same name as an existing column

image

* verify that you can view the two same-named columns in the property list
image

* check out this branch and run the migrations
* go to the column list profile and add in the renamed derived columns (has "derived 1" appended).
* Verify the column is correctly displayed in the inventory list

image

* Also verify that you can no longer create a derived column with an existing column name (e.g., lot_number). 

image

#3439

### DIFF
--- a/seed/tests/test_derived_columns.py
+++ b/seed/tests/test_derived_columns.py
@@ -319,7 +319,7 @@ class TestDerivedColumns(TestCase):
 
         self.property_state_factory = FakePropertyStateFactory(organization=self.org)
 
-    def _derived_column_for_property_factory(self, expression, column_parameters, create_property_state=True):
+    def _derived_column_for_property_factory(self, expression, column_parameters, name=None, create_property_state=True):
         """Factory to create DerivedColumn, DerivedColumnParameters, and a PropertyState
         which can be used to evaluate the DerivedColumn expression.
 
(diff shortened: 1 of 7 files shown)


nllong:
@ebeers-png  -- do you want to check this PR out too?

seed/tests/test_derived_columns.py
```diff
@@ -639,34 +639,49 @@ def test_derived_column_evaluation_with_derived_column_as_source_column(self):
         # -- Setup
         # expression which sums all the parameters
         expression = '$a + 2'
+        derived_column_name = 'dc1'
```
haneslinger, line 642: does this break if you try to get two derived columns?

haneslinger (approved):

ebeers-png:
> @ebeers-png -- do you want to check this PR out too?

It looks great, thank you for all the thorough testing!

nllong:
> > @ebeers-png -- do you want to check this PR out too?
> 
> It looks great, thank you for all the thorough testing!

Thanks for confirming @ebeers-png

Issue #1758, opened 2018-12-21
Allow users to add fields "on the fly"


RDmitchell:
### Expected Behavior
Allow users to add fields (presumably to the JSON blob field) "on the fly" in the program as needed, without importing a file to create those fields.
### Actual Behavior 
n/a
### Steps to Reproduce
n/a
### Instance Information
n/a

RDmitchell:
@nllong / @macintoshpie -- having this functionality came up in a call with Washington DC -- they would really like this ability.

macintoshpie:
Interesting, do you know their use-case? Are there just one or two properties that are missing a column for some reason?

RDmitchell:
@nllong -- do you remember if Andrew Held mentioned what they need it for? 

I can email him to get more specifics if that is useful.

nllong:
They want to add a column such as "User Verified" and have the user edit it on the Inventory Detail page and update the field from a 0 to a 1. I imagine they would then use a DQ rule to add a label to the properties that haven't been verified.

macintoshpie:
Not sure when we'll pick this up, but just to capture some thoughts
- you should only be able to create new columns from the Organizations Column Settings page
- it should be an extra_data column

RDmitchell:
@macintoshpie -- yes, those are what we were thinking

RDmitchell:
I think this issue is still relevant and should be kept open.

isalanglois:
@RDmitchell  I'll move it to the new board

RDmitchell:
see #3780 
Column Settings: getting 520 error when making a new column but column is created

RDmitchell:
@haneslinger / @isalanglois -- see the bug above that I found when testing last week.

RDmitchell:
@haneslinger / @axelstudios -- is this bug still a problem? The related issue is in test and maybe it shouldn't be?

RDmitchell:
See the comments I just added to #3780

RDmitchell:
Depending on the answer to #3780, I will close both of these issues.

RDmitchell:
I am repeating the comment I made in #3780 

Instance: LBNL staging
SHA: 992862f09
Release: 2.17.1
Org: LBNL Test 200

When I added the field in Column Settings, this error was displayed in Developer tools.
![image](https://user-images.githubusercontent.com/6314950/215232872-1041918f-ce1d-4ddd-af76-039e7eb5d170.png)

[...]

PR #3730, opened 2022-12-30
Allow users to create columns

haneslinger:
https://github.com/SEED-platform/seed/issues/1758

not quite sure why it looks ugly.
image

### DIFF
--- a/seed/views/v3/columns.py
+++ b/seed/views/v3/columns.py
@@ -6,6 +6,8 @@
 """
 import json
 
+from django.core.exceptions import ValidationError
+from django.db.models import F, Func, JSONField, Value
 from django.http import JsonResponse
 from django.utils.decorators import method_decorator
 from drf_yasg.utils import swagger_auto_schema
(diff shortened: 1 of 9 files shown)


seed/views/v3/columns.py
```diff
@@ -111,6 +113,46 @@ def list(self, request):
...
+                'message': str(e)
+            }, status=status.HTTP_400_BAD_REQUEST)
+
+        state_model = PropertyState if table_name == 'PropertyState' else TaxLotState
+        states = state_model.objects.filter(organization_id=organization_id)
+        states.update(
+            extra_data=Func(
+                F("extra_data"),
+                Value([self.request.data['column_name']]),
+                Value(1, JSONField()),
```
anchapin, line 146: I noticed that when I create a new column it starts out with a value of 1 for every property across cycles. It seems like that is coming from this line. I would expect the new column to be created without any data in it otherwise you have to go back and edit all of the properties to remove the value of 1.

Additionally, it might be useful to give the user the ability to auto-populate the new column with a certain value specified by the user, but that seems like it would be a new ticket if that's even something that users want. @RDmitchell  can you weigh in on this?

anchapin (changes requested):
I attempted to make the modal less ugly and added a comment that should be addressed but otherwise this looks good.

Issue #3438, opened 2022-08-04
Save Derived Column values to PropertyState


perryr16:

isalanglois:
This will be addressed through #3430

haneslinger:
https://github.com/SEED-platform/seed/issues/3430 is closed but the issue remains. reopening this one

haneslinger:
> ### A potential solution: 
> - save the derived column values to a property state inside a new attribute dictionary called `derived_data`. This would be similar to 'extra_data' on the property_state 
> - after a new derived column is saved, run a method to add derived_data to every property_state attached to a property_view
> - after a property is updated, or new data is uploaded run a method to update or refresh derived data 
> - add a dropdown action on the inventory list for "refresh derived data"

I really like this idea. The fact it mirrors extra data so closely really does us a favor. I'm going to start documenting where logic needs to be added. To be updated.


Add column to `derived_data`:
Update column in `derived_data`:
Delete column in `derived_data`:

Getting properties `derived_data`:
Sorting on `derived_data`:

kflemin:
Some notes on use case:
- new data imported daily
- dozens of derived columns
- Derived columns that reference other derived columns will NOT be allowed
- Investigate when to perform the recalculation (at import time / data change time, or manually in the background)
- If recalculation is not performed in real time, how do we communicate that to the user in the UI? (what visual indicator is there to signal that the derived columns are old values?)
- There could be a 'recalculation' endpoint that could be hit programmatically outside of SEED after imports
- 
@haneslinger will do some benchmarking on a large db to see how 'expensive' caching the fields and recalculating them will be.

haneslinger:
right-o, ive done some benchmarking

- clean org
- 3 derived columns
- uploading portfolio-manager-sample.csv (500+ properties)

| |develop |cached derived columns|
| --- | ----------- | ------ |
|upload file| 3s| 103s|
|map|8s|~2min|
|finish mapping| 53s| ~2min|
| load inventory | ~2s | ~2s|

kflemin:
let's await the import performance updates and revisit this issue!

BenchmarkDC:
Hi Katherine,

We were talking about this on a Clearly BEAM call and I was reviewing the development history. I wanted to clarify - is the plan to fully disallow derived columns referencing other derived columns? Or just not allow caching for derived columns that reference other derived columns?

Andrew, DC

kflemin:
[...]

PR #3732, opened 2022-12-31
Store derived column values

haneslinger:
[Save Derived Column values to PropertyState#3438](https://github.com/SEED-platform/seed/issues/3438)

This a living draft. 

All this PR does is create the field "derived data" in the inventory states and keep it up to date.

The derived_data of states in a given org must update when:
- [x] The state is created.
- [x] A new derived column is created.
- [x]  That derived column is updated, either but changing the expressions or the source columns. 
- [x] The state data changes.

These updates are easy to achieve with signals, but their run times are hard to pin down.


haneslinger:
I compare this branches speed to develop by uploading `covered-building-samples.csv` 500+ properties. both took around 30 seconds. 👍 no noticeable slow down. I have a hard time thinking of a user action that would potentially see a worse slow down.

kflemin:
closing.  we implemented this functionality in #4763

PR #3733, opened 2023-01-03
Fix filter on nulls and empty strings

haneslinger:
#### What's this PR do?
Filter on both "" and null values. 

- pick a column
- ensure the column has a full, null and "" value
- test the "" and !"" filters. 

[#3600](https://github.com/SEED-platform/seed/issues/3600)

image
image
image

### DIFF
--- a/seed/utils/search.py
+++ b/seed/utils/search.py
@@ -432,7 +432,7 @@ def build_view_filters_and_sorts(filters: QueryDict, columns: list[dict]) -> tup
         parsed_filters, parsed_annotations = _parse_view_filter(filter_expression, filter_value, columns_by_name)
 
         # if not "", exclude "" and null
-        if filter_expression[-4:] == '__ne' and filter_value == '':
+        if filter_expression.endswith('__ne') and filter_value == '':
             filter_expression = filter_expression.replace('__ne', '__isnull')
             filter_value = ""  # evals to false
             is_not_null_parsed_filters, _ = _parse_view_filter(filter_expression, filter_value, columns_by_name)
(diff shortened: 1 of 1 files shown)


seed/utils/search.py
```diff
@@ -429,10 +429,24 @@ def build_view_filters_and_sorts(filters: QueryDict, columns: list[dict]) -> tup
...
     annotations = {}
     for filter_expression, filter_value in filters.items():
+        parsed_filters, parsed_annotations = _parse_view_filter(filter_expression, filter_value, columns_by_name)
+
+        # if not "", exclude "" and null
         if filter_expression[-4:] == '__ne' and filter_value == '':
             filter_expression = filter_expression.replace('__ne', '__isnull')
-            filter_value = False
-        parsed_filters, parsed_annotations = _parse_view_filter(filter_expression, filter_value, columns_by_name)
+            filter_value = ""  # evals to false
```
axelstudios, line 437: This line is unnecessary since the above _if_ condition already checked that the value was an empty string

seed/utils/search.py
```diff
@@ -429,10 +429,24 @@ def build_view_filters_and_sorts(filters: QueryDict, columns: list[dict]) -> tup
...
+            is_not_null_parsed_filters, _ = _parse_view_filter(filter_expression, filter_value, columns_by_name)
+
+            parsed_filters &= is_not_null_parsed_filters
+
+        # if exactly "", only return "" and null
+        elif filter_expression[-7:] == '__exact' and filter_value == '':
```
axelstudios, line 443: It would probably be cleaner to use `.endswith()` for both conditions rather than specific substring lengths

axelstudios (changes requested):

axelstudios:
One other unique quirk of the empty/null values is that you can use a filter like `< -100` (any number works) on a text field and it will return exclusively empty string matches.. seems like unexpected behavior, but not a result of this PR

seed/utils/search.py
haneslinger: true, but it mirrors the if else below it and I feel it makes the intention explicit.

axelstudios (approved):

Issue #3734, opened 2023-01-06
Update the error message for the ESPM upload


isalanglois:
Ticket made in response to #3179

RDmitchell:
@haneslinger / @isalanglois -- can we put more details into this issue as to what this issue is about?

RDmitchell:
I can't really test this so am moving it to Done

RDmitchell:
@haneslinger / @nllong -- can we close this?

nllong:
done

PR #3735, opened 2023-01-06
Show portfolio manager errors

haneslinger:
I didn't add any error to the backend, just ensured that the errors that were there get to the frontend.
[Update the error message for the ESPM upload#3734](https://github.com/SEED-platform/seed/issues/3734)


anchapin (approved):
This looks good. I was able to get an error (see screenshot below) and it is more helpful than before.

 
image

RDmitchell:
I guess I will move this to Done, as I don't know how to test it. 

I must say that the error message isn't very illuminating to a non-technical user, but it's probably the best we can do, and at the point, whoever gets this error would presumably need to contact a programmer to decipher what is going on.

Issue #3665, opened 2022-12-06
Custom Reports: Reordering the legend


isalanglois:
Revise the legend so that related lines are adjacent to each other. A related line has the same filter group and aggregation type. See the images below for a comparison of the existing legend order compared to the requested legend order.

Existing Legend Order:

![Image](https://user-images.githubusercontent.com/6326294/206573668-ae50d8f2-9d0f-49ed-88b8-7b696b5e66b4.png)

Requested Legend Order:

![Image](https://user-images.githubusercontent.com/6326294/206573735-947bc1d0-3b62-41cd-b96c-627bd16f1338.png)

haneslinger:
Do we want to reorder the labels then so it goes filter group, agg type, column?  I really think alphabetical order is helpful in this case. Knowing the label differ at their ends make the legend way more readable to me.

anchapin:
@haneslinger I like that suggestion. It makes sense to me to reorder the labels as you indicated and use alphabetical order.

RDmitchell:
Instance: dev1
SHA: 294626cbd
Org: BPS Test

@anchapin / @haneslinger -- it doesn't look like this has been implemented yet, as here is the current order. 

![Image](https://user-images.githubusercontent.com/6314950/211438435-25a05d1c-828e-4414-a874-022ee71cb9e6.png)

I am going to put this in the Q2 project, in the To Do column.

RDmitchell:
One solution to this is to allow the user to control how the legend is displayed, among the choices of Filter Group, Aggregation Type, and metric (?).

isalanglois:
@haneslinger and @anchapin could you follow up on if this has been implemented or the status of this effort?

haneslinger:
Hey robin, I confirmed that my fix was part of the SHA and that it worked. I'm unsure why it wouldn't be for you. I'll afraid I'd need more info. Was this on dev? which org?


<img width="348" alt="image" src="https://user-images.githubusercontent.com/30241543/211653561-c9201f4c-4f57-4d64-8e02-ddcbeb4460c2.png">

RDmitchell:
@haneslinger == see the info in my comment above about the instance and SHA

RDmitchell:
Maybe I need to clear my cache, if the code for fixing this has already been deployed to dev1. 

Could someone confirm that the new code has been deployed to dev1? If that's the case I will empty my cache and retest.

anchapin:
[...]

PR #3740, opened 2023-01-10
Fix custom report legend sort for chrome

haneslinger:
idk why chrome doesn't like `>` but 🤷 
https://github.com/SEED-platform/seed/issues/3665


anchapin (approved):
Assuming the tests pass, it is approved. Thanks for figuring that out quickly.

Issue #3744, opened 2023-01-12
SEED Monthly Data Upload Question - Overwriting Values & "Incoming" count


RDmitchell:
From Dan Eden, via the SEED Users groups.io platform
Daniel Eden, P.Eng., MEL, LEED AP | GRID Product Manager | OPEN Technologies
604-365-1263 | opentech.eco

Hello!

I had two quick questions about the meter data upload to SEED: 

1) Overwriting data: sometimes we need to correct erroneous monthly data by uploading a new file with the corrected data. In the past, when I would upload a new file, it would overwrite the old "bad" data. Now, after the most recent release, I'm getting an error message that there are "overlapping readings" and no meters were "successfully imported". How do I correct old data? 

![image](https://user-images.githubusercontent.com/6314950/212183272-9426eb30-238c-4342-85f4-6b432d10cb1b.png)

2) "Incoming" count: I'm trying to upload 12 months of electricity data and 12 months of nat gas data, but the pop-up modal notes that I'm trying to upload 24 values for each energy type, which I think is an error. Attached is the file that I'm uploading, and see the incoming count in the screenshot above. 
*****
My Reply to that thread

Dan

1) Overwriting meter data: 

I suspect that the meter import does not currently support importing data "over" an existing file. This functionality might need to be added. 

With the release of 2.17.0, you can delete existing meter data, so if it is acceptable to delete the existing data and then import the new data, that release has that functionality. 

I guess the question is whether you need to keep a history of the previously imported data. 

And maybe there are more subtleties to your question that I am not understanding. 

2) Incoming count

I will do some testing with your attached file to see how version 2.17 is handling this.

Robin 
******
Reply from Dan

Hi Robin,

We did a few more checks, and we think we found the error.

I believe that your team recently created a fix to accept the new ESPM export format (where it changed the headers on the monthly tab of the excel workbook), reported in [3179](https://github.com/SEED-platform/seed/issues/3179) and addressed in [3716](https://github.com/SEED-platform/seed/pull/3716).

I don’t think that the fix worked completely. When I try to upload an excel file with the new ESPM excel format, I can finish the meter upload “process”, but I get the two errors listed below (i.e., it notes 24 months instead of 12 months and then also doesn’t upload the meter data noting an “overlapping” issue).

[...]

PR #3745, opened 2023-01-12
Fixed issue with overlapping meter readings with new ESPM format

anchapin:
When we revised the code to accept the new ESPM format for monthly meter usage, there ended up being an issue where the data was being added twice (so 24 months of data instead of 12 months when uploading a year's worth).

#### What's this PR do?
Switches from checking whether the field starts with the header_string to checking whether the stripped field (removed the units and space at the end of the string) is equal to the header_string.

Upload the example file called "example-data-request-response-new-format.xlsx" (saved in seed/tests/data) and see whether it has 48 records for each fuel type or 96 records.

#3744


RDmitchell:
@anchapin -- that was quick !! Thanks for fixing this !

haneslinger (approved):

nllong (approved):
looks good, thanks for moving back to `startswith`. 

Do you want me to merge @anchapin ?

PR #3747, opened 2023-01-13
Remove all extra spaces in headers from XLSX files

nllong:
Files from ESPM can have headers that are inconsistent, e.g., extra space after the name, multiple spaces between words, etc.

#### What's this PR do?
* Update the MCM parser code to strip the spaces and remove multiple spaces for header names only within XLSX files
* Update ESPM test file with these changes for testing

* Verify that the updated test file (seed/data_importer/tests/data/example-pm-data-request-with-meters.xlsx) imports meters correctly

Part of #3744


anchapin (approved):
Everything looks good.

Issue #3610, opened 2022-11-11
Add ability to export column mapping profile to CSV


nllong:
**Is your feature request related to a problem? Please describe.**

The [py-seed project](https://github.com/SEED-platform/py-seed) has the ability to create a column mapping profile from a CSV file. Doing this by hand can be error-prone due to having to match the names as they are in SEED. The CSV file format should look like [this](https://github.com/SEED-platform/py-seed/blob/develop/tests/data/test-seed-data-mappings.csv).

**Describe the solution you'd like**

Add the ability to export the mapping profile to CSV. Eventually, it would be nice to be able to import a mapping profile from the same CSV... but that can be for another day.

RDmitchell:
@nllong -- is this something that can be done through the UI? If yes, where?

RDmitchell:
instance: dev1
SHA: 
Org: City of San Francisco

I went to the Org/Column Mappings link and clicked the Export to CSV icon
https://docs.google.com/spreadsheets/d/1w8xUM6XF6qfz1o-ghH8D5u1MwjaD1mqO/edit?usp=share_link&ouid=103887815394604824348&rtpof=true&sd=true

The file was exported, but looking at it, not all the fields are included, probably because of limitations of Excel
The file stops on line 869 of the spreadsheet, with the field "Retail Store - Open of Closed Refrigeration ..."
![image](https://user-images.githubusercontent.com/6314950/207973937-8c1a5a7f-18a0-4740-ade7-9be061bef77e.png)

However, looking at the column mapping screen in SEED for that org, there are many more fields under Retail Store
![image](https://user-images.githubusercontent.com/6314950/207974270-242c4402-1807-40bc-a84d-8c23e13ff769.png)

And the fields end with Zoo
![image](https://user-images.githubusercontent.com/6314950/207974303-0b111fdf-599c-43a6-96e9-2ca260197c96.png)

Any user with ESPM data will likely have this many fields, as they are the standard "full" set of ESPM fields. 

So we need to figure out a way to accommodate them, either by making 2 files, etc.

RDmitchell:
@isalanglois -- this needs to go back into "in progress" to deal with the problem I found. Thx

isalanglois:
@RDmitchell  thanks for the update, tagging @nllong since it looks like he was working on it before

nllong:
@RDmitchell  - the file linked above is to Potential TAC Partners.

RDmitchell:
[...]

PR #3748, opened 2023-01-13
Sort column mappings in profile before export

nllong:
The column mapping export data was being exported in order of IDs, which was reasonable; however, in orgs with lots of column mappings (~1,000) it appeared that all the data was not being exported. One of the reasons was that the view on the website did not match the view in the exported file.

#### What's this PR do?
* Sort (case insensitive) the column mappings before export

* In SF org (or any other org), go to column mappings
* Export and verify that the to_field is sorted alphabetically and consistent with the results in the SEED UI.

Further addresses #3610 

Here are the results of SF after sorting ... [2020-11-14 org-wide mapping snapshot_mapping_profile.csv](https://github.com/SEED-platform/seed/files/10415371/2020-11-14.org-wide.mapping.snapshot_mapping_profile.csv)


anchapin (approved):
It works as expected and I don't see any issues.

Issue #3756, opened 2023-01-18
Add CRUD operations to meter readings


nllong:
**Is your feature request related to a problem? Please describe.**

MeterReadings are currently only creatable from the import of structured files.

**Describe the solution you'd like**

There needs to be a process to import meter readings directly from the API

PR #3757, opened 2023-01-18
Add CRUD to MeterReading

nllong:

Meter readings can only be imported from data files (green button, BuildingWync, or specifically formatted ESPM formats). Sometimes these files are not available for properties and users will want to import the data via a direct API call. 

#### What's this PR do?
* Adds in the MeterReading view set for creating individual meter readings on a meter
* Create 3 new meter types to handle generic meter types: electric, district chilled water, and custom meter. Prior to this change the district chilled water also included how the water was created, which can be unknown.

* The API methods are not available to the front end. Testing this is easiest through swagger, postman, and/or curl. The unit tests cover the functionality.
* See /v3/properties/{property_pk}/meters/{meter_pk}/readings/ in Swagger

#3756


seed/api/v3/urls.py
```diff
@@ -107,6 +108,9 @@
 properties_router.register(r'meters', MeterViewSet, basename='property-meters')
 properties_router.register(r'notes', NoteViewSet, basename='property-notes')
 properties_router.register(r'scenarios', PropertyScenarioViewSet, basename='property-scenarios')
+# This is a third level router, so we need to register it with the second level router
+meters_router = nested_routers.NestedSimpleRouter(properties_router, r'meters', lookup='meter')
```
nllong, line 112: @anchapin  -- here is the example of the third level route which will also need to be created for /properties/1/scenarios/10/measures

seed/views/v3/meter_readings.py
```diff
@@ -0,0 +1,54 @@
...
+        # the permissions for the user
+        org_id = self.get_organization(self.request)
+        # get the property id - since the meter is associated with the property (not the property view)
+
+        # even though it is named 'property_pk' it is really the property view id
+        property_view_pk = self.kwargs.get('property_pk', None)
+        if not property_view_pk:
+            # Return None otherwise swagger will not be able to process the request
+            return MeterReading.objects.none()
+        property_view = PropertyView.objects.get(pk=property_view_pk)
```
haneslinger, line 33: super nit: this should probably be try/catch-ed

seed/views/v3/meter_readings.py
```diff
@@ -0,0 +1,54 @@
...
+        # Grab the meter id
+        meter_pk = self.kwargs.get('meter_pk', None)
+        if not meter_pk:
+            # Return None otherwise swagger will not be able to process the request
+            return MeterReading.objects.none()
+        self.meter_pk = meter_pk
+
+        return MeterReading.objects.filter(
+            meter__property__organization_id=org_id, meter__property=self.property_pk, meter_id=meter_pk
+        )
```
haneslinger, line 45: again, try/catch-ed

seed/tests/test_meter_views.py
```diff
@@ -147,3 +145,102 @@ def test_update_meter(self):
...
+        }
+
+        response = self.client.post(url, data=json.dumps(payload), content_type='application/json')
+        self.assertEqual(response.status_code, 201)
+        self.assertEqual(response.json()['reading'], 10)
+
+        # now delete the item and verify that there are no more readings in the database
+        detail_url = reverse('api:v3:property-meter-readings-detail', kwargs={'property_pk': property_view.id, 'meter_pk': meter_pk, 'pk': '2022-01-05T05:00:00Z'})
+        response = self.client.get(detail_url, content_type='application/json')
+        self.assertEqual(response.status_code, 200)
```
haneslinger, line 238: what are these lines doing for the test?

haneslinger (commented):
Looking good. What does the Update endpoint look like? Can you update just one reading or must you do them are?

seed/views/v3/meter_readings.py
nllong: This is a filter, so I don't know the advantage of try catching this unless you are concerned about some sort of syntax error. We are verifying above the call that property_pk and meter_pk exist, so the result would just be an empty MeterReading set... which is okay.

seed/tests/test_meter_views.py
nllong: Verifies that the data are returned, but with just a 200. Nothing fancy.

seed/views/v3/meter_readings.py
nllong: updated...

nllong:
> Looking good. What does the Update endpoint look like? Can you update just one reading or must you do them are?

Yup, i added both individual updates and bulk updates. The Swagger page is probably the best view on what one can do now with the meter readings. The one oddity is that since the data are part of the timescale hypertable, the start_time is the primary key.

image

seed/lib/superperms/orgs/models.py
```diff
@@ -161,9 +143,9 @@ class Organization(models.Model):
         'Fuel Oil (No. 5 and No. 6)': 'kBtu (thousand Btu)',
         'Kerosene': 'kBtu (thousand Btu)',
         'Natural Gas': 'kBtu (thousand Btu)',
-        'Other:': 'kBtu (thousand Btu)',
+        'Other:': 'kBtu (thousand Btu)',  # yes, other has a colon at the end.
```
axelstudios, line 146: 🥴

axelstudios (changes requested):
500 errors can result from meter endpoints

axelstudios (approved):

Issue #3788, opened 2023-01-25
Compliance checks do not cast values as needed


nllong:
**Describe the bug**

Compliance reports don't load

**Expected Behavior**

Should be able to visualize the compliance page

**Actual Behavior**

```
Traceback (most recent call last):
  File "/Users/nlong/.pyenv/versions/3.9.7-seed/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/Users/nlong/.pyenv/versions/3.9.7-seed/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/nlong/.pyenv/versions/3.9.7-seed/lib/python3.9/site-packages/sentry_sdk/integrations/django/views.py", line 67, in sentry_wrapped_callback
    return callback(request, *args, **kwargs)
  File "/Users/nlong/.pyenv/versions/3.9.7-seed/lib/python3.9/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/Users/nlong/.pyenv/versions/3.9.7-seed/lib/python3.9/site-packages/rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/nlong/.pyenv/versions/3.9.7-seed/lib/python3.9/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/Users/nlong/.pyenv/versions/3.9.7-seed/lib/python3.9/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/Users/nlong/.pyenv/versions/3.9.7-seed/lib/python3.9/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/Users/nlong/.pyenv/versions/3.9.7-seed/lib/python3.9/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/Users/nlong/working/seed/seed/seed/decorators.py", line 212, in _wrapped
    return fn(self, request, *args, **kwargs)
  File "/Users/nlong/working/seed/seed/seed/utils/api.py", line 149, in _wrapped
    return fn(self, request, *args, **kwargs)
[...]

PR #3789, opened 2023-01-25
Fix bug where compliance calculations were not being cast to the right data types

nllong:
Compliance calculations were not casting the data types before checking causing the front end to not show data.

#### What's this PR do?
* Move the `DATA_TYPE_PARSERS` that was in `search.py` to the Column class. 
* Adds a class method and instance method to cast data from the column. It is only one value at a time at the moment
* Cast `date` as `date` instead of `datetime` (which was happening)
* Write a bunch of test cases for casting data
* Update the pre-commit versions

* Import program target data as a string. This may not be possible from the UI, but it did happen using py-seed
* Set the program metric up
* View the compliance page
* Also, monkey test the search, but the unit tests should cover that

#3788 

We can now view lots of data in the program insights page!
image


seed/utils/search.py
```diff
@@ -287,7 +288,7 @@ def _build_extra_data_annotations(column_name: str, data_type: str) -> tuple[str
         annotations.update({
             final_field_name: Cast(cleaned_field_name, output_field=models.IntegerField())
         })
-    elif data_type in ['number', 'float', 'area', 'eui']:
+    elif data_type in ['number', 'float', 'area', 'eui', 'ghg', 'ghg_intensity']:
```
nllong, line 291: This may resolve some other unexpected issues with casting data via the search.py method.

seed/utils/search.py
```diff
@@ -321,19 +323,6 @@ def _parse_view_filter(filter_expression: str, filter_value: Union[str, bool], c
     :param columns_by_name: mapping of Column.column_name to dict representation of Column
     :return: query object
     """
-    DATA_TYPE_PARSERS: dict[str, Callable] = {
```
nllong, line 324: This was moved to the Column model

kflemin (approved):

axelstudios (approved):

Issue #3780, opened 2023-01-21
Column Settings: getting 520 error when making a new column but column is created


RDmitchell:
Instance: dev1
SHA: 257ecb634
Org: LBNL Test 200

**_I see this option in dev1 but not staging_** ... ??? Is it not to be included in the 2.17.0 release? Or maybe I am just not seeing it (or going crazy !!)

In Organizations / Column Settings, I tried making a new field called "New Field for Robin". 

The spinner spun, and eventually the program returned a 520 error, which seems it is a Cloudflare error (?)
![image](https://user-images.githubusercontent.com/6314950/213839316-e8777c8b-cdfc-45bc-9590-d6a56b844f0a.png)

But it does appear that the new field was created.
![image](https://user-images.githubusercontent.com/6314950/213839390-9872ea16-4c19-49dc-87d0-23d1a051d4c2.png)

RDmitchell:
The field does appear in the Column List Profiles screen for Properties
![image](https://user-images.githubusercontent.com/6314950/213839518-86c2a1de-f464-4029-93bf-595a7afbbb9f.png)

I could also edit it by hand in the Property Detail view
![image](https://user-images.githubusercontent.com/6314950/213839662-e3ee1b61-f066-4030-9909-c511fd554a8d.png)

And the new edits show up in the Property List view
![image](https://user-images.githubusercontent.com/6314950/213839673-0527f665-826d-45de-b958-5b533892bc69.png)

RDmitchell:
@haneslinger / @axelstudios -- is this bug still a problem? The related issue is in test and maybe it shouldn't be? 

Should it be moved to "In Progress"?

haneslinger:
This is still a bug. It will be fixed by https://github.com/SEED-platform/seed/pull/3791 and https://github.com/SEED-platform/seed/pull/3806 both of these PRs are in Review.

haneslinger:
> @haneslinger / @axelstudios -- is this bug still a problem? The related issue is in test and maybe it shouldn't be?

Should it be moved to "In Progress"?

Test is the correct place for  #1758 as it is currently being tested. It and this ticket could be collapsed tho.

RDmitchell:
Instance: LBNL staging
SHA: 992862f09
Release: 2.17.1
Org: LBNL Test 200

When I added the field in Column Settings, this error was displayed in Developer tools.
[...]

PR #3791, opened 2023-01-25
Catch create column error correctly

haneslinger:
@axelstudios this'll give us a actual error message, but is there a way to figure out what the error underneath is before this hits dev1? locally, I get no errors. 

[Column Settings: getting 520 error when making a new column but column is created#3780](https://github.com/SEED-platform/seed/issues/3780)

image

### DIFF
--- a/seed/static/seed/js/controllers/create_column_modal_controller.js
+++ b/seed/static/seed/js/controllers/create_column_modal_controller.js
@@ -37,7 +37,7 @@ angular.module('BE.seed.controller.create_column_modal', [])
         then(() => {
           $uibModalInstance.close();
           $state.reload();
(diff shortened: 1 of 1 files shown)


RDmitchell:
@haneslinger / @axelstudios -- is this bug still a problem? The related issue is in test and maybe it shouldn't be?

haneslinger:
It's a problem, this and https://github.com/SEED-platform/seed/pull/3806 should fix it. I'm just waiting on reviews.

perryr16 (approved):
Approved. The console error is fixed
```
angular.js:15697 TypeError: columns_service.create_column_for_org(...).then(...).error is not a function
```

Issue #3809, opened 2023-01-29
Organization: Column Mapping: What does Add Column do?


RDmitchell:
Instance: LBNL Staging
SHA: https://github.com/SEED-platform/seed/commit/992862f09fd2f9bd7bca3f1403058f53da20e493
Release: 2.17.1
Org: LBNL Test 200

In Organizations / Column Mapping, I can't figure out what Add Column does. If I click on it and type in a new column, it doesn't appear to get saved, and there is no save button. 

Just need to understand it so I can document it. 

![image](https://user-images.githubusercontent.com/6314950/215300844-d9b23a3a-398a-4dbc-af31-01592688bfb3.png)

isalanglois:
@haneslinger Could you look into this one too?

haneslinger:
In traditional seed fashion, [this was a button was added three years ago by somebody who doesn't work here anymore.  ](https://github.com/SEED-platform/seed/pull/2022)

I have to imagine at one point it let you add column mappings (not columns), but this functionality is now broken- for sure at the ui level, maybe deeper down to, idk. 

@nllong, Shoudl we remove it or fix it?

RDmitchell:
@nllong -- I would vote to remove it, but that's just me ... obviously others should weigh in. 

I guess I could query a few existing users (SF, DC, Berkeley) to see if they use it. 

I actually never noticed it before and I suspect others are not using either, particularly if it's not working.

nllong:
yeah, I think the `add column` button can be removed. I really would like to see an easy way to import mappings into this screen from an export. It makes testing these 1000 column mapping organizations a lot easier. The `add column` doesn't really make it easier for a user to setup their column mappings, it just adds another row, which they should do from some sore of CSV import.

RDmitchell:
Instance: dev1
SHA: 26aac34d8

It does appear to be gone ! Thanks !!

PR #3813, opened 2023-01-31
Remove add column in mapping page

haneslinger:
#### What's this PR do?
I didn't remove the function `add_new_column` cause it's used in `copy_csv_headers`. idk if _that_ function works, but that's not what this ticket is about. 🤷 

[Organization: Column Mapping: What does Add Column do?#3809](https://github.com/SEED-platform/seed/issues/3809)


anchapin (approved):
Works as intended. Thanks!

Issue #3776, opened 2023-01-20
Change default behavior of BETTER analysis


anchapin:
Set the default behavior for a BETTER analysis to use the meter data associated with the active cycle instead of all meter data.

RDmitchell:
Are meters currently associated with cycles? There isn't any indication of that in the current Meters view, as far as I can tell. 

And I guess there is another issue for associating Analyses with cycles. 

And if we are having meters be associated with Cycles, seems like we should also make Sensors associated with a Cycle? 

Not sure, just looking at the links under Property Detail and wondering which need to be associated with cycles.

anchapin:
At the moment, meters and sensors are not associated with a cycle, just the property. This is due to the fact that the data from meter reading and data loggers may span across two cycles. Therefore, they exist outside of cycles and are split as needed between cycles in certain situations.

RDmitchell:
Instance: dev1
SHA: 6bec6092b
Org: BPS Test

Does the fact that the program now allows the user to select a meter data range when setting up a BETTER analysis, with the program seeming to default to dates associated with the current cycle, address this issue? 

Inventory List
* Set the Cycle to 2020
* That cycle is set to start on 12-31-2019 and end on 12-30-2020
* Select a property and use the Action pulldown to define a BETTER analysis
* ![image](https://github.com/SEED-platform/seed/assets/6314950/72609199-683e-43d5-ae72-b9a9f9d4869e)
* Clicking on the "Select Meter Data Range" displays the default range associated with the cycle of the current properties
* ![image](https://github.com/SEED-platform/seed/assets/6314950/aa717275-cb9d-4302-ba71-0683d143678a)
* It is also possible to change the date range if desired
* ![image](https://github.com/SEED-platform/seed/assets/6314950/6efa18f6-b59a-4ab2-ade1-ddae42346996)

PR #3819, opened 2023-02-02
Update default meter range selection for BETTER analysis

perryr16:
The current default behavior of a BETTER analysis is to "Use All Meter Data" and there is no option for cycle select.

#### What's this PR do?
Changes the default behavior to "Select Cycle" where users can choose a cycle to be used as a start and end date range. The current cycle is selected by default.

- Create a new BETTER analysis, the positioning and default behavior should be updated to be on the left and selected.
- Create a new BETTER analysis with a specific cycle. Check analysis results.

#3776

OLD
Screenshot 2023-02-02 at 11 34 19 AM

NEW
Screenshot 2023-03-31 at 1 48 10 PM


nllong:
should we do the same for the BETTER time period that we did for the updated EUI? that is, allow for "use cycle period"?

RDmitchell:
@nllong -- seems like it would be good to have that option and be consistent.

Ryoken (approved):

Ryoken:
Appears to work as expected!  It was agreed the further work mentioned should be handled in a new ticket.

axelstudios (approved):

Issue #3801, opened 2023-01-26
Revise EUI modal to include optional start/end dates


anchapin:
Revise the Run Analysis (beta) modal from the Analyses (beta) tab of the Property Detail page to include start and end dates. #3800 was created to revise the api to accept this information. This is giving the user flexibility to set the start and end date otherwise the most recent year of meter data will be utilized like it is currently.

nllong:
I suggest following the pattern setup in the BETTER Analysis pipeline. 

![image](https://user-images.githubusercontent.com/1907354/215651767-d45161ac-d1fd-4975-9769-b00d126c65d9.png)

The only addition that might be nice would be to select "Use Cycle's Date Range" in addition to the "Use All Meter Data" and "Select Meter Data Range"

anchapin:
I was thinking that we should force a minimum of one year period and potentially a maximum of one year as well when the "Select Meter Data Range" is selected. EUI is typically only done based on annual energy usage. What do you think?

Also, if "Use Cycle's Date Range" is selected, what should we do when the cycle is not exactly a year? We could give an average EUI for cases that don't equal a year (i.e. we can take the calculated EUI and multiply it by (365 days / days in calculation range) to convert it to equivalent yearly EUI). Although, the "Annual Coverage %" would need to be factored into this discussion.

haneslinger:
This analysis populates columns, which, of course, are attached to a state which is attached cycle. I think you should not get to choose a date range, but a cycle. If not,  how do we decide what column to populate?

RDmitchell:
I didn't think that the BETTER analyses were associated with cycles ... but maybe I'm wrong. 

It's not obvious when you look at the analysis results that they are associated with a cycle.

RDmitchell:
Maybe for now it would be better to just look for the latest data in meters -- which I think is what we are doing now? And not put it on the user to specify the dates.

anchapin:
Hi @RDmitchell, we just discussed this ticket in an internal meeting today and I wanted to relay the summary of that discussion. We decided to have three options for EUI analysis date range (cycle, most recent 12 months, and custom date range).

RDmitchell:
@anchapin -- thanks for the background.

RDmitchell:
Instance: dev1
SHA: https://github.com/SEED-platform/seed/commit/c94869068feb021cbb258ac4815be65b98dbc5e0 (2.18.0)

This is working

Issue #3800, opened 2023-01-26
Revise the EUI api to allow an input for the start/end dates


anchapin:
Right now the EUI analysis always calculates the EUI based on the most recent year of meter data no matter which cycle is active. This may not always be what the user wants therefore the user should be able to specify the start and end dates. If the dates aren't provided default to the current behavior.

RDmitchell:
Instance: dev1
SHA: c94869068 (2.18.0)

This is working nicely !

PR #3820, opened 2023-02-02
Specify cycle in eui analysis

haneslinger:
https://github.com/SEED-platform/seed/issues/3801 and https://github.com/SEED-platform/seed/issues/3800
image


seed/tests/test_analysis_pipelines.py
```diff
@@ -1001,32 +1001,115 @@ def test_invalid_meters(self):
...
+            end_time=make_aware(datetime(2022, 1, 28, 0, 0, 0), timezone=self.timezone_object),
+            reading=12345,
+            source_unit='kWh',
+            conversion_factor=1.00
+        )
+
+        self.maxDiff = None
+        config = {
+            "select_meters": "select_cycle",
+            "cycle_id": self.cycle.id,
```
anchapin, line 1108: Changing this to use cycle name makes more sense to me.

seed/tests/test_analysis_pipelines.py
```diff
@@ -1001,32 +1001,115 @@ def test_invalid_meters(self):
...
+            end_time=make_aware(datetime(2020, 12, 28, 0, 0, 0), timezone=self.timezone_object),
+            reading=12345,
+            source_unit='kWh',
+            conversion_factor=1.00
+        )
+
+        self.maxDiff = None
+        config = {
+            "select_meters": "date_range",
+            "meter": {
```
anchapin, line 1072: When I select date range I only see "select_meters = date_range" but not the start and end dates. Is that intentional? It would be good to show the start and end dates. See image below.

image

anchapin (changes requested):
Overall, it looks really good. I just had two specific comments related to the analysis card that's created.

seed/tests/test_analysis_pipelines.py
haneslinger: I do not believe cycle name is unique. I can add name for documentation, but name alone will not suffice.

seed/tests/test_analysis_pipelines.py
haneslinger: I did not alter how the analysis are displayed, let me see if I can add it.

seed/tests/test_analysis_pipelines.py
anchapin: Sorry, maybe this isn't the correct place in the code then. I only wanted to change how the analysis is displayed like in my comment below regarding start/end date range in the analysis card. I agree that you should use cycle id for identifying which cycle to use.

seed/tests/test_analysis_pipelines.py
anchapin: Sounds good. Thank you!

anchapin: This looks good now. However, can we have start_date first then end_date? I tried to figure out how to change this but couldn't figure it out quickly.

seed/tests/test_analysis_pipelines.py
anchapin: Thank you for adding the cycle name to the analysis card. Can we remove the cycle id? It's unnecessary for the user.

PR #3822, opened 2023-02-03
Sort the dropdown of meter units

nllong:
In a user's settings page, the list of meter units can be pretty long and as new meters are added, the order of the list needs to be sorted.

#### What's this PR do?
* sort the backend/django return of the org data's meter unit display 

* go to an org
* select setting -> units
* Look at the meter energy units drop down and make sure the values are sorted.

n/a

Before:
image

After:
image


haneslinger (approved):
🧽

PR #3840, opened 2023-02-09
Implement Salesforce integration

kflemin:
## SEED to Salesforce Functionality
Allows organizations to configured their salesforce settings in SEED, add their customized field mappings, and automatically update Salesforce on a set interval

The configuration section is within the org Settings page. "salesforced_enabled" must be checked to enable this functionality.  When enabled, user can enter their salesforce connection information:
Screen Shot 2023-02-09 at 10 18 32 AM

Additional configurations for which labels to use in the workflow and specific field names both on the SEED side and the Salesforce side:
Screen Shot 2023-02-09 at 10 22 04 AM

Configuration for the automatic sync:
Screen Shot 2023-02-09 at 10 18 40 AM

Customized field mappings for the Benchmark object:
Screen Shot 2023-02-09 at 10 18 51 AM

Once all of this is configured, users can go to an individual property detail page and run the update manually:
Screen Shot 2023-02-09 at 10 19 03 AM

Or select a group of properties and run the update manually:
Screen Shot 2023-02-09 at 10 19 21 AM

The update will automatically run daily on all eligible properties if the configurations for hour, and minute are set in the Settings page.

## Notes
- you must run the db migrations to enable this functionality
- new dependencies: seed-salesforce, py-seed, django-celery-beat
- had to update requests to 2.28 due to py-seed dependency mismatch
- for local testing, add this to the end of the current celery start command: -B --scheduler django_celery_beat.schedulers:DatabaseScheduler


seed/static/seed/locales/fr_CA.json
```diff
@@ -765,11 +777,39 @@
...
   "SELECT_DATA_SET_TITLE": "Sélectionner un ensemble de données",
+  "SELECT_DOC_UPLOAD_TITLE": "Sélectionner un document à télécharger",
   "SELECT_THE_TYPE_OF_FILE_TO_ADD_TO_DATASET": "Sélectionnez le type de fichier(s) que vous souhaitez ajouter à {dataset_name}.",
   "SENSORS": "Capteurs",
   "SENSOR_CONTENTS_TITLE": "Confirmer le contenu du fichier",
   "SENSOR_READINGS": "Lectures de capteur",
   "SENSOR_READING_COUNTS_TITLE": "Compte de lecture du capteur",
+  "SF_ACCOUNT_NAME_FIELDNAME_TEXT": "Selectionez le champ qui contient le nom de compte du contact. Ex: 'Organization'",
```
nllong, line 786: so many! good work!

nllong (approved):
Crazy amount of additions to make this happen. Great work!

axelstudios:
@kflemin Here are some of the issues I've found so far, I'll update this as I add more.  Let me know if you start tackling some of them, otherwise I'll start in on it:

- As soon as you hit `Enable Salesforce Integration` and then `Save` you'll get 500 errors from the `salesforce_configs` endpoint. Even if you disable it you'll still get 500 errors.  The issue is that that creates a row with almost entirely `<null>` values in the database, and it can't run `decrypt(item['password'])[0]` when the password is None
- We should not have an endpoint to `list` SalesforceConfigs, that implies that it would return every organizations' config in an array
- All SalesforceConfig endpoints should use `requires_owner` instead of `requires_viewer`, we don't want viewers or members to be able to access decrypted Salesforce passwords
- It allows you to create multiple Salesforce field mappings to the same name and to the same column (needs a unique constraint)
  - Also, not a big deal, but those mappings are saved instantly, whereas the rest of the page has to be explicitly "Saved"
- If you hit the `Enable Salesforce Integration` checkbox and then try to test a Salesforce connection it fails because `Salesforce functionality is not enabled for this organization`. I get that you haven't saved yet, but it seems like you'd still be able to test a connection without saving first.
- We ought to merge the migrations

kflemin:
@axelstudios:

- fixed the 500 related to decrypt pwd
- the list endpoint only returns the org's configs, which is only 1 config possible, but I do use the list endpoint in the frontend to load that form so I don't have to know the ID of the config.
- updated all salesforce_configs views to use 'require_owner' permissions
- added a unique constraint on the salesforce fieldname + seed column
- allow testing connection without saving first
- merged the migrations just for you!😆
I think that's it🤞

axelstudios:
I added some fixes:
- Fixed `undefined` in connection test url, e.g. `/api/v3/salesforce_configs/undefined/salesforce_connection/`
- Fixed 500 error: POST to `/api/v3/salesforce_configs/` would attempt to create multiple configs per org
- Fixed 500 error: POST to `/v3/properties/update_salesforce/` was using a nonexistent key, and had undefined variables
- Fixed error notifications being suppressed on the org settings page
- Running the worker and beat services with the same command is not recommended for production - split into separate processes for production scripts
- Made a slight tweak to the settings layout to match the other checkbox styles, it was moving around on wide screens
- Fixed alignment between field names and field descriptions
- Removed redundant beats arguments

axelstudios (approved):

Issue #3847, opened 2023-02-12
Property footprint does not import from GeoJSON


nllong:
**Describe the bug**

The polygon isn't importing as 'property_footprint' when there is a polygon in a geojson file.

**Expected Behavior**

There should be an entry in the mapping page for one to map the footprint. If mapped to the canonical field of `property_footprint` then it should show up in the inventory.

**Actual Behavior**

Doesn't show any mapping options

**Steps to Reproduce**

Attached to this ticket is a GeoJSON file
* import as GeoJSON (unzip the attached file)
* Verify that there is no "property_footprint" in the mapping page
* This results in no footpring

**Instance Information**

local

**Additional Context**

[ubid_subselection.geojson.zip](https://github.com/SEED-platform/seed/files/10717607/ubid_subselection.geojson.zip)

nllong:
still an issue and a complete mystery :)

nllong:
Still an issue

RDmitchell:
@nllong -- is this issue fixed?

nllong:
no, still not fixed.

PR #3848, opened 2023-02-12
Fix importing of GeoJSON property footprints

nllong:
Footprints were being imported as `bounding_box` but the column was never showing up it the mapping table. Note that `bounding_box` is not the right field if there are coordinates in a GeoJSON file. `bounding_box` comes from UBID decoding.

#### What's this PR do?
Fixes the importing of the footprints from GeoJSON. I think historically we were testing with just CSV files.

See issue #3847. There is a test file in there and some instructions.

#3847 

Fixed version should look like this
image


nllong:
ESPM appears to be down or at least the logging-in is just hanging. Rerun tomorrow morning.

kflemin (approved):

Issue #3818, opened 2023-02-01
Property List: Filter by Label shows more labels than used in the currently inventory


RDmitchell:
Instance: LBNL Staging
SHA: 2ba1e3806
Release: 2.17.1
Org: LBNL 500
Cycle: Test

In the new Property List view, showing the Filter by Label options shows more labels than are currently used in this cycle -- maybe it's showing all the labels across all the cycles? 
![image](https://user-images.githubusercontent.com/6314950/216179861-ac127dc1-eb9d-43e9-b8ef-e38c79f51107.png)

However, the cycle being currently viewed only has 3 labels applied to the records, as shown in the Properties List (Legacy) view. I think this is the appropriate behavior. 
![image](https://user-images.githubusercontent.com/6314950/216180239-fb060e39-c564-44fd-a293-d2f276adffab.png)

Maybe this has changed because of some other functionality, but I think seeing labels that aren't being used in the current cycle is confusing for the user.

RDmitchell:
Instance: LBNL staging
SHA: de3b9d486
Org: LBNL 500

I can't reproduce this issue so I'm closing it.

PR #3858, opened 2023-02-13
Filter cycle specific labels on inventory list

perryr16:
The new Inventory List page's filter auto-complete shows all labels that are associated with any property, in any cycle. This differs from the functionality on Inventory List (legacy) page where only labels associated with the cycle's inventory are present.

#### What's this PR do?
- Adds cycle-specific inventory filters to the returned labels. The prior functionality returned all labels that have been applied to a property regardless of cycle.
- Changes the order of function calls. For context, the filter labels function generated labels and then found the cycle's inventory. However, to accurately filter labels the inventory (`$scope.data`) must first exist. 

- create at least 2 cycles. 
- apply cycle specific labels to a few properties
- apply cross-cycle labels to a few properties
- click the "Filter by Label" input field. The auto-complete should only include labels applied to this cycle.

#3818

NEW
2 cycles with unique and common labels.
Screenshot 2023-02-13 at 3 02 03 PM
Screenshot 2023-02-13 at 3 00 55 PM

OLD
Screenshot 2023-02-13 at 3 07 20 PM


haneslinger:
I found a small bug:

Says I have a filter group, "Called", which simply filters fro properties with the label "Call"
If I go to a cycle where none of the properties have the label Call and apply filter group Call, it'll break.
This is because "Call" was filtered out of labels (as no property has the label call), and the UI got confused.

haneslinger (approved):

axelstudios:
This PR had two regressions:
- Changing the label filter had no effect until another filter was modified
- It only suggested labels _for the current page_ rather than the current cycle. If a record on a different page had a different label it wouldn't appear as an option

It also doubled the inventory fetch requests on change.

@perryr16 I added some fixes:
- When requesting labels for `property_view`s or `taxlot_view`s, `cycle_id` is now an optional path argument to limit `is_applied` ids to the selected cycle
  - This results in a decent performance improvement since listing labels previously required looping over _every_ view to check `is_applied`, rather than just views from the chosen cycle
- When changing cycles, labels that aren't applicable are removed from the filter

axelstudios (approved):

Issue #3859, opened 2023-02-13
In property insights ‘Create a label’ from the legend


haneslinger:

RDmitchell:
@haneslinger / @axelstudios -- I see the "Update Labels" button but what is it supposed to do? 


![Image](https://user-images.githubusercontent.com/6314950/230696593-a8d7649c-e888-4221-bd2f-007a405f9abb.png)

haneslinger:
It allows you to add labels to the visible properties.[ It was requested in the SPEC work](https://docs.google.com/document/d/15hkcIxHxbqKR_EEjAD-r1mEXG0WnD5ppbzWN5LZCFYw/edit)

RDmitchell:
@haneslinger -- but how do you select the properties you want to label?

PR #3860, opened 2023-02-13
Add labels button on property insights page

haneslinger:
[#3859](https://github.com/SEED-platform/seed/issues/3859)


kflemin (approved):

kflemin:
@haneslinger, I forgot to mention, there is an uncaught error that occurs when a user has deselected everything from the chart (or when there is no data) and they try to add a label.

axelstudios (approved):

PR #3866, opened 2023-02-17
Fix Nginx Security

axelstudios:
In our Nginx config we have `add_header` directives at multiple levels, which is BAD - the directives are inherited from the previous level only if there are no `add_header` directives defined on the current level, so our higher level headers were being ignored

#### What's this PR do?
Moves all `add_header`s to the same level

Serve via Nginx, look for all headers


nllong (approved):
amazing

PR #3895, opened 2023-03-14
select cycles instead of start/end dates

### LINKED ISSUE #3624: Switch from start/end years to selecting cycles for compliance programs
Switch from compliance period in years (start and end) to selecting which cycles to include because some organizations have super cycles that they would want to exclude from the compliance period

kflemin:
# Fixes #3624 

 * Requires a database migration to be run
Screen Shot 2023-03-14 at 12 05 11 PM


axelstudios:
I added some fixes:
- Moved all local translation edits to Lokalise and pulled angular/python files
- Fixed selects from showing unavailable options
- Fixed permissions
- Removed delay before showing notifications

axelstudios (approved):

Issue #3919, opened 2023-03-22
Update Salesforce settings so they cannot be saved to the wrong org


isalanglois:

axelstudios:
BETTER & Salesforce configs have been fixed, but I found another page that can save data to the wrong and needs to be fixed: Program Setup

axelstudios:
This issue is done, but is currently in the `test` column (in case we still think that's necessary)

isalanglois:
@RDmitchell  Do you want to test this and close if no other issues arise? Swindler indicated it is complete

isalanglois:
@RDmitchell please test and close or indicate any other concerns

RDmitchell:
@isalanglois -- this isn't something I can test as I am not set up to deal with Salesforce. 

Maybe @kflemin can weigh in on how this can be tested?

axelstudios:
@RDmitchell There are some steps to test in the [PR that fixed this issue](https://github.com/SEED-platform/seed/pull/3925), but it's a bit technical. I'd be in favor of closing this since Katherine already tested the PR

RDmitchell:
@axelstudios -- that's fine. I'll close it.

Issue #3916, opened 2023-03-22
Review issues for verifying BETTER tokens


isalanglois:

axelstudios:
To add to this, it would be great if we could validate BETTER tokens without saving to the database first.  Even though there's an _info_ button with some helpful text about this, it can be confusing to enter a token and click verify only for it to fail unless you save between each attempt

axelstudios:
Another issue - we need to ensure that BETTER endpoints don't have a trailing slash, or you get an API endpoint like this that doesn't work: `https://better-lbnl-staging.herokuapp.com//api/v1`

PR #3925, opened 2023-03-22
Fix saving BETTER/Salesforce data for non-active orgs

axelstudios:
#### What's this PR do?
This PR fixes the ability to save BETTER tokens, Salesforce configs, and Salesforce mappings for a different organization than the currently active one.

Additionally, it fixes a bug with `sync_salesforce` where the `organization_id` was being passed in the POST body rather than the path params.

1. Go to the settings of an organization other than the active one in the top right corner
2. Attempt to verify a BETTER token
3. Attempt to save a Salesforce config and mappings
4. Check the database and confirm that the org ids match the correct org, and not the one in the top right corner

#3919
Related to #3916


kflemin (approved):
looks good!

Issue #3910, opened 2023-03-22
Fix weird download button on the Custom Reports Page.


isalanglois:
At the bottom right of the config that seems to cause errors.  If you're editing the very first report config it shouldn't exist.  It also needs a major UX overhaul

PR #3926, opened 2023-03-22
Fix custom reports download button

haneslinger:
Looks like a big change, but I just added a single div and indented the rest!
#3910
image


kflemin:
@haneslinger, could you add the button inside the "actions area instead of underneath? I think it would look better inside that little frame. thanks!

axelstudios:
@haneslinger I'm definitely glad that this is in the Actions box, but can we set a max width?  It looks crazy on an ultrawide monitor:
image

axelstudios (approved):
Thank you!  Looks great

Issue #3905, opened 2023-03-17
Error importing Meter data from ESPM


RDmitchell:
Instance: seedv2
Org: LBNL Test 200
Cycle: Meter Import Testing

There is an ESPM file that is generating an error when trying to import the meter data. 

It seems that the problem is that ESPM has changed the name of their "ID" from
* Property Id

to

* Portfolio Manager Property ID

I believe @axelstudios indicated that SEED is looking for a field called "Property Id" (notice the lower case "d") in the meter tab, and that is probably what is causing the problem. 

As explained in this doc, once I changed the field name in both tabs to "Property Id", I was able to import the meter data.

https://docs.google.com/document/d/1dNQygyF616CmbXtGwFNXtBg2nVwu3RDopAjimhxqqic/edit?usp=sharing

RDmitchell:
comment from user
*****
I tried changing it from Property ID to Property Id, and it worked!

Thanks for figuring that out. Would be great if SEED told us more information (i.e., more detailed error report) when there is a formatting error.
*****

RDmitchell:
Instance: dev1
SHA: c94869068 (2.18.0)

I am now able to import the original file from Open Tech (in the folder for this issue) that had the new ESPM field names in the meter tab, and the meter data imported without issue.

PR #3927, opened 2023-03-23
Updated the meter import process to work with latest format from EnergyStar Portfolio Manager

anchapin:
It appears that the headers from EnergyStar Portfolio Manager of the meter data export with multiple properties has been changed. The header `Property Id` has been changed to `Portfolio Manager Property ID`.

#### What's this PR do?
Added a check to see which version of the header is being used so both options work successfully. Additionally, the test data file, [example-pm-data-request-with-meters-new-format.xlsx](https://github.com/SEED-platform/seed/blob/e9bfa93fc2fd6e7f5b88334e9ee96d8d90e39c53/seed/data_importer/tests/data/example-pm-data-request-with-meters-new-format.xlsx), was updated to have the new header.

Attempt to import the [updated test data file](https://github.com/SEED-platform/seed/blob/e9bfa93fc2fd6e7f5b88334e9ee96d8d90e39c53/seed/data_importer/tests/data/example-pm-data-request-with-meters-new-format.xlsx) (import once as a spreadsheet to get the properties in SEED and then add the meters associated with that file) and ensure that the meters import correctly.

#3905


axelstudios (approved):
Looks good - for posterity, from the date in Robin's file this change may have happened sometime before 3/7/23, but not worth updating the comment

PR #3940, opened 2023-03-26
Modularized nginx, added gzip, added cache expiration

axelstudios:
We weren't using the `Cache-Control` header for static assets, which caused issues with assets being refreshed after deploying an update.

Better cache-busting solutions exist once we're using bundled assets, but this should help in the meantime.

The browser has been doing [heuristic caching](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#heuristic_caching), and was caching files based on the `Date` and `Last-Modified` headers: `(Date - Last-Modified) / 10`.  So if a file hasn't been modified for a year, and someone loads SEED, and then we push an update immediately after, their browser cache won't re-request the updated file for over a month.

#### What's this PR do?
- Enforces cache-control, requiring static assets to be cleared from the browser cache after one day
- Explicitly disables caching for API responses
- Modularizes Nginx configs so that we can reuse components like security headers and the uwsgi config
- Removes unused SSL config, that should only be present in the web-ssl nginx config
- Enables gzip
  - Before: `8.7 MB` of resources, `8.8 MB` transferred
  - After: `8.7 MB` of resources, `3.0 MB` transferred

- Check that responses have `Content-Encoding: gzip`
- Check that API responses have `Cache-Control: no-cache, no-store`
- Check that `/static/*` file responses have `Cache-Control: max-age=86400` and `Cache-Control: public`
- Check that every response, including error responses like 404s, includes security headers
- Check that the CSP header is returned exclusively for html responses
- Check that BETTER domains are allowed in frames

#3914


nllong (approved):
I built this container locally and tested a full seed workflow. All looked good!

nllong:
It looks like the files will cache expire after one day. Should we have a way to force the local cache to refresh on demand?

axelstudios:
> It looks like the files will cache expire after one day. Should we have a way to force the local cache to refresh on demand?

The only way to do that is with proper cache-busting, where index.html requests `bundle.js?t=1680130128` or similar, so that it knows it needs a new resource.  That needs to be one of the next improvements, possibly with [WhiteNoise](https://whitenoise.readthedocs.io/en/latest/) as well

nllong:
latest updates look good. I say merge!

Issue #3942, opened 2023-03-30
Fix filtering of property columns from tax lot tab


anchapin:
When trying to filter property columns from the tax lot tab of Inventory page a 500 error would occur.

Related to #3426

PR #3943, opened 2023-03-30
Fixed issue with filtering Property columns from Tax Lot tab of Inventory

anchapin:
Previously cross filtering was added for the Property tab of Inventory, but it wasn't also working for the Tax Lot tab.

#### What's this PR do?
Makes the code work in both scenarios.

Import data that includes both properties and tax lots, such as San Jose sample test data, and try filtering a Property column in the Tax Lot tab of the inventory.

#3942


haneslinger (approved):

axelstudios (approved):
Looks good!  No more 500

PR #3947, opened 2023-03-30
Fix Salesforce connection requests for non-superusers

axelstudios:
#### What's this PR do?
Testing a Salesforce connection only worked for _superusers_ and not _organization owners_ because the `organization_id` url param was missing.  This PR fixes the salesforce_connection requests.

1. Ensure that you aren't a superuser
2. Attempt to test a Salesforce config


kflemin (approved):
yay!!

PR #3949, opened 2023-03-31
Fix editing a derived column without changing the name

axelstudios:
Attempting to edit an already-saved column, without changing the name, always resulted in an error because _the name was already taken by itself_.

#### What's this PR do?
- Fixes updating an existing derived column when the name is unchanged
- Fixes `Derived Columns` active nav for all subpages
- Improves the `Create Derived Column` button to set the default type on the editor page based on whether it was clicked when viewing the Property or Tax Lot derived columns
- Improves the `Create`/`Save` button such that it takes you to the correct Property/Tax Lot derived column page on success

- Create a derived column, click Edit, then click Save.  It should now succeed
- Check that the `Derived Columns` nav shows as active while on any of the derived column pages
- Check that the `Create Derived Column` button initializes the Type field depending on which tab you were using when you clicked it
- Check that saving a derived column takes you to the correct tab where it appears

### DIFF
--- a/seed/static/seed/js/controllers/derived_columns_admin_controller.js
+++ b/seed/static/seed/js/controllers/derived_columns_admin_controller.js
@@ -26,10 +26,12 @@ angular.module('BE.seed.controller.derived_columns_admin', [])
       organization_payload,
       derived_columns_payload
     ) {
-
+      $scope.state = $state.current;
       $scope.auth = auth_payload.auth;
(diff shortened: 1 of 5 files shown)


nllong (approved):
Before:

image

After your changes, the derived column can be saved with the same name!
image

Nav stays active ✅
saving takes to correct tab ✅
Property/TaxLot Type is initialized correctly ✅

PR #3952, opened 2023-03-31
Make note event row height dynamic

perryr16:
NoteEvents are set up as a single line `ui-grid`. If a Note text exceeds the screen width it will be truncated with a '...'. (It is copyable with a triple click).

#### What's this PR do?
Attempts to make the ui-grid row height dynamic. Initial research suggests that `ui-grid` `rowHeights` cannot be set automatically ([reference](https://github.com/angular-ui/ui-grid/issues/3239)).

`rowHeight` can be dynamically set with a function.
```javascript
const rowHeight = 25 * Math.ceil(note.text.length / 180)
```
where 25 is the height of a single line and 180 is a rough estimate of number of characters per line on a 1440 px screen (13 inch MacBook). The function could be improved if screen width was taken into consideration.

#### Alternative Solutions 
Instead of working to get row height to be dynamic, we could post a link to the full note on the notes page.

Ultimately this is in need of a more elegant solution

Old
Screenshot 2023-03-31 at 11 37 11 AM

New
Screenshot 2023-03-31 at 10 56 12 AM


perryr16:
Needs further consideration

Issue #3946, opened 2023-03-30
BETTER Analysis -- Getting Unexpected Error Occurred message without any details


RDmitchell:
Instance: dev1
SHA; 8dc72bf0e

See this doc which explains my steps for setting up the BETTER analysis
https://docs.google.com/document/d/1IybGZFPqVlrKvyeTEfj-kRLZldLWU799tp3jEK1DufA/edit?usp=sharing

I believe I have the correct API on BETTER develop. 

The first time I tried, I didn't have Property Type mapped properly and got errors saying I didn't have one of the good property types, ie, Office.

I reimported the data, without deleting the old data, this time mapping "Property Type Self Selected" to the canonical field property_type (I think)

This time, the analysis failed but no information.
![image](https://user-images.githubusercontent.com/6314950/228981686-0c7ad771-eed9-4733-8f76-8618ec0b2344.png)

RDmitchell:
Starting over with a completely new organization, I was able to get the BETTER runs to complete. So not sure what is wrong with the original org (see the doc linked above). Maybe because I imported the meter data twice and when I deleted the properties the meter data didn't get properly deleted? No idea. 

So all 40 properties ran, but I did get this message, which doesn't make sense, because I have 40 properties (so > 30) even though 7 of them said "No reasonable change point model could be found", that would still be 33, which is > 30
![image](https://user-images.githubusercontent.com/6314950/228989507-f48c8573-1a7b-40da-8372-0699de495401.png)

RDmitchell:
But when I click on a Run ID, I am only seeing the Portfolio results, and not also the individual building results -- I thought in the previous iteration of BETTER, when you clicked on a run ID for a specific building, you would see the Portfolio Report first and then the individual report for that building after that, in the same report. 
![image](https://user-images.githubusercontent.com/6314950/228990868-3969dc09-dfb3-4b0d-8b59-24911dbace89.png)

RDmitchell:
And when I click on the "Download link" for a specific building in the Portfolio report, I get a gray screen.
I think that might already be on the Bug Party spreadsheet ?? I think it's Line 62 in that spreadsheet. 

![image](https://user-images.githubusercontent.com/6314950/228991161-9cdcf9bf-bb20-4177-9ef9-eb2c4ddaa4c5.png)

RDmitchell:
See #3891

perryr16:
[...]

PR #3953, opened 2023-03-31
Validate property state before creating BuildingSync document

perryr16:
To generate a BETTER analysis SEED first constructs a BuildingSync document to be used as input for a BETTER analysis. In the creation of this document, the function calls on a given PropertyStates city, state, and postal_code.

There are currently guards against running a BETTER analysis if the PropertyState is missing city or state, but not postal_code

#### What's this PR do?
validates the property states geographic info before creating a buildingsync document.

Run an analysis with no postal code. The error message should be informative.

#3946 

Screenshot 2023-04-04 at 2 02 16 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,46 @@
+# SEED Version 2.18.0
+<!-- Release notes generated using configuration in .github/release.yml at develop -->
+
+## What's Changed
+### New Features 🎉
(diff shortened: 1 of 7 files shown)


RDmitchell:
@perryr16 -- thanks !!

axelstudios (changes requested):
@perryr16 Can we move the error checking to be with the other checks to get all of the errors at once?

axelstudios (approved):

PR #3988, opened 2023-04-20
Fix column list profile GET request syntax

ebeers-png:
#### What's this PR do?

Fixes a syntax bug preventing inventory_service.get_column_list_profile from working correctly.

It's used on the inventory list page when first loaded; without the fix, one of the GET requests will fail with a 301 error.


ebeers-png:
Found another one, figure it might as well go here!

PR #3990, opened 2023-04-21
Improve inventory load times

ebeers-png:

The inventory of an organization with 4000 columns can take a frightfully long time to load (12-15 seconds). This is mainly due to how utils.inventory_filter.get_filtered_results uses models.Column.retrieve_all to fetch every single column in the organization, no matter the profile used.

#### What's this PR do?

I added a parameter for column IDs to retrieve_all. get_filtered_results now queries for the column list profile first, and then uses those column IDs to narrow the search that retrieve_all performs. This way, users see better load times when using a profile that has fewer columns; it's also cut down on the amount of time that even a profile with many columns takes to load. On our server, it's cut down the query time for an organization with 4000 columns by ~10 seconds.

The load times are best compared on a server!


nllong:
@axelstudios  -- I think that the Salesforce tests are failing because this is coming from a fork of SEED. Forks don't have access to all the env vars because it could be a malicious fork where they print env vars to get out secrets.

axelstudios:
> @axelstudios -- I think that the Salesforce tests are failing because this is coming from a fork of SEED. Forks don't have access to all the env vars because it could be a malicious fork where they print env vars to get out secrets.

Confirmed, local testing looks good:

axelstudios (approved):

PR #4012, opened 2023-04-27
Add new `create_property` and `update_with_espm` endpoint

kflemin:
Create a new "Create property" endpoint which creates a property, propertystate, & propertyview via API.  

Needed for 179d web application.


axelstudios (changes requested):
@kflemin I pushed up several fixes for unhandled errors, but there are TONS that remain:

There's no validation for the state object beyond the basic serializer - the user can pass in anything:
- If the state's `organization` and `organization_id` reference a different org than the one used for the query parameter it will accept it regardless, and the state will be inserted for the wrong org
- Should users be able to pass in an entire `extra_data` json object?  This seems like a vulnerability since they could completely exhaust the database storage in no time with massive json objects
- There are tons of state fields the user should _not_ be able to input:
  - `id`
  - `measures`
  - `scenarios`
  - `files`
  - `import_file_id`
  - `organization_id`
  - `data_state`
  - `merge_state`
  - `normalized_address`
  - `long_lat`
  - `centroid`
  - `bounding_box`
  - `hash_object`
  - `created`
  - `updated`
  - `import_file`
  - `organization`
  - and certainly more

docs/source/translation.rst
```diff
@@ -3,7 +3,14 @@ Translating SEED
 
 1. Update translations on `lokalise`_.
 
-2. Copy lokalise.cfg.example to lokalise.cfg. Update API token.
+2. Copy lokalise.yml.example to lokalise.yml. Update API token.
```
nllong, line 6: our lokalise instructions were out of date

seed/tests/test_account_views.py
```diff
@@ -98,13 +97,18 @@ def test_dict_org(self):
             'at_organization_token': '',
             'audit_template_user': '',
             'audit_template_password': '',
-            'at_host_url': 'https://api.labworks.org',
             'salesforce_enabled': False,
             'ubid_threshold': 1
         }
 
         org_payload = _dict_org(self.fake_request, [self.org])
+
```
nllong, line 105: if in your `local_untracked` you have set your own better and AT URLs, then this test fails locally. So let's just check to make sure that it matches a regex.

seed/tests/test_portfoliomanager.py
```diff
@@ -87,7 +88,7 @@ def test_template_list_interface_no_username(self):
         # status should be error
         # message should have "missing username"
         self.assertEqual(400, resp.status_code)
-        data = json.loads(resp.content)
+        data = resp.json()
```
nllong, line 91: I replace these when I see them.

nllong (approved):
@axelstudios  -- this is ready for another review.

nllong:
> @kflemin I pushed up several fixes for unhandled errors, but there are TONS that remain:
> 
> There's no validation for the state object beyond the basic serializer (and I don't think we'd even want to use that serializer for validation since it requires 17 fields, many of which I wouldn't expect to be relevant like `organization_id` and `total_ghg_emissions_intensity`) - the user can pass in anything:
> 
> * If the state's `organization` and `organization_id` reference a different org than the one used for the query parameter it will accept it regardless, and the state will be inserted for the wrong org
> * Should users be able to pass in an entire `extra_data` json object?  This seems like a vulnerability since they could completely exhaust the database storage in no time with massive json objects
> * There are tons of state fields the user should _not_ be able to input:
>   
>   * `id`
>   * `measures`
>   * `scenarios`
>   * `files`
>   * `import_file_id`
>   * `organization_id`
>   * `data_state`
>   * `merge_state`
>   * `normalized_address`
>   * `long_lat`
>   * `centroid`
>   * `bounding_box`
>   * `hash_object`
>   * `created`
>   * `updated`
>   * `import_file`
>   * `organization`
>   * and certainly more

1. This should not be a problem. The org_id is validated on the endpoint.
2. We could enforce only columns that have been added to an org to be allowed in extra_data?
3. The Write Serializer protects all of the "Column.EXCLUDE_.*" objects.

seed/lib/xml_mapping/mapper.py
```diff
@@ -56,20 +56,24 @@ def get_bae_mappings():
     # units field name is the same with " Units" appended.
 
     bsync_assets = BAE.get_default_asset_defs()
-    for item in bsync_assets:
+    for asset in bsync_assets:
+        if isinstance(asset, dict):
```
nllong, line 60: This should never happen with the latest BAE code. @axelstudios  - you probably just needed to run `pip install -r requirements/local.txt`

axelstudios (approved):

axelstudios:
Here are the changes I've made:
- Removed broken `property_id` option
- Fixed extra_data columns to only ones that actually apply
- Changed `merge_state` from 0 to 1 (new)
- Added ValueError handling
- Small cleanup

PR #4020, opened 2023-05-03
Store additional geocoding results

nllong:
During geocoding, there are several additional fields that are returned that can be useful such as neighborhood. These results are now stored in reserved SEED columns (e.g., `Geocoded Neighborhood`, `Geocoded City`, etc.).

#### What's this PR do?
* Fix the pre-commit file to not throw a warning about regexes
* When calling "geocode" persist off the additional data that is collected
* `get_or_create` the additional columns as needed and prevents those columns from being mappable

I had to create a new MapQuest API key since their auth systems seems to have been upgraded. I did need to provide a credit card even for the free account. I used `privacy.com` with a new credit card set with $1/year limit... just in case.

* Create new MapQuest API Key
* Put key into the settings
* Import some addresses with only address and state
* Verify that the additional geocoded data are automatically created

n/a


kflemin:
@nllong verified as working!  had to get off the VPN and refresh the view to "show only populated columns" but then I could see the new geocoded results. 👍👍

Screen Shot 2023-06-20 at 10 00 24 PM

seed/utils/geocode.py
```diff
@@ -198,7 +243,12 @@ def _address_geocoding_results(id_addresses, mapquest_api_key):
 
         response = requests.get(request_url)
         try:
```
kflemin, line 245: One thing: the UI hangs if on VPN...is there a way to catch and handle the SSL error here? 
from the log, error was: 
File "/Users/kflemin/repos/seed-TESTING3/seed/data_importer/tasks.py", line 1379, in _geocode_properties_or_tax_lots
    geocode_buildings(property_state_qs)
  File "/Users/kflemin/repos/seed-TESTING3/seed/utils/geocode.py", line 91, in geocode_buildings
    address_geocoding_results = _address_geocoding_results(id_addresses, mapquest_api_key)
  File "/Users/kflemin/repos/seed-TESTING3/seed/utils/geocode.py", line 244, in _address_geocoding_results
    response = requests.get(request_url)
  File "/Users/kflemin/.pyenv/versions/newseedenv/lib/python3.9/site-packages/requests/api.py", line 73, in get
    return request("get", url, params=params, **kwargs)
  File "/Users/kflemin/.pyenv/versions/newseedenv/lib/python3.9/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/kflemin/.pyenv/versions/newseedenv/lib/python3.9/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
  File "/Users/kflemin/.pyenv/versions/newseedenv/lib/python3.9/site-packages/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
  File "/Users/kflemin/.pyenv/versions/newseedenv/lib/python3.9/site-packages/requests/adapters.py", line 517, in send
    raise SSLError(e, request=request)
requests.exceptions.SSLError: HTTPSConnectionPool(host='www.mapquestapi.com', port=443): Max retries exceeded with url: /geocoding/v1/batch?&inFormat=json&outFormat=json&thumbMaps=false&maxResults=2&json=%7B%22locations%22:%20%5B%7B%22street%22:%20%2215013%20Denver%20West%20Parkway,%20Colorado%22%7D,%20%7B%22street%22:%20%228604%20Sunset%20Blvd,%20California%22%7D,%20%7B%22street%22:%20%221222%20Hunstpoint%20Way,%20Kentucky%22%7D%5D%7D&key=XXXXXX (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1123)')))

kflemin (changes requested):

axelstudios (approved):

RDmitchell:
@nllong -- I am curious why this feature was implemented ... I guess it's good for the user to see the values used for geocoding, but is there more to it than that? 

Also, I didn't have to create a new Mapquest key .. ??

nllong:
If a user doesn't have all the geocoded data, for example they are missing city, then the city will be returned and stored in the `geocoded_city` location. In brief, there is more data in the geocoding that the user might want to have.

PR #4030, opened 2023-05-08
Add ranked sorting to insight property page

haneslinger:
#### What's this PR do?
Adds a hard coded x-axis option, "Ranked" which ranks non-compliant propertyby their non compliance. 
image


seed/static/seed/js/controllers/insights_property_controller.js
```diff
@@ -290,33 +290,49 @@ angular.module('BE.seed.controller.insights_property', [])
...
-            // don't add whisker if data is in range for that metric or it looks bad
-            let add = false
-            metric_type = $scope.configs.chart_metric === 0 ? $scope.data.metric.energy_metric_type : $scope.data.metric.emission_metric_type;
-            if (item['x'] && item['y'] && item['target']) {
-              if ((metric_type === 1 && (item['target'] < item['y'])) || (metric_type === 2 && (item['target'] > item['y']))) {
-                add = true
-              }
-            }
+        // Rank
+        if($scope.configs.chart_xaxis === "Ranked"){
```
haneslinger, line 301: above, L250, we're calculating x, then recalculating here should it be ranked. idk how to only do it once, as  the normal x value must be calculated in the for loop so we have access to the property, and the ranked must happen _after_ so it can sort. any thoughts?

kflemin (approved):
looks great in chrome now!

RDmitchell:
@haneslinger -- I guess I'm being stupid, but I don't understand what the ranking is based on ? Here is an example from the Test BPS org
image

haneslinger:
lol, I had to work with @isalanglois to figure out exactly what was being asked for here. 

It the properties list where the first is the fartherest from compliance and the last is the closest. 

Its not the most intuitive. Any ideas how we can make it better?

RDmitchell:
@haneslinger -- I thought it might be something like that, but then I was confused by the left most point, because the "Distance to Target" line wasn't being displayed, maybe because it's sitting right on the axis. 

I'm not sure how else what is being asked for could be shown. 

It's probably one of those things that we need to get feedback from the users who wanted it to see if it meets their expectations. 

Thanks !!

Issue #4029, opened 2023-05-08
UBID backend CRUD features


perryr16:
The first step in issue #3957 

The UBID model will be a one to many with PropertyStates (and TaxlotStates). A property can have multiple UBIDs, but a UBID could also have multiple properties (addresses). This doesn't need to be many to many as we don't mind if there are duplicate UBIDs with different property names.

fields:
- ID
- UBID
- property
- taxlot
- preferred boolean

RDmitchell:
@perryr16 -- has this issue been resolved? Can it be closed? If not, we should remove the Stale label.

PR #4037, opened 2023-05-12
Add ubid model backend crud functionality

perryr16:
The PropertyState (and TaxLotState) both have a `ubid` field that aligns with a canonical column. Per some new guidelines States can have multiple ubids. To accommodate this a new model `UbidModel` has been created with a many-to-one relationship with PropertyState and TaxlotState. Note that the name of the new model could not be `ubid` to prevent a naming conflict with the existing field.

A UbidModel will have the following fields:
- ubid: string 
- preferred: boolean 
- property: ForeignKey
- taxlot: ForeignKey

#### What's this PR do?
- Incorporates the new UbidModel with backend CRUD functionality. 
- Generates UbidModels when PropertyStates or TaxLotStates have been created/updated using a post_save signal.

#### Known issues 
PropertyState.ubid is one of the matching criteria (ubid, pm_property_id, custom_id_1, and normalized_address). If a user were to upload a file where all fields matched then it would update the propertyView instead of creating a new one. However now that PropertyStates can have multiple UBIDs the matching condition fails if an upload has matching PM Property IDs, but different UBIDs. Should UBID be removed as a matching criteria?

#4029


perryr16:
Functionality included in #4074

PR #4057, opened 2023-05-23
Prevent Automatically Serving Static `index.html` Files

axelstudios:
#### What's this PR do?
1. Prevents automatically serving `/static/**/index.html` files
2. Treats every `/static/**` file, except for js/css/svg, as binary files

1. Browse to `/static/node_modules/angular-dragula/` and make sure the dragula readme page doesn't appear (403 error)
2. Browse to `/static/node_modules/angular-dragula/index.html` and make sure the file downloads rather than rendering and loading js files that we don't want executing
3. Monkey test, make sure the app itself still works as expected and that images load and such

#4039

### DIFF
--- a/docker/nginx/nginx.conf
+++ b/docker/nginx/nginx.conf
@@ -145,6 +145,12 @@ http {
 
         location /static {
             alias /seed/collected_static;
+            types {
+                text/css css;
+                application/javascript js;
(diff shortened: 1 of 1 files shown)


kflemin (approved):
🔥

PR #4066, opened 2023-05-30
Fix Download URLs

axelstudios:
The import-file download buttons on the Dataset Detail page weren't working for any files stored in directories with more than one level of nesting from the media directory.

For example, a PM import stored in `/seed/media/uploads/pm_imports/pm_import_2023-05-30.csv` was incorrectly linking to `/api/v3/media/pm_imports/pm_import_2023-05-30.csv` (missing the `uploads` directory).

#### What's this PR do?
Fixes all download links regardless of nesting and mixed slashes

1. Upload a file normally, and import a file from ESPM
2. With nginx running, check that the download links for both files work as expected


kflemin (approved):
🏆

PR #4068, opened 2023-05-31
Init Accountability Hierarchy

haneslinger:
- Add treebeard
- Add ALI table
  - Give ALI automatically updated path (To be used later)
- Add ALN & ALIT to Orgs
  - Add org property “root” 
  - Add presave to Org that gives it a root
- Add non-nullable ALI to User
- Add non-nullable ALI to Property and Taxlot
  - Add presave to Property and Tablet setting their ALI to root

- Migration so that 
	- all orgs are given root
	- all users, Properties and Tablets are in root

- Add tree page to the org page
- Add AccessLevelViewSet
- set requests ALI in `has_perm_class` (not used yet)
- tests

image
image


haneslinger:
@kflemin @axelstudios The one failing test passes locally 😿 plz review.

haneslinger:
I think this breaks adding an existing user to a new org as the ALI isn't defined. will add soon

kflemin (approved):

haneslinger:
@kflemin might you know why the test is failing? It's not failing for me locally.

Issue #4069, opened 2023-05-31
UBID: validate on creation, decode on update


perryr16:
- When a user creates a ubid, use the sql functions to validate the incoming ubid. If it fails provide an error message 
- any time the preferred ubid changes, decode it (which will update the properties centroid, bounding_box,...)

RDmitchell:
@perryr16 -- can this issue be closed?

Issue #4051, opened 2023-05-19
UBID: Update match merge functions to use a ubid jaccard index threshold to evaluate equivalence


perryr16:
The current process checks for an exact match between incoming and existing ubids. Now that we have Jaccard Index, use a setpoint (ex: 0.44) to evaluate equivalence

RDmitchell:
@perryr16 -- can this be closed, or moved to the Q4 project?

RDmitchell:
@perryr16 -- can this issue be closed?

RDmitchell:
@perryr16 -- actually I see that status = Review, so No to closing !!

Issue #4041, opened 2023-05-12
UBID: Add CRUD frontend


perryr16:
From the [design doc](https://docs.google.com/document/d/1F14NrXYQi-3jZt9RERZaFuRhClzF-XOfuYvFI-aYf6s/edit)
>Frontend (est. 2 weeks):  The second step is to make a frontend that supports the new UBID backend functionality.  CRUD will have to be implemented to manage UBIDs on properties as well as select the primary UBID, probably on the inventory detail page.  Primary UBIDs will have to be made available pretty simply, being listed on the inventory view or displayed on a summary.  Secondary UBIDs will also have to be listed but might not be necessary on the list view or similar where only pertinent info is displayed.

RDmitchell:
@perryr16 -- I am taking the Stale label off this, but maybe it can be closed? Your call !!

RDmitchell:
@perryr16 -- is this still an issue?

perryr16:
Yes, this has been completed

Issue #3957, opened 2023-04-03
UBID design


kflemin:
plan the UBID work for Q3

some notes:

1. add in the SQL commands for UBID functionality
2. refactor UBID and deprecate ULID fields
	1. there can now be multiple UBIDs on a property. one will be labeled "main ubid" or something like that
	2. can have UBIDs on parcels
	3. new "UBID" model in db
		1. need to be able to manage UBIDs on my property
		2. which one do I want to promote as the "main ubid"
		3. import a set of buildings that have ubids, I want to match on ubids, how do I do that? (not a string match, probabilistic since we have multiple ubids. need a 'threshold' for matching ubids)

RDmitchell:
@perryr16 -- should any of this be tested in the UI?

perryr16:
I thought the new UBID UI work had been tested, but I defer to you.

RDmitchell:
I haven't tested it, and don't really know what the changes are. Is it documented anywhere (a spec or something) what changes were made? 

Have the changes been deployed to dev1 or other server for testing?

RDmitchell:
@perryr16 -- is there a spreadsheet or CSV file that has multiple ubids defined for one property that I can use to test?

perryr16:
@RDmitchell I used these test files to test the 'preferred' ubid logic. Each spreadsheet has the same 2 properties with their ubid slightly changed. 
[tf_4029_2_0.xlsx](https://github.com/SEED-platform/seed/files/14842988/tf_4029_2_0.xlsx)
[tf_4029_2_1.xlsx](https://github.com/SEED-platform/seed/files/14842989/tf_4029_2_1.xlsx)
[tf_4029_2_4.xlsx](https://github.com/SEED-platform/seed/files/14842990/tf_4029_2_4.xlsx)
[tf_4029_2_5.xlsx](https://github.com/SEED-platform/seed/files/14842991/tf_4029_2_5.xlsx)
Let me know if you're looking for something different

RDmitchell:
@perryr16 -- thanks ... these should work !

RDmitchell:
instance: dev1
SHA: 
Org: LBNL 245
For the case above, where there were 2 UBIDs separated by a semicolon, the program just imported all the data and is showing it as one UBID. It didn't generated an error
![image](https://github.com/SEED-platform/seed/assets/6314950/32e8016d-9f18-4509-a422-cda117597084)

I think this is a pretty obscure use case, and I don't think we need to support this functionality.

[...]

PR #4070, opened 2023-06-01
Add Ubid frontend features

perryr16:
Now that UBIDs can be compared via the jaccard index, the matching and merging processes must be updated to use a threshold value instead of an exact match. 

#### What's this PR do?

- Generates a new Organization setting for a UBID threshold. The threshold is an org setting that must be between 0 and 1.0. If the jaccard index between an incoming ubid and an existing ubid is above the UBID threshold it will qualify as a match (if the other matching criteria are met)

- Uses signals to generate UbidModels when a new property or taxlot is uploaded with a ubid

- Uses signals to update -State.ubid if a saved UbidModel's "preferred" field is set to true

- Adds frontend CRUD. A new ubid viewer and editor are available from the inventory list actions, inventory detail actions, or under a new UBID inventory detail nav button. The structure is loosley based on the derived column admin and editors. New or updated ubids are validated before they may be saved.

- Adds a ubid jaccard comparison modal to the inventory list actions. It can be triggered when 0-2 properties are selected. If fewer than 2 properties are selected a user may use custom ubids for comparison.

- Renames the ubid_modal to ubid_decode_modal, which allows users to decode a property's ubid

- Allows properties to be merged together while preserving the ubid history.

Run tests, test for functionality.

#3957 
#4051
#4069
#4041 

Screenshot 2023-06-01 at 3 13 56 PM
Screenshot 2023-06-01 at 3 10 26 PM
Screenshot 2023-06-01 at 3 11 38 PM
Screenshot 2023-06-01 at 3 11 09 PM


perryr16:
See PR #4074

Issue #4053, opened 2023-05-19
Multiple Cycle Core Functionality Back End


anchapin:

RDmitchell:
@anchapin -- this is in the Q3 project, labeled Ready to Deploy.

Should it be moved to the Q4 project, with the same label? 

Or closed if it has already been depoyed? 

Thx

anchapin:
@RDmitchell thanks for the reminder. I just moved this and the other issues into Q3 Test.

RDmitchell:
Instance: seeddemostaging
SHA: 22e66f350
Org: LBNL 303

anchapin -- I have tested the front end functionality, and it seems to work well. 

I guess others need to check the backend functionality before it can be closed?

isalanglois:
@anchapin Please provide update on this ticket

Issue #3960, opened 2023-04-03
Implement cross-cycle data import


kflemin:
Allow users to import data from multiple cycles within a single spreadsheet.

RDmitchell:
Instance: seeddemostaging
SHA: 22e66f350
Org: LBNL 303

I tested this with the Sample_MultipleCycles.xlsx file and it worked great !

Issue #3181, opened 2022-03-22
Support importing multiple years of ESPM data from single file (not meters)


laurenliz22:
How they can best import their data from ESPM. They also mentioned importing multiple years in the same file, which sounded like a similar request that we received from Better Buildings.

nllong:
I think this is a request from Better Buildings and ACE. Revisit when those projects start.

RDmitchell:
I took the BETTER label off because it relates to the Better Buildings use case not the BETTER analysis.

nllong:
correct, thanks @RDmitchell

RDmitchell:
Instance: seeddemostaging
SHA: https://github.com/SEED-platform/seed/commit/22e66f350d3a054428ecedcbd37a664d915bac39
Org: LBNL 303

I tested this with the Sample_MultipleCycles.xlsx file and it worked great !

RDmitchell:
@anchapin / @laurenliz22 

I tested with a pretty simple file. 

It might be good to get a real file from Better Buildings or some other source (CERL) with a lot of records and a lot of different Year Ending values to do more robust testing.

PR #4073, opened 2023-06-02
Multiple cycle upload

anchapin:
#### What's this PR do?
Allows you to upload a spreadsheet with data across multiple cycles.

Upload the file '[Sample_MultipleCycles.xlsx](https://github.com/SEED-platform/seed/blob/multiple_cycle_upload/seed/tests/data/Sample_MultipleCycles.xlsx)' and select the 'Multiple Cycle Upload' checkbox.

MultipleCycle Checkbox

Additionally, ensure that the SEED column 'year_ending' is mapped to the "Year Ending" column from the 'Date File Header'.

#3960 #4053 #3181


haneslinger (changes requested):

haneslinger:
Just documenting a known problem: Should a property's year ending not fit into any existing cycles, the API 500s and the UI hangs.

haneslinger:
idk if this is intentional, but If you selected multi-cycle and a cycle, it will only upload the properties that are part of the selected cycle

image


```
>>> PropertyView.objects.values("cycle__name")
<QuerySet [{'cycle__name': '2018 Calendar Year'}]>
```

seed/data_importer/match.py
```diff
@@ -72,14 +125,17 @@ def match_and_link_incoming_properties_and_taxlots(file_pk, progress_key, sub_pr
...
     """
+
     from seed.data_importer.tasks import pair_new_states
 
     import_file = ImportFile.objects.get(pk=file_pk)
-    progress_data = ProgressData.from_key(progress_key)
     update_sub_progress_total(100, sub_progress_key)
 
+    progress_data = ProgressData.from_key(progress_key)
+
```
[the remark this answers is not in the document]
anchapin: I wasn't really sure how to handle the `progress_data` related pieces of the code. I ended up moving this and the associated other lines of code into the `_by_cycle` method because I was having a hard time figuring out how to pass the `merged_linked_property_views` and `merged_linked_taxlot_views` back to the main method.

seed/data_importer/tasks.py
```diff
@@ -1580,7 +1620,7 @@ def _get_field_from_obj(field_obj, field):
 
 
 @shared_task
-def _map_additional_models(ids, file_pk, progress_key):
+def _map_additional_models(ids, file_pk, progress_key, cycle=None):
```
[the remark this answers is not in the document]
anchapin: What would you suggest? Should I make a helper function again like the `match_and_link_incoming_properties_and_taxlots` function in 'match.py'?

seed/data_importer/tasks.py
```diff
@@ -1603,28 +1643,62 @@ def _map_additional_models(ids, file_pk, progress_key):
...
+        progress_data.total = (
+            1  # geocoding
+            + len(id_chunks)  # map additional models tasks
+            + 2  # match and link
+            + 1  # finish
+        )
+        progress_data.save()
+
+        for id_chunk in id_chunks:
+            property_states = PropertyState.objects.filter(id__in=id_chunk).prefetch_related('building_files')
```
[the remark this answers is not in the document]
anchapin: These chunks were existing and I wasn't sure how to check whether it was doing anything or not. Any suggestions?

seed/data_importer/match.py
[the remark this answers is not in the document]
haneslinger: I mean just this line moving from L80 (now L131) to here?

seed/data_importer/tasks.py
```diff
@@ -1334,7 +1334,21 @@ def geocode_and_match_buildings_task(file_pk):
         # Loop through the property_state objects.
         for property_state in property_states:
             # Find the cycle that corresponds with property_state year_ending.
-            cycle = Cycle.objects.filter(end__year=property_state.year_ending.year, organization_id=property_state.organization_id).first()
+            # _log.error( '+++++++++++++++++++++')
+            # _log.error(property_state.year_ending.year)
+            # _log.error(property_state.year_ending.month)
+            # _log.error(property_state.year_ending.day)
```
haneslinger, line 1340: 🧽

haneslinger (approved):
sweet!

seed/views/v3/import_files.py
```diff
@@ -627,23 +627,21 @@ def start_save_data(self, request, pk=None):
                     pk)}, status=status.HTTP_400_BAD_REQUEST)
 
         cycle_id = body.get('cycle_id')
+        multiple_cycle_upload = body.get('multiple_cycle_upload', False)
```
haneslinger, line 630: I would think `cycle_id == "year_ending"` would be enough to commucate we are doing multicycle, is there a reason we  need `multiple_cycle_upload` too?

seed/migrations/0196_remove_time_from_cycle.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 3.2.19 on 2023-06-09 13:43
```
haneslinger, line 1: This was tested on a db with cycles in it, right? I'm surprise it worked.

seed/data_importer/tasks.py
```diff
@@ -1325,20 +1323,58 @@ def geocode_and_match_buildings_task(file_pk):
...
+                cycle = Cycle.objects.filter(
+                    end__gte=property_state.year_ending,
+                    start__lte=property_state.year_ending,
+                    organization_id=property_state.organization_id
+                ).first()
+            # Check if cycle is none
+            if cycle is None:
+                cycle = Cycle.objects.get(
+                    pk=import_file.cycle_id
+                )
```
haneslinger, line 1349: Isn't there a good chance this is "year_ending"?

seed/data_importer/tasks.py
```diff
@@ -1701,6 +1740,10 @@ def pair_new_states(merged_property_views, merged_taxlot_views, sub_progress_key
 
     sub_progress_data.step('Pairing Data')
     view = next(chain(merged_property_views, merged_taxlot_views))
+    # Check if 'view' is an array, then
+    # use the first element, otherwise use 'view' as is
+    if isinstance(view, list):
+        view = view[0]
```
haneslinger, line 1746: Can you explain why this was added?

seed/data_importer/tasks.py
```diff
@@ -1768,6 +1811,12 @@ def pair_new_states(merged_property_views, merged_taxlot_views, sub_progress_key
         # if pv.state.lot_number and ";" in pv.state.lot_number:
         #     pdb.set_trace()
 
+        # Check if 'pv' is an array, then
+        # use the first element, otherwise use 'pv' as is
+        if isinstance(pv, list):
+            if len(pv) > 0:
+                pv = pv[0]
+
```
haneslinger, line 1819: Can you explain why this was added?

haneslinger (commented):
couple of questions.

seed/data_importer/tasks.py
anchapin: This is no longer necessary. Thanks for reminding me to take this out.

seed/data_importer/tasks.py
anchapin: This is no longer necessary. Thanks for reminding me to take this out.

seed/views/v3/import_files.py
anchapin: Now that we are letting the users set a default cycle I'm using cycle_id to save the default cycle. Therefore, I decided to add a new boolean field to import_file to save the fact that we are doing multicycle. I'm open to suggestions though.

seed/data_importer/tasks.py
anchapin: No, this will be the default cycle now. I'm no longer using "year_ending" as a flag for cycle_id.

anchapin:
Note that tax lots don't work (in that they only get added to the first cycle listed in the upload file) but it doesn't break SEED. This seems sufficient for now since the first known customer for this feature indicated that they aren't planning to use tax lots initially.

There is an issue with "Year Ending" just being a year value like "2019" though. It won't find a matching cycle and always use the default cycle.

axelstudios (approved):

RDmitchell:
@anchapin -- does it matter what the default cycle is set to when the Multiple Cycle box is checked? 
image

RDmitchell:
@anchapin -- I think never mind to the default cycle question -- it appears that is the cycle used if the Year Ending field is blank.

RDmitchell:
@anchapin -- so I guess the date in the Year Ending field just has to be equal to or between the dates defined in the cycle, right? so the record with year ending of 1/1/2021 would be associated the the 2021 cycle if the cycle were defined as 01/01/2021 - 12/31/2021, correct? That seems to be the way the program is interpreting it. Just need to make sure I understand the mechanics so I can document it. Thx

ebeers-png:
What is the expected result when there are cycles with overlapping start/end ranges, and a property could be associated with either one? Also, is there any plan to enable this feature for taxlots as well? BEAM just merged in SEED 2.19.0, so if either of these questions were addressed in a later version, let me know!

haneslinger:
Hey @ebeers-png! Great question. The short of it is there's alot to fix up about this feature, but it's been lost under and flood of other requests 🤷🏻 

 As of current, should multiple cycles apply, we simply use the first. ([code](https://github.com/SEED-platform/seed/blob/Create-technologyLibrary/seed/data_importer/tasks.py#L1614))
 
 Also, while enabling this feature for taxlots should be a matter of strategic coping and pasting, it just hasn't been done

haneslinger:
@kflemin

kflemin:
@ebeers-png @haneslinger I created a new issue for implementing this for taxlots: https://github.com/SEED-platform/seed/issues/4781

As for the cycle selection, we currently just selected the first cycle returned that matches the criteria for start-date/end-date. We are open to suggestions for improving this functionality though...feel free to open a ticket and we'll take a look!

PR #4075, opened 2023-06-02
INclude x and y in property insights tables

haneslinger:
image


RDmitchell:
It seems like it might make sense to add the text from the 2 axis to the column names, so in the example above the "X" column header would be something like "X: Latitude" and the Y column would have the header "Y: Site EUI <units>"

perryr16 (approved):
Works as expected!

PR #4085, opened 2023-06-08
Fix categorical data on reports scatter plot

haneslinger:
image


kflemin (commented):
looks great, works great!

kflemin (approved):

PR #4094, opened 2023-06-09
Fix User Creation Bugs

axelstudios:
#### What's this PR do?
Fixes multiple bugs:
1. It was not possible to use the Admin `Create a user` form when also specifying a new organization to create
2. The same `Create a user` form failed to update the organization list after creating a new organization
3. The `create_default_user` script failed to add new users to the specified organization if it already existed, leaving the new user orphaned

1. Check that the `Create a user` form works as expected when specifying a new organization name, and then check the organization dropdown at the top right for the newly created org
2. Check that an organization owner (non-superuser) can still invite new members to their organization
3. Run `manage.py create_default_user`, specifying and existing organization name, and check that the new user was added to the organization as an owner

#4095

### DIFF
--- a/seed/lib/superperms/orgs/decorators.py
+++ b/seed/lib/superperms/orgs/decorators.py
@@ -32,6 +32,14 @@ def requires_parent_org_owner(org_user):
     )
 
 
+def requires_owner_or_superuser_without_org(org_user):
+    """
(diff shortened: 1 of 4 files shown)


kflemin (approved):
this worked for me!

Issue #4076, opened 2023-06-05
ESPM Imports Have Broken Source Types


axelstudios:
**Describe the bug**
The `source_type` column in the `data_importer_importfile` table incorrectly shows a tuple, and the wrong tuple at that:
![image](https://github.com/SEED-platform/seed/assets/411466/2f9d9177-1e7d-4d1f-8f2b-7d1bde2f1720)

Worse yet, all ESPM imports in production have the wrong source_type, and it varies:
![image](https://github.com/SEED-platform/seed/assets/411466/92447fa7-1425-458b-9844-0cb3abf4e61e)

**Expected Behavior**
The value should be `Portfolio Raw`.  Eventually we want to migrate this column to an unsigned type for enum strictness

**Actual Behavior**
The value is the tuple for `Assessed Raw`

**Steps to Reproduce**
Import a file from ESPM

nllong:
No testing needed, database update only.

PR #4096, opened 2023-06-13
Fix ESPM Source Types

axelstudios:
ESPM imports have never used the correct source type:
- The `data_importer_importfile` table should always show `Portfolio Raw` in the `source_type` column when `source_program` is `PortfolioManager`.
  - Instead, due to an enum order issue, it has always been `Assessed Raw` or even worse `(2, 'Assessed')`.
- The `seed_propertystate` table should always show initial import records tied to ESPM imports as `source_type=1` (`PORTFOLIO_RAW` for extra_data only) and `source_type=3` (`PORTFOLIO_BS` for mapped imports).
  - Instead, due to the same enum order issue, it has always been `0` (`ASSESSED_RAW`) and `2` (`ASSESSED_BS`).

#### What's this PR do?
- Fixes ESPM import `source_type` fields for the two tables mentioned above.
- Removes the `source_type` column from the `columnmapping` table
- Enforces enum strictness of the `source_type` column on the `propertystate` table
- Updates the `create_default_user` script to add the user to the specified organization if the organization already exists
- Enforces strictness on other repetitiously hardcoded source_types (`GreenButton`, `PM Meter Usage`, etc)

1. Using the _develop_ branch, import records from ESPM
2. Inspect the two tables mentioned above, observe that the `source_type` columns are incorrect
3. Checkout this branch, run the migrations, observe the database changes identifying Portfolio Manager as the data sources

#4076 

#### Note
In the future, we need to migrate the `data_importer_importfile` table to using numeric enums for strictness, this PR is a stepping stone to that end


nllong (approved):
I searched for all cases where the following were used the source type names and it looks like they were mostly converted to reading from the `SEED_DATA_SOURCES` dict. There are a few "BuildingSync Raw" floating around that could be updated, but not a big deal.

```
Assessed Raw
Portfolio Raw
Assessed
Portfolio
BuildingSnapshot
BuildingSync Raw
GreenButton
PM Meter Usage
```

There are a few cases in data_importer/tasks.py where we have the below code chunk which we could probably simplify and DRY up.

```
source_type_dict = {
    'Assessed Raw': ASSESSED_RAW,
    'Portfolio Raw': PORTFOLIO_RAW,
    'BuildingSync Raw': BUILDINGSYNC_RAW
}
```

There are also cases where we have `SensorMetaData` and `SensorReadings`, but those aren't in the `SEED_DATA_SOURCES`, which probably should be. And metadata is one word -- so it should be `SensorMetadata`? 

And `seed/data_importer/models::ImportFile` has the source_type field, but it doesn't actually use `SEED_DATA_SOURCES`. I think this is related to your `Note` in the PR.

Tested with local database--no problems.a

I am running on a production dump, but it is still running (50 minutes so far)

Before: 
image


after:
image

image

seed/migrations/0197_pm_source_type.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def fix_source_type(apps, schema_editor):
+    pm_imports = apps.get_model('data_importer', 'ImportFile').objects.filter(source_program='PortfolioManager')
+    property_states = apps.get_model('seed', 'PropertyState').objects.filter(
+        import_file__in=pm_imports,
+        source_type__isnull=False,
+    )
+
+    for state in property_states.filter(source_type=ASSESSED_RAW):
```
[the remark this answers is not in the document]
nllong: 4 hours on local dump of production... not too bad.

seed/data_importer/tasks.py
```diff
@@ -729,31 +729,31 @@ def finish_raw_save(results, file_pk, progress_key):
...
 
-    if import_file.source_type in ['PM Meter Usage', 'GreenButton'] and progress_data.summary() is not None:
+    if import_file.source_type in [SEED_DATA_SOURCES[PORTFOLIO_METER_USAGE][1],
+                                   SEED_DATA_SOURCES[GREEN_BUTTON][1]] and progress_data.summary() is not None:
         import_file.cycle_id = None
 
         new_summary = _append_meter_import_results_to_summary(results, progress_data.summary())
         finished_progress_data = progress_data.finish_with_success(new_summary)
-    elif import_file.source_type == "SensorMetaData":
+    elif import_file.source_type == SEED_DATA_SOURCES[SENSOR_METADATA][1]:
```
nllong, line 744: nice

seed/data_importer/tasks.py
```diff
@@ -1428,12 +1428,7 @@ def map_additional_models(file_pk):
     if import_file.cycle is None:
         _log.warning("This should never happen in production")
 
-    source_type_dict = {
-        'Assessed Raw': ASSESSED_RAW,
-        'Portfolio Raw': PORTFOLIO_RAW,
-        'BuildingSync Raw': BUILDINGSYNC_RAW
-    }
-    source_type = source_type_dict.get(import_file.source_type, ASSESSED_RAW)
+    source_type = SEED_DATA_SOURCES_MAPPING.get(import_file.source_type, ASSESSED_RAW)
```
nllong, line 1431: perfect

seed/migrations/0197_pm_source_type.py
```diff
@@ -1,22 +1,18 @@
...
     )
 
-    for state in property_states.filter(source_type=ASSESSED_RAW):
-        state.source_type = PORTFOLIO_RAW
-        state.save()
-
-    for state in property_states.filter(source_type=ASSESSED_BS):
-        state.source_type = PORTFOLIO_BS
-        state.save()
+    property_states.filter(source_type=ASSESSED_RAW).update(source_type=PORTFOLIO_RAW)
```
nllong, line 14: ahh, much better... yeah, this will be mmmuuch faster.

seed/models/models.py
```diff
@@ -33,7 +35,21 @@
     (BUILDINGSYNC_RAW, 'BuildingSync Raw'),
     (GREEN_BUTTON, 'GreenButton'),
     (PORTFOLIO_METER_USAGE, 'PM Meter Usage'),
+    (SENSOR_METADATA, 'SensorMetadata'),
```
nllong, line 38: A migration isn't needed because we don't actually reference `SEED_DATA_SOURCES` in our models, right?

nllong (approved):
cool, thanks for cleaning up the remaining items. I can test this against a prod dump, but will need to restore one. Let me know if you already did this...

seed/models/models.py
axelstudios: @nllong We do, but it's _never_ been used: https://github.com/SEED-platform/seed/blob/bugfix/4076-espm-source-type/seed/models/column_mappings.py#L116

The prod db only has null values in that column.  Instead of the 0196 migration in this PR, I'd prefer to just delete that column.

axelstudios: Just update the PR to remove the column from the `columnmapping` table, and added enum enforcement to the `propertystate` table

seed/migrations/0197_pm_source_type.py
axelstudios: Yup.  Just had to confirm that we weren't reimplementing the `save` method on any of these models

PR #4108, opened 2023-06-21
AH test case class, decorator, and compliance_metric tests

kflemin:
Add helper methods for testing ALI-related permissions
Add decorator @has_perm_class('requires_root_member_access') that can be used to test "setup" type API endpoints
Update compliance_metrics endpoints and tests


haneslinger (approved):
Nice

PR #4127, opened 2023-07-05
Add permissions to ColumnMappingProfileViewSet

haneslinger:


haneslinger:
@kflemin! I just realized I permission-ed to many to these  endpoints. Which ones shouldn't be? retrive and suggestions i think?

kflemin:
@haneslinger I think filter, csv, and suggestions can be less permissioned. Otherwise looks good!

kflemin (approved):

PR #4139, opened 2023-07-12
Use iterator() when rehashing to save memory

nllong:
The `0200_rehash` migration was failing due to memory pressure when migrating the production database with over 14E6 states. 

#### What's this PR do?
Convert the `.all()` call to `.all().iterator()` to prevent all records from returing at one time, rather force a database iterator.

This migration still takes *a long* time to run--around 10 hours. We might need to parallalize this in the future.

If CI passes, then there are no syntax errors. The PR does not change the core functionality.

n/a


perryr16 (approved):
Between `count()` and `iterator()` this seems like a simple way to reduce memory usage

axelstudios (approved):

nllong:
> Between `count()` and `iterator()` this seems like a simple way to reduce memory usage

yeah, I think the memory pressure probably started with the `len(PropertyState.objects.all()`... because it would need to load the `.all()` to calculate `len`

PR #4159, opened 2023-07-20
add error handling for sf edge case

kflemin:
Tiny change that will hopefully better support SFE salesforce use case


nllong (approved):
looks good. @axelstudios  -- I think this will have to go in the next release? unless we want to patch the new patch already.

axelstudios (approved):

PR #4161, opened 2023-07-22
Permission tests for salesforce workflow

kflemin:
- Permission tests for salesforce workflow
- moving update salesforce test to salesforce file for ease of testing (even though that view lives in the properties file)


seed/tests/test_salesforce_views.py
```diff
@@ -356,3 +352,260 @@ def test_no_sync_when_disabled(self):
...
+        # setup labels
+        self.ind_label = Label.objects.create(
+            name='Indication Label', super_organization=self.org
+        )
+        self.violation_label = Label.objects.create(
+            name='Violation Label', super_organization=self.org
+        )
+        self.compliance_label = Label.objects.create(
+            name='Compliance Label', super_organization=self.org
+        )
```
haneslinger, line 376: these labels were also created in `setUp()`

haneslinger (commented):
🤔 the docker image failed to build, I restarted it.

seed/tests/test_salesforce_views.py
kflemin: @haneslinger  sorry, not sure what you mean here...this is the setUp function

PR #4174, opened 2023-08-03
Set the MEDIA_ROOT path back to original value during test

nllong:
The `test_media.py` file mucks with the location of the MEDIA_ROOT. If there is a test that runs after this that write data to the MEDIA_ROOT, then it will still be pointing to the updated location (which is the `/tmp` dir). If the subsequent test needs to read out of MEDIA_ROOT it will fail due to the "file not in django working path" error.

#### What's this PR do?
After running the test, in the tearDown, set the MEDIA_ROOT back to the previous value.

CI only

N/A

### DIFF
--- a/seed/tests/test_media.py
+++ b/seed/tests/test_media.py
@@ -50,6 +50,7 @@ class TestMeasures(TestCase):
     @classmethod
     def setUpClass(cls):
         # Override MEDIA_ROOT as the temporary dir (kind of a bad way to do this but ya know)
+        cls.previous_media_root = settings.MEDIA_ROOT
         cls.temp_media_dir = tempfile.TemporaryDirectory()
         settings.MEDIA_ROOT = cls.temp_media_dir.name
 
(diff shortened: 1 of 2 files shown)


seed/tests/test_media.py
```diff
@@ -50,6 +50,7 @@ def setUp(self):
     @classmethod
     def setUpClass(cls):
         # Override MEDIA_ROOT as the temporary dir (kind of a bad way to do this but ya know)
+        cls.previous_media_root = settings.MEDIA_ROOT
```
nllong, line 53: this is part 1 of the real change

seed/tests/test_media.py
```diff
@@ -87,6 +88,8 @@ def setUpClass(cls):
 
     @classmethod
     def tearDownClass(cls):
+        # set the media directory back
+        settings.MEDIA_ROOT = cls.previous_media_root
```
nllong, line 92: this is part 2 of the real change

kflemin (approved):
♥️

PR #4188, opened 2023-08-14
[FAILING] Don't call get_queryset in decorators

haneslinger:
One test should fail, but it's the one that tests that get_queryset _is_ called in the decorators. 

The fact nothing else fails brings into question why it must be called.


haneslinger:
@nllong @kflemin repinging this, can I remove the failing tests.

nllong:
doing clean up. This is very stale and not sure if this is still needed. rebase and create new PR with justification if this is still needed.

Issue #4197, opened 2023-08-16
Ambiguous Analysis Error Messages Exclude Details


axelstudios:
**Describe the bug**
When an analysis run fails the only information shown to the user is the `user_message`, which can be as ambiguous as `Unexpected error from BETTER service.`

**Expected Behavior**
The reason why the analysis failed should be clear to the user.

**Actual Behavior**
The actual error in `debug_message` is not shown.

RDmitchell:
@axelstudios -- can this issue be closed?

RDmitchell:
instance: dev1
SHA: 070d4598b

BETTER errors now contain much more information about the problem. For example, I chose buildings without meter data to run BETTER, and the program showed the message that there was not good meter data. 
![image](https://github.com/SEED-platform/seed/assets/6314950/acb733c8-3efc-4032-a612-fc2f38374652)

PR #4198, opened 2023-08-16
Show Analysis `debug_message` If Available

axelstudios:
#### What's this PR do?
If an analysis result contains `debug_message` content then the UI will present the results in the format `${user_message} - ${debug_message}`.

1. View (or create) an analysis run that fails with additional `debug_message` information
2. Look for the debug message to be appended to the default user message

#4197


kflemin (approved):
🏆

PR #4202, opened 2023-08-17
Add Permissions to PropertyViewSet

haneslinger:
@axelstudios @kflemin the create test just won't work. I don't think it's reading the args
```
django.db.utils.IntegrityError: null value in column "cycle_id" violates not-null constraint
DETAIL:  Failing row contains (171, null, null, null).
```
why?


haneslinger:
@kflemin re-pinging you on these failing tests

seed/tests/test_property_views_viewset.py
```diff
@@ -0,0 +1,96 @@
...
+        self.login_as_child_member()
+        resp = self.client.patch(url, params, content_type='application/json')
+        assert resp.status_code == 404
+
+        # root member can
+        self.login_as_root_member()
+        resp = self.client.patch(url, params, content_type='application/json')
+        assert resp.status_code == 200
+
+    @skip("doesn't work???")
```
kflemin, line 84: do we still want this one like this?

kflemin (approved):

seed/tests/test_property_views_viewset.py
haneslinger: I would like it to work. Do you know why it's not?

haneslinger: @kflemin repinging this. idk why this doesn't work

PR #4206, opened 2023-08-23
Optimize list analyses endpoint

haneslinger:
Previously, `GET /api/v3/analyses/` returned a whole propertystate per `AnalysisPropertyViews` of which there are many. We did this the compute view display_name client side. now we compute that in the backend, greatly decreasing the amount of fetched and sent data. 

We were also querying for each `AnalysisPropertyViews`'s outputfiles one at a time. We are now prefetching them with the `AnalysisPropertyViews`

on a test db with 400 `AnalysisPropertyViews`, `GET /api/v3/analyses/` used to take ~4.5 secs and returns ~75 KB. It now returns < .4 secs and returns ~10KB.  I believe both of these performance gain will scale linearly with the number of `AnalysisPropertyViews`.

### DIFF
--- a/seed/serializers/analysis_property_views.py
+++ b/seed/serializers/analysis_property_views.py
@@ -8,12 +8,11 @@ from rest_framework import serializers
 
 from seed.models import AnalysisPropertyView
 from seed.serializers.analysis_output_files import AnalysisOutputFileSerializer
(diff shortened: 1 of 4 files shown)


kflemin (approved):
nice speed up!

Issue #4207, opened 2023-08-24
Implement EEEJ Analysis


kflemin:

RDmitchell:
@kflemin -- Github automatically closed this issue, and you do have a PR that seems to address it, so it it ok to keep it closed?

RDmitchell:
Instance: staging
SHA: 0dbce5ad6 (2.20.1)
Org: BPS Test DC

@kflemin -- I have looked at the PR for this issue, since there isn't any information in the issue about how this works. 

I have a few questions. 

I used the org that has real DC data, so the addresses are good, and the mapping of the buildings works. 

I turned off everything except Property Points, Census Tracts and Highlight DACs.

If the Census Tract is shaded in blue, presumably that means it is a DAC? 

It might be worth adding a legend or some sort of information to the user so they know that? Maybe even just a Tooltip for the "Highlight DACs" checkbox text? And possibly also for the "Census Tracts" text, to say that the census tracts are outlined (I know it's obvious once you check the box, but sometimes it's nice to tell the user what is going on rather than forcing them to experiment, although in this case, experimenting is very easy). 
![image](https://github.com/SEED-platform/seed/assets/6314950/aba498d9-34c8-4031-aea6-194d36c5df9b)

RDmitchell:
@kflemin -- also, is there a field added to Property to indicate DAC yes or no? So that I could filter on that field and see all the properties that are in DACs?

RDmitchell:
@isalanglois -- I need to document this in the release notes for 2.20, and so can you give me some background about how the information about DACs might be used? Would it be used to identify areas for energy retrofits, etc?

RDmitchell:
@kflemin -- I am trying to document for users how the EEEJ feature works, and I want to make sure I have this information correct. 

It seems like SEED is using the US Census Bureau data to define the census tracts
https://geocoding.geo.census.gov/geocoder/geographies/onelineaddress?form to compare)

And the 
![image](https://github.com/SEED-platform/seed/assets/6314950/660a876c-cdee-40a5-8efe-0f52646335ec)
for the DAC definitions
 https://screeningtool.geoplatform.gov/en/#3/33.47/-97.5 to compare)

Is that correct? Thx !!

RDmitchell:
@kflemin -- another question -- does SEED make a DAC field with a Yes or No in it so that those buildings can be filtered in the Inventory List and Detail? 

If not, we should add that.

RDmitchell:
[...]

PR #4208, opened 2023-08-24
EEEJ Analysis Functionality

kflemin:
#### What's this PR do?
New EEEJ Analysis 

Add a few buildings (must contain address_line_1, city/state or postal_code)
Select buildings -> Run an Analysis -> select the EEEJ Analysis
Ensure that the analysis is successful
Check that the census tract field identified is correct (can use https://geocoding.geo.census.gov/geocoder/geographies/onelineaddress?form to compare)
Check that the DAC designation is correct (can use https://screeningtool.geoplatform.gov/en/#3/33.47/-97.5 to compare)

#4207

Screen Shot 2023-08-23 at 9 52 41 PM
Screen Shot 2023-08-23 at 9 52 47 PM


nllong (commented):
Worked pretty well! A few things, some that can probably be pushed out until later.

1. We should have the analysis use the lat/long if it is already in the database, even if the lat/long is manually entered. 
2. Addresses can be wrong and the Census geocoder seems to be more sensitive than engines like google. (e.g., 1101 Madison St vs 1101 Madison Ave.) Google will find the correct location whereas the census encoder will misplace the location.
3. results!

image

4. I love the link to ejscreen, would be nice to download and keep this in SEED and just show as part of the results.
image

nllong (commented):
love the udpate to use the lat/long if given... it fixed the coffee shop. Geocoding by a mistyped address was 53035090900

image

nllong (approved):
This is working for me. I pushed a couple clean ups and there are few comments that can be removed. 

Merge once you review the comments and the tests pass!

💥💯💥

kflemin:
@nllong: Adding a note here that the EJScreen Report is an html report and cannot easily be saved on the SEED side, so we are just providing the link to it for now. We could investigate a way to say it as a PDF in the future with some kind of html to pdf exporter module.

nllong:
The PDF request has come up from Berkeley several times. They want to be able to generate a PDF of the results (BETTER or otherwise), and then push that PDF to a field in Salesforce. Just adding notes, nothing to do at the moment.

PR #4210, opened 2023-08-28
Move analyses endpoint

haneslinger:

The endpoint to get all a properties analyses used to be `/v3/analyses/{id}?property_id={id}`, same as the regular "list analyses" endpoint but with an arg. I've removed the arg, and created the endpoint `/v3/properties/{id}/analyses`. This vibes better with the AH permissions. 

I keeped the `v3/analyses/{id}/views/` endpoints in place tho. What do you think of that choice?


kflemin (approved):
I like it! :)

PR #4211, opened 2023-08-30
Fix 179d ESPM update endpoint to run in foreground with Celery enabled

nllong:
The import process relies heavily on cascading delay/background tasks. To allow a full path through the existing background tasks in the foreground on a system that has celery enabled is complicated. 

#### What's this PR do?
* creates two new methods for saving raw data and mapping in the foreground. 

There is definitely some areas of improvement with the entire Celery workflow, but this will not be done until the AH work is complete.

* locally start SEED with the always eager flag set to false. 
* test the `Update from ESPM` method. It should now return successfully and update the SEED property

n/a

### DIFF
--- a/seed/data_importer/tasks.py
+++ b/seed/data_importer/tasks.py
@@ -596,6 +596,52 @@ def _data_quality_check_create_tasks(org_id, property_state_ids, taxlot_state_id
     return tasks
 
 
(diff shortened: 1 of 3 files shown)


kflemin (approved):

kflemin:
🔥

PR #4221, opened 2023-09-11
Fix refresh_metadata

haneslinger:
Fixed up refresh_metadata so that 1. it can handle taxlots and properties, 2. it only needs view ids, 3. it updates everything in one transaction. 

@nllong, Should rename the action, endpoint, and task, too? What to?


perryr16 (approved):

nllong:
how about rename to "reset date time to now"?

PR #4223, opened 2023-09-13
Fix display name

haneslinger:


kflemin (approved):
works great!

RDmitchell:
@haneslinger -- what did this fix? Was it something that the user would see?

haneslinger:
On the analysis page, in the "by property" tab, the display names of the properties was broken. This fixes that. 

So the user will see it, but it's more a fix than a change.

haneslinger:
nope, it's been there

RDmitchell:
Looking at that tab, it's a bit confusing. I think I might write up an issue to see if it can be enhanced to possibly be more useful. 

Thanks

Issue #4218, opened 2023-09-11
Allow AnalysisPropertyView to have a missing PropertyView


perryr16:
**Describe the bug**
`Analysis` and `PropertyView` are related through a joins table `AnalysisPropertyView` that has a M2M relationship to both `Analysis` and `PropertyView`. When a propertyview is deleted the analysis should still exist, but the AnalysisPropertyView needs to allow for broken connections to either `PropertyView` or `Analysis`.

**Expected Behavior**
Allow the join table to be missing one of the two 'edges' (property view or analysis). If an AnalysisPropertyView is missing an edge, show a warning on the analysis. Ex: "Analysis was created with a deleted property"

When batch deleting properties there is a delete analyses checkbox that should be removed. This is the only location the endpoint is triggered so the related backend code should also be removed
![Screenshot 2023-09-11 at 9 34 38 AM](https://github.com/SEED-platform/seed/assets/58446472/fd77ed15-d690-4e34-bcca-0f753cfb000c)

RDmitchell:
@perryr16 -- has this been fixed?

perryr16:
Yes, I believe we can close this

PR #4227, opened 2023-09-19
Allow analysis property view related properties to be null

perryr16:
Analyses have a many-to-many relationship with Properties and PropertyStates via a join table AnalysisPropertyView. If a property is deleted, its Analyses and AnalysisPropertyViews are cascade deleted. 

#### What's this PR do?
This PR allows AnalysisPropertyView to have null values for both Property and PropertyState. If an analysis is related to a deleted property, the property appears as a warning icon + 'Unknown'. There is an existing tooltip that will show 'Property no longer exists' when hovering over the unknown property. This PR also removes the checkboxes 'Also delete x analyses' and 'Also delete x batch analyses' from the property delete modal.

Upload properties, create analyses, delete some of those properties, and check the analyses.

#4218

Screenshot 2023-09-19 at 9 56 28 AM

Screenshot 2023-09-19 at 9 55 12 AM

Screenshot 2023-09-19 at 9 55 44 AM

Screenshot 2023-09-19 at 9 55 55 AM


haneslinger:
Can you add the little triangle here too?
image

haneslinger (approved):

perryr16:
The analyses page now shows warning icon if property DNE
Screenshot 2023-09-20 at 10 24 55 AM

PR #4235, opened 2023-09-22
Fix user creation bug

perryr16:
There are 2 locations new users can be created
1. /app/#/accounts/{org_id}/members > invite new member
2. /app/#/profile/admin (only if current user is a superuser)

#### What's this PR do?
Adds ALI info to the superuser 'create a user' form. If a new organization is selected the ALI fields are set to undefined and disabled. The result user will be assigned to the root of the new org.

As a superuser create a new user from both members and admin pages.

Screenshot 2023-09-22 at 3 29 57 PM

### DIFF
--- a/seed/static/seed/partials/new_member_modal.html
+++ b/seed/static/seed/partials/new_member_modal.html
@@ -13,35 +13,35 @@
         <div class="form-group"  ng-class="{ 'has-error' : new_member_form.last_name.$invalid && !new_member_form.last_name.$pristine }">
             <label class="control-label col-sm-3" for="newMemberLastName" translate>Last Name:</label>
             <div class="col-sm-8">
-                <input type="text" class="form-control" name="last_name" placeholder="{$:: 'Enter last name' | translate $}" ng-model="user.last_name" required>
+                <input type="text" class="form-control" id="newMemberLastName" name="last_name" placeholder="{$:: 'Enter last name' | translate $}" ng-model="user.last_name" required>
                  <p ng-show="new_member_form.last_name.$invalid && !new_member_form.last_name.$pristine" class="help-block" translate>Last name is required.</p>
             </div>
         </div>
         <div class="form-group"  ng-class="{ 'has-error' : new_member_form.email.$invalid && !new_member_form.email.$pristine }">
             <label class="control-label col-sm-3" for="newMemberEmail" translate>Email Address:</label>
             <div class="col-sm-8">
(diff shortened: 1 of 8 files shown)


seed/static/seed/partials/new_member_modal.html
```diff
@@ -13,35 +13,35 @@ <h4 class="modal-title" id="newMemberLabel" translate>Invite a New Member</h4>
         <div class="form-group"  ng-class="{ 'has-error' : new_member_form.last_name.$invalid && !new_member_form.last_name.$pristine }">
             <label class="control-label col-sm-3" for="newMemberLastName" translate>Last Name:</label>
             <div class="col-sm-8">
-                <input type="text" class="form-control" name="last_name" placeholder="{$:: 'Enter last name' | translate $}" ng-model="user.last_name" required>
+                <input type="text" class="form-control" id="newMemberLastName" name="last_name" placeholder="{$:: 'Enter last name' | translate $}" ng-model="user.last_name" required>
```
perryr16, line 16: Changes in this file are primarily clean up

seed/views/v3/users.py
```diff
@@ -192,31 +193,22 @@ def create(self, request):
...
-                return JsonResponse({'status': 'error', 'message': 'valid arguments for role are [viewer, member, '
-                                                                   'owner]'},
-                                    status=status.HTTP_400_BAD_REQUEST)
-
-            OrganizationUser.objects.filter(
-                organization_id=org.pk,
-                user_id=user.pk,
-                access_level_instance_id=access_level_instance_id,
-            ).update(role_level=get_role_from_js(role))
-
```
perryr16, line 219: `org.add_member()` has a default `role` value of `ROLE_OWNER`. This caused errors when attempting to create. The role is later applied in an `OrganizationUser` query, however due to the error the code never reached this point

seed/tests/test_admin_views.py
```diff
@@ -110,7 +110,7 @@ def test_add_user_existing_org(self):
             'first_name': 'New',
             'last_name': 'User',
             'email': 'new_user@testserver',
-            'role_level': 'ROLE_MEMBER',
+            'role': 'member',
```
perryr16, line 113: This test has been broken for a while, but due to a few stacked errors it 'passed'. In /v3/users.py, the `def create` function first calls `org.add_member` without passing a `role` argument triggering the default argument 'owner'. The role passed in the request is applied after the user is created `if body.get('role')...`. Because 'role' was not a key this code was skipped and the new user remained an 'owner'

haneslinger (approved):

PR #3899, opened 2023-03-15
Create property timeline view

perryr16:
#### Any background context you want to provide?
See #3769

#### What's this PR do?
- Adds timeline features 
- Refactors scenario and measure layout
- Migration to retroactively create events for existing databases
- Auto generates events when a Note, Analysis, or Building File is created
- Notes and Analyses have a cascade delete relationship with events 

#### How should this be manually tested?
setup:
- populate a single property across multiple cycles 

Notes:
- create a note in cycle A, create a note in cycle B. Check the timeline view and ensure notes are in the correct cycles
- delete a note and ensure the timeline event has also been deleted 

Analyses:
- create an analysis in cycle A and in cycle B. Check the timeline for the correct placement 
- delete an analysis and ensure the event has also been deleted 

AT Uploads:
- upload 2 building sync files for cycle A. upload a building sync file for Cycle B. Ensure the scenarios and measures table at the bottom of the property detail shows the correct AT files for the selected cycle.
- Check the timeline for ATEvents 
- Go to property detail and delete a scenario. Check the timeline has been updated for the removed scenario.

Timeline:
- toggle "collapse all" and "expand all". Confirm expected functionality 
- sort by "Date, Oldest to Newest", sort by "Date, Newest to Oldest"
- Use the event selection to deselect a type of event. Ensure the type is not listed in the timeline.
 

#### What are the relevant tickets?
#3769 Create Property Timeline Table
#3843 Create property timeline view UI
#3768 Create a new navigation item in property detail for Property Timeline
#3890 Edit Scenarios and measures on Property Details Page
#3777 Add a view set for measures

#### Screenshots (if appropriate)
[Mockups](https://xd.adobe.com/view/9516edb6-76a6-4a19-8b74-203c7f32385e-c045/grid/)

Property Detail Scenarios and Measures
Screenshot 2023-03-20 at 11 35 56 AM

Timeline Collapsed
Screenshot 2023-03-20 at 11 33 12 AM

Timeline Expanded
Screenshot 2023-03-20 at 11 34 42 AM

Event Selection
Screenshot 2023-03-20 at 11 35 15 AM

PR #4240, opened 2023-09-25
Fixed Issues with Scenario/Measure CRUD Operations

axelstudios:
These changes are cherry picked from a64092639fcb4647d0d5d4c257a56858561f2f46 on the `bps-feature-develop` branch that was never merged to develop (but all of the BPS changes were merged in with https://github.com/SEED-platform/seed/pull/3899/)

#### What's this PR do?
- Fixed 500 errors when the input datatypes did not match the expected model types
- Fixed 500 error when using PUT with a nonexistent scenario
- Removed 3 fields from the scenario input model that always resulted in 500 errors: `measures`, `property_state`, and `reference_case`
- Removed unused pagination inputs on measures endpoints
- Fixed listing measures for a scenario, if the scenario exists with zero measures it now returns an empty array rather than a 404
- Fixed measure input model, `recommended` is a boolean rather than a string
- Updated schema generation to allow OpenAPI enum types and document the enum options
- Added enum validation to scenario/measure endpoints. Strings and ints are both accepted and validated
  - Before, this was resulting in 500 errors if invalid options were sent, like `false` or `0` or other invalid enum integers.  Subsequently, any endpoint that pulled in another model with an invalid integer would start returning 500 errors.

### DIFF
--- a/seed/models/property_measures.py
+++ b/seed/models/property_measures.py
@@ -157,7 +157,9 @@ class PropertyMeasure(models.Model):
             return None
 
         if isinstance(impl_status, int):
-            return impl_status
+            if impl_status in [(t[0]) for t in cls.IMPLEMENTATION_TYPES]:
+                return impl_status
+            return None
 
         value = [y[0] for x, y in enumerate(cls.IMPLEMENTATION_TYPES) if y[1] == impl_status]
         if len(value) == 1:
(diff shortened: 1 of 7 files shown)


seed/models/property_measures.py
```diff
@@ -157,7 +157,9 @@ def str_to_impl_status(cls, impl_status):
             return None
 
         if isinstance(impl_status, int):
-            return impl_status
+            if impl_status in [(t[0]) for t in cls.IMPLEMENTATION_TYPES]:
+                return impl_status
```
perryr16, line 161: 👍

seed/views/v3/property_measures.py
```diff
@@ -22,8 +23,15 @@ class PropertyMeasureViewSet(SEEDOrgNoPatchNoCreateModelViewSet):
     """
     serializer_class = PropertyMeasureSerializer
     model = PropertyMeasure
+    pagination_class = None
     orgfilter = 'property_state__organization_id'
 
+    enum_validators = {
```
perryr16, line 29: 👍

perryr16 (approved):

PR #4243, opened 2023-09-25
Add smaller EEEJ data files for testing

kflemin:
Add smaller datafiles for testing out EEEJ functionality, rather than importing the real datafiles which is time consuming.


seed/lib/geospatial/eeej.py
```diff
@@ -28,8 +28,14 @@ def import_hud():
     # property_name comes from PROJECT_NAME or PROPERTY_NAME_TEXT columns
     """
     # Note: instead of update_or_create, it might be better to clear and start over
-    HUD_DATA_PATH_HOUSING = os.path.join(settings.BASE_DIR, 'seed/lib/geospatial/', 'Public_Housing_Developments.xlsx')
-    HUD_DATA_PATH_MULTIFAMILY = os.path.join(settings.BASE_DIR, 'seed/lib/geospatial/', 'Multifamily_Properties_-_Assisted.xlsx')
+    # Use smaller files to test with
+    settings_env = os.environ.get('DJANGO_SETTINGS_MODULE', None)
+    if settings_env == 'config.settings.test':
```
nllong, line 33: love this

nllong (approved):
this will help the test spin up time for sure. I think it won’t address the py-seed timeout because that isn’t running in “test” move. Maybe we should also allow reading in another env var such as “ONLY_LOAD_SMALL_EEEJ_DATA=true” or something like that. 

We have a solution for now…

Issue #4220, opened 2023-09-11
AH - Only root users can write to property columns on analysis creation


perryr16:
### Current Functionality 
Any user within an org can run an analysis on any property, regardless if they are the root user of this property. The analysis will write data to the property state. Ex: An EUI analysis will write values to columns `Fractional EUI` and `EUI Coverage`

### Desired Functionality 
Anyone can create an analysis, but only a root user can write data to the property's columns

Anyone can run and create analyses, however, if the current user is not the root owner of the property then the analysis should not write columns to that property state.

### Note
Hold off on this for now.
This feature must be built on top of the AH/develop branch.

RDmitchell:
@perryr16 -- is this issue still relevant?

perryr16:
@RDmitchell Yes, this is still relevant

RDmitchell:
thx !

PR #4251, opened 2023-09-27
Allow only property owners to write to property columns on analysis creation

perryr16:
`in progress...`

#4220


perryr16:
Needs further discussion

PR #4254, opened 2023-09-27
Add envvar to specify loading test EEEJ dataset

kflemin:
set the EEEJ_LOAD_SMALL_TEST_DATASET env var to True if you would like to only load the small EEEJ dataset for testing.
If the env var does not exist, it gets set to False by default everywhere but when using the 'test' settings file, in which case it is set to True.

This can be used in workflows such as py-seed.


nllong (approved):
i like this approach much better! That way we can set this in py-seed.

PR #4264, opened 2023-10-02
Add documentation for cleaning up conflicting column names during migration

nllong:
The column name constraint migration needed more instructions on how to cleanup the database. These *could* be converted over to a script, but for now I really think we need the human judgment on 1) age of the organization, 2) the active use of the organization, 3) the raw vs mapped impact.

#### What's this PR do?
* Adds instructions to our migrations.rst file

* verify that the steps make sense

n/a


docs/source/migrations.rst
```diff
@@ -54,7 +54,19 @@ Version 2.20.0
 
 Version 2.19.0
 --------------
-- There are no special migrations needed for this version. Simply run `./manage.py migrate`.
+- Run `./manage.py migrate`.
+- There is a new migration in this release that requires column names to be unique across `organization`, `table_name`, and `is_extra_data`. This migration will fail if there are duplicate column names. If you have duplicate column names, you will need to manually fix them in your database before running the migration. The following steps will help you identify and fix the duplicate column names:
+    - Check the organization. If it is a deprecated org to gauge the impact of the change. Often this issue arose in older organizations when units were not part of the columns. The old mapping columns were not upserts with the units, so typically the columns impacted are the ones with units.
```
kflemin, line 59:
```suggestion
    - Check the organization age to gauge the impact of the change. If it is a deprecated org, impact of the change will be low. Often this issue arose in older organizations when units were not part of the columns. The old mapping columns were not upserts with the units, so typically the columns impacted are the ones with units.
```

docs/source/migrations.rst
```diff
@@ -54,7 +54,19 @@ Version 2.20.0
...
+    - Query the `seed_column` table for the organization and column name displayed on the screen (e.g., `organization_id = 300 and column_name = 'Source EUI (kBtu/ft2)'`). If there is no `table_name` set, it is likely an import file column name and can easy be cleaned up without causing issues. In such cases, there will be two rows, and you want to keep the one with the `units_pint` column set.
```
kflemin, line 60:
```suggestion
    - Query the `seed_column` table for the organization and column name displayed on the screen (e.g., `organization_id = 300 and column_name = 'Source EUI (kBtu/ft2)'`). If there is no `table_name` set, it is likely an import file column name and can easily be cleaned up without causing issues. In such cases, there will be two rows, and you want to keep the one with the `units_pint` column set.
```

kflemin (approved):

PR #4270, opened 2023-10-03
Fix edge case for map label filtering

axelstudios:
Filtering properties or taxlots by label on the map page works as expected for small databases, but once property/taxlot ids are no longer identical to view ids the map view filters against the wrong ids.

#### What's this PR do?
Filters properties and taxlots using the view id instead of the property/taxlot id


nllong (approved):
sweet, this looks good. dev1 looked good too.

Patch the release?

RDmitchell:
@nllong / @axelstudios -- should I try to monkey test this?

PR #4276, opened 2023-10-09
Fix mapping of BETTER's measures with empty strings to false

nllong:
There appears to have been an update with BETTER's API where measure recommendations are either `''` or `True`. 

#### What's this PR do?
Hande the case where the measure recommendation is an empty string

Run a property with better (I used `better.lbl.gov`)

n/a


kflemin (approved):

RDmitchell:
Instance: dev1
Org: JCC

Tested running the Offices with good meter data and the BETTER runs are now simulating. 

Thanks @nllong !

PR #4280, opened 2023-10-12
Decrease BETTER polling rate 5x, increase timeout 5x

axelstudios:
BETTER's API can begin limiting API requests for long periods of time once a certain threshold of requests have been sent for a specific API key.

#### What's this PR do?
Reduces the polling rate for BETTER results from every second to every 5 seconds, and increases the timeout from 1 minute to 5 minutes.

Monitor backend network requests when performing a BETTER analysis.


nllong (approved):
Sweet, works as expected. This will save a lot of requests hitting the BETTER api and prevent normal users from getting throttled.

PR #4281, opened 2023-10-13
Enable natural sorting of Inventory List columns

axelstudios:
After switching from the front to back-end for sorting Inventory List columns we lost the ability to _naturally_ sort.

#### What's this PR do?
- Creates a new Postgres collation that Char/Text fields can opt-in to that indexes using natural sort.
- Extra_data columns with a data_type of `string` or `None` are handled specially to use natural sorting.

Sort any of the columns affected by the migration in this PR, or any string-based extra_data column, especially data that sorts differently when naturally sorted (e.g. `[1, 5, 10]`)

NaturalSort


seed/migrations/0210_natural_sort.py
```diff
@@ -0,0 +1,219 @@
...
+    dependencies = [
+        ('seed', '0209_auto_20230929_0959'),
+    ]
+
+    operations = [
+        CreateCollation(
+            'natural_sort',
+            provider='icu',
+            locale='en@colNumeric=yes',
+        ),
```
axelstudios, line 16: This is where the magic happens

nllong: wow, never new that was a thing! Go postgres!

nllong (approved):
This worked as expected, great update!


image

PR #4282, opened 2023-10-13
Add failing logic to AT export if AT credentials are invalid

perryr16:
If a user's AT credentials were invalid the page would hang

#### What's this PR do?
Fails gracefully if credentials are invalid

Inventory List > selection > Actions > Export to Audit Template. Once with invalid credentials, and once with valid credentials

Screenshot 2023-10-13 at 1 17 34 PM


docker-compose.dev.yml
```diff
@@ -22,7 +22,6 @@ services:
     environment:
       - DJANGO_SETTINGS_MODULE=config.settings.docker_dev
       - DJANGO_LOG_LEVEL
-      - EEEJ_LOAD_SMALL_TEST_DATASET=True
```
perryr16: Should the docker-compose.dev.yml include this env var?

nllong:
@perryr16  -- do you know why I would be getting this error? I know that my username and PW are right. Is there something to do with the org that I have?

perryr16:
Are you using my org token? Is your `AUDIT_TEMPLATE_HOST` set to  `'https://buildingenergyscore.energy.gov'`?

docker-compose.dev.yml
nllong: I think so because that is what we use for our GitHub regression testing, right? We want that to run faster.

nllong (approved):
worked! but note that we need to fix allowing passwords with `%` symbols.

PR #4288, opened 2023-10-19
Feature to enable Redis for ElastiCache

dhaley:
#### What's this PR do?

Fixes bug with CELERY BROKER connection string to Elasticache Redis

In Django shell, print CELERY_BROKER_URL to make sure it is correct when using Docker.py settings.

No

### DIFF
--- a/config/settings/docker.py
+++ b/config/settings/docker.py
@@ -150,7 +150,7 @@ LOGGING = {
     'loggers': {
         'django': {
             'handlers': ['console'],
-            'level': os.getenv('DJANGO_LOG_LEVEL', 'DEBUG'),
+            'level': os.getenv('DJANGO_LOG_LEVEL', 'ERROR'),
         },
     },
 }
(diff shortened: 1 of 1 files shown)


config/settings/docker.py
```diff
@@ -34,16 +34,20 @@
...
-DEBUG = False
+DEBUG = os.environ.get('Debug', 'False')
 COMPRESS_ENABLED = False
 # COMPRESS_STORAGE = 'compressor.storage.GzipCompressorFileStorage'
 # COMPRESS_OFFLINE = True
 
 # Make sure to disable secure cookies and csrf when using Cloudflare
-SESSION_COOKIE_SECURE = False
-CSRF_COOKIE_SECURE = False
+SESSION_COOKIE_SECURE = os.environ.get('SESSION_COOKIE_SECURE', 'True')
```
nllong, line 43: shouldn't this be defaulted to `False` if not found. Otherwise our existing deployments will break.

config/settings/docker.py
```diff
@@ -34,16 +34,20 @@
     if not locals().get(loc):
         raise Exception("%s Not defined as env variables" % loc)
 
-DEBUG = False
+DEBUG = os.environ.get('Debug', 'False')
```
nllong, line 37:
```suggestion
DEBUG = os.environ.get('Debug', False)
```

config/settings/docker.py
dhaley: Yes. I just fixed this.

config/settings/docker.py
dhaley: This should be fixed now.

nllong (approved):

PR #4291, opened 2023-10-19
Fix cycle deletion modal

axelstudios:
We recently added the `ng-strict-di` directive for strict dependency injection checking

#### What's this PR do?
Fixes the cycle deletion modal that was missing the `organization_service` injection

Click the `Delete` button for any cycle, ensure that the modal appears


nllong (approved):
sweet, delete cycle worked.

PR #4292, opened 2023-10-19
Delete PropertyStateViewSet

haneslinger:


nllong (approved):
This looks good. Since we have so many warnings about not using the endpoint, then just removing makes sense.

PR #4300, opened 2023-10-21
Update list of BETTER's measures returned

nllong:
Somewhere in the recent past, the list of measures from BETTER was updated. This includes adding a few new ones and renaming/removing others. Existing databases will still have the old fields but new runs in the old database will add in the new ones.

List of measures were taken from looking at this dict: https://github.com/LBNL-ETA/better_lbnl/blob/756f101292083bb053af06fd152ea7e1135878df/better_lbnl/building_analytics/models.py#L549

#### What's this PR do?
* Update the list of measures from BETTER

* Create a new property with meter data (import the better test data file)
* Select a property and run better
* Make sure the new measures are in the columns
* 

n/a

image

image

Note that the previous results were also 0 and 1.
image


axelstudios (approved):

axelstudios:
Ran before and after, measure recommendation columns went from 9 to 15 👍

PR #4364, opened 2023-10-26
Fix showing derived column values on the inventory list page

axelstudios:
#### What's this PR do?
Fixes showing derived column values.  The issue was a strict type comparison between a string and a number, that should've both been numbers

Check that derived column values appear on the inventory list page


kflemin (approved):
it works! thank you!

PR #4374, opened 2023-10-31
Add healthcheck to docker web container

axelstudios:
#### What's this PR do?
Adds `healthcheck` to the `seed_web` container so that Docker understands the health status of the dependencies.

1. Run SEED using `docker compose`
2. Wait for the health status of the `seed_web` container to transition from `Starting` to `Healthy`
3. Stop the `seed_celery` container
4. Wait at most 60 seconds, and check that the container status is now `Unhealthy`
5. Start the `seed_celery` container
6. Wait at most 60 seconds, and check that the container status is now back to `Healthy`

#### Screenshots
image

### DIFF
--- a/docker-compose.local.yml
+++ b/docker-compose.local.yml
@@ -83,6 +83,12 @@ services:
       - seed_media:/seed/media
     ports:
       - "80:80"
+    healthcheck:
+      test: curl -f http://localhost/api/health_check/ || exit 1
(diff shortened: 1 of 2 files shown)


perryr16 (approved):
Works as expected. Able to see the transition from healthy -> unhealthy -> healthy

Issue #4375, opened 2023-10-31
Investigate solution for supporting redis cluster


kflemin:
How can we support redis cluster? (for high-availability deployment)

- update to django-redis?
- update redis-py to at least 4.1.0 (but that creates issue with django-redis-cache)
- update to Django v4 with built-in redis caching?

PR #4376, opened 2023-11-01
Redis backend update

axelstudios:
The dependency we were using to leverage Redis as the Django/Celery cache (`django-redis-cache`) effectively hasn't been updated since 2020, and requires `redis-py < v4.0`. Redis-py introduced cluster-mode in v4.1.0.

#### What's this PR do?
Drops the `django-redis-cache` dependency in favor of `django-redis`, which is compatible with current versions of `redis-py`. This PR also adds `hiredis` for performance.

Once we migrate to Django v4 we can use Django's built-in Redis support going forward.

1. `pip install -r requirements/base.txt`
2. Update your local config (usually `local_untracked.py`), **see the migration steps below** to understand what to change
3. Test that SEED and Celery work as expected

#4375

## BREAKING CHANGE
This updated dependency requires users and deployments to modify their `CACHES` config. [Django v4 uses the same syntax](https://docs.djangoproject.com/en/4.2/topics/cache/#redis), so this is a necessary step regardless.

### Migration Steps
1. Update your dependencies with `pip install -r requirements/base.txt`
2. Update the CACHES `BACKEND` property to `django_redis.cache.RedisCache`
3. Update the CACHES `LOCATION` property to match the redis-py native URL notation for connection strings, including the `redis` protocol and database number. e.g. `redis://localhost:6379/1`
    1. Since the `CELERY_BROKER_URL` setting must also be in the same format, it may be helpful to configure that setting first and then reference it in the caches `LOCATION` parameter.

#### Example migration
In your Django settings file, replace existing code that looks like this:
```python
CACHES = {
    'default': {
        'BACKEND': 'redis_cache.cache.RedisCache',
        'LOCATION': 'localhost:6379',
        'OPTIONS': {'DB': 1},
        'TIMEOUT': 300
    }
}

CELERY_BROKER_URL = 'redis://%s/%s' % (
    CACHES['default']['LOCATION'], CACHES['default']['OPTIONS']['DB']
)
```

with this:

```python
CELERY_BROKER_URL = 'redis://localhost:6379/1'

CACHES = {
    'default': {
        'BACKEND': 'django_redis.cache.RedisCache',
        'LOCATION': CELERY_BROKER_URL,
    }
}
```

For advanced connections, the connection string can also specify SSL, username, and password, e.g. `rediss://username:password@localhost:6379/1`

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -394,6 +394,7 @@
     "unmerge",
     "unmerges",
     "unpair",
+    "Unpickles",
     "unresolvable",
     "untracked",
     "uom",
--- a/.cspell.json
+++ b/.cspell.json
@@ -116,6 +116,7 @@
     "edgecase",
     "eeej",
     "ejscreen",
+    "ELASTICACHE",
     "energystar",
     "enums",
     "env",
--- a/config/settings/docker.py
+++ b/config/settings/docker.py
@@ -24,7 +24,7 @@ SMTP_ENV_VARS = ['EMAIL_HOST', 'EMAIL_PORT', 'EMAIL_HOST_USER',
 # The optional vars will set the SERVER_EMAIL information as needed
 OPTIONAL_ENV_VARS = ['AWS_ACCESS_KEY_ID', 'AWS_SECRET_ACCESS_KEY', 'AWS_SES_REGION_NAME',
                      'AWS_SES_REGION_ENDPOINT', 'SERVER_EMAIL', 'SENTRY_JS_DSN', 'SENTRY_RAVEN_DSN',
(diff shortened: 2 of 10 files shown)


kflemin:
@nllong what do you think?

nllong (approved):
Looks good. Let's bring this up on our next call that users will need to update their local untrack files before we merge down.

axelstudios:
```[tasklist]
### TODO
- [ ] Let users know they'll have to update their local configs once this PR merges
```

kflemin:
@dhaley will test this out around Feb12th to make sure it works on stratus before merging.

PR #4378, opened 2023-11-02
Permission

haneslinger:
Any idea on how to test here? celery take it a little hard, but I just need to know what the celery task was passed. I could mock it if not for the fact we are going though the http client.


haneslinger:
@kflemin repinging

kflemin:
note to review data quality check permissioning when we implement the cross-cycle data quality checks

PR #4394, opened 2023-11-08
Test disallow nonroot users to create columns

haneslinger:
White's a non root user, blacks a root user. See that white Can't create column "Color" but can use it after black created it. 

image
image
image


haneslinger:
@perryr16 could you help me with these front end tests?

kflemin (approved):

PR #4410, opened 2023-11-17
Fix co2 analysis column naming

ebeers-png:
Very tiny bug fix - when creating a column during the CO2 analysis, it was naming the column with a tuple instead of a string.


nllong (approved):
nice catch, thanks for fixing!

Issue #4137, opened 2023-07-11
Add error messages to aln and column creations.


haneslinger:
As of https://github.com/SEED-platform/seed/pull/4134, one cannot
 create a column named after an accessLevelname nor visa versa. 

This is only implemented in the backend, not the frontend. Please give the frontend some nice warnings and/or errror messages.

RDmitchell:
@haneslinger -- has this issue been resolved?

haneslinger:
Yes, thank you

PR #4417, opened 2023-11-27
4137 fix anl error message

haneslinger:
https://github.com/SEED-platform/seed/issues/4137
image


kflemin (commented):
Can you change the notification text to say:
Access level names cannot match SEED column names: ['City']

Otherwise looks good!

PR #4423, opened 2023-11-28
Add count to default reports

haneslinger:
Truly the dumbest, least efficient way to do this, but it's the least invasive and if/esle-y


kflemin (approved):
count aggregate chart looks good!

PR #4425, opened 2023-11-29
Property Insights Tooltip Refinement

### LINKED ISSUE #3741: Property Insights: One more tooltip refinement
Instance: dev1
SHA: 294626cbd
Org: BPS Test
Program: BPS Compliance - Energy only
Cycle: 2020

Here is what the tool tip looks like now, which is a big improvement based on fixes from #3594 
image
image

The only thing that is a bit mysterious to the uninitiated user are what the numbers on the first line mean, which was the 1st bullet in the original description in #3594 

The values displayed reflect the X and Y axis.

However, in this case, because the two numbers are separated by a comma and there is also a comma in the Gross Floor Area, they are a bit difficult to distinguish.

One suggestion would be to add descriptors in front of each, and possibly put them on separate lines? I know that starts to be a bit bulky but I think that is outweighed by having something that the user can easily figure out.

So something like

Gross Floor Area: 76,319
Site EUI: 62
Property - Medstar POB South Tower 

Another option would be to just eliminate any commas in either value and then separate by a comma, without field names in front of the values

This probably doesn't have a high priority at this point, but I wanted to keep track of it in an issue, so I will add the Ice Box label to it. If we get user feedback about this, we can reprioritize it.

kflemin:
Related Issue: #3741 

Adds labels to X and Y values in the tooltip:

Screen Shot 2023-11-29 at 11 52 16 AM


haneslinger (approved):

RDmitchell:
instance: dev1
SHA: 
Org: BPS Test

Perfect ! Thanks !

image

Issue #4433, opened 2023-12-05
Speed up /analyses with outputfiles


haneslinger:

RDmitchell:
@haneslinger -- is this fixed? Should it be tested? Presumably this is deployed to dev1?

RDmitchell:
The status is "Review" so maybe it hasn't been deployed to dev1 yet.

haneslinger:
I think this should be on dev

RDmitchell:
@haneslinger / @kflemin -- how should this be monkey tested, ie, is this something I can test through the UI?

kflemin:
@RDmitchell, I think all you can do here is open up the developer tools in your browser and select the 'network' tab. then load the analyses page and ensure that it takes on the order of 5seconds (rather than > 10 seconds).  This can probably just get closed too.
![Screenshot 2024-02-14 at 3 25 08 PM](https://github.com/SEED-platform/seed/assets/2205659/733718ec-2f7c-4e6d-8e13-d56212d74755).

RDmitchell:
instance: dev1
SHA: d5d60eaac
Org: LBNL 245
Analysis loaded in < 5 seconds.

PR #4434, opened 2023-12-05
Forgo evil evil serializer

haneslinger:
#4433

### DIFF
--- a/seed/views/v3/analyses.py
+++ b/seed/views/v3/analyses.py
@@ -160,7 +160,6 @@ class AnalysisViewSet(viewsets.ViewSet, OrgMixin):
                 {
                     "id": view.id,
                     "display_name": view.display_name,
-                    "parsed_results": view.parsed_results,
                     "analysis": view.analysis_id,
                     "property": view.property_id,
                     "cycle": view.cycle_id,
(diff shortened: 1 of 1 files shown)


kflemin (approved):

kflemin:
did it speed it up?

haneslinger:
I created ~100 output_files. The old code ran in ~1 sec, the new in <1 sec. negligible with that number of rows. However, the end point use to run a query for each analysis_property_view/AnalysisOutputFile pair, which accounted for the bulk of the run time. and now it's done in a single query.

It's worth noting this endpoint is still probably returning more data than needed, but this should address the main problem.

RDmitchell:
@haneslinger -- you can test using the BPS Test - DC org -- it has 3,000+ records

haneslinger:
It's a pain to get a prod db on my local instance and it makes testing really time consuming. If @kflemin thinks it beneficial to run more test before merging I can, but I think it might be best to simply get the change to prod quickly.

That said, @axelstudios said "[BPS Test - DC org] seed_analysisoutputfile_analysis_property_views has 869 rows". I have 52 locally. If BPS Test - DC org take 5 min, you'd think mine would take at least 10 sec?

ugh, let my see if the "parsed data" field is good to be removed. that'll cut down on the size.

RDmitchell:
@haneslinger -- no worries -- I'll test when this is deployed to dev1 !!

haneslinger:
right-o, removed "parsed_results", didn't need it. decimated the size of the response.

PR #4437, opened 2023-12-06
Additional analyses performance improvement

axelstudios:
#### What's this PR do?
Improves a very long, slow query, by ~17x

Previously, `GET /api/v3/analyses/` took `~13.52s`, now it takes `910ms` - `4.19s` for the same large dataset

Request the above URL and check the response timing

#4433


axelstudios:
Replaced by #4438

PR #4440, opened 2023-12-08
Add Cross Cycle Tabular View for Portfolio Summary Goals

perryr16:
Create a new view for displaying cross-cycle tabular data in a format requested by Better Buildings.

#### What's this PR do?
- A new Portfolio Summary page to view individual and aggregated EUI improvement between 2 cycles.
- A New Goal database table with the following attributes: 
   - Organization
   - Baseline Cycle
   - Current Cycle
   - Access Level Instance 
   - Area Column
   - EUI Column 1
   - EUI Column 2 (optional)
   - EUI Column 3 (optional)
   - Name
   - Target Percentage
   
- Access Level Instance controls: Users cannot view properties or access level names that are 'above' themselves in the tree. For example, a parent user can see its own and its children's properties. A child user can only see its properties, not its sibling's or parent's.

- EUI Columns: Any column that has an EUI data type may be selected as a Goal EUI column. Extra data fields are eligible if their data type has been changed, however, derived columns are not. A primary column must be selected, however secondary and tertiary columns are not required. If a primary column data entry for a property state is empty the next EUI column will be used. If none is found it is ignored.

- Goals cannot be created that violate database or UI constraints. 
   - Names must be unique
   - Cycles must be unique
   - Baseline cycle must proceed Current cycle 
   - Access Level Instance must be within a user's tree permissions
   - Columns must be unique
   - Target % must be integer 0-100

- Speed and Optimization: Serializing 100 `PropertyView`s for frontend use takes approximately 1 second. `Cycle`s can contain thousands of `PropertyView`s leading to several minute loading times. Pagination has been introduced to reduce load times to less than 10 seconds, however, it requires two separate database queries to retrieve the correct set of properties from the two `Cycle`s. The first query will retrieve the first 50 `PropertyView`s that meet the sorting and filter requirements from the Portfolio Summary table for a certain cycle. Then using the list of `Property` IDs from the first response, a second database query is executed to retrieve `PropertyView`s from the other cycle using the shared `Property` ID value as a database filter. Because there are two queries, SEED can only sort or filter on one `Cycle` at a time. In the event of a conflict (a filter on a baseline column and a current column), the newer filter will be used

- Table Features
   - Sorting: Supports sorting on a single column.
   - Filtering: Supports multiple filters on shared cycle columns, however, the table can only filter on either baseline or current specific columns. If there is a conflict the newer filter will used.
   - Access Level Instances can be hidden 
   - EUI values are found based on Goal EUI column priority. If the primary EUI is missing for a `PropertyState` the secondary EUI column will be used. If both are missing, the tertiary EUI column will be used.
   - Columns not moveable for readability.
   - Labels are displayed per Cycle
   - CSV downloads of the current page

- Create an ALI tree 
- Create 3 cycles 
- Import the same base set of properties to each cycle, however, the EUI values for each cycle should be varied. Optionally include an extra data column.
- Go to Insights > Portfolio Summary > New Goal 
- Create a few goals. Ensure safeguards are in place - Names must be unique, cycles must be unique, baseline cycle must precede current cycle, access levels are limited to users tree permissions, EUI columns must be unique, EUI columns 2 and 3 are optional, target percentage must be an integer 0-100
- load the new goals, and ensure quick load times. check pagination, filtering, and sorting.

#### Sample Test Files
ALI Tree: * must first make 'Sector', 'SubSector', and 'Partner' Access Levels (+ > Access Level Tree > Edit/Add Access Levels)
[ali_tree_better_buildings_lite.xlsx](https://github.com/SEED-platform/seed/files/13789378/ali_tree_better_buildings_lite.xlsx)

Baseline Data
[AH_BB_1000props120.xlsx](https://github.com/SEED-platform/seed/files/13789387/AH_BB_1000props120.xlsx)

Middle Data
[AH_BB_1000props100.xlsx](https://github.com/SEED-platform/seed/files/13789388/AH_BB_1000props100.xlsx)

Current Data
[AH_BB_1000props70.xlsx](https://github.com/SEED-platform/seed/files/13789389/AH_BB_1000props70.xlsx)

#4456

Screenshot 2023-12-12 at 2 03 17 PM

Screenshot 2023-12-12 at 2 17 22 PM

Screenshot 2023-12-12 at 2 51 31 PM

Screenshot 2023-12-12 at 2 16 56 PM


kflemin (approved):
🏆

RDmitchell:
Wow this looks really great. 

So I guess when this gets deployed, it will be available to everyone, not just Better Buildings, right?

perryr16:
Yeah, anyone with access to AH will have access to this

PR #4442, opened 2023-12-12
display x axis label for ranked distance to compliance

kflemin:
X-axis label was missing for "Ranked distance to compliance" on property insights page. 
It is there now:
Screen Shot 2023-12-12 at 4 19 37 PM


nllong (approved):
most excellent

Issue #3371, opened 2022-07-14
Filters - Include multiple "and" / "or" / "exclude"


isalanglois:
Harry requested this for Maryland user, we need to communicate with them once this task is complete

Contact: Maryland

RDmitchell:
It might be worth talking to the folks at  the Maryland office to see how they were trying to filter the data to make sure we understand the use case.

isalanglois:
@RDmitchell  That's a good idea. Let's talk about it tomorrow in the meeting

RDmitchell:
@isalanglois -- I moved this from Q4 to Q3, in case Maryland needs it before the final quarter. Someone might want to check with Maryland to see when they might want to start using SEED.

ebeers-png:
I've recently added this to BEAM, based on the mockup Veronique made in Google Docs (linked above). Would you like the code for it contributed back to SEED? It replaces the current AND/OR/EXCLUDE bar in the inventory list, and can be saved in filter groups. 

![image](https://github.com/SEED-platform/seed/assets/53191386/44fb19b3-aaf1-4bad-8a33-ef02c1ba37ae)

haneslinger:
PR: https://github.com/SEED-platform/seed/pull/4449

RDmitchell:
instance: dev1
SHA: bcbec0272
Org: BPS - DC

Seems to work well 
![image](https://github.com/SEED-platform/seed/assets/6314950/32922d7f-dcb0-448f-861f-b8a8e125ed04)

PR #4449, opened 2023-12-19
Labels: Added complex filtering logic

ebeers-png:
#### What's this PR do?

Adds more complex label filtering logic to the inventory list, by allowing users to filter the inventory by label on AND, OR, and EXCLUDE all at once. 

The complex label logic can be saved to filter groups, and migrations will update pre-existing filter groups to the new setup. 

#### Remaining work

Before merging this into develop, the inventory list menu should be re-organized to make room for the two new rows, which take up a lot of screen space.

The inventory map page, which also uses label filtering, was left unchanged, since it doesn't (yet) use Filter Groups and will work fine without the complex label logic. But it can be easily updated if needed, since its own code is already based on the inventory list's.

#3371

(Ignore the missing icons)

image


seed/migrations/0212_move_filtergroup_labels.py
```diff
@@ -0,0 +1,40 @@
+# Generated by Django 3.2.18 on 2023-12-19 16:15
```
haneslinger, line 1: 💯

haneslinger (commented):
This is great! thank you! I have only a few notes, and a tiny commit that adds a test I would like to see passing. I'm happy to address them, if you like.

@kflemin, how to you want to handle that white space in the ui @ebeers-png mentioned above?

@ebeers-png, Have you noticed any different in query speed?

ebeers-png:
@haneslinger Thank you for the review and suggestions! Added the catch for erroneous data to the update method, and I was able to simplify _get_label_views() a little more as well. 

Re: query speed, I haven't really noticed a difference in inventory load times. But that's mainly because SEED calculates included/excluded building ids on the frontend, in inventory_list_controller (lines 1110-1128). That's not the case for the custom report insights though, which uses _get_label_views() in models/data_views to get the right buildings. So I'd have to check whether the old code (with `set.intersection(...)` everywhere) is faster than what you suggested (though I doubt it).

haneslinger:
The failing test is not this PRs fault. I'm addressing it now

Issue #3904, opened 2023-03-17
Unmerging records removes them from the cycle they belong to


RDmitchell:
From Veronique, Clearly Energy

Unmerging records in a cycle takes those newly separated records out of any cycles, which is a problem.

RDmitchell:
@isalanglois / @nllong -- this issue should get some priority, as it was submitted by Clearly Energy

RDmitchell:
@nllong / @isalanglois -- seems like this might have a fairly high priority since Clearly Energy reported it.

RDmitchell:
see also #3836

RDmitchell:
@kflemin -- based on our call today with BEAM, we need to figure out where we are with this issue.

PR #4452, opened 2023-12-20
Fix: Use old cycles during unmerge

haneslinger:
In contrast to https://github.com/SEED-platform/seed/issues/3904, here's the new workflow

Okay, so we've got a property present in two cycles, one has been merged into:
image
image

see?
image

Now if we unmerge the merged one:
image

We've one got one new property:
image


haneslinger:
@kflemin @axelstudios the failing test here imply this intented behavoir. why? Is it still?

kflemin:
I don't have any historical context for this one. @nllong? @axelstudios?   Basically, what we are trying to do here is: when we unmerge 2 properties, we want to associate one of the unmerged properties back to the "old" existing property and create a new property for the other unmerged property.  This is in contrast to what happens now:  when a property is "unmerged" into 2, it creates 2 new properties.

haneslinger:
@kflemin @axelstudios repinging this

kflemin:
**Resolution:**
- assign meters to both properties
- 1 of the unmerged properties should retain the history
- in the notes of the property that has no history, add a note "unmerged from X" so we still retain a link to the history (db table note)

haneslinger:
@kflemin In the note, what field should I use the reference the other Property?

kflemin:
> @kflemin In the note, what field should I use the reference the other Property?

Let's reference the property view ID (since users don't have access to property IDs directly).

kflemin (approved):

Issue #4453, opened 2023-12-20
Defining AH levels: Getting column name errors when the columns don't exist


RDmitchell:
instance: dev2
SHA: b3c6784d2
Org: LBNL 3033

I am trying to define Access Levels. 

I am setting this up on a completely new organization that I just created, I have not imported any files. 

I try to set up the first level, called "Accountable Organization Name", which is the name of a field in the file that I am GOING to import, but have NOT imported yet. 

But when I click the Save button, the program gives the error "are column names". 
![image](https://github.com/SEED-platform/seed/assets/6314950/ef357170-d04d-496d-9493-0edc406ccdc4)

Why does SEED think this is an already defined column, when I haven't imported any data yet? 

Also, I know I was told that the access levels couldn't be field names, but then I was also told that the access levels had to match the field names in the imported data, which makes complete sense to me in terms of mapping. 

So I am VERY confused !!

RDmitchell:
And now, when I click on Organizations / Column Settings (I think it was Settings not Column Mapping but now I am not completely sure -- it was one or the other), I am getting this error
![image](https://github.com/SEED-platform/seed/assets/6314950/e4aab4e9-a9c2-4179-80fe-4f73388a9969)

RDmitchell:
And now I can't get to the Inventory screen -- I am getting a "You don't have permission to perform this action" error message if I click on the Inventory nav button.
![image](https://github.com/SEED-platform/seed/assets/6314950/4d1c1100-8833-4fbd-a2d6-67a78daba04e)

I deleted all the access levels, but that didn't help
![image](https://github.com/SEED-platform/seed/assets/6314950/f0c0ef7d-2bc2-4cf4-a11f-e31641afb8ff)

haneslinger:
https://github.com/SEED-platform/seed/pull/4455

haneslinger:
Alright, so we have a lovely little onion of errors here.

1. The "are column names" error appears because the check is currently done cross org, and should be done interorg. https://github.com/SEED-platform/seed/pull/4455 addresses that.
2. That means _some org does_ have columns named "Accountable Organization Name", thats "Army - BPS".
3. idky about the column page error,  looking into it.

RDmitchell:
@haneslinger -- I just deleted my last 2 comments, because I think they were due to @anchapin working on the server.

I don't think they had anything to do with this issue. 

However, I am still confused about access level names and column/field names. 

[...]

PR #4455, opened 2023-12-21
Fix column names errors

haneslinger:


haneslinger:
tests fixed https://github.com/SEED-platform/seed/pull/4457

RDmitchell:
@haneslinger -- so does this mean issue #4453 should be set to either "Review" or "Ready to Deploy" so that at some point it can be moved to the dev server for testing?

haneslinger:
This definitely addresses part of https://github.com/SEED-platform/seed/issues/4453, but i think that ticket is about more.

kflemin (approved):

Issue #3328, opened 2022-06-23
Sensors: Add spinners or progress bars when displaying the data takes a long time, in particular just getting to the Sensors page


RDmitchell:
**Is your feature request related to a problem? Please describe.**

Right now in Sensors, when you create a data logger, add sensor fields or add actual sensor data, the program take a long time to refresh the page to see the changes, and the user doesn't know what's going on. 

This also happens when you refresh your browser

Here is a movie that shows the problem 
https://drive.google.com/drive/folders/1JlsRQYQNZXvhz0olGrPDlfV9RoxaGNjI?usp=sharing

**Describe the solution you'd like**

Let the user know that the program is still working on displaying the data, either with a spinner or a progress bar or something. 

**Describe alternatives you've considered**

spinner or progress bar

Instance: dev1
SHA: 5f181142b
Org: Bureau of Indian Education Test Account

axelstudios:
@RDmitchell Since we added pagination to the sensors page to speed up all actions is this issue resolved now?

RDmitchell:
@axelstudios -- I'll test this. Thanks

RDmitchell:
@axelstudios -- I'm not sure the new pagination has any affect on this. 

On dev, org == Cureau of Indian Education Test Account
![image](https://user-images.githubusercontent.com/6314950/195715017-fc74fd2c-7812-4f88-b472-5dd44fe11444.png)

There is only 1 Property in this org
![image](https://user-images.githubusercontent.com/6314950/195715179-d7d88985-441c-4585-b5ca-f7e083f20305.png)

To get to the Sensors section, go to the Detail view for that property, then click on Sensors at the top, to get to the Sensors data display. 

It takes 15-20 seconds for the program to display the Sensor data, which is where I was wanting to have the spinner, so the user knows that the program is doing something.
There are 10 data loggers
![image](https://user-images.githubusercontent.com/6314950/195715768-d754a586-f9f7-45e0-9618-6587356da592.png)
Each Data logger has 9 Sensors
![image](https://user-images.githubusercontent.com/6314950/195716168-b868db91-9f22-4157-baef-b7f942761a47.png)
Then there is the Sensor Readings section, which I see does have pagination. However, it doesn't seem to have helped speed up the display of the rest of the data on the Sensor page. 
It think it would still be nice to have a spinner when waiting to have the entire Sensor data displayed. 
[...]

PR #4461, opened 2024-01-04
Fix and optimize sensors

haneslinger:
vroom vroom https://github.com/SEED-platform/seed/issues/3328


haneslinger:
@axelstudios why aren't these passing.

PR #4464, opened 2024-01-05
Sensor reading performance improvement

axelstudios:
#### What's this PR do?
Removes a long, unnecessary `COUNT` query, which cuts query time by 40%

Browse to the sensors page for an org with, ideally, millions of readings

#3328

### DIFF
--- a/seed/utils/sensors.py
+++ b/seed/utils/sensors.py
@@ -51,8 +51,7 @@ class PropertySensorReadingsExporter():
         if self.showOnlyOccupiedReadings:
             sensor_readings = sensor_readings.filter(is_occupied=True)
 
-        # order by id **greatly** speeds this up (cause of indexing, I think
-        timestamps = sensor_readings.distinct('timestamp').order_by("timestamp", "id").values_list("timestamp", flat=True)
+        timestamps = list(sensor_readings.distinct('timestamp').order_by('timestamp').values_list("timestamp", flat=True))
         paginator = Paginator(timestamps, per_page)
         timestamps_in_page = paginator.page(page)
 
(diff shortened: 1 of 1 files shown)


seed/utils/sensors.py
```diff
@@ -51,8 +51,7 @@ def _usages_by_exact_times(self, page, per_page):
         if self.showOnlyOccupiedReadings:
             sensor_readings = sensor_readings.filter(is_occupied=True)
 
-        # order by id **greatly** speeds this up (cause of indexing, I think
```
haneslinger, line 54: you had me make a commit _just_ to add this comment, why are you removing it?

seed/utils/sensors.py
```diff
@@ -78,9 +77,7 @@ def _usages_by_exact_times(self, page, per_page):
                 for sensor in self.sensors
             }
 
-            sensor_readings = SensorReading.objects.filter(timestamp__range=[earliest_time, latest_time], sensor__in=self.sensors)
-            if self.showOnlyOccupiedReadings:
-                sensor_readings = sensor_readings.filter(is_occupied=True)
```
haneslinger, line 83: Why isn't this needed anymore?

seed/utils/sensors.py
axelstudios: Because that change is gone now in favor of a better one - also, you merged your PR before it was reviewed/approved, so I couldn't make the change there.

seed/utils/sensors.py
axelstudios: It's already handled at the top, no need to do it twice

haneslinger (approved):

PR #4468, opened 2024-01-08
Move program setup to modal on insights pages

kflemin:
To prepare for the portfolio (cross cycle) summary coming soon to the insights tab, the program setup page was moved to a modal accessible from both the property insights and program insight pages. 

Screen Shot 2024-01-08 at 4 31 10 PM

Screen Shot 2024-01-08 at 4 31 21 PM

Screen Shot 2024-01-08 at 4 30 53 PM

To test:
ensure that you can still CRUD programs from both the insights pages and that the page behaves property when dismissing the modal (for example, if you delete the program that is currently displayed on the page, it should select another program to display)

### DIFF
--- a/seed/static/seed/js/controllers/insights_program_controller.js
+++ b/seed/static/seed/js/controllers/insights_program_controller.js
@@ -12,15 +12,35 @@ angular.module('BE.seed.controller.insights_program', []).controller('insights_p
   'compliance_metric_service',
   'spinner_utility',
   'organization_payload',
+  'filter_groups',
   'cycles',
+  'property_columns',
   'auth_payload',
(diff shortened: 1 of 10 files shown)


perryr16 (changes requested):
Overall this is great and makes the workflow much simpler. 

One suggested change. When a program is selected in the modal, and the modal is closed, that program should be loaded on screen. Currently whatever program was displayed is reloaded.

kflemin:
@perryr16, I wasn't sure which way to go for that. I'll implement what you suggested!

perryr16 (approved):
Approved! 🏆

nllong (approved):
nice, modality!

Issue #4430, opened 2023-12-05
Inventory: Map: Display is painfully slow for 3000+ records


RDmitchell:

haneslinger:
@axelstudios  assigning part 2 to you

PR #4469, opened 2024-01-09
Retrieve all map data at once

haneslinger:
maps use to get data pages  like this
image
now it does it like this
image
yay!

Best speed up for https://github.com/SEED-platform/seed/issues/4430 will be view-location based filtering, but this is nice to.


kflemin (approved):
works great!

Issue #4308, opened 2023-10-26
Double refresh issue - Inventory List Filtering


kflemin:
Inventory List Filtering - Double refresh is still there. Reported by Alex.

RDmitchell:
@kflemin -- is this something I can test through the UI? Look at the PR, it seems like maybe something has to be done on the back end in order to see the problem, and/or the fix?

kflemin:
@RDmitchell, there's not really anything for you to test on this one...maybe just that the inventory page still loads for you after this fix.

RDmitchell:
@kflemin -- assuming this is deployed to dev1, I can load the inventory page

But I do see a very subtle second refresh of the data, very quickly after the first .. is that what this is about? 

I guess we should make sure this fix has been deployed to dev1?

RDmitchell:
instance: dev1
SHA: d5d60eaac
Org: LBNL Test 200

@kflemin -- When I click the browser refresh (Chrome), SEED displays the upper half of the screen, and then refreshes again and displays the upper half of the screen with the data table / tabs included. 

So there are 2 refreshes, but I'm not sure that is what this issue is about.

PR #4480, opened 2024-01-12
Fix inventory list double refresh on load

perryr16:
Within the `inventory_list_controller` there are 10 code blocks that call the `load_inventory` function, and only 4 that are called outside of a modal close.
    1. `filterUsingLabels`
    2. `update_cycle`
    3. `gridApi.core.on.filterChanged`
    4. `gridApi.core.on.sortChanged`

`filterUsingLabels` is responsible for the initial page load. If the ui-grid table has existing filters or sorts, their respective functions `gridApi.core.on.filterChanged` and `gridApi.core.on.sortChanged` will each be called leading to 3 possible table loads.

#### What's this PR do?
1. Combines the `.filterChanged` and `sortChanged` into a single call `filterOrSortChanged` as they use the same code.
2. On initial load, check for filters or sorts. If they exist, ignore the first call to `filterOrSortChanged`

- Place a `console.log('load_inventory')` statement in the function `load_inventory`. 
- Add a few sorts and filters. Refresh and track the number of logs.

Try to further adjust the table (pinned columns, hidden columns, label filters...) and make sure it still loads only once

#4308


nllong:
- [ ] Make sure to update the display name of the PR to be explicit

haneslinger (approved):
sick

Issue #4479, opened 2024-01-10
International Character Support


kflemin:
SEED currently strips international characters (such as French accents).  Need to support international characters

- requested by Open Tech

RDmitchell:
@axelstudios / @nllong -- don't know how to test this. 

And I guess it's not deployed to dev1 yet anyway.

nllong:
This is now on dev1, sorry for the delay/miscommunication!

RDmitchell:
Thanks Nicholas!!

dreneden1:
We'll test this asap! Thank you again :)

RDmitchell:
instance: dev1
SHA: d5d60eaac
Org: LBNL 245

created this spreadsheet
![image](https://github.com/SEED-platform/seed/assets/6314950/45493bc7-0247-472a-9c32-9b46c570e074)
Imported it 
Mapping -- characters were displayed properly
![image](https://github.com/SEED-platform/seed/assets/6314950/ed446e57-1f41-467b-92be-6c9e5706f1b0)
Mapping Review -- characters were displayed properly
![image](https://github.com/SEED-platform/seed/assets/6314950/b231f335-8a86-41a8-924b-5263c7086b3a)
List View -- characters were displayed properly
![image](https://github.com/SEED-platform/seed/assets/6314950/265536fb-c887-460b-ae29-dfc4ff9a0b3a)
Detail View -- characters were displayed properly
![image](https://github.com/SEED-platform/seed/assets/6314950/637cc90a-f0f7-4e58-8183-8780ed71e6d3)

@kflemin -- is this a good enough test? I am going to close it but you can always reopen it for more testing if these characters aren't enough.

PR #4484, opened 2024-01-15
Support unicode characters by replacing unidecode with new normalize method

nllong:
Back in [2016](https://github.com/SEED-platform/seed/pull/1145) we addd the unidecode library to fix unicode issues with the data. That worked well until now, where we need to keep diacritics/accent marks and further support the arabic character set. 

#### What's this PR do?
* Remove unidecode
* Create new method to normalize set of characters that would prevent reasonable matches (e.g., mdash, fancy quotes, etc).
* Use the unicodedata.normalize method to force unicode characters to combine the letter and diacritic together. Using the NFC (Normalization Form Composition) setting which has extended functionality.

* unit tests
* import unicode data (new test file forthcoming)
* UI testing by inserting unicode characters. A great test is to edit a matching field to insert a unicode character, then import a new dataset with that unicode character in the matching field.

#4479


seed/lib/mcm/cleaners.py
```diff
@@ -33,6 +34,46 @@
...
+    ord('•'): '*',
+    ord('⁄'): '/',
+    ord('×'): 'x',
+    ord('⁓'): '~',
+    # mdash, ndash, horizontal bar
+    ord('–'): '-',
+    ord('—'): '--',
+    ord('―'): '-',
+    ord('¬'): '-',
+    # guillemets?
```
nllong, line 56: @kflemin  -- should we normalize on these too?

kflemin: you could normalize them to single and double quotation marks

PR #4493, opened 2024-01-18
AH - update dialog text on Upload ALIs modal and collapse help section by default

kflemin:
- hide cycle msg on modal for ali import
- default access tree help section to collapsed
- update upload access levels dialog text to warn users that the import may take several minutes

Screen Shot 2024-01-18 at 3 34 26 PM


seed/static/seed/partials/data_upload_modal_progress.html
haneslinger, line 1: not your fault, this file is wild

haneslinger (approved):

PR #4495, opened 2024-01-19
Disallow manual merge of records with different alis

haneslinger:
try the merge two things in different alis?

image


RDmitchell:
what is an alis ?

haneslinger:
@kflemin, do you know why the tests are erroring? It says 
```TypeError: catching classes that do not inherit from BaseException is not allowed```
and I do create new Error but they inherit from Exception.

haneslinger:
> what is an alis ?

Access Level Instances

RDmitchell:
Thx !!

kflemin (approved):
🏆

Issue #4494, opened 2024-01-19
Unit conversion does not work for extra data fields


perryr16:
There are 4 different data types outside the normal float, int, string... These are eui, area, ghg, and ghg_intensity. PropertyState fields associated with these four types are given a pint unit and saved in the db as `<Quantity (value, 'unit')>`. Before displaying values on the Inventory List table the quantity values are converted based on org display unit settings. 

example: 
- the db stores 'Area' data as `ft**2`, 
- a property state has a `gross_floor_area = <Quantity(10, 'ft ** 2')>`
- if the org display unit for area is `ft**2` the `gross_floor_area` is displayed as `10` in the Inventory List table
- if the org display unit for area is `m**2` the `gross_floor_area` is displayed as `0.9` in the Inventory List table

Extra data fields are not given a pint unit and are saved in whatever format they were imported as (string or number). Because they don't have pint units they are ignored when the unit conversion takes place allowing data to be misrepresented.

RDmitchell:
instance: dev1
SHA: 3.0.0-beta.0.d60822140
Org: LBNL 345
Cycle Extra Data Units Testing

How should this be manually tested?

- upload property data with both canonical area columns and extra data area columns (or eui, ghg, ghg_intensity)
- on column settings change the data type of the extra area field to area
- on org settings, change the default display units of area from ft2 to m2
- ensure inventory list values are updated (this should also be visible on the mapping page of the import process)

What are the relevant tickets?
https://github.com/SEED-platform/seed/issues/4494

Testing Steps
* Imported file with extra data fields for Gross Floor area and Site eui
* Set the two extra data fields to have appropriate units
![image](https://github.com/SEED-platform/seed/assets/6314950/a696dd47-b18a-4a4a-9775-a6cdbad3d27f)
* Inventory list shows the new units for the extra data fields
![image](https://github.com/SEED-platform/seed/assets/6314950/3aa36445-d551-460d-9602-8cbf72c4ba11)
*Reimported the file and the units show up in the mapping
![image](https://github.com/SEED-platform/seed/assets/6314950/d1d5dbf4-a09c-4db5-8331-24a89213ca2a)

RDmitchell:
@perryr16 -- looks great !

PR #4496, opened 2024-01-19
Apply unit conversion to extra data

perryr16:
There are 4 different data types outside the normal float, int, string... These are eui, area, ghg, and ghg_intensity. PropertyState fields associated with these four types are given a pint unit and saved in the db as <Quantity (value, 'unit')>. Before displaying values on the Inventory List table the quantity values are converted based on org display unit settings.

Extra data fields are not given a pint unit and are saved in whatever format they were imported as (string or number). Because they don't have pint units they are ignored when the unit conversion takes place allowing data to be misrepresented.

#### What's this PR do?
Adds a pint unit to extra data columns that have been assigned a data type in column settings.

- upload property data with both canonical area columns and extra data area columns (or eui, ghg, ghg_intensity)
- on column settings change the data type of the extra area field to area
- on org settings, change the default display units of area from ft2 to m2
- ensure inventory list values are updated (this should also be visible on the mapping page of the import process)

#4494

Inventory List
Screenshot 2024-01-19 at 2 54 58 PM
Screenshot 2024-01-19 at 2 54 37 PM

Mapping Screen
Screenshot 2024-01-19 at 3 39 06 PM
Screenshot 2024-01-19 at 3 38 11 PM


perryr16:
Timing note:
properties `/filter` endpoint timing on develop: ~0.8 s
properties `/filter` endpoint with new changes: ~1.0 s

kflemin (approved):

Issue #4321, opened 2023-10-26
Program Setup - don't save invalid program


kflemin:
Program Setup should not save an invalid program after clicking the `New Program` button. Reported by Alex.

perryr16:
This should wait until #4468 is merged.

RDmitchell:
instance: dev1
SHA: d5d60eaac
Org: BPS Test

Modal seems to be working well 
![image](https://github.com/SEED-platform/seed/assets/6314950/dc4382f1-90d2-48dd-938a-8e14537e55b9)

PR #4499, opened 2024-01-23
Fix program setup's new program button action

perryr16:
The "New Program" button on the program insights config page would save invalid programs to the db on each click. 

#### What's this PR do?
New programs are only created when they have passed validation via the 'Save Changes' button

Go to Program Insights config modal and click "New Program" multiple times. Nothing should happen after the first click.

#4321


kflemin (approved):
works great...thanks for fixing!

Issue #4502, opened 2024-01-24
AH: Getting error adding label based on hierarchy filter


RDmitchell:
instance: dev2
SHA: a79e79785
Org: LBNL 406

I believe I have successfully imported records according to the AH I set up for this org. 

This is the root, so it can see all the records
![Image](https://github.com/SEED-platform/seed/assets/6314950/05d0f8ea-0d52-43ca-bcf7-0a9b17c7be5d)

I filtered on AH County for "Alameda" and 3 records were returned
![Image](https://github.com/SEED-platform/seed/assets/6314950/d3a58bc2-6bb5-4d6f-9bc4-b6aa372a579c)

I then checked those three records, and used the Action menu to add the label "Alameda County" to those records, and click the Create Label button. But the program returns an error
![Image](https://github.com/SEED-platform/seed/assets/6314950/d29ef198-21a2-46ff-8ff7-32c7114827f2)

However, it appears that the label was actually created. 
![Image](https://github.com/SEED-platform/seed/assets/6314950/9e038e79-8f3c-4519-8194-bcd735518f95)

RDmitchell:
instance: dev2
SHA: d30082bbc
Org: LBNL 406

filtering on AH City = El Cerrito (which results in all the records), then adding a label that I made on the fly called "El Cerrito", the label was properly applied without errors.

PR #4510, opened 2024-01-26
AH - Fix label creation UI

kflemin:
This PR resolves the bug described in #4502

As an org owner: user can create a new label and assign it to properties:
Screenshot 2024-01-26 at 12 36 07 PM
Screenshot 2024-01-26 at 12 36 13 PM

As an org owner: user can assign labels to the properties they have access to, but not create new labels:
Screenshot 2024-01-26 at 1 14 47 PM

### DIFF
--- a/seed/static/seed/partials/update_item_labels_modal.html
+++ b/seed/static/seed/partials/update_item_labels_modal.html
@@ -3,7 +3,7 @@
     <h4 class="modal-title" id="manageLabelsModalLabel" translate>Add/Remove Labels</h4>
   </div>
   <div class="modal-body">
-    <div class="newLabelInput" ng-if="is_ali_root" style="margin-top: 0">
+    <div class="newLabelInput" ng-if="is_ali_root" style="margin-top: 0" ng-if="is_ali_root">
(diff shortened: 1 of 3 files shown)


seed/static/seed/partials/update_item_labels_modal.html
```diff
@@ -3,7 +3,7 @@
     <h4 class="modal-title" id="manageLabelsModalLabel" translate>Add/Remove Labels</h4>
   </div>
   <div class="modal-body">
-    <div class="newLabelInput" ng-if="is_ali_root" style="margin-top: 0">
+    <div class="newLabelInput" ng-if="is_ali_root" style="margin-top: 0" ng-if="is_ali_root">
```
perryr16, line 6: looks like `ng-if="is_ali_root"` is already part of the "newLabelInput"

perryr16 (changes requested):

seed/static/seed/partials/update_item_labels_modal.html
kflemin: ah! sorry i missed that. i'll clean it up

perryr16 (approved):

PR #4512, opened 2024-01-31
Restrict leaf user permissions for goals

perryr16:
Per Sydney, partners should only have read access to goals.

#### What's this PR do?
- Adds frontend logic to prevent leaf users from accessing goal configs 
- Adds a backend decorator to prevent leaf users from performing write actions on goal views (create, update, delete)

- As root user create/update/delete goals 
- As a middle tree user (not root or leaf) create/update/delete goals
- As a leaf (partner) user attept to create/update/delete goals. Your permission should be denied.

- In a new org with a single user with an access level tree depth of 1, create/update/delete goals

Leaf
Screenshot 2024-01-31 at 12 41 03 PM
Root
Screenshot 2024-01-31 at 12 41 36 PM
Middle
Screenshot 2024-01-31 at 12 40 21 PM
Root & Leaf
Screenshot 2024-01-31 at 12 49 05 PM

### DIFF
--- a/seed/docs/views.py
+++ b/seed/docs/views.py
@@ -64,7 +64,7 @@ def faq_page(request):
                 faq_data[category_name].append(parsed_faq._asdict())
 
     if not request.user.is_anonymous:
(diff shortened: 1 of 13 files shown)


kflemin (approved):
👍

PR #4519, opened 2024-02-06
Handle ali links during import

haneslinger:
1. Edit `match_merge_link` so it takes it state instead of view
2. Edit `match_merge_link`so it checks ali info
3. Edit import process so it communicates links
4. Edit import process so it can deal with incoming records with no ali, that don't merge but link
5. Edit  import process so it doesn't link records with different alis
6. so many tests


kflemin (approved):
🔥

kflemin (approved):

kflemin:
@haneslinger you can merge this one when you are ready!

Issue #4144, opened 2023-07-14
Error if applying labels if they already exist (?)


RDmitchell:
Instance: production
SHA: de3b9d486
Org: ID = 407

From @nllong 

*****
@RDmitchell was demoing to Maryland. She tried to apply labels to, i’m guessing, buildings that already had the label applied. The error was:

web
File "/usr/lib/python3.9/site-packages/django/db/backends/utils.py", line 75, in _execute_with_wrappers
web
return executor(sql, params, many, context)
web
File "/usr/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
web
return self.cursor.execute(sql, params)
web
File "/usr/lib/python3.9/site-packages/django/db/utils.py", line 90, in __exit__
web
raise dj_exc_value.with_traceback(traceback) from exc_value
web
File "/usr/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
web
return self.cursor.execute(sql, params)
web
django.db.utils.IntegrityError: duplicate key value violates unique constraint "seed_propertyview_labels_propertyview_id_statusla_9136d976_uniq"
web
DETAIL: Key (propertyview_id, statuslabel_id)=(3336285, 6408) already exists.

looks like we need to check if the label is already applied?

RDmitchell:
This might have been a situation where the label was applied to the first 100 labels, and then I tried to apply it to the entire filtered set.

When I do this on an org with 7,000+ records, I get a 520 error
![image](https://github.com/SEED-platform/seed/assets/6314950/80ac252b-2a4f-42b1-9b6d-66244eb22748)

kflemin:
+1 for San Francisco!

RDmitchell:
@nllong -- can we bump this up in priority to fix it since SF is having this issue?

RDmitchell:
I added this to the Q2 bug fix project

kflemin:
@RDmitchell, have you run into this bug recently? I have tried to replicate it locally but am not able to...

nllong:
I've tried to recreate this too, but no luck. SF has definitely hit this issue too. I tried locally with and without celery, and tried running on production using SF data, but nope. I don't see how they are hitting this issue.

kflemin:
I have a potential fix i'll make a PR for...hopefully it prevents the error in the future.

RDmitchell:
[...]

PR #4523, opened 2024-02-09
ignore duplicates when adding labels

kflemin:
ignore "duplicate errors" when adding labels to properties.
related issue: #4144


nllong (approved):
might help...

Issue #4106, opened 2023-06-20
Filtering: Invalid data type error


RDmitchell:
Instance: seedv2
SHA: de3b9d486
Org: BPS Test - Robin

When trying to filter on Gross Floor Area, to include > and <, SEED displayed an error. But it still performed the correct filtering, and the second time I did that filter in the same session, the errors did not appear. 
![image](https://github.com/SEED-platform/seed/assets/6314950/f0dbecbe-e4cd-4365-91a2-a7de786d4af8)

Here is how Gross Floor Area is defined in Column Settings
![image](https://github.com/SEED-platform/seed/assets/6314950/c41e319a-9bd2-4ba1-9251-fd0829998abb)

RDmitchell:
Instance: dev1
SHA: c8a8e06fc

This is still an issue

RDmitchell:
@kflemin -- I added this to the Q2 bug fix if there is time. Thx

RDmitchell:
Instance: dev1
SHA: 
Org: BPS Test

No error when creating the multi-criteria filter !!

![Image](https://github.com/SEED-platform/seed/assets/6314950/426e9c39-bb09-4b85-a4ed-4e56e602d817)

Issue #3970, opened 2023-04-07
Filtering: Type > or < before entering number sometimes gives an error


RDmitchell:
Instance: dev1
SHA: c94869068 (2.18.0)

When filtering using > and <, if you type just one of those characters, before typing a number, SEED produces an error. 

If you type fast enough, ie, > 200, then the error is not displayed. 

Seems like it would be nice to not have that error appear, if the user (me in this case) was pondering what range to filter on (!)

![image](https://user-images.githubusercontent.com/6314950/230683567-8fde5f2c-1f79-4cc5-87c4-a238820074aa.png)

RDmitchell:
see #4106 which is closed -- same issue

PR #4527, opened 2024-02-13
Inventor List - Ignore incomplete filters containing just an operator

kflemin:
Do not display an error to user when a filter only contains an operator (>, <, =, etc.). Assume that the user is still typing.

Type an operator in a filter field on the inventory list page and ensure there is no red notification that pops up.

#3970, #4106


perryr16 (approved):
Simple fix!

A future task could be to prevent the inventory from reloading if a filter is an opreator

Issue #4418, opened 2023-11-27
Map: bubble showing the property name or address is very far from the point


RDmitchell:
Instance: seedv2
SHA: 0dbce5ad6

When looking at the Inventory Map, and clicking on a point, the map is very far away from the point. The arrow should be right on the point so that it is easy to see what the value is. Right now it is so high that it is off the map, and you have to move the map to see the entire text

![image](https://github.com/SEED-platform/seed/assets/6314950/40b5d501-cd89-4bd1-b940-f5e36988e7b8)

Moving the map, the descriptor bubble is still very far away from the point. 
![image](https://github.com/SEED-platform/seed/assets/6314950/ae5e2dc9-750d-416c-b212-7f04528b2552)

RDmitchell:
Seems like we might want to fix this for the EEEJ analysis, etc.

RDmitchell:
@kflemin -- if this is not hard to do, I think it would be nice to fix this in the bug round

haneslinger:
@axelstudios you did this page right? Can I give this to you?

PR #4528, opened 2024-02-13
Inventory Map - Fix positioning of the popup tooltip to be closer to the pin

kflemin:
#### What's this PR do?
Moves the tooltip closer to the pin on the map

#4418 

used to be like this:
Screenshot 2024-02-13 at 3 08 56 PM

now is like this:
Screenshot 2024-02-13 at 3 08 03 PM


perryr16 (approved):
🥇

PR #4535, opened 2024-02-15
Fix AH/develop migration order

perryr16:
#### What's this PR do?
reorders migrations


axelstudios:
Cherry-picked into `AH/develop` 👍

Issue #4093, opened 2023-06-09
Add a warning when you try to create a suborg


haneslinger:
It's going away soon

RDmitchell:
@haneslinger -- is this issue still relevant?

RDmitchell:
@haneslinger -- is this issue still relevant?

RDmitchell:
@haneslinger -- is this issue still relevant? Or has it been resolved?

haneslinger:
still relevant.

RDmitchell:
instance: dev2
SHA: 9495faa33
Org: BPT Test

@kflemin -- I was able to make a sub org and I didn't see the deprecation message ... Wonder if I did something wrong? This is for an org without any AH defined -- does that make a difference? 

![image](https://github.com/SEED-platform/seed/assets/6314950/aa375a85-b4a1-488e-abed-bc2ebe5c5cfb)

kflemin:
I don't see it either...it must not have gotten deployed yet. sorry!

RDmitchell:
@kflemin -- should we change the status to "Ready to Deploy", rather than test?

kflemin:
@RDmitchell, this is now on dev2!

RDmitchell:
Instance: dev1
SHA: 
Org: LBNL 200

Good message!
![Image](https://github.com/SEED-platform/seed/assets/6314950/464d0510-ff52-4bbd-ba74-63ca146fc4de)

PR #4536, opened 2024-02-15
AH - Add deprecation warning when user creates a Sub Organization

kflemin:
Add a warning to the create sub organization modal
#4093 

Screenshot 2024-02-15 at 3 21 07 PM


perryr16 (approved):
approved!

RDmitchell:
@kflemin -- this is probably only on dev2 right now, right?

RDmitchell:
instance: dev2
SHA: 9495faa33
Org: BPT Test

I was able to make a sub org and I didn't see the deprecation message ... Wonder if I did something wrong? This is for an org without any AH defined -- does that make a difference? 

image

PR #4537, opened 2024-02-15
Reject import file if it contains braces

perryr16:
#### Background
Curly braces `{ }` are an angular reserved operator used to pass variables from js to html, if braces are included in incoming file headers angular becomes confused and misrepresents data.

#### What's this PR do?
Rejects import files if the headers contain curly braces `{ }`. 
Invalid files will be rejected before the mapping screen.

example invalid header file (any file with braces in a header will do)
[12props_special_characters.xlsx](https://github.com/SEED-platform/seed/files/14312847/12props_special_characters.xlsx)

#4511

Screenshot 2024-02-15 at 3 59 16 PM


kflemin (approved):
works great!

PR #4538, opened 2024-02-17
Update the default dates on data quality rules

nllong:
When creating a new org, the default dates for a few of the rules (e.g., sale_date, release_date, year_built) where from 2020 or 2019.

#### What's this PR do?
* Update so when creating a new org, the default rules will be 2024.

* Create a new organization
* Go to the Data Quality page
* Verify that the fields with dates and years are now 2024.

n/a

### DIFF
--- a/seed/models/data_quality.py
+++ b/seed/models/data_quality.py
@@ -203,7 +203,7 @@ class Rule(models.Model):
             'data_type': TYPE_DATE,
             'rule_type': RULE_TYPE_DEFAULT,
             'min': 18890101,
-            'max': 20201231,
+            'max': 20241231,
             'severity': SEVERITY_ERROR,
             'condition': RULE_RANGE,
         }, {
@@ -232,7 +232,7 @@ class Rule(models.Model):
             'data_type': TYPE_DATE,
             'rule_type': RULE_TYPE_DEFAULT,
             'min': 18890101,
-            'max': 20201231,
+            'max': 20241231,
             'severity': SEVERITY_ERROR,
             'condition': RULE_RANGE,
         }, {
@@ -241,7 +241,7 @@ class Rule(models.Model):
             'data_type': TYPE_DATE,
             'rule_type': RULE_TYPE_DEFAULT,
             'min': 18890101,
-            'max': 20201231,
+            'max': 20241231,
             'severity': SEVERITY_ERROR,
             'condition': RULE_RANGE,
         }, {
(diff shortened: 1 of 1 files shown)


RDmitchell:
@nllong -- is there a way to have the program figure out what the current year is and make that default, so in 2025 we don't have to change it by hand?

nllong:
> @nllong -- is there a way to have the program figure out what the current year is and make that default, so in 2025 we don't have to change it by hand?

Yes, we can do this, but it require a bit more rework of the defaults section. I'd be happy to do it, but this fixes the issue for now.

nllong:
nvm, I just updated it to be dynamic...

RDmitchell:
@nllong -- thanks !!

kflemin (commented):
@nllong, what about the default cycle that gets created with a new org? 
Screenshot 2024-02-19 at 3 47 42 PM

kflemin (approved):

nllong:
> @nllong, what about the default cycle that gets created with a new org? Screenshot 2024-02-19 at 3 47 42 PM

good idea, I can add this update too!

nllong:
> > @nllong, what about the default cycle that gets created with a new org? Screenshot 2024-02-19 at 3 47 42 PM
> 
> good idea, I can add this update too!

Actually, this is already done... it is just the current_year -1. Which is okay, right? Or do we want this to be just the current year? It is strange because most cycles are mid year, (June 30). So we will always be off half the year.  I say we leave it for now.

RDmitchell:
instance: dev1
SHA: bcbec0272
Org: LBNL 633

I made a new org and looked at the dates in the DQ world, and they seem to be still maxed out at 2020, with the starting dates being **1889** ???

image

@kflemin -- seems like maybe this hasn't been deployed to dev1 yet?

axelstudios:
@RDmitchell This fix is deployed to dev1 currently, and appears to working as expected
image

RDmitchell:
Instance: dev1
SHA: 77b9e4afa

When I make a new rule, and set the type to Date, then click the calendar, April 2024 appears
image

@nllong -- Is that what this fixes? For existing rules, nothing changes

image

kflemin:
@RDmitchell it was my understanding that this was for new rules.  I don't think we can auto-update existing rules.

RDmitchell:
@kflemin -- Ah, yes, that makes sense. 

I will close the issue

Issue #4045, opened 2023-05-12
Allow owners to change the ALI of users below them.


haneslinger:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

RDmitchell:
@haneslinger -- is this issue still relevant?

haneslinger:
kind of, @nllong, If I'm a member who's not in the root, should I have the ability to CRUD the users beneath me?
@kflemin and I had imagined a new role "user manager", who is a member that can CRUD users but can't change org stuff the owner can. That's sorta something that's been put on the back burner, but I think we still want it?

PR #4552, opened 2024-02-28
Allow change of ali

haneslinger:
[#4045](https://github.com/SEED-platform/seed/issues/4045)
image


kflemin (commented):
this isn't working for me...it looks fine in the UI and I see no errors in the console, but if I refresh the page after making a change (and saving the change), the values revert back to their original values.

haneslinger:
@perryr16 @axelstudios How do I fix the front end tests?

perryr16:
The issue seems related to simultaneous promise execution. By calling `update_role` and `update_ali` back to back the `user.access_level_instance` can be overwritten to its original if `update_role` is slower than `update_ali`. 

The simplest way to fix this is to call them asyncronously
```
    $scope.update_user = (user, user_edits) => {
      // 1. update role 
      // 2. update ali
      organization_service.update_role(user.user_id, $scope.org.id, user_edits.role)
        .then(() => {
          refreshRoleStatus();
          return organization_service.update_ali(user.user_id, $scope.org.id, user_edits.access_level_instance.id)
        })
        .catch((data) => {
          $scope.$emit('app_error', data);
        })
    }

```
I'd be a bit of a refactor but I'd consider merging this logic into a single `OrganizationUser` update endpoint.

PR #4554, opened 2024-02-29
Handle empty cycles on portfolio summary

perryr16:
#### What's this PR do?
Handles the case when the current cycle (or baseline cycle) has no data.

`Note:`
Things break down if baseline and current cycles have no overlapping data. The property filter can only filter one cycle at a time, meaning the second filter request uses the property ids from the first. If there are no overlapping properties the second filter will return empty. Fixing this would be very difficult.

create a goal with an empty baseline cycle
create a goal with an empty current cycle 
create a goal with empty current and baseline cycles

Current Cycle Empty
Screenshot 2024-02-28 at 5 28 36 PM
Baseline Cycle Empty
Screenshot 2024-02-28 at 5 28 08 PM
Both Cycles Empty
Screenshot 2024-02-28 at 5 27 32 PM


kflemin (approved):

RDmitchell:
instance: dev2
SHA: f3d92cbff
Org; BB Full Run

I set up a Portfolio Summary where botht eh Baseline and Current Cycles are empty and the program doesn't show any data, which seems appropriate, and doesn't give an error, which also seems appropriate. 

image

So I am going to close this issue

PR #4558, opened 2024-03-07
remove eol Docker version.

dhaley:
#### What's this PR do?

Adapt buildspec.yml and Makefile so that codebase can be built by multiple Codeubild jobs to produce Docker image artifacts in multiple repos.

Adds support for optionally using HTTP2 with environment variable at docker build time.

Test builds in Codebuild.

`	docker build -t $(REPO):$(TAG) \
	  -f Dockerfile \
      --build-arg NGINX_LISTEN_OPTS=$(NGINX_LISTEN_OPTS) \
		./
`

New allocation for seedcerl deployment in Stratus Cloud.


axelstudios:
@dhaley I don't think the `http2` directive will do anything here, since HTTP/2 only works over HTTPS. This Nginx config is only used for uWSGI and special handling for a few routes - We have a separate Nginx container for HTTP/2 with SSL termination and security headers, but when we're not using that we just use CloudFlare in front of this config for HTTP/3

PR #4561, opened 2024-03-08
Allow sensor deletion

haneslinger:
I also moved the sensors endpints/tests

image


kflemin (approved):
I fixed styling on the data logger PR #4563, so merge that one first before this one.

kflemin:
@haneslinger, could you resolve these conflicts so we can merge it in? thanks!

PR #4562, opened 2024-03-11
Allow data logger update

haneslinger:
image


RDmitchell:
@haneslinger -- is there an issue that this PR belongs to?

PR #4571, opened 2024-03-18
Feature/upgrade codebuild http2

dhaley:
#### Support for multiple project builds and optional HTTP2 

Adapt buildspec.yml and Makefile so that codebase can be built by multiple Codeubild jobs to produce Docker image artifacts in multiple repos.

Adds support for optionally using HTTP2 with environment variable at docker build time.

Test builds in Codebuild or manually.

	docker build -t $(REPO):$(TAG) \ -f Dockerfile \ --build-arg NGINX_LISTEN_OPTS=$(NGINX_LISTEN_OPTS) \ ./

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,12 @@
+ARG NGINX_LISTEN_OPTS
+
 # AUTHOR:           Clay Teeter <teeterc@gmail.com>, Nicholas Long <nicholas.long@nrel.gov>
 # DESCRIPTION:      Image with seed platform and dependencies running in development mode
 # TO_BUILD_AND_RUN: docker-compose build && docker-compose up
 
 FROM alpine:3.14
+
+ARG NGINX_LISTEN_OPTS
 
 RUN apk add --no-cache python3-dev \
         postgresql-dev \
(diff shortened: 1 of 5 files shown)


axelstudios:
Copying this comment from the previous PR:

> I don't think the http2 directive will do anything here, since HTTP/2 only works over HTTPS. This Nginx config is only used for uWSGI and special handling for a few routes - We have a separate Nginx container for HTTP/2 with SSL termination and security headers, but when we're not using that we just use CloudFlare in front of this config for HTTP/3

dhaley:
> Copying this comment from the previous PR:
> 
> > I don't think the http2 directive will do anything here, since HTTP/2 only works over HTTPS. This Nginx config is only used for uWSGI and special handling for a few routes - We have a separate Nginx container for HTTP/2 with SSL termination and security headers, but when we're not using that we just use CloudFlare in front of this config for HTTP/3

Alex, technically HTTP2 works without SSL termination in the AWS ECS environment as we have a lot of sites using that configuration.  The advantage of having NGINX configured with HTTP2 when the TargetGroup is also HTTP2 is that no errors are generated in the TargetGroup Healthcheck.  Without nginx being configured for HTTP, then the TargetGroup Healthcheck receives a 400 error causing the healthcheck to fail.

The default on the container build is still to just use HTTP1.1 so this would be a non-breaking change.  But the Docker ARG is needed for the Stratus Environment to allow the Healthcheck to pass now that all of our TargetGroups and Load-balancers have been upgraded to default to HTTP2.

I can supply more info or screenshots if you want to see more specifics of the Stratus production environment.  If this change is controversial, I could break this pull request into two requests since there is also a commit to support multiple projects using this codebase in Codebuild.  But taking the HTTP2 support out of this request will cause me to have to do significant work for the seedcerl project to create TargetGroups and load-balancers in Stratus that are downgraded to HTTP1.1.

axelstudios:
@dhaley Gotcha, that's good enough for me if this is something that tangibly benefits the ECS setup

kflemin (approved):

Issue #4569, opened 2024-03-15
AH - hide member details from leaf users


kflemin:
on the organization members page, do not show non-root members & viewers other people's info.  The members page should only show the members within their ALI.  They should not be able to see details about root-level members.

PR #4572, opened 2024-03-18
Limit member visibility to users within access level instance

perryr16:
ICF requested that the members page should only show users that are at their own level or below.
i.e. starbucks partners should only be able to see starbucks partners, not admin users, and not other partner users
#### What's this PR do?
Uses ALIs to filter out users with higher permissions

create a few members at different access levels. 
Log in and navigate to 'members' page. Only users in your ali or below are visible.

#4569
Admin
Screenshot 2024-03-18 at 3 02 46 PM
Sector Level
Screenshot 2024-03-18 at 3 00 48 PM
Partner Level
Screenshot 2024-03-18 at 3 02 03 PM


kflemin (approved):
🏆

Issue #4575, opened 2024-03-19
Bug - Menu org change does not redirect to new org


perryr16:
### Issue 
If a user changes their org from the menu org selection 
![Image](https://github.com/SEED-platform/seed/assets/58446472/adb25813-da05-44b6-b84a-c21554abb624)

the url does not update and the user is left looking at the original organization data, not the newly selected organization data.

### Suggested action: 
If a new org is selected and the url contains an org_id argument, update the argument which should trigger a page reload with the correct data.

RDmitchell:
Instance: dev1
SHA: 3.0.0-beta.0.d60822140

I am not seeing the fix. 

I log in as lbnl, with lots of associated orgs
![Image](https://github.com/SEED-platform/seed/assets/6314950/451b4aac-8e55-4ce8-9e86-d1641353f6da)

 I click on another org
![Image](https://github.com/SEED-platform/seed/assets/6314950/4a641572-30dd-4371-941c-fc3087df94ae)

I am still on lbnl
![Image](https://github.com/SEED-platform/seed/assets/6314950/c26faba8-9ce7-4864-b8f3-eee2b3bba54b)

perryr16:
@RDmitchell that may be a different bug. 

This ticket refers to when you are on an org setting page, like "Cycles" and you change org via the menu. The cycle page should update with the new data and the url updated with the new org id (/accounts/<org id>/cycles).

My dev1 access doesn't have multiple orgs, so I can only test on dev2. Below is my testing procedure.

https://github.com/SEED-platform/seed/assets/58446472/8c393605-0250-46f3-8d21-6d8e1ff06a40

RDmitchell:
Instance: dev1
org: lbnl

@perryr16 -- I can't even change my org to see the cycles for different orgs. So it doesn't seem like I can even test this

I will make a separate issue for that.

RDmitchell:
the new issue where I can't change orgs from any location in SEED
#4618

PR #4576, opened 2024-03-19
Redirect on org change if necessary

perryr16:
When changing the organization from the menu, the page did not redirect to that org's data if the url contained an `organization_id` argument.

#### What's this PR do?
checks for the presence of the `organization_id` arg and updates it if necessary

With multiple orgs navigate to any of the organization settings pages. Change the org from the dropdown. The page data should reflect the newly selected org.

#4575

Screenshot 2024-03-19 at 1 17 45 PM


kflemin (approved):
I love this so much!! thank you! 🚚

Issue #4305, opened 2023-10-26
Delete Button not working after typing text - Inventory List Filtering


kflemin:
Inventory List Filtering - Label filter allows you to type in random text, and the delete button doesn't do anything if you've typed text

PR #4591, opened 2024-03-27
Clear invalid label tags

perryr16:
The label filters (and, or, exclude) are based attached to the model via a `<tags-input>` and `<auto-complete>` elements. If a valid label is selected, its value is attached to the `<tags-input>`'s ng-model. If invalid text is entered, it is not attached to the ng-model and not available to reset in $scope. This makes clearing the text difficult.

#### What's this PR do?
After resetting the ng-model's labels, the trashcan icon will also re-render the `<tags-input>` element, clearing any remaining text. 

It's a little hacky, but it's cleaner than trying to locate the correct HTML element within <tags-input>, clear its contents, and convince the element that it is now in a 'valid' state.

Add a few labels. Filter on those labels. Add some random text to the filter fields. Hit the trash icon.

#4305 

https://github.com/SEED-platform/seed/assets/58446472/5b71532b-5f35-4ce6-adc6-066d1fbcbaba


nllong (approved):
worked for me!

Issue #3986, opened 2023-04-19
Enable filtering of columns that appear in both inventory types


RDmitchell:
Instance: dev1
SHA: https://github.com/SEED-platform/seed/commit/8dc72bf0e2983bb470c887b10d4e59461754ab2b
Org: LBNL Test 200

**Property tab:**

**DOES NOT WORK:**
* Filtering on Jurisdiction Tax Lot ID (and other Tax Lot fields) for "" (blank) -- it shows no records
* Can't sort Tax Lot fields
* Filtering on ; (semicolon) in Jurisdiction Tax Lot ID (to see records with multiple tax lots) returns no records, when it should return 3
* This is useful so that you can see all the records that have more than one Tax Lot associated with a property* 

**Tax Lot tab**

**DOES NOT WORK:**
* Filtering on Jurisdiction Tax Lot ID for "" (blank) or !"" (not blank) fields
* Sort on Property Fields
* Filtering on ; (semicolon) in PM Property ID (to see records with multiple tax lots) returns no records, when it should
  This is useful so that you can see all the records that have more than one property associated with a tax lot

RDmitchell:
See history of this issue from previous issue #3499

PR #4600, opened 2024-03-29
Enable blank filters for other inventory columns

perryr16:
Filtering for blank values on the inventory list does not work for the other inventory type.
ex: on property list, filter for properties with no jursidiction_taxlot_id

#### What's this PR do?
- Determines if a blank filter in the other inventory column exists
- Return the combination of -views that do not have related `taxlotproperty`s AND -views that matched the blank filter

With related taxlots and properties, filter for blanks on the other inventory columns

#3986
This PR does not address semicolon filtering, which is also specified in the ticket

https://github.com/SEED-platform/seed/assets/58446472/84d6d6c6-5718-4620-8c7a-8e873687d424


kflemin (approved):
works great!

Issue #4429, opened 2023-11-30
Derived Column page fails silently when incorporating other derived columns


axelstudios:
**Describe the bug**
Attempting to create a derived column using _another_ derived column fails silently when clicking `Create`.

**Expected Behavior**
Errors should be reported to the user, and existing derived columns should not be allowed to appear in the column selection list

**Actual Behavior**
Silent failure with a console error pointing to `check_for_circular_definition`

**Steps to Reproduce**
1. Create a derived column
2. Attempt to create a second derived column that incorporates the previous one

perryr16:
Hey @axelstudios, I agree there is a bug here, but I wanted to check on the presence of derived columns in the column selection. The  feature was requested via [this ticket](https://github.com/SEED-platform/seed/issues/2760), so I'm thinking we leave it in and improve the error handling. Thoughts?

PR #4601, opened 2024-04-01
Fix circular derived column error handling

perryr16:
#### What's this PR do?
- fixes a bug causing the app to fail silently when a new derived column referenced an existing derived column.

- create a derived column, `dc1`
- create a second derived column, `dc2 = dc1`. Initially this failed silently, however this should be a valid action.
- edit `dc1` to `dc1 = dc2`. This should cause a circular warning to the user.

#4429 

https://github.com/SEED-platform/seed/assets/58446472/58a9bc52-5c28-4483-95e9-4bbca99eea56


kflemin:
@perryr16, we do not want to allow derived columns that are made up of other derived columns.

perryr16:
@kflemin that goes against the feature that was introduced in #3305. Is the feature no longer needed?

perryr16:
This should be reworked to prevent users from defining derived columns using other derived columns

PR #4610, opened 2024-04-04
Geocode UBIDs using centroid

axelstudios:
Prior to this change, when importing a file with a UBID we were setting the PropertyState's latitude and longitude to the center of the UBID _**bounding box**_, which isn't necessarily representative of the footprint.

#### What's this PR do?
- Sets the PropertyState's latitude and longitude to the center of the UBID _**centroid**_.
- This PR also fixes floating-point rounding errors for lat/long values when geocoding from a UBID by rounding to 12 decimal places. There's no risk of loss of precision, because at its highest precision the area encompassed by the coordinate would be _12,392 square nanometers_ at the equator.

1. Upload a file with a UBID like `85FPPRRC+XH9-7-8-3-6`
2. Confirm that the lat/long is `39.7424125, -105.178515625`, NOT `39.7424625, -105.178484375`

The red dot represents the lat/long SEED was setting, the center the centroid (blue pin) is what is now used:
image

### DIFF
--- a/seed/tests/test_utils_ubid.py
+++ b/seed/tests/test_utils_ubid.py
@@ -107,7 +107,7 @@ class UbidUtilMethods(TestCase):
             self.assertAlmostEqual(coord[0], known_property_centroid[index][0])
             self.assertAlmostEqual(coord[1], known_property_centroid[index][1])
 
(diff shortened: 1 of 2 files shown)


nllong (approved):
sweet, that fixed it

image

PR #4611, opened 2024-04-04
Fix mapping previews for files with a single column

axelstudios:
We've had a bug _for a long time_ that prevented seeing the first five rows during mapping of a file with a single column of data

#### What's this PR do?
Fixes the five-row preview when mapping a file with a single column of data

1. Create a file with a single column of data, e.g.:
    ```csv
    Custom ID 1
    120243 E True Lane
    95373 E Peach Avenue
    95864 SW Cottonwood Court
    155371 W Cherry Loop
    180821 S Tyler Lane
    ```
2. Upload the file and go to the mapping page
3. The first-five-rows preview should show the full values, not just the first letter of each

Before and after:

Single-Column-Import


kflemin (approved):
👍

Issue #4606, opened 2024-04-02
Inventory List / Map: UBIDs not displaying


RDmitchell:
instance: dev1
SHA: 77b9e4afa

When importing data with valid UBIDs, the UBIDs are not being displayed on the Inventory  List / Map view. 

However, they are displaying on the Inventory Detail view

Here is an example where the UBID is valid as shown in the PNNL UBID tool
![image](https://github.com/SEED-platform/seed/assets/6314950/a9d2a304-8706-4a2a-9942-d24f8d09816f)

But it is not being displayed in the Inventory List Map view
![image](https://github.com/SEED-platform/seed/assets/6314950/a120c624-f696-45bb-bc42-6e56818ad91c)

But is is displaying in the Inventory Detail view for the same property
![image](https://github.com/SEED-platform/seed/assets/6314950/46fe17cb-3c7a-426a-8aa1-bdd816cc020a)

perryr16:
### Background
The issue appears to be related to cycle selection on the `/inventory_map_controller.js`

The map calls the `inventory_service.get_properties` function, which hits the backend `/properties/filter` endpoint. If no cycle is specified in the request, then the first cycle for the org is assigned. 

The map controller calls the `get_properties` function with `undefined` as the cycle argument, leading to incorrect cycle data being displayed.


### Suggested Changes
```
-       const pagination = await fn(1, chunk, undefined, undefined).then((data) => data.pagination);
+       const pagination = await fn(1, chunk, $scope.cycle.selected)cycle, undefined).then((data) => data.pagination);


-       const page_promises = page_numbers.map((page) => fn(page, chunk, undefined, undefined).then((data) => {
+       const page_promises = page_numbers.map((page) => fn(page, chunk, $scope.cycle.selected_cycle, undefined).then((data) => {
```

perryr16:
@RDmitchell -- I'm hoping you can test something for me. My assumption is that the map is displaying the properties from the alphabetically first cycle in the org. Can you confirm that your alphabetically first cycle does not have properties with ubids? If it doesn't could you either add a few ubids to that cycle, or change the name of a cycle with ubids to appear first alphabetically.

RDmitchell:
Instance: dev1
SHA: 3a0fb4476
Org: LBNL 245

UBID rectangle now displaying in Inventory List Property tab view (as well as in the Detail view)
![Image](https://github.com/user-attachments/assets/cddc2ba0-d6a6-4598-8628-7e814d13cda7)

PR #4612, opened 2024-04-05
Fix merge modal for all cycles

axelstudios:
When requesting filtered properties with the `/api/v3/properties/filter/` endpoint, if you leave out the `cycle` param the backend defaults to the first alphabetical cycle

#### What's this PR do?
- Makes the cycle id explicit, so that when you're on the Inventory List page, select records, and open the merge modal, it will now work for all cycles and not just the first alphabetical cycle
- Fixes the cycle toggle on the map page, it was also affected by this issue
- Adds a fallback on the front-end, such that if a cycle isn't passed when filtering properties it will use the most-recently selected cycle

1. In an org with multiple cycles, go to the Inventory List page and select any cycle that is not alphabetically sorted first
2. Select a few records, and then open the Merge modal
3. Ensure that the data appears as expected

1. Go to the Inventory Map page
2. Toggle between different cycles, ensure that different (and correct) data appears for each cycle

### DIFF
--- a/seed/static/seed/js/controllers/inventory_list_controller.js
+++ b/seed/static/seed/js/controllers/inventory_list_controller.js
@@ -648,7 +648,7 @@ angular.module('BE.seed.controller.inventory_list', []).controller('inventory_li
             const data = new Array(selectedViewIds.length);
 
             if ($scope.inventory_type === 'properties') {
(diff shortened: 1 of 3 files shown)


perryr16 (approved):
super

perryr16:
This PR fixes #4606

PR #4613, opened 2024-04-05
Fix ubid bugs

perryr16:
This is in response to some bugs Robin encountered while testing the UBID work

#### What's this PR do?
- Fixes a bug that caused the import process to hang if ubid was the only matching criteria 
- Fixes a bug that did not preserve access level instances on merges

The following test files include the NREL FTLB, the NREL Cafe, and an invalid property without a ubid. Each file increases the ubid footprint.

1. set matching criteria to only ubid
2. upload a file with ubids [3props_ubids_0.xlsx](https://github.com/SEED-platform/seed/files/14910546/3props_ubids_0.xlsx)
3. in org settings change the ubid threshold to 0.1 to allow matching 
4. upload a file with slightly different ubids [3props_ubids_1.xlsx](https://github.com/SEED-platform/seed/files/14910547/3props_ubids_1.xlsx)
5. properties should have been merged and preferred ubid set to the incoming ubid (see inventory detail ubid)
6. change the org setting ubid threshold to something more strict, like 0.9 
7. upload a file with slightly different ubids [3props_ubids_4.xlsx](https://github.com/SEED-platform/seed/files/14910548/3props_ubids_4.xlsx)
8. new views have been created as incoming properties did not meet matching criteria 
9. change the org setting ubid_threshold back to 0.1 
10. upload a file with slightly different ubids [3props_ubids_5.xlsx](https://github.com/SEED-platform/seed/files/14910549/3props_ubids_5.xlsx)
11. all properties with overlapping ubids should have been merged and ubid history preserved 
12. unmerge a merged property, check ubid history for each property

### DIFF
--- a/seed/data_importer/match.py
+++ b/seed/data_importer/match.py
@@ -843,20 +843,19 @@ def check_jaccard_match(ubid, state_ubid, ubid_threshold, matching_criteria):
     """
     Use jaccard index between an incoming ubid and an existing state_ubid to determine if states are 'matching'
 
-    :param ubid: sring, incoming ubid
+    :param ubid: string, incoming ubid
     :param state_ubid: string, existing state's ubid
     :param ubid_threshold: float, organization's ubid_threshold
     :param matching_criteria: dict, organization's matching criteria with ubid removed
     """
-    # If state_ubis is None and ubid is the only matching_criteria, no match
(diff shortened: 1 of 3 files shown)


kflemin (approved):
🎉

PR #4619, opened 2024-04-09
Speed up get labels

haneslinger:
speeds up `/api/v3/taxlots/labels\` and `/api/v3/properties/labels` by:

- bulking and the serializer
- computing `is_applied` in a constant instead of linear number of queries
- changing `if inventory` to `if inventory != []:` (probably the most impactful)

With the CERL data, it used to take 6-7s, now takes < 1s.


kflemin (approved):
🎉 works great! nice speedup!

PR #4621, opened 2024-04-10
Speed up taxlot property serializer

haneslinger:
now, instead of doing queries for every property returned to get their meters and ali, we just do it once at the top. 

on the CERL data, the `properties/filter` endpoint was taking >2 secs. It now takes ~1 sec.


kflemin (approved):
🏆

PR #4623, opened 2024-04-10
Fixes bug to hide edit form on save

perryr16:
From the inventory detail, after saving inventory edits the from remained in edit mode. I went back to PRs from 2023 and experienced the same behavior

#### What's this PR do?
- toggles a scope variable that hides the form.

- inventory detail 
- edit 
- make edits 
- save 
- confirm 
- 'edit' button should replace 'save' and 'cancel' buttons


seed/static/seed/js/controllers/inventory_detail_controller.js
```diff
@@ -475,7 +476,7 @@ angular.module('BE.seed.controller.inventory_detail', []).controller('inventory_
      * save_item: saves the user's changes to the Property/TaxLot State object.
      */
     const save_item_resolve = (data) => {
-      $scope.$emit('finished_saving');
```
perryr16, line 478: This emit statement looks like it was related to hiding the form, but I went back several years and couldnt find any reference to `'finished_saving'` outside of these two `$emits`

Honestly, I'm not sure how this worked before.

kflemin (approved):

PR #4625, opened 2024-04-11
Speed up filter groups

haneslinger:
remove a lot of unneeded queries from this endpoint. not a huge diff, ~.5sec, but a wins a win. 

before 
image
after
image


kflemin (approved):
👍

Issue #4624, opened 2024-04-11
"Only Show Populated Columns" CLP when both Properties and Taxlots are present in the cycle


kflemin:
**Describe the bug**

When an org has both Properties and Taxlots (and they are paired together), selecting the Action "Only show populated columns" will result in an inventory table that is completely empty (just the headers are present), even though the number of properties is not 0 and can be seen on the top right of the table.  Removing some of the opposite inventory type columns from the Column List Profile created will fix this issue, but I cannot figure out the reason why some columns cause this behavior to happen.

Testing with the san jose datasets in the tests/data folder, we have found that removing the following columns from the respective CLPs fixes the issue:

From the PROPERTIES column list profile, remove the following TAXLOT columns:
- Parcel ID
- Taxlot Object ID
- Lot Number
- Number of Addresses
- Number of Floors
- Owner Postal Code
- Latitude
- Longitude

From the TAXLOT column list profile, remove the following PROPERTY columns:
- Property Object ID
- Latitude
- Longitude

**Expected Behavior**

The table should render correctly no matter what columns are added to the column list profile.

**Actual Behavior**

![Screenshot 2024-04-11 at 1 13 19 PM](https://github.com/SEED-platform/seed/assets/2205659/9124c0bd-7b8c-4c87-8a73-7aa0f9c4d883)

**Steps to Reproduce**

     1. Import the san jose taxlot data
     2. Import the san jose property data
     3. Go to the property inventory page. Notice that the table renders correctly
     4. From the actions dropdown select "only show populated columns". Notice that the table no longer renders 
     5. Do the same on the taxlot inventory page.
     6. Edit the column list profile to remove the columns listed above. Notice that the table renders correctly again.

**Instance Information**
we have seen this issue locally on develop branch as well as on dev2

axelstudios:
This was a bug that occurred when values of `related` columns were _null_

PR #4626, opened 2024-04-11
Handle null values in related columns

axelstudios:
#### What's this PR do?
Related values could be null, when we were expecting strings. This change fixes that issue

1. Create a new database
2. Import `covered-buildings-sample.csv` with taxlot and property mappings
3. Check that the inventory list loads

#4624


kflemin (approved):
🤩🤩🤩🤩

PR #4630, opened 2024-04-12
Speed up derived columns

haneslinger:
tightening up those queries

before:
image

after:
image


perryr16 (approved):
Nice, this halved my `.evaluate` response time

PR #4636, opened 2024-04-15
Speed up stats

haneslinger:
IMPORTANT
----
the lastest commit, `alt`, is a version that'll only run quick when you're extra columns are sparely populated (i think), like on dev1, BPS- DC, cycle 2016. it's odd,the commit before that might be better

alot stuff we weren't using anymore, however, most improvement comes from rewriting the queries. 

cuts it in half, but it still takes along time. 

Far and away, the most expensive part it getting the count of states that have each column populated. I rewrote the query to be more DB heavy, which sped it up, but it's still rather expensive it runs this query for every column (diff column name, of course):
>SELECT COUNT(*) AS "__count" FROM "seed_propertystate" WHERE ("seed_propertystate"."id" IN (SELECT U0."state_id" FROM "seed_propertyview" U0 INNER JOIN "seed_property" U2 ON (U0."property_id" = U2."id") INNER JOIN "orgs_accesslevelinstance" U3 ON (U2."access_level_instance_id" = U3."id") WHERE (U0."cycle_id" = 647 AND U3."lft" >= 1 AND U3."rgt" <= 2 AND U2."organization_id" = 412)) AND NOT ("seed_propertystate"."total_ghg_emissions" IS NULL)); args=(647, 1, 2, 412)

and this query for every extra data column:
> SELECT COUNT(*) AS "__count" FROM "seed_propertystate" WHERE ("seed_propertystate"."id" IN (SELECT U0."state_id" FROM "seed_propertyview" U0 INNER JOIN "seed_property" U2 ON (U0."property_id" = U2."id") INNER JOIN "orgs_accesslevelinstance" U3 ON (U2."access_level_instance_id" = U3."id") WHERE (U0."cycle_id" = 647 AND U3."lft" >= 1 AND U3."rgt" <= 2 AND U2."organization_id" = 412)) AND "seed_propertystate"."extra_data" ? 'geojson_Maxlat' AND NOT (("seed_propertystate"."extra_data" -> 'geojson_Maxlat') = 'null')); args=(647, 1, 2, 412, 'geojson_Maxlat', 'geojson_Maxlat', 'null')

now, while I dont think there's way around performing a group by for every column, I bet there a way to make a temporary table out of the bit they all share, the bit that gets the relevant states.

> FROM "seed_propertystate" WHERE ("seed_propertystate"."id" IN (SELECT U0."state_id" FROM "seed_propertyview" U0 INNER JOIN "seed_property" U2 ON (U0."property_id" = U2."id") INNER JOIN "orgs_accesslevelinstance" U3 ON (U2."access_level_instance_id" = U3."id") WHERE (U0."cycle_id" = 647 AND U3."lft" >= 1 AND U3."rgt" <= 2 AND U2."organization_id" = 412))

 I just dont know how.


seed/views/v3/analyses.py
```diff
@@ -300,176 +298,32 @@ def stats(self, request):
...
                 "status": "success",
                 "total_records": views.count(),
-                "number_extra_data_fields": len(extra_data_count),
-                "column_settings fields and counts": count_agg_dict,
-                "existing_extra_data fields and count": extra_data,
-                "property_types": property_types,
-                "year_built": year_built_list,
-                "energy": energy_list2,
-                "square_footage": square_footage_list2,
+                "number_extra_data_fields": columns.filter(is_extra_data=True).count(),
```
haneslinger, line 325: this field has changed a little.
it used to be number of extra data columns with at least one appearance in a state in this cycle, now it's just number of extra data columns, period.

kflemin:
documenting the current plan for this ticket: awaiting feedback from @ebeers-png. Otherwise we will go with alt, deploy to dev2 and do some testing there with orgs that have many extra data fields.

axelstudios:
I updated this to only need 1 query for extra_data columns, and 1 query for canonical columns, resulting in an 8.7x speed up over the improvements that Hannah already implemented.

In total, the stats request is now 35x faster than it used to be, and only takes 436ms to load for a large org

axelstudios (approved):

PR #4641, opened 2024-04-17
ensure is_ali_root is populated on redirects

kflemin:
Sometimes some of the items are not shown in the Actions dropdown on the inventory list (the one I notice is "Only Show Populated Columns").  Refreshing the page fixes the issue, but this is annoying.
I tracked the issue down to menu.user.is_ali_root not always being populated on some redirect issues.

To test:
- go to inventory list page
-  ensure you see 'only show populated columns' from actions dropdown
- switch the org to a different org in the top right of the page. the page will refresh
- open up the actions dropdown and ensure AGAIN that 'only show populated columns' is still there

### DIFF
--- a/seed/static/seed/js/controllers/inventory_list_controller.js
+++ b/seed/static/seed/js/controllers/inventory_list_controller.js
@@ -90,6 +90,9 @@ angular.module('BE.seed.controller.inventory_list', []).controller('inventory_li
     };
     $scope.organization = organization_payload.organization;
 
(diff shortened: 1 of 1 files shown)


perryr16 (approved):
Works great, good catch

kflemin:
@perryr16 I think there are other cases (I've seen it with the column list profiles page and the edit buttons not showing up until refresh) so let me know if you run into one and we'll fix it as we go along! :)

PR #4647, opened 2024-04-25
alternative way to get used cols for sharing page

kflemin:


kflemin:
@perryr16 what do you think of this way to get the populated columns?

perryr16 (approved):
Works great. Tested with columns generated from a CO2 analysis

PR #4654, opened 2024-04-30
Speed up ali tree

haneslinger:
the built in tree `bulk_dump` was slow, so I wrote a faster one. `api/v3/organizations/2/access_levels/tree/` on the CERL data used to take 13secs. It now takes <.5sec.


kflemin (approved):
this is much faster! awesome!

PR #4657, opened 2024-05-03
Add 2 factor authentication to login 

perryr16:
#### What's this PR do?
Uses an open source, well rated, free library [django-two-factor-auth](https://github.com/jazzband/django-two-factor-auth/) to add token authentication via any authenticator app (google authenticator, microsoft authenticator, etc...) or via email.

Users will only be prompted to enable 2FA on their first login following the introduction of this code. Opting in or out is at the user level and users can always access the 2FA settings from /profile  > Two Factor.

If a user chooses "Token Generator" then a QR code is displayed and can be scanned by any number of auth apps. Once scanned, users can enter the token to login

If a user chooses "Email" then a token will be sent via email to the associated account.

The UI for this library uses Django templates and does not enter the angular layer. To customize the Django templates they have been imported into the seed/landing/templates/two_factor directory and edited. To make navigation changes to the response from django-2-f-a, the response must be caught in a custom django view (see /CustomLoginView.py)

1. With token generator
Login with any user > Enable 2Fa > Select "Token Generator "
Scan QR code with auth app of choice and enter token 
User should be logged in 

2. With email
Go to profile > Two Factor > disable two factor authentication 
then re-enable two factor and select "Email". An email will be sent out, if testing locally it will likely be found in the seed_web logs.
Enter the token
User should be logged in 

Screenshot 2024-05-03 at 9 47 25 AM

3. Disable 2fa 
Go to profile > two factor > disable two factor authentication 
log out 
log in 
user should be logged in without 2FA prompts.
 

#4657


kflemin:
@perryr16, this is great!  Documenting a few issues I ran into:

1. first, when I choose the 'email' verification method, I see a whole lot of errors in the logs. A lot of these:
"AttributeError": type object RequestContext has no attribute  'cancel_url', 'is_ali_root', 'ali_leaf', .... etc.

2. I've also run into this error in the logs:
    raise VariableDoesNotExist("Failed lookup for key "
django.template.base.VariableDoesNotExist: Failed lookup for key [method] in <EmailDevice: default (email@nrel.gov)>
2024-05-14 15:23:51 DEBUG    Exception while resolving variable 'method' in template 'two_factor/core/setup.html'.
  File "/Users/kflemin/.pyenv/versions/seed-10/lib/python3.10/site-packages/django/template/base.py", line 829, in _resolve_lookup
    current = current[bit]
TypeError: 'EmailDevice' object is not subscriptable

3. On the login screens, there are 2 things that we need to change:  
- The first one is that we do not want to give them a "create my account" button.
- The second one is that we have lost the 'forgot your password' functionality.

This is the old login screen:
Screenshot 2024-05-14 at 4 16 06 PM

And the new ones:

Screenshot 2024-05-14 at 4 16 01 PM
2205659/b8ceb482-080c-4f43-8bd9-a201c7a4a75a)

Maybe if you just turn "Create my account" into "Forgot my password" it might be good enough?

thanks!

PR #4663, opened 2024-05-07
Flip axis

haneslinger:
the problem with flipping is x and y are now different on each chart..... should I flip the other one to?

image


kflemin:
@haneslinger, something is not quite right with these axes...let me know if you are seeing the same thing or if it's a weird local caching issue:

Screenshot 2024-06-13 at 3 41 09 PM

The best way I can explain what I'm seeing is...the tick marks on the axes haven't flipped yet maybe?

haneslinger:
I think I fixed it, but idk whats happened to this branches history

haneslinger:
heres a new one thats less confusing https://github.com/SEED-platform/seed/pull/4693

Issue #4549, opened 2024-02-28
Improve the Import Process


kflemin:
Per Peer Review feedback, upgrade data mapping and matching processes to make the workflow more intuitive. Include further integration of data types and dual unit support. Need to improve the performance of the importing process. Look at better scaling of worker nodes. 

More details:
 1. Profile the import process (steps 1 through 6 on the UI) - some steps seem to get skipped while others take a long time to run
    - also profile each celery task and see which ones take the longest to run
    - a specific example that is taking a long time (1 second for each imported record): https://github.com/SEED-platform/seed/blob/develop/seed/data_importer/match.py#L724
    
 3. Investigate if there are things that can be skipped completely. for example:  
    - Geocoding:  if the org doesn't have a mapquest key or lat/lng/UBID are not provided, skip the geocoding process
    - Linking: if there's only one cycle, can we skip the whole linking process completely?
    - Pairing: if there are no taxlots in the org and no taxlots in the import, can we skip the pairing process completely?
    - Matching:  are there improvements we can do here (matching within and without cycles?)
 4. Can we improve the Progress API endpoint to be more fault tolerant:
   - frontend: keep retrying on error?
   - backend: reset the TTL of each key when it updates the progress values?
 

@axelstudios

axelstudios:
The mapping step has room for improvement, but by far the biggest bottleneck is after you hit `Confirm mappings & start matching`:
![Start Matching](https://github.com/SEED-platform/seed/assets/411466/d00077c4-1a89-4439-aaa5-86e39698fd4b)

Anecdotally, when uploading a file with 145,921 rows to an existing organization with 145,921 matching rows these are the celery tasks and their timing:

- 1x `seed.data_importer.tasks._geocode_properties_or_tax_lots` (35s)
- 1,460x `seed.data_importer.tasks._map_additional_models` (1m 8s)
- 1x `seed.data_importer.match.match_and_link_incoming_properties_and_taxlots` (2d 13h 55m 9s)
- 1x `seed.data_importer.tasks.finish_matching` (0.3s)

Total matching time: 61.94 hours (2 days 13 hours 56 minutes 33 seconds)

When going through [this loop](https://github.com/SEED-platform/seed/blob/develop/seed/data_importer/match.py#L724) it takes approximately 1.5 seconds for each loop iteration (each record in the import file)

RDmitchell:
@axelstudios / @kflemin -- I am going to test importing on staging the next few days. 

[...]

PR #4669, opened 2024-05-13
Comprehensive performance improvements

axelstudios:
There was _significant_ overhead to loading every django html request - e.g. just loading index.html on seedv2 takes 2.5 seconds before any data comes through or any other requests can be made, and this happens any time you load/refresh the page.

#### What's this PR do?
1. Completely removes the aforementioned overhead via `COMPRESS_CACHEABLE_PRECOMPILERS` - the problem was the `npx sass` command was running and generating a new css file on every single load, resulting in hundreds of thousands of files in the CACHE directory and multiple gigabytes. Now it only runs on change (500x faster).

    |              | **Before** | **After** |
    |--------------|------------|-----------|
    | **Overhead** | 2.5s       | 5ms       |

1. Enables Django compression using brotli, bundles all individual static files into logical groups, and only loads the files necessary for the specific context (so instead of making 251 requests on load now it only makes 23). **Compression is only used in production**, but you can enable it in a development environment. This change should also permanently fix stale browser cache issues (8x faster):
    - Landing only loads `landing.css`
    - SEED only loads `vendors.css`, `seed.css`, `vendors.js`, and `seed.js`

    |                      | **Before** | **After** |
    |----------------------|------------|-----------|
    | **Data Transferred** | 8.1 MB     | 1.9 MB    |
    | **Load Time**        | 1.71s      | 224ms     |

1. HUGELY improves performance when importing files with UBIDs into an org that already has records with UBIDs. For instance, if you import the 512-record `covered-buildings-sample.csv` to a new org, then upload it again with a single column changed and click `Save Mappings` (30x faster):

    |             | **Before** | **After** |
    |-------------|------------|-----------|
    | **Queries** | 798,549    | 33,142    |
    | **Time**    | 1h 57m 40s | 3m 52s    |

1. Switches AngularJS tests to use Puppeteer instead of the deprecated Protractor+WebDriver setup (just run `npm test`, but keep in mind that it's expecting SEED to be running on localhost:80). This PR also switches to automatic test paths instead of hardcoded paths
1. Cleans up all landing files to remove unnecessary and (lots of) deprecated code
1. Fixes serving the FAQ page on Windows
1. Suppresses noisy `Created ubid_model id` logs
1. Sorts org members on the Accounts page in ascending order by 'last_name', then 'first_name', or otherwise email (previously the order was arbitrary)
    image
1. Renames `inventory_settings` and `inventory_detail_settings` to `inventory_column_list_profiles` and `inventory_detail_column_list_profiles` respectively, since the previous names were very ambiguous
1. Makes the inventory list labels column pixel-perfect with borders consistent with the other fixed columns, and removes the broken sort button from the notes_count & meters_exist_indicator column headers
1. Formats large numbers with thousand-separator commas throughout the app, which is especially useful for orgs with large amounts of data
1. Makes clicking the SEED Platform text logo in the top left corner navigate instantly in all but one case
1. Updates frontend dependencies

1. Run `pip install -r requirements/base.txt`
2. To test using compression locally, modify your `local_untracked.py` file with the following setting:
```py
COMPRESS_ENABLED = True
```
3. Monkey test, ensure everything works as before

Tangentially related to #4549


kflemin (approved):
this is⚡️⚡️⚡️⚡️!!!

PR #4677, opened 2024-05-22
On click go to property

haneslinger:


kflemin (commented):
this works great! one tiny but worthwhile enhancement to this would be to add the property display field to the tooltip.  it's the one that's defined here:

Screenshot 2024-05-28 at 3 13 22 PM

kflemin (commented):
@haneslinger, this looks and works great on the left-side chart!  On the histogram though, I get a message in the console and at the top of the page that the transition failed.  Are we supposed to be able to click on the histogram?

Screenshot 2024-05-30 at 3 00 30 PM
Screenshot 2024-05-30 at 3 00 39 PM
Screenshot 2024-05-30 at 3 00 50 PM

kflemin (approved):
🏆

RDmitchell:
instance: dev1
SHA: 87dd50ae1

@kflemin -- I guess this is to allow the user to navigate to an individual property in the Default Reports scatter plot, which it does. 

However, what is the 2nd item that you added about adding the Property Display field 

I have set the Property Display Field to be Property Name
image

That field exists and is populated

If I click on that point on the graph
image

It takes me to this property record, where the property name exists

But that is not displayed in the tooltip for that record. 

Maybe that should be made into a new issue, because the other functionality of going from a point on the scatter plot to an individual property record is working.

Issue #4678, opened 2024-05-24
Update BETTER to version 1.7


kflemin:
Version 1.7 of BETTER will be released on production 6/30.  Ensure that SEED is updated to work with the new version

RDmitchell:
I guess that will include being able to model Public Library (or whatever the official BETTER building type that is).

Thanks !!

RDmitchell:
Ah, yes, I see Public Library in the PR ! Thanks again !!

RDmitchell:
Need to make sure that for testing, the 2 dev servers point to the BETTER develop server, which they don't currently. 

![image](https://github.com/SEED-platform/seed/assets/6314950/bdf8217a-4119-4c32-a0b5-3aac01db36d2)

PR #4679, opened 2024-05-24
BETTER v1.7 updates

kflemin:
Updates to support BETTER version 1.7

BETTER anticipated release schedule:
  - currently on BETTER dev server
  - on staging 6/15
  - on prod 6/30

Updates:
1. updated ee_measures, both the list and the structure to use TOKEN rather than name
2. updated path to results in results array (a element named 'assessment_results' was introduced in the 'assessment' hash)
3. added support for Public Library and Restaurant

To test, ensure that you are pointed to the BETTER DEV server (https://better-lbnl-development.herokuapp.com/).  Run analyses and ensure they run successfully and you can get results.

#4671, #4678

DO NOT MERGE until BETTER v1.7 is deployed on the BETTER production server.


RDmitchell:
It looks like seed dev1 is pointing to the BETTER production server right now - might want to change that to the BETTER dev server? 

image

RDmitchell:
Looks like seeddemostaging is set to the BETTER staging server.
image

RDmitchell:
@kflemin -- so it seems like the 2 dev servers need to be set to point to the BETTER develop server. Thanks !!

kflemin:
@RDmitchell, yep, we'll have to make that change when we deploy these updates for sure!  this one will be a bit tricky to coordinate testing on the server since the changes are not compatible with the earlier versions of BETTER, but we will make it work!

RDmitchell:
@kflemin -- thanks and have a good 3 day weekend !! Time to stop working !!

axelstudios (approved):

PR #4683, opened 2024-05-28
Replaced missing ESPM report

axelstudios:
The CI was relying on ESPM report `2807325`, which appears to no longer exist

#### What's this PR do?
Creates a replacement

Tests should pass


kflemin (approved):
Awesome, thanks for fixing!! No idea what happened to the original "SEED_Test - Single Property" Template!

Issue #4071, opened 2023-06-01
Add Column Uniqueness Exclusions


axelstudios:
**Is your feature request related to a problem? Please describe.**
Importing data from ESPM, and then adding custom data within SEED (like Salesforce IDs) via manual edits will cause the record to be an in-exact match for every future import rather than an ignored duplicate.

**Describe the solution you'd like**
The Column Settings page should have an additional column with a checkbox to allow specific columns to be excluded from hashing.  That way the `hash_object` can be identical to future ESPM imports, even if changes were made in SEED to excluded columns.

RDmitchell:
@axelstudios / @kflemin -- this is a problem for the City of Berkeley because they have 2 custom fields that they edit and that causes the duplicate data to be imported every time they update their ESPM data. 

This has caused them to have to edit the ESPM data file outside SEED and only include data that has changed, which is definitely less than ideal. 

Can we bump up the status of this so it gets fixed sooner rather than later ? Thanks !!

haneslinger:
https://github.com/SEED-platform/seed/pull/4687 heres a necessary first step

RDmitchell:
Ammon from the City of Berkeley is happy to demo the issue they are having if that would be useful.

PR #4687, opened 2024-06-03
Exclude related inventory type columns from the hash

haneslinger:
#4071


kflemin (approved):
🎉

seed/data_importer/tasks.py
```diff
@@ -1682,11 +1682,16 @@ def finish_matching(result, import_file_id, progress_key):
 
 
 def hash_state_object(obj, include_extra_data=True):
-    def add_dictionary_repr_to_hash(hash_obj, dict_obj):
+    def add_dictionary_repr_to_hash(hash_obj, dict_obj, inventory_type):
         if not isinstance(dict_obj, dict):
             raise ValueError("Only dictionaries can be hashed")
 
+        excluded_keys = Column.objects.filter(is_excluded_from_hash=True, table_name=inventory_type.__name__).values_list(
```
kflemin, line 1689: @haneslinger I think this needs to be filtered by organizationID too?

seed/models/columns.py
```diff
@@ -1356,12 +1358,20 @@ def retrieve_db_field_name_for_hash_comparison():
 
         :return: list, names of columns, independent of inventory type.
         """
-        result = []
-        columns = Column.retrieve_db_fields_from_db_tables()
-        for c in columns:
-            result.append(c["column_name"])
+        fields = inventory_type._meta.fields
+
+        excluded_columns = list(
```
kflemin, line 1363: @haneslinger  same with this one...should it also get filtered by org id?

kflemin:
@haneslinger do you think there's a possibility to rework the code a bit so that we could somehow prefetch the included hash columns (for taxlots or properties) before looping through each property and hashing them? I think it's adding a lot of extra queries since it runs on each property or taxlot.

kflemin:
closing since this work is included in https://github.com/SEED-platform/seed/pull/4769

PR #4690, opened 2024-06-12
Add parallelization to import tasks

perryr16:
To import a file of 500 records with matching fields, but different "notes" took 90 seconds. After improvements, importing took 16 seconds.

#### What's this PR do?
- Moves a db query outside of a loop - saved 25 seconds 
- Chunks incoming data and runs task `match_and_link_incoming_properties_and_taxlots_by_cycle` in parallel chunks, using the number of celery workers to determine the number of parallel tasks. Results are aggregated at the conclusion of all tasks. With 5 tasks, all tasks complete within 16 seconds.

upload files, monitor flower

Screenshot 2024-06-12 at 10 52 38 AM

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -21,8 +21,10 @@ django-extensions==3.1.3
 django-model-utils==4.3.1
 
 # Time zones support - Do not update these without doing significant testing!
-pytz==2018.7
-python-dateutil==2.7.3
+# pytz==2018.7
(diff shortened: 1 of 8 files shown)


perryr16:
Parallelizing the entire match_and_merge task breaks when duplicate properties exist in an import file. We need to be more precise if we are to use parallel tasks

PR #4691, opened 2024-06-12
Allow org setting default report axis options

haneslinger:
image


kflemin:
@haneslinger, here's another one with a conflict i'm not sure how to fix

kflemin:
@haneslinger could you take a look at these unittest failures? maybe something changed in the reports to cause this? thanks!

RDmitchell:
instance: dev1
SHA:
Org: JCC

This seems to be working

PR #4692, opened 2024-06-13
Speed up UBID merging

perryr16:
#### What's this PR do?
- Refactors a query within states_to_views to query against all states within the cycle, while the original queried against all property states regardless of cycle. Depending on the size of the db this could be a very large number of states that \
- Improves ubid merging

Time savings for 2 files with 500 matching records: 
develop: 90s 
this branch: 45s

### DIFF
--- a/seed/data_importer/match.py
+++ b/seed/data_importer/match.py
@@ -3,7 +3,7 @@
 SEED Platform (TM), Copyright (c) Alliance for Sustainable Energy, LLC, and other contributors.
 See also https://github.com/SEED-platform/seed/blob/main/LICENSE.md
 """
-
+import logging
 import datetime as dt
 import math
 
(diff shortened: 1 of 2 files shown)


perryr16:
This really only improves uploads if UBID is the only non `None` matching criteria in the upload file.

seed/data_importer/match.py
```diff
@@ -595,19 +595,25 @@ def states_to_views(unmatched_state_ids, org, access_level_instance, cycle, Stat
         if "ubid" in matching_criteria:
             check_jaccard = bool(matching_criteria.get("ubid"))
             ubid = matching_criteria.pop("ubid")
+            region = ubid.split("+")[0]
```
perryr16, line 598: checking region is not a 100% guarantee. Might want to put this particular change on hold

perryr16:
see https://github.com/SEED-platform/seed/pull/4704

PR #4693, opened 2024-06-13
New flip axis

haneslinger:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,346 +1,146 @@
 {
   "name": "seed",
   "version": "3.0.0-beta.0",
-  "lockfileVersion": 3,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "name": "seed",
-      "version": "3.0.0-beta.0",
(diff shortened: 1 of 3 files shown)


RDmitchell:
instance: dev1
SHA: 87dd50ae1
Org: JCC

in the Insights/Default reports, I don't see any options for the X Axis and only 1 (Count) for the Y Axis.
image

How do these get populated? 

On production, for the same org, I see lots of options for those two pulldowns
image
image

RDmitchell:
@haneslinger / @kflemin -- it seems like maybe there is a place to set the fields to be displayed in these pulldowns, but I can't seem to find it. https://github.com/SEED-platform/seed/pull/4691

Can you point out to me where to set these?

RDmitchell:
@haneslinger / @kflemin -- NEVER MIND _- found them in Default Display Fields.

RDmitchell:
Hmm, fairly tedious to select them one by one. Maybe a future issue might be to allow the user to select multiple fields at a time.

RDmitchell:
This appears to be working.

PR #4695, opened 2024-06-14
Add cross cycle data quality checks

perryr16:
This feature is based on the Better Buildings "Data Review Report" checks.

#### What's this PR do?
This PR adds the ability to run a function that dynamically determines if property pairs within a goal have passed a default set of checks. If the property pairs do not pass the checks a label associated with the current cycle property view and goal will be applied, and the data will be excluded from the Portfolio Summary calculations.

Changes in this PR
- Adds a new function to run cross-cycle (goal) data quality checks triggered from the Portfolio Summary actions dropdown
- Adds a UI displaying the results of the Goal data quality checks
- Adds 12 new default goal data quality rules
- Modifies the join table between Property Views and Labels (PropertyViewLabel) to include an optional relationship to the Goal model
- Rules have been modified to include a boolean field "cross_cycle" to determine if a rule should be applied across multiple cycles or within a single cycle. Default is false.
- Goal rules use a goal_id and a data_type to determine which column value to use for the check.
- Labels that have been applied to a goal will not appear in the inventory list.
- The ability to create new Goal rules has been disabled.

- Create a goal 
- Portfolio Summary calculations should be empty as "Passed Checks" is default false
- From the Portfolio Summary action dropdown, select "Run Data Quality Checks"
- A results modal will appear at when the checks have finished. 
- The Passed Checks should have been updated
- If a property pair has failed checks a label will be applied to the current property. Labels assigned during a Goal data quality check will not be visible on the inventory list.

Screenshot 2024-07-01 at 10 47 08 AM

Screenshot 2024-07-02 at 10 25 10 AM

Screenshot 2024-06-27 at 9 39 39 AM


kflemin (approved):
🔥

PR #4700, opened 2024-06-20
Add 2 Factor Authentication

perryr16:
#### What's this PR do?
Adds Two Factor Authentication features to the login process using [Django-Two-Factor-Auth](https://github.com/jazzband/django-two-factor-auth) as a base. The Django-Two-Factor Auth (django-2fa) library is built on top of the [django-otp library](https://django-otp-official.readthedocs.io/en/stable/). 

The login process is handled through django-2fa, however two-factor method configuration is handled locally on SEED with the help of [qrcode](https://pypi.org/project/qrcode/) and [pyotp](https://pypi.org/project/pyotp/) libraries.

Users who have logged in for the first time following these changes will be redirected to the two factor profile page where they can set their method of choice. Following log ins will direct to the SEED homepage.

This PR introduces a new org setting to require all users to use a two factor authentication method. If this method is set, all active users who do not already have a 2fa method set will be assigned email verification with the option to switch to any token auth app. User's part of multiple orgs will be required to use 2fa if any parent org 2fa requirement is set.

To set up the Authenticator App Token method, users must scan a generated QR code and verify the code from the authenticator app. If the code is not verified, a default method will be assigned based on the 2fa requirement org setting.

- first login should redirect to the two factor profile page
- select email, log out, log in, an email should have been sent, use that to verify the log in or click the 'resend email' button.
- select token, but fail to verify the token, and click cancel. The auth method should switch to deisabled or email
- select token, scan the qr code, verify the token, or click re-generate qrcode to refresh the process. When logging back in, users will be prompted to enter the token from the authenticator app.

#4581

Screenshot 2024-06-24 at 9 51 15 AM

Screenshot 2024-06-24 at 9 36 26 AM

Screenshot 2024-06-24 at 10 38 37 AM


kflemin (approved):
🎉

PR #4704, opened 2024-06-24
Performance refactor merge ubids

perryr16:
#### What's this PR do?
- Refactors a query within states_to_views to query against all states within the cycle, while the original queried against all property states regardless of cycle. Depending on the size of the db this could be a very large number of states
- Improves ubid merging and early exits if possible

benchmarking for 2 files with 500 matching records: 
develop: 80s 
this branch: 70s


axelstudios:
Incorporated into https://github.com/SEED-platform/seed/pull/4718

PR #4726, opened 2024-07-26
Inventory groups

ebeers-png:
We're pushing groups back to SEED from BEAM to support development of group compliance and systems/services. A lot of the original code is pretty overcomplicated/overwrought, so feel free to edit as much of it as you'd like. The groups page added to the inventory list navigation can also be repurposed for adding and managing utilities.

#### What's this PR do?
Adds inventory groups to SEED. Groups are simple, manual groupings of buildings that work independently of cycle (a view added to a group in one cycle will still be in the group in another cycle), and contain either properties or taxlots, but not both. No other functions are associated with groups right now.

Specific elements added by this PR:
- InventoryGroup and InventoryGroupMapping models, serializer, support for merging/unmerging buildings
- On the inventory list, adds a new tab in the navigation for creating, renaming, and deleting groups. Page is otherwise empty.
- On the inventory list, adds a row under the "View by Property/Taxlot" tabs for group tabs (if any groups exist) that the user can switch between.
- On the inventory list and inventory detail pages, updates the Action menu with a "Add to/remove from group" option for adding/removing buildings from groups.

Groups can be created three ways: 
1. Select buildings on the inventory list, choose the action "Add to/remove from group," and create a new group in the modal window.
2. Go to an inventory detail page, choose the action "Add to/remove from group," and create a new group in the modal window.
3. Go to the "Group" tab on the inventory list page and create a group there.

Add and remove buildings using the action menu on either the inventory list page or an inventory detail page. Swap between groups using the tabs, and try out filters, ensuring that it properly filters the subset of buildings in the group.

image
image


haneslinger:
hey, thank you so much. there are a couple of changes we need (adding in AH, a few constraints...), so I think I'll build a PR based off of this work, if that's okay?

ebeers-png:
You're welcome to commit to this PR as well, but either is fine with me!

Issue #4694, opened 2024-06-14
Inventory List: Tax Lot Tab: Can't move columns


RDmitchell:
instance; dev1
SHA: 
Org: LBNL 200
Cycle: 2018 Compliance Cycle

There are records mapped to the Tax Lot tab.

In that tab, when I try to move fields to the right or left with my mouse, they don't move. 

Also it seems like the Unpin option isn't working for the Tax Lot tab

We need the same functionality in the Tax Lot tab as in the Property tab

RDmitchell:
Instance: Staging
SHA: 17f3257f4
Org: LBNL 200

Still not working.

@axelstudios reports that it is fixed but not merged in

RDmitchell:
Here is the outstanding PR: https://github.com/SEED-platform/seed/pull/4731

RDmitchell:
instance: dev1
SHA: 87dd50ae1
Org: LBNL 200
Cycle:  2018 Compliance Cycle

Working great ! Thanks !

PR #4731, opened 2024-08-01
Fix bug preventing tax lot columns from moving

perryr16:
#### What's this PR do?
Because tax lots do not have meters, the meters' column was `undefined`, causing errors when attempting to move the inventory list columns

Upload tax lots. Move the inventory list tax lot columns

#4694


haneslinger (approved):
Nice one!

Issue #4177, opened 2023-08-04
UBID: Allow importing multiple UBIDs in the same field


RDmitchell:
instance; seeddemostaging
SHA: 22e66f350
Org: LBNL 101
Commercial Real Estate for sale in El Cerrito-PortreroFixed-UBID.xlsx

I have generated 2 slightly different UBIDs, from 2 different PNNL UBID creation tools, for the same property. 

I would like to import them both into SEED at the same time. 

I tried adding them both to the UBID field in the import file, separated by a semicolon, but upon import they were both added to the same UBID line in the table. 
![image](https://github.com/SEED-platform/seed/assets/6314950/5dc2a3b4-f17e-492c-9c6d-0c92f8c998bf)

RDmitchell:
The file in question is here
SEED Shared Folder\Software Development\Issues\Issue 4177

RDmitchell:
And I guess right now there is a way to add the 2nd UBID "by hand" by copying and pasting the 2nd UBID into the UBID UI implementation

RDmitchell:
instance: seeddemostaging
SHA: ee056e67e
Org: LBNL 21

This worked great, as long as the UBID field was not set to a matching field. 

I will write up a separate issue for that, ie, probably in the initial mapping step, if UBID is set to matching and any record has more than one UBID, we give some sort of error to the user that they can't have more than one UBID in the same field if it is a matching field.

PR #4732, opened 2024-08-02
Allow multiple ubids during import

perryr16:
#### What's this PR do?
Allows import files to contain multiple UBIDs separated by semicolons (;). UBIDs separated by other symbols will be treated as a single UBID

Upload a file with multiple UBIDs per property separated by semicolons

#### What are the relevant tickets
#4177


haneslinger:
Looks good to me!

the mapping page only show one ubid, but its up to you if that feels right
image

haneslinger (approved):

perryr16:
The logic to split ubids occurs during the mapping step, so by the time the user sees the mapping results the state has already been created, and the ubids have been separated with the last set as preferred. It felt like the least invasive approach, but I'd be happy to look into showing all ubids on the mapping results.

RDmitchell:
@perryr16 -- How does the preferred UBID get picked? Does that happen in the mapping step? 

If it is happening automatically, I think it might be confusing to the user to just see the final UBID in the mapping results screen, if they didn't understand that only one was selected. 

I'm not sure the best way to show the multiple UBIDs on that mapping results screen, but it might be worth thinking about? 

And if it's too complicated at this point, you could go with what you have. 

And, on the other hand, I am pretty sure that most users are not going to want to really understand the UBID world, and not bothering them with the multiple UBIDs when reviewing the mapping results is probably fine !!!

perryr16:
@RDmitchell the first listed UBID will be assigned as the preferred UBID. Displaying all the UBIDs on the mapping as a ';' separated list is possible, but it's a little ugly. I'd vote to table that feature until requested.

RDmitchell:
@perryr16 -- tabling until we get user feedback is fine !!

PR #4733, opened 2024-08-05
Clean up default reports

haneslinger:
- correct x y in org settings
- correct x y in tool tip
- round bin labels
- sort bins in asc order

image


kflemin (commented):
🏆

kflemin (approved):

RDmitchell:
Instance: dev1
SHA: c646212e5
Org: BPS Test - Robin

The scatter plot on the left doesn't seem to be working -- the x-axis is not set up correctly.
image

RDmitchell:
@kflemin -- it doesn't look like this is quite fixed.

RDmitchell:
@kflemin -- this is how the scatter plot used to work (from my Getting Started guide)

The scatter plot would show a dot for each individual building, according to how the axes were defined. You could see the basic details about the building by hovering over the dot, and these days, we would probably want to implement the ability, as in other parts of the program, where the user could click on the dot and go to the detail view for that building. 

Seems like we need to make a separate issue for this, since this is just a PR and doesn't seem to have an issue associated with it? 

image
image
image

PR #4740, opened 2024-08-08
Enable data type setting during data import

crutan:
#### What's this PR do?

Adds data type definition to import step, also adds the capability for omitted columns to be saved as omitted to column mapping profiles

Data type setting can be tested by uploading a CSV with extra data columns and defining their types.  Omissions can be tested by adding a column to a datafile that you wish to omit - Tax ID or something similar, and then updating your column mapping profile with the omit column checked.  Then re-upload that file and apply the column mapping profile to see that the column is now omitted.

###  Deploys in concert with: https://github.com/SEED-platform/py-seed/compare/4701-data-type-integration-during-data-import?expand=1


kflemin (approved):
works great!

Issue #4737, opened 2024-08-06
Filtering: Add AND and OR to pulldown list names for clarity


RDmitchell:
Instance: staging
SHA: 17f3257f4

The current filtering by labels is more sophisticated and allows much better filtering, but I am constantly trying to understand the meaning of each, ie, which is the AND and which is the OR choice. 

This is probably very low priority, and this request can be rejected (!) but I am wondering if it would make sense to add those "qualifiers" to the labels of the pulldowns, such as the following: 

![image](https://github.com/user-attachments/assets/a5e1184c-dace-42b8-988b-c856eb4486e8)

Or it could be a ? icon with that information could be added instead of adding those terms to the pulldown list label itself.

RDmitchell:
instance: dev1
SHA: 87dd50ae1

Done ! Perfect!
![image](https://github.com/user-attachments/assets/88e26bc9-f558-4708-8fb9-7931995019b0)

PR #4742, opened 2024-08-08
update filter AND OR placeholder text

kflemin:
On Inventory List page, Update placeholder text on "AND" and "OR" filters to be more intuitive.
Screenshot 2024-08-08 at 3 44 06 PM

Relevant Ticket: #4737


kflemin:
@RDmitchell what do you think of the placeholder text in the screenshot above? Let me know if you'd like me to change the wording

perryr16:
Just a thought, would we also want to change the exclude label placeholder? Something like "Exclude all selected labels".

RDmitchell:
@perryr16 -- I think your suggestion of adding the extra text to Exclude would be good, if it's not too late. And if it is too late, no worries, just leave it as is for now !

perryr16 (approved):
🏆

Issue #4701, opened 2024-06-21
Data Type Integration during Data Import


kflemin:
**Problem 1**- Save Data Types during Data Import
On data import, the data mapping screen looks like the following:
![import1](https://github.com/SEED-platform/seed/assets/2205659/206ef37e-7bb0-44df-af3e-9d822f654d3c)

We need to provide the user with the ability to save the data types of the incoming data. If the file is a CSV, the data type is unknown, if it is Excel then it is typically correct. Regardless, the user needs to have the ability to select String, Integer, Number. This will improve importing and error checking. 
The possible data types can be found in the columns settings page. For example, it looks like this for extra_data columns:
![import2](https://github.com/SEED-platform/seed/assets/2205659/ee530d39-c926-45c6-9f04-32a4bf7247c4)
 
**Problem 2 -** Number/Float Redundancy
We need to remove the redundant Number/Float. Should probably just be Number. This will likely require a database migration to map existing Floats -> Number.
**Problem 3** 
If the field is a known unit type, e.g., EUI, GHG, then we should probably add a separator to denote those, because they are special.
**Problem 4**
The columns mapping page looks like this:
![import3](https://github.com/SEED-platform/seed/assets/2205659/aaa1c689-1c6f-4bab-8677-e56c1a14a565)

 Is this a problem? Do we need to add any of the datatype functionality here? Just check/doublecheck.
**Problem 5**
Last, the SEED API allows for a file to be posted to create a "column mapping profile". The file format needs to be updated.
 https://github.com/SEED-platform/py-seed/blob/develop/tests/data/test-seed-data-mappings.csv
 Add column for "omit" mapping. Verify that if "omitted", that the column data won't be stored to the database and that it will be recalled correctly in the column mapping page represented in Problem 1 above. 
**Problem 6**
Add column for data type. Need to catch cases where the column data type changes because it may require that the data get an updated type in the database. If this is the case, then I think we want to error and not allow the user to change the datatype.

PR #4744, opened 2024-08-09
Add messages for reasons the save button on column mappings is disabled

crutan:
#### What's this PR do?

Adds messaging to the save button tooltip to help explain why the button may be disabled for column mapping profile changes.

View the column mappings page and make changes to extant profiles

#4701

### DIFF
--- a/seed/static/seed/js/controllers/column_mappings_controller.js
+++ b/seed/static/seed/js/controllers/column_mappings_controller.js
@@ -45,6 +45,7 @@ angular.module('SEED.controller.column_mappings', []).controller('column_mapping
 
     $scope.mappable_property_columns = mappable_property_columns_payload;
     $scope.mappable_taxlot_columns = mappable_taxlot_columns_payload;
+    $scope.profile_action_message = '';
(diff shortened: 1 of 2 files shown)


perryr16 (changes requested):
The `profile_action_message` kept getting overwritten to "" by the `profile_action_ok` function. Maybe something like this?
```
    $scope.save_button_tooltip = () => {
      if (!$scope.changes_possible) return 'No changes'
      else if ($scope.header_duplicates_present()) return 'Duplicate headers present'
      else if ($scope.empty_units_present()) return 'Empty units present'
      else if (!$scope.profile_action_ok('update')) return 'Unknown profile type'
      else return 'Save'
    };
```

Also, it's an existing bug, but it looks like line 180 in columng_mappings.html needs to check for ghg and ghg_i.

crutan:
Thanks, Ross.  Included the simplified errors, and I have added ghg and ghg_i to the danger class check.

perryr16 (approved):
works great!

Issue #4749, opened 2024-08-14
AT workflow: better handling of the paginated submissions response


kflemin:
The Audit Template API response is paginated, with a per_page default of 50. For orgs with lots of submissions, we are currently only getting the first 50 submissions (sorted with ascending submission date).  

SEED needs to handle this pagination better to retrieve ALL submissions.

PR #4750, opened 2024-08-15
Add ability to retrieve all Audit Template submissions

perryr16:
The Audit Template get submissions endpoint is paginated and can only fetch a single type of submission statuses. The max per_page value is 100.

#### What's this PR do?
- calls a loop that will increase the page number each iteration until the returned submissions are less than `per_page`.
- allows multiple audit template status types to be saved to the org as a comma separated string. For each valid type, a request to AT will be made and added to the submissions

- import submission data for a city with more than 100 submissions.
or
- if city has less than 100 submissions, the `per_page` variable on Line 161 in /audit_template/audit_template.py can be changed to a value less than the total submissions (ex: per_page = 2). 

#4749
Org Settings 
Screenshot 2024-08-20 at 6 24 54 AM


kflemin (approved):
🎉 works great!

PR #4753, opened 2024-08-16
Handle BETTER terms and conditions error

perryr16:
BETTER v1.7.1 will be released on 8/31/24. The release includes an update that requires users to accept terms and conditions before using the API.

#### What's this PR do?
- Refactors the error handling for the analysis modal. If an error is returned, the modal will display the error and remain open until the user closes the modal.
- Rewords the help text in org settings to direct users to create a BETTER account

Attempt to either a) validate BETTER token from org settings or b) create a better analysis, before and after the user has accepted terms

#4751

Screenshot 2024-08-16 at 12 51 35 PM
Screenshot 2024-08-16 at 12 51 14 PM


kflemin (approved):
🚀

PR #4754, opened 2024-08-20
Doubled rehash performance

axelstudios:
#### What's this PR do?
Updates the rehash loops (in recent migrations and via manage.py) to be _at least_ twice as fast using three techniques:
1. Prepared statements
2. Skipping database writes when rehashing wouldn't change a record
3. Bypassing Django's save hooks (which also prevents modifying the state's `updated` value)

When rehashing 1,946,731 PropertyStates:
- If every record needs to be updated, the runtime goes from `30m 25s` ➡️ `14m 52s` (**2x faster**)
- If none of the records need to be updated, rehashing only takes `30m 25s` ➡️ `4m 16s` (**7x faster**)

##### Bug Fixes
- Fixes the `manage.py rehash` command from loading all PropertyStates and TaxLotStates into memory
- Fixes `hash_state_object` being called twice for every state

Run `manage.py rehash` before and after changing to this branch (nothing in the database should change)


perryr16 (approved):
Saw 3x speed improvement when rehashing 1500 records. 🏅

kflemin (approved):

PR #4756, opened 2024-08-21
Speed up step 6

haneslinger:
The most important changes are in seed/data_importer/match.py. I rewrote `link_states` to 1) uses a view lookup table and 2) only do linking, not merging.

`seed/utils/match.py` has a few little changes
- rename `matching_criteria_column_names` to `matching_criteria_column_names`
- `_link_matches` doesn't need `matching_views` to be ` query anymore
- `match` returns `all_matching_views` instead of `matching_views_in_cycle` and `matching_views_out_of_cycle`

on a fresh db, importing ~4000 properties when from 5.6 mins to 2.7 mins.
on a db that already had couple thousands states, importing ~4000 properties when from 7.2 mins to 2.4 mins.

On prod, this file has been taking hours, so its hard to say to what degree this will improve performance, but I hope the gains will grow dramatically.


seed/data_importer/match.py
```diff
@@ -728,32 +746,40 @@ def states_to_views(unmatched_state_ids, org, access_level_instance, cycle, Stat
...
     unlinked_states = []
-
-    batch_size = math.ceil(len(states) / 100)
     for idx, state in enumerate(states):
+        # get matches
+        existing_state_matches = match_lookup.get(tuple(getattr(state, c) for c in tuple_values), [])
+        existing_views_matches = [view_lookup[x["id"]] for x in existing_state_matches]
+
+        # ensure ali is correct
+        view = ViewClass.objects.filter(state_id=state.id, cycle_id=cycle.id).select_related(f"{class_name}__access_level_instance").first()
```
perryr16, line 766: Could this query could be sped up by using states realtion to view
`state.propertyview_set.filter(cycle_id=cycle.id).select_related(f"{class_name}__access_level_instance").first()`

side bar: why is PropertyState a 1-M with PropertyView? It seems a 1-1 relationship would be more appropriate ([see line 407 in models/properties.py](https://github.com/SEED-platform/seed/blob/develop/seed/models/properties.py#L407))

haneslinger: I''m unsure if that speeds it up, but it can't hurt!

> side bar: why is PropertyState a 1-M with PropertyView? It seems a 1-1 relationship would be more appropriate (error log if ps has more than 1 pv)

My fake answer is that hypothetically, technically, kinda sorta, two views of a single property could want to be represented by the same state, should there only be one record of info for both 2012 and 2013 or something, idk. You'd have to do that programmatically, the UI (and API I think) would never allow that.

My real answer is I believe the previous owners of this codebase didn't know how to code 1-1s. If you look for it you start to see those oversights around.

perryr16: Yeah, I didn't see any meaningful difference while testing both versions of that query.

Thanks for both answers :), I fear the latter may be true.

haneslinger:
@kflemin There's some sorta transaction error going on in the tests. I dont understand, do you?

kflemin (approved):

perryr16 (approved):

Issue #4675, opened 2024-05-21
200 error trying to import file by logging into ESPM from SEED


RDmitchell:

RDmitchell:
Instance: production
SHA: 1dc901360
Org: LBNL 200
Data set: ESPM Import Test

The program got further, to Geocoding, and then appeared to get stuck there

![image](https://github.com/SEED-platform/seed/assets/6314950/a4596ad4-16b9-403a-b2ae-ca2afaed35e1)

It seems that Geocoding is causing a 200 error. This was importing the ESPM report as an Excel file, not using the SEED / ESPM login feature

![image](https://github.com/SEED-platform/seed/assets/6314950/91216d46-aa02-4aed-8cea-7f586c9f8153)

*************************
Instance: dev1
SHA: f5fc3b772
Org: LBNL 200
Data set: EPA Test Import

Mapping
![image](https://github.com/SEED-platform/seed/assets/6314950/ec5b68cc-e706-4a65-9233-a73428bd8886)
Mapping Review
![image](https://github.com/SEED-platform/seed/assets/6314950/630900be-5476-4a7d-a8e7-1aa2d108316a)
Finding Inventory Matches -- 200 error
![image](https://github.com/SEED-platform/seed/assets/6314950/69e95c6c-d94f-4307-99bb-4f25216207a0)

RDmitchell:
It would appear that the MapQuest API key is invalid. 
![Image](https://github.com/SEED-platform/seed/assets/6314950/6c99246e-b37f-4a4c-896b-dbf3dbc60b60)

But I don't really want to have geocoding on anyway.

RDmitchell:
nstance: dev1
SHA: f5fc3b772
Org: LBNL 200
Data set: EPA Test Import

If I uncheck the geocoding fields, the import gets past the Geocoding 200 error. 

Got through the entire import
![Image](https://github.com/SEED-platform/seed/assets/6314950/b9e91bc9-1314-441c-905b-0500a57f8e59)

I guess we need a way to bypass the geocoding step in mapping if there are problems with the geocoding settings? Either the API or the fields set for geocoding? 

Need to give the user some information about how to deal with this.

RDmitchell:
Instance: production
SHA: 1dc901360
Org: LBNL 200
EPA Test Import cycle and data set

Unchecked the Geocoding fields in Column Settings, and then the program could import the Custom Report Template from the ESPM login in SEED. 

Seems like we need to have more user help about this problem.

[...]

Issue #4674, opened 2024-05-21
Show field mapping for ESPM imports via logging in to ESPM account


RDmitchell:

RDmitchell:
Also, I am wondering what happens in this case when there are not appropriate matching fields defined between the existing data and the data imported from ESPM. I haven't really tested the direct ESPM import, so I am not sure how the fields are mapped. I will try testing that, although our test ESPM account doesn't seem to have very good data in it.

RDmitchell:
I have made an issue folder and have put a sample ESPM file there, from the SEED testing ESPM account

https://drive.google.com/drive/folders/1GbVV_F4iPyNQr79Muc0yf_PAX_NEidaB?usp=drive_link 

Not sure it will be useful but it was generated from the ESPM Custom Report Template called 

SEED City Benchmarking Report

which has 4 records in it

PR #4762, opened 2024-08-28
Improve handling of failed ESPM imports 

perryr16:
If an org MapQuest API key was invalid, the import process would stall and the mapping suggestions would not be saved.

#### What's this PR do?
- puts a catch statement on the `uploader_factory.check_progress_loop_main_sub` function.
- always display existing mapping suggestions, even if matching had not previously completed
- translations
- refactors for clarity

Upload with an invalid MapQuest API.
1. Enter an invalid MapQuest API key. (if the API key is blank, geocoding is skipped)
2. Upload ESPM data: Data > Import Portfolio Manager Data > Log in with SEED_City_test > select SEED City Benchmarking Report > go through mapping/matching.
3. User should be presented with an informative error message.
4. After the failed import, go to Data > click on the dataset name > Data Paring for the failed import > the mappings should appear but are still editable as it was unsuccessful.

#4675 #4674

Screenshot 2024-08-27 at 1 41 53 PM


seed/static/seed/js/controllers/mapping_controller.js
```diff
@@ -896,11 +896,7 @@ angular.module('SEED.controller.mapping', []).controller('mapping_controller', [
 
     const init = () => {
       $scope.initialize_mappings();
-
-      if ($scope.import_file.matching_done) {
-        display_cached_column_mappings();
-      }
```
perryr16, line 902: This was an intentional decision at one point.

kflemin (approved):

PR #4763, opened 2024-08-28
Add derived data

haneslinger:
To test:
1. Create derived columns
2.  Look at those columns, they should be empty!
3. call the action "Update derived data on them". They will turn yellow, and update.
4. edit a field that'll change the derived data and rerun step 3. 

TO COME: calling the action automatically.


perryr16 (changes requested):
Overall, this is promising! A few notes
- Looks like the tax lots update_dervied_data endpoint isn't configured yet.
- Property derived columns are not populated when viewing associated properties on the tax lots tab. (Same for tax lots on the properties tab)
- The evaluation seems pretty quick for 1000 properties. What are you expecting for an evaluation time on a large dataset? I ask because I think a progress bar and an auto refresh following the completion of the task would be nice for the user.
- It's super small, but the modal could use some padding.

haneslinger:
@perryr16 theses are good notes.

> Looks like the tax lots update_dervied_data endpoint isn't configured yet.
Property derived columns are not populated when viewing associated properties on the tax lots tab. (Same for tax lots on the properties tab)

Will fix

> The evaluation seems pretty quick for 1000 properties. What are you expecting for an evaluation time on a large dataset? I ask because I think a progress bar and an auto refresh following the completion of the task would be nice for the user.

Its in the background, no need for progress bar. Run with celery to see

> It's super small, but the modal could use some padding.
👍🏻

perryr16:
Let me rephrase, if you'd like to have an auto-refresh instead of requiring the user to refresh manually, you could use the `uploader_service.check_status_loop` to monitor the Celery task and trigger a page or UI-grid reload when it’s complete.

haneslinger:
@perryr16 I think I understand. Would you be able to link an area in the code I can ape?

perryr16:
@haneslinger, for sure.

If you take a look at the [delete_cycle_modal_controller.js](https://github.com/SEED-platform/seed/blob/develop/seed/static/seed/js/controllers/delete_cycle_modal_controller.js#L89) you should be able to track the logic. 

Essentially, the view generates a `progress_key` and asynchronously kicks off a background task which will perform the logic and update that progress object. While the task is running, the view immediately returns the `progress_key`. The frontend hits `check_progress_loop` with the given `progress_key` every 750ms to measure the percentage complete.

perryr16 (approved):
Moved frontend suggestions to a new issue #4811

PR #4769, opened 2024-09-04
Exclude Column From Hash - Front End  work and Rehash Job

crutan:
#### What's this PR do?

This PR adds the ability to flag a column for exclusion from the hash used to compare states at import.  This allows users the ability to ignore columns that may contain unreliable data when comparing.

The changes include a new endpoint for updating multiple columns in one request, as well as the ability to rebuild the hash on all properties effected by a change to the columns in the hash.

Go to organization->column settings, and then flag a column for exclusion from the hash.  The modal will display a progress bar while rehashing is done.

https://github.com/SEED-platform/seed/issues/4071
includes work from this PR: #4687


haneslinger:
looking good! 

About hitting refresh, I get this warning, even if the job it done.
image

I have `CELERY_TASK_ALWAYS_EAGER` set to true, maybe it's fine with celery on? what are you workin with?

crutan:
> About hitting refresh, I get this warning, even if the job it done.
> I have `CELERY_TASK_ALWAYS_EAGER` set to true, maybe it's fine with celery on? what are you workin with?

Needed to empty onbeforeunload() so the browser skips that.  Done.  Celery task setting shouldn't matter here.

kflemin (approved):

PR #4782, opened 2024-09-10
Improve handling of 2 factor authentication requirement

perryr16:
If a user was a member of multiple organizations, and one of those organizations required 2-factor authentication, the 2-factor 'disabled' button was still available on other organizations.  If 'disabled' was selected it would appear to save, but the selection would revert to the previous selection.

#### What's this PR do?
- Checks all organizations that the user belongs to (excluding super_user access) for a 2fa requirement. If any orgs require 2fa the option will be grayed out and help text indicating which orgs require 2fa is shown.
- Adds a SEED Home button to 2fa landing.

- Add a user to org1 and org2
- Require 2fa in org1
- Attempt to disable 2fa in org2

Screenshot 2024-09-10 at 11 30 36 AM


kflemin (approved):
works great! thanks!

PR #4794, opened 2024-09-12
Fix Django autoreload

axelstudios:
Using a fresh Python virtual environment, SEED no longer runs thanks to a bug introduced in the `importlib_metadata` dependency used by Django

- https://stackoverflow.com/questions/78977665/django-autoreload-raises-typeerror-unhashable-type-types-simplenamespace
- https://github.com/python/importlib_metadata/issues/506

#### What's this PR do?
Locks `importlib_metadata` to the previous version, v8.4.0, until the issue has been resolved

1. Run `pip install -r requirements/base.txt`
2. Check that Django starts successfully


kflemin (approved):
🔥

Issue #4800, opened 2024-09-17
Update sessions to expire after a certain about of time


kflemin:
Implement a solution that would allow us to set the session expiration time per SEED server via an environment variable or similar mechanism.  Certain SEED instances may have stricter requirements for session expiration due to inactivity (ex: 30minutes)

PR #4801, opened 2024-09-17
Added `COOKIE_EXPIRATION` environment variable

axelstudios:
We were using `SESSION_EXPIRE_AT_BROWSER_CLOSE`, which in "normal" cases causes the user to be logged out every time they close their browser. However, in modern browser environments with Chrome and Firefox most users have their browsers set to resume their session upon reopening, which causes sessions to be valid until the CSRF cookie expires (**1 year**).

#### What's this PR do?
Adds support for a `COOKIE_EXPIRATION` environment variable that limits cookie validity to the Django default of two weeks, unless overridden

1. Logout/Login
2. Check the `Application/Cookies` section of the browser's developer tools, and ensure that the expiration is two weeks from now instead of `Session`

#4800


kflemin (approved):
🔥

Issue #4719, opened 2024-07-12
Investigate/Scope ability to move properties between access levels


kflemin:
There is a potential use case where a property might need to get moved to a different access level (provided the user has access to do so).  Investigate/scope out this case

ebeers-png:
I definitely think it's useful to be able to change a property's access level within SEED. As an existing user I'm used to changing a building's data within SEED, so I was confused when I wanted to move buildings into a newly-created lower access level and couldn't. I didn't want to create and upload brand new spreadsheets just for updating their access level either, in the event that my spreadsheets didn't match whatever data I had in the organization already.

I haven't looked at how access levels are actually implemented, though, so I understand if moving properties between them is not really allowable.

haneslinger:
from a server side perspective, it should be as simple as altering a properties ali. I'm trying to sniff out any use cases where that would cause a problem, like any possible fall out from a user having then losing access to a property. I think its fine.

@kflemin, what would the ui be?

kflemin:
I think this should be something that is available in the "Actions" dropdown on the inventory list page (that way you can bulk-change).  We could also make it available in the Actions dropdown of a property's detail page.  I envision a modal opening with dropdowns to select Access Level and Access Level Instance.

PR #4802, opened 2024-09-18
Add move properties option to inventory actions dropdown

crutan:
#### What's this PR do?

Adds a move properties option to the dropdown menu on the inventory screen "Move properties between ALIs".  The option is selectable only if properties have been selected in the grid.  When chosen, the option will display a modal allowing the user to select from their available ALIs as a destination to move the properties.  Once the move is complete, the grid will refresh.

In an instance with multiple ALIs, move items between them using the selection menu and the modal.

#4719 

Select option:
image

Modal:
image


kflemin:
@haneslinger can you review this one since it touches AH?

seed/views/v3/properties.py
```diff
@@ -1758,6 +1758,63 @@ def delete_inventory_document(self, request, pk):
...
+        org_id = self.get_organization(request)
+        ids = request.data.get("property_view_ids", [])
+        ali_id = request.query_params.get("access_level_instance_id", None)
+
+        if not ali_id:
+            return JsonResponse({"status": "error", "message": "Target ALI not found"}, status=status.HTTP_404_NOT_FOUND)
+
+        # filter ids based on request user's ali
+        ali = AccessLevelInstance.objects.get(pk=request.access_level_instance_id)
+        target_ali = AccessLevelInstance.objects.filter(pk=ali_id, lft__gte=ali.lft).first()
```
haneslinger, line 1791: I'm a little confused by this line, can you explain?

seed/views/v3/properties.py
```diff
@@ -1758,6 +1758,63 @@ def delete_inventory_document(self, request, pk):
...
+        """
+        org_id = self.get_organization(request)
+        ids = request.data.get("property_view_ids", [])
+        ali_id = request.query_params.get("access_level_instance_id", None)
+
+        if not ali_id:
+            return JsonResponse({"status": "error", "message": "Target ALI not found"}, status=status.HTTP_404_NOT_FOUND)
+
+        # filter ids based on request user's ali
+        ali = AccessLevelInstance.objects.get(pk=request.access_level_instance_id)
```
haneslinger, line 1790: It might be worth checking the user has access to this ali.

You should be able to do this with a decorator like this: https://github.com/SEED-platform/seed/blob/4cf3e362f14059faff77aef1252f66a61e9cc266/seed/views/v3/analyses.py#L80

haneslinger (approved):
Back end looks good! two small comments

seed/views/v3/properties.py
crutan: The intention here is to ensure that the user has access to the ALI they are targeting with the move.  the ALI on line 1790 is the user's assigned access level instance, and the ali_id is the intended target ALI's pk.  The filter attempts to load the target ali by comparing their left sides.

seed/views/v3/properties.py
crutan: Oh, yes, that is much more straightforward than what I am doing.  Thank you.  I need to move the access_level_instance_id into the post body for this, but that is not a big deal.

crutan: And that is now done.  Removed the confusing bits with the lft__gte filter in favor of the decorator.  Thanks, Hannah!

kflemin (approved):
🏆

RDmitchell:
@kflemin -- do users know what an ALI is ?? I'm not sure I do -- Access Level <?> ... what does the I stand for?

seed/views/v3/properties.py
haneslinger: yay! for future reference, to ensure an ALI is within ones subtree, the lft must be greater *and* the rgt must be less.

Take for example the tree below. If we only compare lft, it would seem that Duodenal was a child of gastric.
image

crutan: Thanks, that helps make it clear.

Issue #4809, opened 2024-09-20
Add filter groups to the default reports page


kflemin:

RDmitchell:
Instance: dev1
SHA: 3a0fb4476
Org: LBNL 5055

Works great !

Without the filter group, all the data for the cycle is displayed
![image](https://github.com/user-attachments/assets/1cf27367-6478-44ee-9e64-854bbb8c17a7)

With the filter group, only the records for that filter are displayed, ie, 32
![image](https://github.com/user-attachments/assets/2f87a2e5-6f1a-42ba-bd85-b45d8e57eee3)

PR #4812, opened 2024-09-24
Add filter groups to default reports page

crutan:
#### What's this PR do?

Adds a selection widget for filter groups to the default reports page.

Create a filter group that narrows your data set.  Run a report with no filter group selected, observe the number of properties, then select a filter group and update the charts - you should see your property count decrease according to the filter group conditions

#4809 

image
image


seed/views/v3/organizations.py
```diff
@@ -850,6 +854,15 @@ def get_raw_report_data(self, organization_id, access_level_instance, cycles, x_
...
 
+        if filter_group_id:
+            filter_group = FilterGroup.objects.get(pk=filter_group_id)
+            if filter_group.query_dict:
+                qd = QueryDict(mutable=True)
+                qd.update(filter_group.query_dict)
+                columns = Column.retrieve_all(org_id=organization_id, inventory_type="property", only_used=False, include_related=False)
+                filters, _annotations, _order_by = build_view_filters_and_sorts(qd, columns, "property")
+                all_property_views = all_property_views.filter(filters)
+
```
haneslinger, line 865: Does this handle the filter group's label filtering too? I think you have to do something like in[ this function ](https://github.com/SEED-platform/seed/blob/Fix-element-statistics/seed/models/data_views.py#L31), I I bet you can get it a little cleaner

seed/static/seed/js/controllers/inventory_reports_controller.js
```diff
@@ -56,6 +57,10 @@ angular.module('SEED.controller.inventory_reports', []).controller('inventory_re
     $scope.access_level_instance_id = null;
     $scope.users_access_level_instance_id = user_service.get_access_level_instance().id;
 
+    filter_groups_service.get_filter_groups('Property').then((filter_groups) => {
+      $scope.filter_groups = filter_groups;
+    });
```
haneslinger, line 62: This could be done in `seed.js` too but I dont really think it matters 🤷🏻

haneslinger (changes requested):
@crutan, im sure I speak for the whole team when I say im really pleased and impressed with how quickly you have come up to speed on SEED and the work you've been contributing! I only have one note

seed/views/v3/organizations.py
crutan: It sure doesn't.  Thanks for pointing that out.

seed/static/seed/js/controllers/inventory_reports_controller.js
crutan: I've moved this into seed.js - I think that makes good sense.

seed/tests/test_organization_views.py
```diff
@@ -283,4 +283,5 @@ def test_filtered_report(self):
 
         url += f"&filter_group_id={self.office_filter_group.id}"
         resp = self.client.get(url, content_type="application/json")
+        print(resp.json()["data"])
```
haneslinger, line 286: remove

seed/tests/test_organization_views.py
```diff
@@ -268,7 +268,7 @@ def setUp(self):
         self.office_filter_group = FilterGroup.objects.create(
             name="office",
             organization_id=self.org.id,
-            inventory_type=1,  # Property
+            inventory_type=0,  # Property
```
haneslinger, line 271: 👍🏻

seed/models/filter_group.py
```diff
@@ -21,12 +24,47 @@ class FilterGroup(models.Model):
...
-    def filtered_property_views(self, columns=[], views=None):
-        qd = QueryDict(mutable=True)
-        qd.update(self.query_dict)
-        filters, _annotations, _order_by = build_view_filters_and_sorts(qd, columns, "property")
-        filtered_views = views.filter(filters)
+    def views(self, views, columns=[]):
+        if VIEW_LIST_INVENTORY_TYPE[self.inventory_type][1] == "Property":
+            return self._filtered_property_views(views, columns)
+        elif VIEW_LIST_INVENTORY_TYPE[self.inventory_type][1] == "Tax Lot":
+            return self._filtered_taxlot_views(views, columns)
```
haneslinger, line 31: It's not strictly needed, just an fyi, if you want to only have one function, you could do something like [this](https://github.com/SEED-platform/seed/blob/Fix-update-display-name/seed/utils/match.py#L257), where the class is variable

haneslinger (approved):
Noice

seed/models/filter_group.py
crutan: Oh, perfect, thank you for the pointer.  I'll implement that and collapse these down into a single function.

kflemin (approved):
🎉

PR #4819, opened 2024-09-27
Fix rehash migrations prior to `is_excluded_from_hash`

axelstudios:
If you had a database that was on a migration earlier than [migration 0223](https://github.com/SEED-platform/seed/blob/develop/seed/migrations/0223_water_use_columns.py) and attempted to migrate using the latest codebase, then 0223 would fail because #4769 changed the rehash function to expect a new column to exist.

#### What's this PR do?
- Changes the rehash function to check if [migration 0225](https://github.com/SEED-platform/seed/blob/develop/seed/migrations/0225_column_is_excluded_from_hash.py) has run yet, and if not it uses the older rehash logic
- Adds a performance improvement so that an organization's columns are fetched only once instead of for every Property/TaxLot State (19% faster than the improvements added in #4754)

1. Checkout SEED v3.0 and create a new database
2. Import some data
3. Checkout the develop branch
4. Run migrations - there should be 2 rehash migrations (the first uses the old logic, and the second uses the new logic)

### DIFF
--- a/seed/data_importer/tasks.py
+++ b/seed/data_importer/tasks.py
@@ -18,7 +18,7 @@ from collections import defaultdict, namedtuple
 from datetime import date, datetime
 from itertools import chain
 from math import ceil
-from typing import Union
+from typing import Optional, Union
 
 from _csv import Error
 from celery import chain as celery_chain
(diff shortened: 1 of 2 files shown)


kflemin (approved):
🏆

PR #4821, opened 2024-09-27
Fix `hiredis` not found error

axelstudios:
Any new docker builds (or virtual environments) created within the last 5 hours have failed to start celery due to the error `name 'hiredis' is not defined`.

This is because `redis` v5.0.8 had the dependency `hiredis>=1.0.0`, but v5.1.0 included the breaking change `hiredis>=3.0.0`. `redis` is a sub-dependency of our dependency `django-redis`, which has a _very_ liberal requirement of `redis>=4.0.2`.

#### What's this PR do?
Updates `hiredis` to v3.0.0 to match the [requirement from the sub-dependency `redis`](https://github.com/redis/redis-py/blob/v5.1.0/setup.py#L59) to resolve celery launch errors

1. Create a new virtual environment
2. Run `pip install -r requirements/base.txt`
3. Ensure Django and celery both start successfully


kflemin (approved):
🍩

Issue #4811, opened 2024-09-23
Add progress bar to derived data fetch


perryr16:
Add a progress bar to track the derived data update progress. After it has completed, refresh the inventory

PR #4825, opened 2024-10-01
Add progress bar to derived data update

perryr16:
#### What's this PR do?
- adds a progress bar when update derived data is called 
- auto refreshes after derived data has been updated

Upload a bunch of properties, create a bunch of derived columns, update derived data

#4811 

https://github.com/user-attachments/assets/14363f54-32de-40b9-be05-777b3738d366


haneslinger:
wait, so can the user still click away from this process if they want? It could take a really long time, no? If they click away, will there be any indication that those columns are still updating?

perryr16:
Interesting. So the user can click away, but there's no indication the process is still running. Either we could make the modal static, forcing the user to wait until it's done, or I could add back the header styling. What's your preference?

haneslinger (approved):

Issue #4817, opened 2024-09-25
Default Reports - Add a statistics table


kflemin:
On the Default Reports page, under the charts, add a table listing a few statistics about the data in the chart.  We would like to display the following for each axis: 
- sum
- mean
- min & max
- 5% (5th percentile)
- 25% (25th percentile)
- 50% (median)
- 75%
- 95%

Additionally, if a user has selected a particular access level instance and that instance is not a leaf node, add rows in the table to display the statistics for each Access Level Instance included in the selected instance level (just 1 level down).  If there is no access level selected or the org has no access levels, then the table would have a single row.
<img width="978" alt="Screenshot 2024-09-25 at 1 29 24 PM" src="https://github.com/user-attachments/assets/8b86fa0a-6857-4530-b3b3-0943be5bbd0b">

RDmitchell:
instance: dev1
SHA: 3a0fb4476
Org: LBNL 5055

I am seeing this in my test account, although I don't have AH set up other than root
![Image](https://github.com/user-attachments/assets/611c7e40-d8d6-413b-a67a-c155319655e1)

@kflemin -- is seeing this good enough to close this issue, or does it need to be tested in an org with AH set up?

RDmitchell:
Instance: dev1
SHA: 389985212
Org: BPS - DC

The statistics table is working for Default reports, but I'm wondering about the value of the "Sum" table ... in the example shown above, and which I also tested, the program is summing the field "Year Built", which doesn't seem to make too much sense to me. 

Not sure how to decide which values to sum, but it might be worth adding some user control over that? 

![Image](https://github.com/user-attachments/assets/6699e763-0353-4c1d-a1bd-d44dc5a841ce)

PR #4831, opened 2024-10-07
Build axis data rollup table

crutan:
#### What's this PR do?

Adds an additional data view to the default reports page to display rolled up data for each axis at the current ALI plus 1 level down into that ALIs children

#4817 

image


kflemin (commented):
🏆

Issue #4836, opened 2024-10-08
"Delete Organization" button should have a confirmation


devinbrady:
**Is your feature request related to a problem? Please describe.**

When a superuser is on the Admin page, they have a red Delete Organization button. When you click the button, deletion starts right away. 

**Describe the solution you'd like**

There should be a confirmation for this destructive action. 

**Additional context**

The admin page is probably used by very few people. Still, there is a possibility that an errant click could destroy data.

RDmitchell:
@devinbrady - great !! I have been deleting orgs and have meant to make an issue for this, so thanks !!

RDmitchell:
instance: dev1
SHA: 3a0fb4476

This message now appears when you click on Delete Organization 
![image](https://github.com/user-attachments/assets/c7fb3cae-ed8c-4ca4-add8-235ad29ef7d1)

Thanks !!

PR #4851, opened 2024-10-16
Fix organization delete - ensure there is a chance to confirm and that it works in one step.

crutan:
#### What's this PR do?

Reworks the organization deletion job in celery to ensure the property/taxlot deletions happen before the organization deletion is attempted.

Adds a modal confirmation for organization deletion, migrates the inventory progress bar to that modal.

Create an org, load some inventory, then delete the organization.

#4814 #4836 

image


haneslinger:
this is all so much cleaner. Can you explain where the bug was?

crutan:
> this is all so much cleaner. Can you explain where the bug was?

I can try.  This is based on a lot of experimentation and reading of the celery docs - and I may well be misunderstanding them.

In short, I think the chain() method, while techincally working correctly, is not aware of the underlying tasks generated in the task it fires off - `_delete_organization_inventory` doesn't do any deletion itself, but instead schedules a slew of tasks to delete chunks of inventory.  It appears that the chain only waits until that `_delete_organization_inventory` task finishes before movign on, and when it does there is usually still inventory in the process of being deleted so the ` Organization.get().delete()` call fails.

So I effectively single-threaded it.  This will mean that orgs with boatloads of properties/taxlots will take longer to delete - but I'm not entirely sure how core of a use-case that is.  This is doing deletes in chunks of 100.  We could certainly increase the chunk size there - postgres is fast and could handle a bigger set of IDs without any issue.

Issue #4858, opened 2024-10-16
Default Reports - use the min/max ranges from the stats table to set axes range


kflemin:
For YearBuilt specifically, the charts are tough to read b/c all datapoints are bunched around 1980-2024 while the axis stretches all the way to 0.  Investigate using the min/max range returned for the stats table as the range for the axes.

crutan:
I have properties in my db from 1880-2012, I tweaked the chart setup to use min/max values from the actual data, and it looks like this:

![image](https://github.com/user-attachments/assets/bd72f5ce-3081-420f-bbbb-bafa9e2b4dca)

The only downside I see here is that the first and last in the series are those half-bubbles at the beginning and end of the chart.  I could keep it like this, or add some cushion (1% of the mean value or something) to the beginning and end?

PR #4859, opened 2024-10-17
Ensure year_built labels do not have commas, adjust width of scatterchart

crutan:
#### What's this PR do?

Fixes the range for numeric x-axis values to be the actual range of data +/- a half percent cushion

Fixes year_built labels to be strings, so the system will not add commas after the thousands digit.

Run a chart where the x-axis is year built, note the range and labels.

#4858

image


kflemin (approved):
🏆

Issue #4852, opened 2024-10-16
Portfolio Summary Page: Allow sorting on additional columns


kflemin:
Add ability to sort on "pass checks" "notes", "questions", "new construction" columns (everything after % improvement)

PR #4862, opened 2024-10-18
Updates to portfolio summary

perryr16:
The following is based on user feedback.

#### What's this PR do?
- Allows users to sort on portfolio summary columns that represent data from related models (`goal_note` and `historical_note`).
- Fixes bug that ignored 0 values in data quality checks, leading to false "passed checks"
- Excludes baseline new builds from summary calculations

- Attempt to sort on Question, Resolution, Historical Note, and Passed Checks
- Set EUI to 0 for a property, rerun data quality checks. Property should fail and a label applied
- Set New Build or Acquired to true for a property, baseline calculations should update to ignore this property (looking at the change in total area is easiest), while current calculations remain unchanged. 
#4852
#4583


seed/static/seed/js/controllers/portfolio_summary_controller.js
```diff
@@ -820,7 +821,13 @@ angular.module('SEED.controller.portfolio_summary', [])
 
           if (sort.direction) {
             // remove the column id at the end of the name
-            const column_name = name.split('_').slice(0, -1).join('_');
+            let column_name;
+            if (name.includes('.')) {
```
perryr16, line 825: is this not specific enough? can columns have periods?

seed/utils/search.py
```diff
@@ -507,3 +507,40 @@ def build_view_filters_and_sorts(
...
+        column_name = column_name.replace("goal_note", "goalnote")
+
+    boolean_column = column_name == "property__goalnote__passed_checks"
+    target: Union[bool, str]
+    if boolean_column:
+        target = False
+    else:
+        target = ""
+
+    related_model = Case(When(**{column_name: target}, then=Value(1)), default=Value(0), output_field=IntegerField())
```
perryr16, line 538: By defautl, django will follow these patterns
- asc: None, "A", "B", "Z"
- desc: "Z", "B", "A", None
I thought this was counterintuitive and if a user sorted a text column like historical notes, "A" should be at the top, and blanks should below "Z". If this doesnt matter I can simplify this function by removing the `related_model` annotation

kflemin: I like the way you implemented it!

kflemin (approved):

PR #4863, opened 2024-10-21
Add modal to update meter connections

haneslinger:
- add "include_systems" param to the get groups endpoint
- add modal to change a meters connection
image


haneslinger:
included and merge in https://github.com/SEED-platform/seed/pull/4868

Issue #4847, opened 2024-10-10
Export insights graphs/reports


kflemin:
Export simple reports from insights tab in CSV and PNG format

RDmitchell:
Instance: dev1
SHA:
Org: LBNL 5055

I think this is a good start. 

* Property Insights has an Export button which exports a PNG file of the graph
* Default Reports has an Export Data button to export the data to an XLXS file and an Export Charts to a PNG file.
* Custom Reports has a Download chart button to download a PNG of the graph

And the PNGs and XLSX files look good.

PR #4869, opened 2024-10-28
Add export charts to default reports

haneslinger:
https://github.com/SEED-platform/seed/issues/4847
image


crutan (approved):
Looks good to me - I think the unittests will succeed on re-run - they did for me earlier.

PR #4885, opened 2024-11-21
Portfolio summary move to backend

perryr16:
ICF reported sorting issues on the goal_note columns, leading to duplicate and unsorted results.

#### What's this PR do?
- Moves many of the portfolio summary formatting functions to the backend. The property retrieval function was originally reused from the inventory list, however as the portfolio summary has grown in complexity, having a dedicated function for property retrieval will simply the requests.
- Fixes the portfolio summary sorting on goal_note columns (question, passed checks...). The duplicate records were related to a SQL join when ordering on a 1-many field.

- create multiple goals with the same properties. Attempt to sort on all goal_note and historical columns (compare with develop)
- This introduces a lot of new changes. Double check all existing functionality. There should be no differences in functionality

[#4886](https://github.com/SEED-platform/seed/issues/4886)


seed/utils/search.py
```diff
@@ -509,39 +509,44 @@ def build_view_filters_and_sorts(
     return new_filters, annotations, order_by
 
 
-def build_related_model_filters_and_sorts(filters: QueryDict, columns: list[dict]) -> tuple[Q, AnnotationDict, list[str]]:
```
perryr16, line 512: The old function was the source of the sorting bugs because ordering by a related field in a 1-to-many relationship creates a SQL join, which duplicates rows when a property is part of multiple goals or goal_notes.

The new function works by querying and sorting the related models (goal_note & historical_note) independently, then mapping them back to property views. This avoids sorting-related issues at the parent level, simplifies the query, and resolves the duplication bug.

kflemin (approved):

PR #4888, opened 2024-11-22
Add derived columns to program setup

haneslinger:


kflemin:
can you double check my fix? derived cols were not evaluating in the compliance calcs on the insights pages b/c they don't have a numeric datatype...this is meant to fix that and was working for me.

kflemin (approved):

RDmitchell:
@kflemin / @haneslinger -- is there an issue associated with this PR?

kflemin:
I don't think so

RDmitchell:
@kflemin -- I don't really understand what this PR does -- does Program refer to making a program in Insights, so that now it is possible to have derived columns in those definitions?

RDmitchell:
@kflemin -- never mind. I think I understand. 

I have created a derived column, called TotalSiteEnergy, and I now see it in the list of fields when defining the energy metric in an Insights program.

image
image

PR #4891, opened 2024-12-05
Add transactions template to portfolio summary

perryr16:
Better Buildings has several data report templates they plan to use. One of which is a Transactions template used to track eui per transaction

Sample Transactions Template: [2024_DataReviewReport_TransactionTemplate_Returning Franchisees_4.8.2024.xlsb.zip](https://github.com/user-attachments/files/18029697/2024_DataReviewReport_TransactionTemplate_Returning.Franchisees_4.8.2024.xlsb.zip)

#### What's this PR do?
It incorporates the BB transaction template into the existing Portfolio Summary. The Goal model has been updated to include a "type" field that has choices of "standard" or "transaction", and a nullable "transactions_column". Regardless of goal type, the same endpoints are hit, but the frontend will interpret them differently based on the "type" field. The transactions template differs from the standard by using an additional "transactions" column to calculate the weighted EUI per restaurant transaction.

The fields included in the standard goal are: Area, kBtu, EUI, Area Change, EUI Change
The fields included in the transaction goal are: Area, kBtu, Transactions, EUI(s), EUI(t), Area Change, Transactions Change, EUI(s) change, and EUI(t) change.

Sample Data
[12props_eui_2021_2022_2023.xlsx](https://github.com/user-attachments/files/18029550/12props_eui_2021_2022_2023.xlsx)
The above has 12 properties across 3 different cycles (2021-2023). After creating those cycles, perform a multicycle upload.

To test, create a few goals of both types and confirm expected functionality.

#4892 

Screenshot 2024-12-05 at 1 43 24 PM


kflemin (approved):
💰

PR #4896, opened 2024-12-11
Add show all labels description text and translation

kflemin:


RDmitchell:
@kflemin -- is this something I can test?

PR #4901, opened 2024-12-12
Add or create labels during upload

perryr16:
#### What's this PR do?
Allows users to upload property and taxlot labels. Incoming values in a csv's "Property Labels" or "Tax Lot Labels" column will be processed and applied as a label. If a label does not exist, one will be created.

Labels are applied to a `PropertyView`, however during the upload process `PropertyStates` are created (mapping) several steps before a `PropertyView` is applied (matching). As a result, the `PropertyState` has to carry the incoming label information in a new text field `PropertyState.incoming_labels` until it can be applied to the related `PropertyView`.

From the inventory list, export a set of properties to a csv or spreadsheet. Then in a new cycle, upload that exported file, confirm the correct labels are applied.

#4789


kflemin (approved):

kflemin (changes requested):
@perryr16 This is almost working!

1.  I've noticed that nothing happens with the labels if we are importing data into a cycle that already has data in it (performing a merge).  No new labels are created, and the property labels are not added to the properties. Doesn't matter if the existing property had a label already or didn't. It does work as designed in an empty cycle though.
2.  When creating a new label via import (and really any time), can you make the default be "show in list = true" instead of false?  It's confusing otherwise.
3.  Can you also remove the first line of the csv inventory export that lists some information about filters? It breaks the subsequent CSV import and is not comprehensive anyway. I think just listing the headers directly will be better.

perryr16:
Most recent commits address the following
- new labels are defaulted to "show in list"
- labels are created/added during a merge import into a cycle with existing data

Outstanding issue / edge case
- incoming states flagged as duplicate will not have their labels processed. This is usually not an issue, however if a label is deleted from the existing property or organization, a new label will not be applied because the incoming state is recognized as a duplicate and is ignored. A fix that will not influence import speed will require further development.
- To recreate: import file with new label. Delete label. Reimport file.

PR #4904, opened 2024-12-13
Reduce dockerfile-dev build time in half with `uv` and `pnpm`

anchapin:
I started using `uv` locally to replace `pip` and noticed that it reduces the time to install packages by about half. Additionally, I noticed that when the dockerfile is built, that the `pip install -r requirements/local.txt` step was taking most of the time.

#### What's this PR do?
I looked up how to switch to `uv` from `pip` and implemented those changes. I also looked up if there was an equivalent speed up that can happen for replacing `npm install` and tried using `pnpm`. These changes are only included in the dockerfile-dev file since likely needs to be rebuilt the most often compared to dockerfile which is used for deploying seed.

Switch to the branch and use the following command: `docker-compose stop && docker-compose rm -f && docker-compose -f docker-compose.yml -f docker-compose.dev.yml build` and see how long the build process takes. For me it was about 150 seconds compared to about 300 seconds on the develop branch.

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -1,6 +1,8 @@
 FROM node:20-alpine3.16 AS node
+# Use a multi-stage build to first get uv
+FROM ghcr.io/astral-sh/uv:0.2.12 as uv
 
-FROM alpine:3.14
+FROM alpine:3.16
 
 COPY --from=node /usr/lib /usr/lib
 COPY --from=node /usr/local/lib /usr/local/lib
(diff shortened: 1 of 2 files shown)


anchapin:
Note that I updated the requests version number because I was getting an error with the old version. I'm not sure if this is really needed though.

anchapin:
I also revised the version of alpine to 3.16 to match what node is based on. I was seeing errors related to this but again not sure if this change is necessary.

axelstudios:
> I also revised the version of alpine to 3.16 to match what node is based on. I was seeing errors related to this but again not sure if this change is necessary.

The problem with Alpine 3.16 is that it install Python v3.10 instead of v3.9. That problem should be going away soon with the dependency updates

PR #4917, opened 2024-12-20
Fix  upgrade recommendation

haneslinger:


kflemin (approved):

laurenliz22 (approved):
I think this looks like everything we talked about.  Thanks Hannah!

Issue #4846, opened 2024-10-10
Allow user to add a Property or a Taxlot from the UI


kflemin:
Currently on SEED, the only way to add a building or taxlot is via import. We want to allow users to add a new building or taxlot via a form. 

This work should be scoped (design doc first).

perryr16:
[spec doc](https://docs.google.com/document/d/1zOP6fh2FMG5uNsQ01tlw3e2HLq9M2owLUeCzLugCDos/edit?tab=t.0#heading=h.vk8ozpxuhnd)

PR #4918, opened 2024-12-20
Create inventory form

perryr16:
Currently, the only way to introduce a new property or tax lot is through a data upload.

#### What's this PR do?
Introduces a new page available from the inventory nav to create a property or taxlot via a form. When the form is submitted, the backend creates a stub state, view, and property before using the existing "update" endpoint (inventory detail edit feature) to apply the configured state data passed from the frontend. By using the existing "update" endpoint, we can reuse the same matching/merging/linking functions.

notes: 
-  If a new property includes a jurisdiction tax lot id, a taxlot with that id and any other taxlot fields will be created, then linked through a new TaxLotProperty.
- Because a property create is treated as an "update" and not a data upload, the new property will not have a historical entry (building file) on the property detail.
- If a property with a taxlot is created, there will be 2 backend requests, one for a property create and one for a taxlot create.
- If a created property matches a property within the cycle, it is treated as an update and not a merge.

- Test form functionality: add row, delete row, column list profile, reset...
- Test validation: duplicate rows, missing matching criteria
- Create a new property/taxlot with some canonical and extra data fields
- Create a property/taxlot that matches an existing record in cycle
- Create a property/taxlot that matches an existing record out of cycle
- Create a property with an associated taxlot via jurisdiction tax lot id. The link should appear on the property detail under "Paired Tax Lots"

#4846

Screenshot 2025-01-08 at 10 58 46 AM


nllong:
was branched, updated, and merged. closing this one.

PR #4927, opened 2025-01-15
Remove server settings from the session cookie

axelstudios:
The `sessionid` cookie should have only the minimal session-specific data necessary, and not server settings that can become out-of-sync if stored in session tokens

#### What's this PR do?
Overrides the `two_factor.views.core.LoginView.render` method to make the backend setting `INCLUDE_ACCT_REG` available to the view template without storing it in each user's session cookie

1. In `local_untracked.py` toggle the `INCLUDE_ACCT_REG = True` variable and check that the login page correctly renders (or hides) the `Create my Account` button
2. Optionally inspect the session cookie data and ensure the `include_acct_reg` is no longer included. Copy your `sessionid` cookie value and run the following commands:
   ```bash
   # shell
   export DJANGO_SETTINGS_MODULE="config.settings.local_untracked"
   ```
   
   ```python
   # python
   from django.conf import settings
   from django.core import signing

   cookie_value = "YOUR_SESSIONID_COOKIE_VALUE"

   decoded = signing.loads(
       cookie_value,
       key=settings.SECRET_KEY,
       salt="django.contrib.sessions.backends.signed_cookies",
   )

   print(decoded)
   ```
   
   The new session data (for a logged-out user) should show the following:
   ```json
   {
     "wizard_custom_login_view": {
       "step": "auth",
       "step_data": {},
       "step_files": {},
       "extra_data": {},
       "validated_step_data": {}
     }
   }
   ```

### DIFF
--- a/seed/landing/forms.py
+++ b/seed/landing/forms.py
@@ -5,22 +5,8 @@ See also https://github.com/SEED-platform/seed/blob/main/LICENSE.md
 
 from django import forms
 from django.contrib.auth.forms import UserCreationForm
-from django.utils.translation import gettext_lazy as _
 
 from seed.landing.models import SEEDUser
-
-
-class LoginForm(forms.Form):
-    email = forms.EmailField(
-        label=_("Email"),
-        help_text=_("ex: joe@company.com"),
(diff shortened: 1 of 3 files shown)


perryr16 (approved):
Works great 🏆

Issue #4928, opened 2025-01-16
Groups: Can't seem to add inventory records to group


RDmitchell:
instance: staging
SHA: 3a0fb4476
Org: LBNL 200
Cycle: BETTER Testing

I created a Group from the Group link on the Inventory page
![image](https://github.com/user-attachments/assets/ec6841c4-0d51-419b-9c77-fa3156f35451)

In this org, there is only one AH level, ie, "root", so when I am logged in to that org, seems like I should be able to do any action
![image](https://github.com/user-attachments/assets/2c11cbf0-9b48-4cab-8c00-60c6d008c0b2)

In the BETTER Testing cycle, I checked the first record, and selected the "Add to / Remove from Groups" option in the Actions pulldown.
![image](https://github.com/user-attachments/assets/a2ca8ee7-e27a-4836-b270-f1006ae89e72)

When I click on the Add button in the Add/Remove Groups, I am not allowed to add it, and a box appears saying "Access level mismatch"
![image](https://github.com/user-attachments/assets/3ca6c214-99e7-45f6-9aa8-c293b8107f0a)

Probably user error, but I would have thought this was the path to get this to work.

RDmitchell:
instance: dev1
SHA: f4282854c
Org: LBNL 200
Cycle: 2023 Compliance Cycle

@kflemin -- This is almost working but it's still a bit buggy. 

Here is what happens. 

In the Inventory List view, I select a property, then select the Add to / Remove from Groups option, which opens this screen. I make a new group
![Image](https://github.com/user-attachments/assets/183611da-06a5-4492-bcac-d56e08369c2f)

The new group is added to the list in the modal, and there is an Add button that is "active". I click on that button to add the property to the group
![Image](https://github.com/user-attachments/assets/72bfa687-de43-4d23-8825-366e0dcfb83d)

The program "hangs" and the Done box is "grayed out". A line about autofocus appears in the Console tab of the developer tools, and the Add button for that new group is not highlighted.
![Image](https://github.com/user-attachments/assets/c601daa6-d91b-4d03-803a-13f8dc3a6d04)

If I click the non-highlighted Add button for the new group with the new building to add to the group, then the Done button becomes active, and I can click it
![Image](https://github.com/user-attachments/assets/fd595f37-4292-4935-a644-48b9ac8faf74)

Now in Property Groups, the new group is displayed and the property has been added to it. 
[...]

Issue #4908, opened 2024-12-17
SEED groups - 500 error when trying to create a new group after selecting inventory records 


kflemin:
Double check that users can create groups after selecting properties and selecting Add/Remove from group from the actions menu, that adding a new group from that modal works.   See comments: https://github.com/SEED-platform/seed/issues/4849

Also double check that the functionality works when an org has an AH. (I couldn't add 2 properties in the same ALI from that modal).

RDmitchell:
@kflemin -- I see you already made an issue for this. Perfect !

RDmitchell:
Here is a screenshot showing what I did -- I selected 3 properties, then clicked on the Actions menu and selected the Add to/Remove from Groups option, when in the Create new group input box, I typed a new group and clicked on the Create Group button, which generated a 500 error
![image](https://github.com/user-attachments/assets/3b652f51-d149-4853-829a-32b6ae16ef8d)

RDmitchell:
instance: dev1
SHA: f4282854c
Org: LBNL 200
Cycle: 2023 Compliance Cycle

I can make the group now.

PR #4932, opened 2025-01-21
Fixes inventory groups add inventory bug

perryr16:
There was an argument mismatch between the frontend and backend when fetching the ALIs associated with a selection for the groups modal. The frontend passed view_ids while the backend interpreted them as inventory_ids (property_ids or taxlot_ids). On small DBs this wasn't noticeable as the view_ids and property_ids aligned. However, in larger dbs, the views and the properties did not align and the backend would return zero ALI matches. This led to the frontend form being disabled as it expects the ALI count to equal 1, not 0.

#### What's this PR do?
- This PR corrects the django db logic by converting the view_ids to inventory_ids before filtering for access level instances. 
- Fixes smaller syntax and swagger bugs

From the inventory list, select a few properties with the same access level and add them to a group. Check for console errors.

I was not able to recreate the 500 error, however I'm hopeful this will correct it.
#4908 
#4928

The below screenshot was taken from develop after attempting to add a single property to a group. The warning is misleading as a single property would have exactly 1 ALI
Screenshot 2025-01-22 at 2 10 00 PM

### DIFF
--- a/seed/views/v3/access_levels.py
+++ b/seed/views/v3/access_levels.py
@@ -575,7 +575,7 @@ class AccessLevelViewSet(viewsets.ViewSet):
         if not view_ids:
             return JsonResponse({"status": "success", "access_level_instance_ids": []})
 
-        inventory_type, InventoryClass = ("taxlot", TaxLot) if inventory_type == "taxlots" else ("property", Property)
+        InventoryClass = TaxLot if inventory_type == "taxlots" else Property
         inventory_ids = InventoryClass.objects.filter(views__id__in=view_ids, organization_id=organization_pk).values_list("id", flat=True)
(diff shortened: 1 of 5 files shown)


seed/views/v3/access_levels.py
```diff
@@ -111,7 +111,6 @@ def descendant_tree(self, request, organization_pk=None):
     @has_perm_class("requires_owner")
     @action(detail=False, methods=["POST"])
     @swagger_auto_schema(
-        manual_parameters=[AutoSchemaHelper.query_org_id_field()],
```
perryr16, line 114: org id is automatically included as it appears in the url. "api/organizations/{org_id}/add_instance"

kflemin (approved):

PR #4955, opened 2025-03-18
Fix groups meters property column

haneslinger:
the property display name column now 1. works with extra data and 2. links. I wanna add a screen shot, but its army data, so, dm me i guess, @kflemin


kflemin (approved):
works great!

PR #4959, opened 2025-03-25
Fix ESPM download, change generateData from POST to GET

haneslinger:
ESPM changed their api! This broke our [tests](https://github.com/SEED-platform/seed/actions/runs/13996604162/job/39200257219?pr=4956):
image

specifically, this line [broke](url):
image

I tried calling the same endpoint on [portfoliomanager.energystar.gov](portfoliomanager.energystar.gov), and it appears to be a get
image

I changed it to a get and it worked!


nllong (approved):
This is great, thanks!

PR #4982, opened 2025-05-09
Fix bsyncr xml timeseries ordering

kflemin:
- Ensures that timeseries data in the bsyncr XML generation is sorted by date
- Also adds 'EndTimestamp' to the Timeseries element in the XML


nllong:
are these ESPM errors in the CI known?

PR #4991, opened 2025-05-27
Update redis key timeout to 15 minutes

kflemin:
We ran into some issues importing GeoJSON files. Issues seem to stem from losing a redis key during the geocoding step of the import process. The default timeout was 5 minutes...this PR increases it to 15 minutes.


axelstudios:
Expiration looks good 
image

PR #4993, opened 2025-05-29
upgrade from Alpine 3.14 to 3.18

dhaley:
This change addresses a critical vulnerability flagged by AWS Inspector:

IN-DISCONTINUED-001 – ALPINE_LINUX_3_14

The base image was previously using Alpine Linux 3.14, which reached end of support in November 2022. Amazon Inspector now considers this OS discontinued and no longer actively supported or maintained. Findings related to this are informational only, and upgrading is required to restore full Inspector coverage and ongoing security patching.

What's this PR do?

This PR updates the base image from Alpine 3.14 to Alpine 3.18 in both Dockerfile and Dockerfile.ecs.
Alpine 3.18 is the lowest currently supported version, providing security updates with minimal risk to system compatibility.

How should this be manually tested?

Build both Dockerfile and Dockerfile.ecs, then verify that SeedWeb application functionality is operational.

Basic functionality validation should include:

Confirming successful container build.

Launching the application locally or in ECS.

Validating web service endpoints and key app workflows.

Ensuring no regressions in authentication, database connectivity, or frontend rendering.

PR #5003, opened 2025-06-06
Remove dev dependencies

dhaley:

This fix for Dockerfile.ecs doesn't install node dev dependencies into the prod ECS image used in AWS.

#### What's this PR do?

It removes all critical vulnerabilities including a KEV and reduces high vulnerabilities from 45 to 12 and medium from 46 to 22.


axelstudios:
Since there's only one line changed, I'm including that with the #4993 PR

PR #5027, opened 2025-07-08
Add gfa to facilities plan

haneslinger:


kflemin:
@haneslinger, I think you just have a few tests to update now that you are calculating running gfa sum.

PR #5061, opened 2025-08-22
Fix export race condition

perryr16:
Reported Bug
- Large inventory exports (15000+ records) were failing with a 404 at the `/v3/cahce_entries/<id>` endpoint.

Expected Behavior
- When data is exported, it kicks off a progress object. When the progress object finishes, the frontend will request cached data.

Issue
- The backend considers the progress finished when `progress_data.finish_with_success()` is called, but the frontend treats any progress object with progress 100 as finished. 
- So with very large datasets the progress was hitting 100 before the backend was ready, causing the frontend to request data that did not exist yet.

#### What's this PR do?
Caps the running progress at 90% before the backend calls `progress_data.finish_with_success()` which will then set the progress to 100%. This ensuring the cache data has been set and avoids any frontend race conditions.

Export more than 15 thousand records


kflemin (approved):
🏆

PR #5065, opened 2025-09-04
Improve SEED export to import process

perryr16:
SEED ran into a few issues while importing data that had been previously exported

#### What's this PR do?
- Improves column matching to prevent duplicates
   - at present, there is a column constraint that prevents duplicate extra data columns, but a scenario exists where extra data columns are created with the same name as canonical columns. Not knowing the consequences of removing this constraint, logic has been added to more reliably match incoming columns to existing columns. For example, in the existing code an incoming column named `custom id 1` would not be mapped to the existing column `Custom ID 1`, and a new extra data column would be created causing confusion. Matching is now based on the incoming column matching the column_name or display_name of a canonical column (case insensitive).

- Handles a GEOSGemotry error: 
   - If a GOESGeometry object is initialized with a blank polygon (`POLYGON (())`) then it will throw a reader error. A guard has been put in place to default to an empty polygon (`POLYGON EMPTY`) that can be processed
   - Handled Error: django.contrib.gis.geos.error.GEOSException: Error encountered checking Geometry returned from GEOS C function "GEOSWKTReader_read_r".

- Conditionally adds a `"Property Footprint"` column on import. 
   - A "property_footprint" column was always added during a geoJSON import. This would cause a duplicate column during import if an incoming file contained a "Property Footprint" or "property_footprint" 

   - Existing property footprints will be used instead of always generating new ones. This was problematic if an incoming property has a property footprint, but no coordinate info.

- Exports (csv and geoJSON) do not include notes or labels unless the "Include Notes" checkbox is marked.
   - Exports could end up with duplicate Property Notes columns, as there is a canonical column called Property Notes as well as a Property Notes field that represented the text within a property's Notes model. 
   - Reduces the chance of an unnecessary Property Labels column
   - If included, the exported property's Notes model is now called "Property Notes Exported" to avoid confusion.

- GeoJSON exports now exclude -State metadata fields that can lead to redundant or recursive extra data fields during import
   - Excluded Fields: Access Level Names, groups_indicator, id, labels, merged_indicator, meters_exist_indicator, notes_count, property_labels, property_notes, property_state_id, property_view_id, related, taxlot_labels, taxlot_notes, taxlot_state_id, taxlot_view_id, 

- Improves error messaging during mapping 
   - Instead of a catch all duplicate headers error message, there are now two messages. One for SEED headers, and one for incoming Data File Headers.
   - If there are duplicate SEED Headers, the user can modify the header and continue
   - If there are duplicate Data File Headers, the user will have to modify the import file and try again
 
Do a series of import -> export geojson -> import -> export geojson -> import with sample data, as well as existing data.

Sample Data:  /seed/tests/data/nrel-site.geojson


perryr16:
see #5094, #5096

RDmitchell:
@perryr16 / @kflemin -- is there an issue associated with this PR that I can use for testing?

perryr16:
@RDmitchell, no, there's no official issue for this PR. This logic got split into 2 separate PRs (listed above) and each should have general info and a basic testing procedure in their descriptions. If it's helpful, I could generate an issue.

RDmitchell:
OK, I'll look at the PRs for testing instructions. Thx !

PR #5074, opened 2025-09-10
Fix inventory list date filters

perryr16:
Partial dates were causing the inventory list filter to fail. For example, filtering a date/dateime field with ">2025" would fail when trying to convert it to a datetime object.

#### What's this PR do?
Adds a `parse_date` function that can convert partial date strings to datetime objs.
- 2025
- 2025-01 (or 2025/01)
- 2025-01-01
- Full ISO format
Allows for the following comparisons: `=, !=, <, >, <=, >=`. However, all canonical datetime values are stored in the database as timezone-aware UTC datetimes with millisecond precision, direct equality comparisons (`=, !=`) are unreliable. It's recommended users use a date range to narrow in on a specific date (ex: `>2025-01-01, <2025-01-02`).

If an invalid or unexpected filter is caught (ex: `@#$%^`), an error message will appear, the spinner is hidden, and the filter is ignored.

Test the above partial dates and various operators on columns of data type date and datetime.
Input an invalid filter.

Extra Data:
Screenshot 2025-09-10 at 10 15 40 AM
Canonical:
Screenshot 2025-09-16 at 9 14 09 AM

### DIFF
--- a/seed/models/columns.py
+++ b/seed/models/columns.py
@@ -9,7 +9,6 @@ import locale
 import logging
 import os.path
 from collections import OrderedDict
-from datetime import datetime
 from typing import Any, Callable, Literal, Optional
 
 from django.apps import apps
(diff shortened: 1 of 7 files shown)


kflemin (approved):
🔥 
Screenshot 2025-09-16 at 9 46 00 AM

RDmitchell:
@perryr16 -- should this date format also work for this type of filtering? It doesn't seem like it does

image

There are dates < 1/31/2024
image

but filtering to see them doesn't seem to work. 
image

Issue #5079, opened 2025-09-12
Inventory: Filtering using multiple criteria doesn't seem to be working


RDmitchell:
Instance: production / staging
SHA:  b26466665
Org: BPS Test - Robin
Cycle: 2019 Compliance Cycle

Want to see the records that have Property Type = Police Station AND Fire Station. 

If I add labels to each type, I see that there are 7 records between the two types, 3 Fire stations, and 4 Police stations

<img width="1088" height="358" alt="Image" src="https://github.com/user-attachments/assets/3b762e22-bc1d-485d-bdc4-d463d6a5fcb6" />

But if I just want to use the filter boxes at the top of the columns, that doesn't seem to work. 

If I type **_fire_** in the filter box, the 3 records with property type = **_fire station_** are returned

<img width="1086" height="251" alt="Image" src="https://github.com/user-attachments/assets/67ca62d3-1625-46bc-b891-ce2786633240" />

If I type **_police_** in the filter box, the 4 records with property type = **_police station_** are returned

<img width="1086" height="270" alt="Image" src="https://github.com/user-attachments/assets/19841fd1-187f-4936-a73b-9f023be5008f" />

If I then add a comma and fire, so that the filter box is **_police, fire_** then only the 3 records with property type **_fire_** are returned

<img width="1082" height="268" alt="Image" src="https://github.com/user-attachments/assets/dce74de5-8a3e-4b15-9cdf-667ce84ffef4" />

If I add a third filter parameter of **_medical_**, then only the medical offices are returned, instead of all 3 property types.

<img width="1090" height="236" alt="Image" src="https://github.com/user-attachments/assets/49fb864c-222b-4d7f-991a-11c7c5dadbc1" />

RDmitchell:
This is potentially the AND filter condition

Is there an OR filter option in the filter input box?

RDmitchell:
instance: dev1
SHA: 7db009a58
Org: organization_id=406

If I filter on "Aquarium", I get one result

<img width="309" height="94" alt="Image" src="https://github.com/user-attachments/assets/ceb8ed76-325e-4e67-996b-615b0b4894ae" />

If I filter on "Financial Office", I get one result

<img width="296" height="132" alt="Image" src="https://github.com/user-attachments/assets/1322314d-d4a7-4b03-98d2-2b068b732423" />

 If I filter on both, separated by a comma, I get no results

[...]

PR #5091, opened 2025-09-18
Handle multiple inventory list filters

perryr16:
The inventory list filter should handle comma separated filters as multiple filters connected by an AND operator.
Screenshot 2025-09-18 at 2 24 36 PM
However, only the last filter was applied (ex: Custom ID 1 contains '3').

The frontend built the query_param string as a dictionary. This caused any duplicate keys to overwrtie one another. For example, if a user was trying to return all Custom ID 1's with a 1 and a 2
```
params = {
   custom_id_1__icontains=1,
   custom_id_1__icontains=2,
}
```
however, the query params would only reflect the last condition
```
params = {
   custom_id_1__icontains=2
}
```

#### What's this PR do?
- Concats values for identical keys. Ex. `custom_id_1__icontains=1`, `custom_id_1__icontains=2` becomes `custom_id_1__icontains=1,2`
- The backend will split those filter values apart and `AND` them together when filtering.
- Preserves the ability for `Goals`, `FilterGroups`, and the inventory_service's `get_properties`/`get_taxlots` to use the `build_view_filters_and_sorts` function.

Enter multiple conditions with the same operator into a inventory list filter. Ensure results are expected

#5079


kflemin (approved):
🏆

PR #5096, opened 2025-09-22
Improve matching logic to avoid duplicate columns

perryr16:
#### What's this PR do?
At present, there is a column constraint that prevents duplicate extra data columns
``` 
models.UniqueConstraint(fields=["organization", "column_name", "table_name", "is_extra_data"], name="unique_column_name"),
```
But this does not prevent extra data columns from being created with the same name as canonical columns. Not knowing the consequences of removing this constraint, logic has been added to more reliably match incoming columns to existing columns. 

For example, in the existing code an incoming column named `custom id 1` would not be mapped to the existing column `Custom ID 1`, and a new extra data column would be created named `custom id`. 

Matching is now based on the incoming column matching the `column_name` or `display_name` of a canonical column (case insensitive).

Import a file with variations on existing columns. For example, your import file headers could be
PM PROPERTY ID, custom id 1, address_line_1, Address Line 2, created, Updated.

All should map to their respective canonical columns. Created and Updated should be ignored


kflemin (approved):

RDmitchell:
@perryr16 / @kflemin -- any issue associated with this PR, or instructions for testing?

PR #5099, opened 2025-09-24
Include footprint in geojson export geometry

perryr16:
Footprint data was included in the "features" section of a geojson export instead of the "geometry" section. The resulting geojson did not plot any footprint data, only UBID bounding box, centroid, and center point (lat_long).

The issue was introduced when `property_footprint` and `taxlot_footprint` were added as canonical columns in [a 2019 migration](https://github.com/SEED-platform/seed/blob/develop/seed/migrations/0100_auto_20190228_1239.py#L87-L98). 

The geojson export function attempted to capture footprint data by looking for a property state key called exactly "property_footprint", however after it was converted to a column, it would appear as "property_footprint_{column id}".

Note: when serialized, columns have an additional attribute,  "name", which is 
```
"column_name" + _ + "column_id"
pm_property_id_123
```

#### What's this PR do?
Finds the `property_footprint` and `taxlot_footprint` columns and uses their serialized "name" to find the matching footprint data and adds it to the geojson geometries.

- Import data with property footprints as lat long coordinates
- Export data to geojson
- Open exported geojson in geojson.io
- Confirm that property footprint is plotted


kflemin (approved):

RDmitchell:
@kflemin -- so I guess this isn't deployed to staging yet, right? I thought it was, and told Maryland to test this functionality. But if it isn't deployed there yet, I will tell them not to do that test. Thx !

Issue #5109, opened 2025-10-03
Default Reports: Scatter plot not working (??) - comparing staging to production for the same org


RDmitchell:
This could always be user error (!) so it would be good if someone else could test to see if they are experiencing this. 

Instance: Staging
SHA: 7db009a58

Instance: Production
SHA: b26466665

Org for both: BPS Test - Robin

Staging has been updated to the latest patch release, and then the production DB was moved over to staging for testing. 

Looking at the Default reports for the BPS Test - Robin org, trying to plot exactly the same thing, the bar charts and statistics are identical between the two servers, but the scatter plot does not seem to be plotting the correct information. 

Here is what is showing on the production server, which I believe is correct

<img width="1799" height="1098" alt="Image" src="https://github.com/user-attachments/assets/7cdd8f52-8955-46ee-841c-96a7f4c5766d" />

Here is what is showing on the staging server, where the scatter plot doesn't seem to have the correct results

<img width="1834" height="1024" alt="Image" src="https://github.com/user-attachments/assets/3ade382d-9ea8-40e3-bd53-8e8a2029b943" />

RDmitchell:
@kflemin / @axelstudios / @perryr16 -- maybe one of you could look at this next week. 

Always very possible it's me and not SEED, so it would be good if one of you could test in your world.

PR #5111, opened 2025-10-06
Reset scale bounds on report change

perryr16:
Default report scales could get out of sync when changing between reports. Zooming in or out would set the default scales for any future report.

#### What's this PR do?
Resets the x & y max scale values on report change

Repeat the below procedure for both scatter and bar charts.
- Create 2 reports
- Load report 1
- Zoom in or out
- Change to report 2
- The report x & y scale min/max should appear meaningful (the reset zoom should always bring all data back into view).

#5109


seed/static/seed/js/controllers/inventory_reports_controller.js
```diff
@@ -629,12 +628,16 @@ angular.module('SEED.controller.inventory_reports', []).controller('inventory_re
...
               yAxisTickFormat: $scope.yAxisSelectedItem.axisTickFormat
             };
 
-            // new chartJS chart data
+            // yAxisVars.axisMin and axisMax are unused and will remain 'undefined'.
+            // They may be placeholders for future functionality.
+            // Leaving them undefined is ideal, as it lets Chart.js automatically determine the proper min/max values.
             $scope.scatterChart.options.scales.y.min = $scope.yAxisSelectedItem.axisMin;
+            $scope.scatterChart.options.scales.y.max = $scope.yAxisSelectedItem.axisMax;
             $scope.scatterChart.options.scales.y.type = $scope.chartData.chartData.every((d) => typeof d.y === 'number') ? 'linear' : 'category';
```
perryr16, line 641: I haven't found any code that would set `.axisMin` or `.axisMax`. But I'm leaving it in, in case I've missed something, or it's the base for future functionality. It shouldnt matter because resetting the scales to `undefined` fixes the issue.

kflemin (approved):

PR #5119, opened 2025-10-15
Show previously mapped data types

perryr16:
#### What's this PR do?
- Adds a "to_data_type" field to cached column mappings that can populate the data type field when reviewing an existing mapping

`NOTE`: This is forward only. It does not work with existing import file mappings.

- Import a file
- Go to data
- Click dataset name
- Click Data Mapping of newly imported file
- Ensure data types are populated

#5119


perryr16:
Functionality included in #5127

Issue #5040, opened 2025-07-31
Revise Hierarchy Permissioning for Overall Org Data


kflemin:
When a "member" or "viewer" of an organization that has a hierarchy views the Organizations page, we should restrict the information that they have access to.  Specifically, we should:
1. hide the total # of properties and taxlots in the org
2. hide the organization owners (?)

PR #5124, opened 2025-10-20
Limit organization data based on user access level

perryr16:
#### What's this PR do?
Limits org data based on user role and access level info. The following keys have been modified
- "owners": If a user is a non owner, the list is empty
- "cycles": The number of properties and taxlots in the "cycles" key is limited to those that the user has access to based on their access level instance.

- view organizations page with an owner and member/viewer. Owners list should reflect the current user role
- view the organizations page with a root and leaf user. Number of properties/taxlots should reflect the current user ALI

#5040

Owner/root in both orgs
Screenshot 2025-10-20 at 1 32 32 PM
Owner/root in "bb", member/leaf in "demo"
Screenshot 2025-10-20 at 1 30 12 PM


kflemin:
@perryr16 this is great! could you make one small modification: if the user is a viewer, don't show all the owners? but if they are a member, they can see the owners. thanks!

kflemin (approved):

PR #5133, opened 2025-11-06
Update buildingsync to version 2.7.0 and add federal bldg fields

kflemin:
- Updated BuildingSync version to 2.7.0 (was 2.6.0)
- Added FederalBuilding fields to buildingsync export for CERL workflow


seed/audit_template/audit_template.py
```diff
@@ -598,6 +600,123 @@ def _build_measures_element(em, property_id, building_id):
     ]
 
 
+def _build_federal_element(em, state, org):
```
haneslinger, line 603: I can take a crack at shortening this function if you like, but it works!

haneslinger (approved):

Issue #5088, opened 2025-09-17
Inventory List: Can't sort fields that are not associated with the tab being viewed, ie, can't sort Tax Lot fields in the Property Tab


RDmitchell:
Instance: Staging
SHA: b26466665

This seems new, but at this point I can't actually remember how the program used to work. 

But I think it used to be possible, in each tab (Property and Tax) to sort (ascending / descending) by any field in the inventory list view, even if the field was from the other "tab". 

Right now, you can only sort by fields that are associated with the tab being viewed.

For example, in the View by Property tab, it is only possible to sort by the Property fields, not the Tax Lot fields

<img width="735" height="299" alt="Image" src="https://github.com/user-attachments/assets/769d3961-393c-460e-985f-de5f252dd173" />

**_It should be possible to sort by any field displayed in the tab, regardless of whether it is a property or tax lot field._**

PR #5142, opened 2025-11-21
Sort on related columns from inventory list

perryr16:
Sorting has previously been disabled for related columns on the inventory list ui-grid

#### What's this PR do?
`_parse_view_sort` / `build_view_filters_and_sorts` now recognize columns that live on the related inventory (tax-lot fields on the property grid, or vice versa). When you sort on one of those columns, we build a per-record value from the related join and hand it back to the queryset, so the backend actually honors the sort order instead of ignoring it. For string-y related fields we run the value through the same “natural sort” treatment as first-party columns, which keeps mixed numeric strings in human order (1, 2, 10 instead of 1, 10, 2).

Added tests

Upload taxlots and properties, sort on related fields

#5088


perryr16:
The failing test should be addressed by PR #5146

PR #5156, opened 2025-12-18
Revert "Disable ability to remove matching criteria fields once inventory is added"

haneslinger:
…tory is added (#4222)"

This reverts commit 37d1da5f9b1f820fca13c594102bcb32197bfbc2.

If an org does not use Accountability Hierarchy, then they will (again) be allowed to remove a matching criteria field.  If an org does use Accountability Hierarchy, then they will NOT be able to remove a matching criteria field. 
Matching criteria fields can be added in both cases.


kflemin (approved):
🏆

PR #5176, opened 2026-02-20
Custom SMTP backend

JesseMBarr:
Corporate networks sometimes use internally signed certificates for smtp servers. In this case, Django either needs the certfile/keyfile, or to ignore the ssl context. This works in a dev environment with self-signed certificates as well.

#### What's this PR do?
- Added a custom smtp email backend that allows the user to toggle off TLS context.
- This bypasses certificate verification for smtp servers with internally signed (or self-signed) certificates.
- Realistically this should only be used in a development or highly trusted environment.
- This backend can easily be expanded on.
- Added an environment variable to toggle the TLS context (defaults to true)
- Added some logic for handling boolean environment variables, specifically smtp related for now.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
       - id: check-toml
       - id: check-xml
       - id: check-yaml
+        exclude: '(pnpm-lock\.yaml|pnpm-workspace\.yaml)'
       - id: debug-statements
       - id: end-of-file-fixer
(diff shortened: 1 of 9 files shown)


JesseMBarr:
I can't apply labels, but feature is probably the most accurate, though this PR also fixes a small bug when reading boolean environment variables.

axelstudios:
Looks good!

axelstudios (approved):

PR #5258, opened 2026-07-06
Fix error when logging out

nllong:
When clicking logout, the system would return an error. This branch fixed the logout-to-login error caused by /account/login/ rendering the two-factor login template without the context SEED expected.

The concrete error was a Django template variable lookup issue, especially the stale {{ username }} hidden input in the login template. The added test guards against django.template DEBUG errors on GET /account/login/ and checks that the page uses the real two-factor field id="id_auth-username" instead of name="username": seed/landing/tests.py:26.

The fix also makes /account/login/ route to CustomLoginView before the stock two-factor URLs: config/urls.py:66, and supplies default context for other_devices, backup_tokens, and cancel_url: seed/landing/views.py:224.

#### What's this PR do?
* fix the logout

* there is a new test added
* log in and then log out, make sure the terminal doesn't have any errors

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -63,7 +63,8 @@ urlpatterns = [
     path("api/token/refresh/", TokenRefreshView.as_view(), name="token_refresh"),
     path("api/version/", version, name="version"),
     path("api/", include((api, "seed"), namespace="api")),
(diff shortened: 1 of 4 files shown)


kflemin (approved):
🏆

PR #5259, opened 2026-07-06
Fix NLR name

nllong:


kflemin (approved):
looks good! i fixed a few alliance references too.

PR #5261, opened 2026-07-07
Fix migrations when restoring prod database, sort deterministically when ordering property|tax states

nllong:
When trying to restore the production database, a couple issues appeared with the migrations around primary key constraints.

#### What's this PR do?

- Fixed prod migrations by restoring missing primary key constraints on orgs_organization, orgs_organizationuser,
    seed_cycle, and seed_goal.
- Restored the missing unique_user_for_organization constraint and added guarded repair migrations so existing prod DBs can  migrate without failing on later foreign keys.
- Also fixes a small edge case of non-determinism in tests/operation where two records that are created at the exact same time could return different ordering. Updated by including the 'id' in the order by ( list(StateClass.objects.filter(id__in=state_ids).order_by("updated", "id").values_list("id", flat=True))

ideally restore prod database and make sure things hold together.


kflemin (approved):
tested locally on dev1 database 👍

PR #5269, opened 2026-07-14
fix: add --ignore-scripts to pnpm install in Dockerfile.ecs

dhaley:
## Summary

Fixes the ECS Docker build failure caused by pnpm preinstall scripts running during `pnpm install` and attempting network access (fetching internal certs) that isn't available in CodeBuild.

## Changes

- **Dockerfile.ecs**: Add `--ignore-scripts` to `pnpm install` to prevent preinstall hooks from running during Docker build. Add `--config.verify-deps-before-run=false` to the subsequent `pnpm build` (with comment explaining why).
- **Dockerfile.ecs.ci**: New CI-safe variant of the ECS Dockerfile (no internal cert fetch, no private registry access) for validating the production build in GitHub Actions.
- **.github/workflows/ci.yml**: Add job to build Dockerfile.ecs.ci on PRs touching Dockerfile.ecs or related files.
- **seed/tests/test_portfoliomanager.py**: Skip PM test class when credentials are unavailable.

## Review feedback addressed

- Documented why `--config.verify-deps-before-run=false` is required (preinstall hook skip)
- Fixed typo: "shared volunteers" -> "shared volumes"

Replaces #5252 (was opened from a fork branch that CodeBuild cannot access).

PR #5271, opened 2026-07-16
fix: ECS Docker build - replace corepack, pin Python 3.13, add pnpm --ignore-scripts

dhaley:
## Summary

Fixes the ECS Docker build failure caused by pnpm preinstall scripts running during `pnpm install` and attempting network access that isn't available in CodeBuild.

## Changes

- **Dockerfile.ecs**: Replace `corepack enable` with `npm install -g pnpm@11.0.8`. Add `--ignore-scripts` to `pnpm install`. Add `--config.verify-deps-before-run=false` to the Angular build step. Pin `uv sync --python 3.13` for aarch64/ARM Alpine compatibility.
- **Dockerfile.ecs.ci**: CI-safe variant (identical to Dockerfile.ecs minus internal cert fetch) for GitHub Actions validation.
- **.github/workflows/ci.yml**: Add `ecs-docker-build` job gated on relevant file changes. Add disk space cleanup step to unit test jobs.
- **seed/tests/test_portfoliomanager.py**: Skip PM test class when credentials unavailable.
- **.python-version**: Unchanged (stays at 3.14 for local development).

Replaces #5269 (had a messy git history from a bad merge revert that broke CI).

## Why corepack was removed

Corepack + pnpm have an unresolvable version conflict in node 24:

1. `corepack enable` installs a pnpm shim
2. The shim downloads the latest pnpm (e.g., 11.13.1)
3. pnpm reads `packageManager: pnpm@11.0.8` from package.json and rejects the mismatch
4. No env vars suppress this - it's pnpm's own enforcement, not corepack's

Fix: install pnpm@11.0.8 directly via `npm install -g`, bypassing corepack entirely.

## Why Python 3.13 in ECS Dockerfiles

This is an **aarch64 (ARM) specific issue**. Proven in #5272:
- `docker-build-no-cache` on x86_64: **passes**
- `docker-build-no-cache-arm` on aarch64: **fails**

Root cause:
1. uv's managed `cpython-3.14.4-linux-aarch64-musl` is built with **clang** and embeds `--rtlib=compiler-rt` in sysconfig
2. uv's managed `cpython-3.14.4-linux-x86_64-musl` is built with **gcc** and does NOT have this flag
3. `doublemetaphone==1.2` has no cp314 wheels on PyPI (only cp310-cp313)
4. On aarch64 Alpine: source build fails because g++ doesn't understand `--rtlib=compiler-rt`
5. On x86_64 Alpine: source build succeeds because no incompatible flag is present

This affects:
- Docker Desktop on Apple Silicon (macOS) - containers run aarch64
- ARM-based CodeBuild/ECS compute
- Any `--platform linux/arm64` cross-build

To reproduce on an x86_64 Linux machine:
```bash
docker buildx build --no-cache --platform linux/arm64 --tag seed:test-arm64 .
```

The fix pins `--python 3.13` in `uv sync` inside the ECS Dockerfiles only, which uses pre-built cp313 musllinux wheels (available for both architectures). `.python-version` stays at 3.14 for local development.


.python-version
```diff
@@ -1 +1 @@
-3.14
+3.13
```
axelstudios: I would really prefer not to downgrade Python globally for all developers and deployments - Python 3.14 works fine with Alpine and our dependencies, and has been deployed successfully (and [doublemetaphone builds](https://github.com/SEED-platform/seed/actions/runs/29459844376/job/87500829196#step:7:708))

axelstudios (changes requested):

dhaley:
main Dockerfile on develop branch now compiles for ARM and has appropriate tests, so this PR is not needed.

PR #5286, opened 2026-07-25
Fix column mapping export to handle omitted fields

kflemin:
column mapping profiles with omitted fields were crashing during export


nllong (approved):
looks amazing!

